PR #55, opened 2020-02-05
fix JSON struct

Efrat19:
Came across this and figured out you might want to fix the docs as well


Fenny:
Hi @Efrat19, we just moved our documentation to https://github.com/gofiber/docs 2 minutes ago :smile:

I will fix this typo right away, thanks for your contribution!

PR #72, opened 2020-02-07
Added context Bind support, it will try to bind JSON into variable

m1ome:


cgiacomi:
That is really cool, was actually thinking of that.
If I may add a small suggestion, it would be really great if the unmarshaling was done based on the 'content-type' passed in the header, so that if the content-type was for example 'application/xml' it would use an XML binder and so on for the different types. Just a suggestion.

Fenny:
@m1ome, @cgiacomi, thanks for the input! @koddr, can you join the discussion?

We want to keep the Fiber API in line with Expressjs ( minimalism ).

> If I may add a small suggestion, it would be really great if the unmarshaling was done based on the 'content-type' passed in the header, so that if the content-type was for example 'application/xml' it would use an XML binder and so on for the different types. Just a suggestion.

I do like this idea, maybe we could discuss to use `BodyParser()` instead of `Bind`.
https://github.com/expressjs/body-parser

Let me know what you think.

Until this feature is implemented, you can unmarshal json using your favorite parser:

```golang
	jsoniter "github.com/json-iterator/go"
)

type data struct {
	Foo string `json:"foo"`
}

func main() {
	app := fiber.New()

	// => {"foo": "bar"}
	app.Post("/test", func(c *fiber.Ctx) {
		var d data
		jsoniter.UnmarshalFromString(c.Body(), &d)
	})

	app.Listen(8080)
}
```

koddr:
@Fenny I agree with `BodyParser()`. It's more understandable, than `Bind()`.

Main idea of Fiber — unix way — instead of all-in-one multi-tool. 

> Each thing must perform one function, but do it extremely well.

Fenny:
Add `BodyParser` to fiber, https://fiber.wiki/#/context?id=bodyparser

Planned for the next release :+1:

cgiacomi:
Yeah the idea of BodyParser is great, and like mentioned it would keep the framework very lightweight.

PR #88, opened 2020-02-10
Use colorable for Windows

mattn:
image

### DIFF
--- a/application.go
+++ b/application.go
@@ -20,6 +20,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/mattn/go-colorable"
 	"github.com/valyala/fasthttp"
 	"github.com/valyala/fasthttp/reuseport"
 )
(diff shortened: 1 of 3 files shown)


Fenny:
@mattn, nice catch thanks!

koddr:
@Fenny it's okay, but it adds another addiction that I don't think is necessary at all. 

Does we really need to enter an extra package so that on Windows one line is not colored white? IMHO, absolutely not. **Fiber** is minimalism. I understand, why we use `jsontier` — it's fastest package than `encode/json`, but this... uh?

> And, btw, what is this console? CMD? PowerShell?

mattn:
I use cmd.exe. I don't mind you revert this and remove escape sequences.

keob:
@mattn You can try use powershell core，looks great😀

mattn:
Since powershell is too heavy to start-up, I don't use it.

FYI, Well, this issue will not be resolved eventhough you use powershell or powershell core. If you want to solve this, the console must be set ENABLE_VIRTUAL_TERMINAL_PROCESSING using API.

See https://docs.microsoft.com/en-us/windows/console/setconsolemode?redirectedfrom=MSDN#ENABLE_VIRTUAL_TERMINAL_PROCESSING

mattn:
image

keob:
@mattn 😅😅, thanks😀

PR #93, opened 2020-02-10
Remove escape sequences

mattn:
If you want to remove dependencies to packages Windows specific, feel free to merge this,

See https://github.com/gofiber/fiber/pull/88#issuecomment-584027033


Fenny:
It's fine for now, we might remove all colors in the future.

PR #113, opened 2020-02-11
German README

MarvinJWendt:
Hi, first of all congratulations on such an awesome hype 🚀 

I have translated the whole `README.md` to German.

This is a draft because I want to look over it again tomorrow, so that I can find mistakes, if there are any that I don't see now.
If somebody else finds some, feel free to write them down :)


Fenny:
Looks good! Thanks for contributing @MarvinJWendt

Issue #126, opened 2020-02-12
🤔 Is there any way for handling JSON error?


1995parham:
**Question description**
I have used `gofiber` for creating a REST API with JSON. The `JSON` function returns error so I must handle it every time in responses. Is there any better way of doing this?

**Code snippet (optional)**

```go
package main

import "github.com/gofiber/fiber"

func main() {
  app := fiber.New()
  app.Get("/json", func (c *fiber.Ctx) {
    if err := c.Status(200).JSON("Hello World"); err != nil {
      panic(err)
    }
  })
}
```

koddr:
@1995parham thx for issue.

We're working around easy way to error handling right now on Fiber. Stay tuned! 😉

1995parham:
@koddr Thanks for your response 🙌 I will wait.

Fenny:
@1995parham, thanks for submitting your issue.
`JSON` sets the body to `""` and `Content-Type: application/json` by default.
The status code is always 200 by default unless you set it. ( less code :) )
```go
package main

import "github.com/gofiber/fiber"

func main() {
  app := fiber.New()
  app.Get("/json", func (c *fiber.Ctx) {
    // When working with strings, nothing could go wrong
    // You are not required to check for errors, specially when no structs are serialized
    c.JSON("Hello World");
    // When you work with structs, its a good behavior to check for errors
    // If the JSON serialization goes wrong, it just sets the body to ""
    // For now you can check for errors to display a custom message
    if err := c.JSON(&user); err != nil {
      c.Status(500).JSON("Something went wrong")
    }
  })
}
```
We are working on a better way to handle global errors.

1995parham:
@Fenny I agree with your solution, we can return an error from the handler. If you want, I can create a PR for it.

Fenny:
@1995parham, PR's are always welcome :+1:
I have time to work in the afternoon :+1:

1995parham:
I will create PR so we can work on it 🙌

Fenny:
@1995parham, me and @koddr are actually discussing some kind of `recover` feature.
This has nothing to do with the `c.Next(err)`, but we would like to hear your thoughts on this.

```go
package main

import (
	"./fiber"
)

func main() {
  app := fiber.New()
  
  app.Get("/", func(c *fiber.Ctx) {
    panic("I panic")
  })
  
  // app.Recover, app.Error... ?
  app.Panic(func(c *fiber.Ctx) {
[...]

PR #128, opened 2020-02-12
Add Recover

Fenny:
https://github.com/gofiber/fiber/issues/126
```go
//
// Recover ( disabled by default, unless a custom Recover handler is set )
// We might enable Recover by default with "500 Internal Server Error" in the future
//
func main() {
  app := fiber.New()
    panic("I panic")
  })
  app.Recover(func(c *fiber.Ctx) {
    c.Status(500).Send(c.Error()) // => 500 "I panic"
  })
  app.Listen(3000)
}
```

```go
//
// Context error
//
package main

import "github.com/gofiber/fiber"

func main() {
  app := fiber.New()
  app.Post("/api/register", func (c *fiber.Ctx) {
    if err := c.JSON(&User); err != nil {
      c.Next(err)
    }
  })
  app.Get("/api/user", func (c *fiber.Ctx) {
    if err := c.JSON(&User); err != nil {
      c.Next(err)
    }
  })
  app.Put("/api/update", func (c *fiber.Ctx) {
    if err := c.JSON(&User); err != nil {
      c.Next(err)
    }
  })
  app.Use("/api", func(c *fiber.Ctx) {
    c.Set("Content-Type", "application/json")
    c.Status(500).Send(c.Error())
  })
  app.Listen(1337)
}
```

### DIFF
--- a/application.go
+++ b/application.go
@@ -30,7 +30,7 @@ import (
 
 const (
 	// Version : Fiber release
-	Version = "1.4.4"
+	Version = "1.5.0"
 	// Website : Fiber documentation
 	Website = "https://fiber.wiki"
 	banner  = "\x1b[1;32m" + ` ______   __     ______     ______     ______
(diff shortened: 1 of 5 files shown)


1995parham:
@Fenny Thanks 🙌 I will use it in my project!

koddr:
@1995parham it would be great to some feedback after using and testing this features 👍

PR #131, opened 2020-02-12
Add OSX passing TravisCI badge, Re-organized flags

koddr:


.github/README.md
```diff
@@ -234,7 +237,7 @@ func main() {
 
 ## 👍 Contribute
 
-If you want to say **thank you** and/or support the active development of `fiber`:
+If you want to say **thank you** and/or support the active development of `Fiber`:
```
MarvinJWendt: I think this should be changed in every language then, for consistency 😄

koddr: Yep, I worked for RU translate, but it's not done now. Therefore, EN readme is our front and need to be awesome 😉

PR #147, opened 2020-02-16
Deprecate old functions

Fenny:


request.go
```diff
@@ -310,7 +311,7 @@ func (ctx *Ctx) IPs() []string {
 }
 
 // Is : https://fiber.wiki/context#is
-func (ctx *Ctx) Is(ext string) bool {
+func (ctx *Ctx) IS(ext string) bool {
```
koddr: Why `Is` on uppercase?

koddr: oh, my PR #148 fails because this change 😞

PR #148, opened 2020-02-16
Update some tests

koddr:
- application_test
- request_test


koddr:
@Fenny this fail, because method switch name (`Is -> IS`), see PR #148

Issue #178, opened 2020-02-22
🔥 Consider strong typing for API


alex-shpak:
**Is your feature request related to a problem?**
Hi!

In Go (IMO) it is considered bad practice to use `interface{}` if strongly typed API can be implemented.
I understand that idea is to have similarity with the `expressjs` library, but now developer has to visit documentation/source to identify what are proper params types for the routing function.

Also I think strongly typed API is one of things developers look after when choosing framework. It also can decrease amount of runtime checks and will show errors/mistypes during compile time.

**Describe the solution you'd like**
As an example for routing function:
```go
func (app *App) Get(
    path string, 
    handler func()(ctx *Ctx), 
    middlewares ...func()(ctx *Ctx)
) App* {
    ...
}
```

**Describe alternatives you've considered**
Leave it as it is ¯\\_(ツ)_/¯

Fenny:
Fiber started as a personal project (coming from node) to rewrite all my web apps into Go.
I tried to stick to the same workflow as Express by using interfaces, this is indeed not the Go standard.

I do not want to break the current API by changing this, it could be possible for `v2` and I would love to hear your thoughts on this.

alex-shpak:
I see.
Well, it is up to library maintainers to decide. I feel like having non strongly typed API can turn off some Go developers. Would be nice to have `¯\_(ツ)_/¯`.

In my case I added `fiber` to dependencies, started to write `main` function to try around and hit `app.` autocompletion, saw `args ...interface{}` and was like: damn... now I need to go find docs.

And not to mention I'm a member of strongly typed camp. 😄

hyingreborn:
agree，Fiber made it easy for me to migrate from nodejs to Go。
The reason I chose to learn Go instead of TypeScript:
- A strong typed language can reduce errors and catch exceptions more easily
- Moderate development efficiency and adequate operational performance
- Cross-compile into binaries for easy distribution and deployment

Now, it's true that docs are often needed

Fenny:
I have discussed this with @koddr, and we want to share our conclusion on this matter.

Most methods that use the `...interface{}` argument are used for `app.METHOD` to register routes.
We think that this is straightforward because this is the first thing you learn visiting Fiber or coming from Express.

Now there are some `*Ctx` methods that also accept `interface{}` as an argument:
`func (ctx *Ctx) Body(args ...interface{}) string`
`func (ctx *Ctx) Cookies(args ...interface{}) string`
[...]

Issue #159, opened 2020-02-20
How do i use other flag parse tools ,like kingpin🤔 


dominicqi:
**Question description**
`fiber.New()` calls `flag.Parse` 。but when i use `kingpin`,they are in conflict.
i think the `fiber.New()` may should not call  `flag.Parse()` . What do you think？
Could there be another solution?Sorry, I'm new to golang.
**Code snippet (optional)**

```go
package main

import (
	"github.com/gofiber/fiber"
	"gopkg.in/alecthomas/kingpin.v2"
	"os"
)

func main()  {
	var testFlag string
	application := kingpin.New("demo", "demo")
	application.Flag("test","kinping flag").Default("test").StringVar(&testFlag)
	application.Parse(os.Args[1:])
	server := fiber.New()
	server.Listen(8080)
}
```

Fenny:
Interesting, thnx for the bug report. I will look into it asap.

dominicqi:
> Fiber and other frameworks rely on `flag.Parse`. This problem could be on `Kingpins` end. Does it relate to this issue [alecthomas/kingpin#194](https://github.com/alecthomas/kingpin/pull/194)?
> 
> If you could provide an error log, we might be able to help out.

I edit the code snippet. if no flag set  the server start success.  but if i start with flag
`
./test --test=kingpin
flag provided but not defined: -test
Usage of ./test:
  -child
        is child process
  -prefork
        use prefork
`

dominicqi:
I'm a little confused why `prefork` and `child` setting by flag?  why not like other  parameters  like 'Banner' or 'Engine.xxx'.  If someone use `fiber` , they have to  set the flags,if they want change .It sounds a little strange.

Fenny:
This is what I found regarding your issue: https://stackoverflow.com/questions/29699982/go-test-flag-flag-provided-but-not-defined

Did you try https://github.com/golang/go/issues/31859#issuecomment-489889428https://github.com/golang/go/issues/31859#issuecomment-489889428
```go
var _ = func() bool {
    testing.Init()
    return true
}()
```

Fiber uses flags to control child processes https://github.com/gofiber/fiber/blob/master/app.go#L301
And one of our features is to run `./fiber -prefork`

dominicqi:
> @dominicqi I created a PR with a possible fix, could you try it again with the master branch?

It not work. 
`if flag.Lookup(name) == nil ` always be true.
`Lookup` find flag reg by `flag.Var`。
[...]

Issue #154, opened 2020-02-18
🔥 Add some middleware


hyingreborn:
**Is your feature request related to a problem?**

**Describe the solution you'd like**
Looger、Recovery(is necessary in the development)
Ggzip(Currently Static() provides compression, but cannot be customized)
CORS、CSRF、JWT Auth etc...

**Describe alternatives you've considered**
Are there plans for these middleware?

**Additional context**

Fenny:
We are focusing on v2 at the moment, when this is released we have more time to work on the middleware repository. Until then, PR's are more than welcome!

PR #184, opened 2020-02-24
March v1.7.2

Fenny:
March **v1.7.2**

**[NEW]** app.Settings.Immutable / _n/A_
**[NEW]** middleware.Cors https://github.com/gofiber/fiber/issues/154
**[NEW]** middleware.Secure https://github.com/gofiber/fiber/issues/154
**[NEW]** middleware.Redirect https://github.com/gofiber/fiber/issues/154

**[UPDATE]** *Cookie
**[UPDATE]** ctx.Cookie(*Cookie)
**[UPDATE]** fasthttp/websocket > `v1.4.2`

**[FIX]** Immutable values https://github.com/gofiber/fiber/issues/185
**[FIX]** Stronger typed API https://github.com/gofiber/fiber/issues/178
**[FIX]** Argument conflict https://github.com/gofiber/fiber/issues/159
**[FIX]** Convert Recover error
**[FIX]** Bundle template dependencies


app.go
```diff
@@ -72,19 +72,27 @@ type (
 	}
 )
 
-var prefork = flag.Bool("fiber-prefork", false, "use prefork")
-var child = flag.Bool("fiber-child", false, "is child process")
+var prefork bool
+var child bool
 
-// New: https://fiber.wiki/application#new
+func init() {
```
koddr: but `init()` is bad practice for testing, as I get comment on Dev.to 🤔

utils.go
```diff
@@ -136,6 +137,16 @@ func getBytes(s string) (b []byte) {
 	return *(*[]byte)(unsafe.Pointer(&s))
 }
 
+// https://stackoverflow.com/questions/49193480/golang-flag-redefined
+func regBoolVar(p *bool, name string, value bool, usage string) {
```
golangcibot: `regBoolVar` is unused (from `deadcode`)

utils.go
```diff
@@ -136,6 +137,16 @@ func getBytes(s string) (b []byte) {
...
+	if flag.Lookup(name) == nil {
+		flag.BoolVar(p, name, value, usage)
+	}
+}
+func getBoolFlag(name string) bool {
```
golangcibot: `getBoolFlag` is unused (from `deadcode`)

app.go
alex-shpak: My gut feeling is that it was better with `fiber` prefix to avoid collision with end-user flags. Or maybe no flags at all and pass everything via `Settings`?

group.go
```diff
@@ -12,113 +12,113 @@ type Group struct {
...
 
-// Group ...
-func (grp *Group) Group(newPrfx string, args ...interface{}) *Group {
-	var prefix = grp.prefix
-	if len(newPrfx) > 0 && newPrfx[0] != '/' && newPrfx[0] != '*' {
-		newPrfx = "/" + newPrfx
+// Group : https://fiber.wiki/application#group
+func (grp *Group) Group(prefix string, handlers ...interface{}) *Group {
+	var oldPrefix = grp.prefix
+	if len(prefix) > 0 && prefix[0] != '/' && prefix[0] != '*' {
```
alex-shpak: Will there be unexplained behaviour if somebody will pass `*/abrakadabra` as `prefix`? Maybe compare entire `prefix` with `*`?

Fenny: Our current wildcard logic is that everything behind * is irrelevant, everything behind a `*` is not matched.

alex-shpak: Okay 👌

app.go
Fenny: We need the flags because we are working with separate child processes using prefork. They do not share memory. I solved the `flag.Parse()` conflict looking at this solution: https://github.com/valyala/fasthttp/blob/master/prefork/prefork.go#L45,L60

context.go
```diff
@@ -119,7 +119,7 @@ func (ctx *Ctx) Accepts(offers ...string) (offer string) {
 	if len(offers) == 0 {
 		return ""
 	}
-	h := ctx.Get(fasthttp.HeaderAccept)
+	h := ctx.Get(HeaderAccept)
```
koddr: Maybe simplify to:

```go
if h := ctx.Get(HeaderAccept); h == "" {
	return offers[0]
}
```

for more beauty 😉

ulissesferreira:
Sorry guys, was taking a look at the update and approved changes without meaning too 😭. The CORS middleware comes in handy

context.go
Fenny: But we use the `h` value later on, if we wrap it inside an `if statement` we lose context of the value and get `h is undefined` :<

Fenny:
> Sorry guys, was taking a look at the update and approved changes without meaning too sob. The CORS middleware comes in handy

We have to delay middleware to `v1.8.1`, it needs a bit more time

PR #196, opened 2020-02-29
support for range header

hanFengSan:
Support for range header
https://expressjs.com/en/api.html#req.range
Don't support the option of "combine". If you think it is necessary, I can add it later.

### DIFF
--- a/context.go
+++ b/context.go
@@ -560,7 +560,7 @@ func (ctx *Ctx) Query(key string) (value string) {
 // Range : https://fiber.wiki/context#range
 func (ctx *Ctx) Range(size int64) (rangeInfo RangeInfo, err error) {
 	rangeStr := string(ctx.Fasthttp.Request.Header.Peek("range"))
-	if rangeStr == "" || strings.Index(rangeStr, "=") < 0 {
+	if rangeStr == "" || !strings.Contains(rangeStr, "=") {
 		return rangeInfo, fmt.Errorf("malformed range header string")
 	}
 	data := strings.Split(rangeStr, "=")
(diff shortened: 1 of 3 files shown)


context.go
```diff
@@ -548,11 +558,45 @@ func (ctx *Ctx) Query(key string) (value string) {
...
 
 // Range : https://fiber.wiki/context#range
-func (ctx *Ctx) Range() {
-	// https://expressjs.com/en/api.html#req.range
-	// https://github.com/jshttp/range-parser/blob/master/index.js
-	// r := ctx.Fasthttp.Request.Header.Peek(HeaderRange)
-	// *magic*
+func (ctx *Ctx) Range(size int64) (rangeInfo RangeInfo, err error) {
+	rangeStr := string(ctx.Fasthttp.Request.Header.Peek("range"))
+	if rangeStr == "" || strings.Index(rangeStr, "=") < 0 {
```
golangcibot: S1003: should use !strings.Contains(rangeStr, "=") instead (from `gosimple`)

Fenny:
@hanFengSan, thanks for the PR!
Everything looks good, but I got two suggestions before I we can merge.
- I think that the size argument should be optional
```Range(size int64) => Range(size ...int64) ```
When left blank, it should return all ranges available.
```Range: pages=30-40, 60-70```
```c.Range() -> { Type: 'pages', Ranges: [ [ 30, 40 ], [ 60, 70 ] ] }```
- How does it behave if a the range header contains `bytes=0-`

Awesome PR, let me know what you think.

PR #247, opened 2020-03-28
Fixed inconsistency in middleware examples in README

ochrist2:
After copying this example from the logger middleware
```
        // Optional logger config
	config := logger.LoggerConfig{
		Format:     "${time} - ${method} ${path}\n",
		TimeFormat: "Mon, 2 Jan 2006 15:04:05 MST",
	}

	// Logger with config
	app.Use(logger.New(config))
```
and running it I got the error `./main.go:18:12: undefined: logger.LoggerConfig`, changing `.LoggerConfig` to `.Config` fixed the issue as I imagine there was a name change on the API recently.

Similarly, in the example for the recovery middleware there is a `recover.LoggerConfig` instead of `recover.Config`. 

I've renamed all occurances of `LoggerConfig` in the README files to `Config` to fix these issues so the examples can be copy and pasted without errors. I thought I'd fix it for you guys instead of just telling you about it, my apologies if I've got something wrong.

### DIFF
--- a/.github/README.md
+++ b/.github/README.md
@@ -448,7 +448,7 @@ func main() {
   app := fiber.New()
 
   // Optional recover config
-  config := recover.LoggerConfig{
+  config := recover.Config{
     Handler: func(c *fiber.Ctx, err error) {
 			c.SendString(err.Error())
 			c.SendStatus(500)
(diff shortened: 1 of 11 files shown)


Fenny:
Your awesome!

PR #252, opened 2020-03-31
v1.8.431

Fenny:
**[FIX]** `c.XHR()` now checks for in case sensitive `xmlhttprequest`

### DIFF
--- a/app.go
+++ b/app.go
@@ -29,8 +29,8 @@ const Version = "1.8.43"
 // Map is a shortcut for map[string]interface{}
 type Map map[string]interface{}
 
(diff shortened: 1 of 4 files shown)


app.go
```diff
@@ -29,8 +29,8 @@ const Version = "1.8.43"
 // Map is a shortcut for map[string]interface{}
 type Map map[string]interface{}
 
-// Fiber denotes the Fiber application.
-type Fiber struct {
+// App denotes the Fiber application.
+type App struct {
```
renanbastos93: Please, decide. Will be to use `*fiber.Fiber` or `*fiber.App`?

PR #263, opened 2020-04-07
Fixing the routing wildcard example

agbaraka:


Fenny:
@agbaraka thanks for your PR! If you could fix this example in every translated `readme` I can merge!

PR #266, opened 2020-04-07
Fix listen message printing when error

UJPrasad:
Fixed,

```
Fiber xx listening on xx
```

Even if there was an error

### DIFF
--- a/app.go
+++ b/app.go
@@ -339,10 +339,7 @@ func (app *App) Listen(address interface{}, tlsconfig ...*tls.Config) error {
 	}
 	// Create fasthttp server
 	app.server = app.newServer()
-	// Print listening message
-	if !isChild {
-		fmt.Printf("Fiber v%s listening on %s\n", Version, addr)
-	}
+
 	var ln net.Listener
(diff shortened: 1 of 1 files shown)


Fenny:
Will this still print `Fiber xx listening on xx` even if there is no error? 
`app.server.Serve` is a blocking function.

UJPrasad:
@Fenny I've made a few changes, I guess this way we get net/listener errors leaving the user to manage the serve errors from fasthttp?

@renanbastos93 I guess the current PR doesn't need much of test as it still getaways with the errors from fasthttp's Serve?

Thanks!

renanbastos93:
> @Fenny I've made a few changes, I guess this way we get net/listener errors leaving the user to manage the serve errors from fasthttp?
> 
> @renanbastos93 I guess the current PR doesn't need much of test as it still getaways with the errors from fasthttp's Serve?
> 
> Thanks!

It was a question to instigate more your mind but I believe that it's good to force think for we being to build the big things. hehehe
You made a good job.

Thanks for contributing.

PR #280, opened 2020-04-19
added example for route middleware

saiumesh535:
Added route level middleware example


Fenny:
@saiumesh535 this would be a good example in our [recipes](https://github.com/gofiber/recipes) repo!

PR #288, opened 2020-04-23
Added ctx.QueryMulti()

monzarrugh:
QueryMulti returns all values for the query string parameter in the url.

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -600,6 +600,16 @@ func (ctx *Ctx) Query(key string) (value string) {
 	return getString(ctx.Fasthttp.QueryArgs().Peek(key))
 }
 
+// QueryMulti returns all values for the query string parameter in the url.
+func(ctx *Ctx) QueryMulti(key string) (values []string) {
+	valuesBytes := ctx.Fasthttp.QueryArgs().PeekMulti(key)
+	values := make([]string, len(valuesBytes))
+	for i, v := range valuesBytes {
+		values[i] = getString(v)
+	}
+	return values
+}
+
 // Range returns a struct containing the type and a slice of ranges.
 func (ctx *Ctx) Range(size int) (rangeData Range, err error) {
 	rangeStr := string(ctx.Fasthttp.Request.Header.Peek(HeaderRange))
--- a/ctx.go
+++ b/ctx.go
@@ -603,7 +603,7 @@ func (ctx *Ctx) Query(key string) (value string) {
 // QueryMulti returns all values for the query string parameter in the url.
 func(ctx *Ctx) QueryMulti(key string) (values []string) {
 	valuesBytes := ctx.Fasthttp.QueryArgs().PeekMulti(key)
-	values := make([]string, len(valuesBytes))
+	values = make([]string, len(valuesBytes))
 	for i, v := range valuesBytes {
 		values[i] = getString(v)
 	}


ctx.go
```diff
@@ -600,6 +600,16 @@ func (ctx *Ctx) Query(key string) (value string) {
 	return getString(ctx.Fasthttp.QueryArgs().Peek(key))
 }
 
+// QueryMulti returns all values for the query string parameter in the url.
+func(ctx *Ctx) QueryMulti(key string) (values []string) {
+	valuesBytes := ctx.Fasthttp.QueryArgs().PeekMulti(key)
+	values := make([]string, len(valuesBytes))
```
renanbastos93, line 606: https://github.com/gofiber/fiber/pull/288/checks?check_run_id=613350878#step:4:19
> [line 606 : column 9] - no new variables on left side of :=

thomasvvugt:
Hi @monzarrugh !

Thanks for opening this Pull Request. I think it's a nice addition and I do see use cases of having multiple query parameters sharing the same key. However, I was unable to find any official documentation of having duplicate key entries in query strings.

Commonly seen when having the same identifier and data type in query string, is adding the model name, i.e. `/logs?user_id=1&server_id=2`.

Because this pull request does not share a default or common use case, I would suggest using the pointer to FastHTTP's RequestCtx `ctx.Fasthttp.QueryArgs().PeekMulti(key)`, rather than implementing this function into Fiber itself.

Let us know if you have any questions!

monzarrugh:
> Hi @monzarrugh !
> 
> Thanks for opening this Pull Request. I think it's a nice addition and I do see use cases of having multiple query parameters sharing the same key. However, I was unable to find any official documentation of having duplicate key entries in query strings.
> 
> Commonly seen when having the same identifier and data type in query string, is adding the model name, i.e. `/logs?user_id=1&server_id=2`.
> 
> Because this pull request does not share a default or common use case, I would suggest using the pointer to FastHTTP's RequestCtx `ctx.Fasthttp.QueryArgs().PeekMulti(key)`, rather than implementing this function into Fiber itself.
> 
> Let us know if you have any questions!

I added it because I found the need for it for my own use actually.

The use cause would be if you would implement filtering by multiple values for the same key (categories) for example.
`/products?cat=shoes?cat=shirts?cat=bags`

Fenny:
Hi @monzarrugh and welcome, thank you for trying to improve fiber by making a PR that contains a possible new feature for fiber.

I do agree with @thomasvvugt, [RFC 3986](http://tools.ietf.org/html/rfc3986#section-3.4) has a section `3.4. Query`, which refers to key=value pairs, but nothing is said on how to interpret order and duplicate fields and so on. 

This means you may do what you like but at the same time duplicate keys have no standard when it comes to the order of the values (`first-given`, `last-given`, `array-of-all`,), in addition to the server keeping the order of the query string, there is also the question about the browser sending them in DOM (or some other fixed) order.

I did some research on how other HTTP back-end's parse duplicate queries and found this report from [OWASP](https://www.owasp.org/images/b/ba/AppsecEU09_CarettoniDiPaola_v0.8.pdf) that analyzed HTTP Parameter Pollution attacks. `Page 9` clearly shows that many frameworks are inconsistent when it comes to parsing duplicate queries. Some only use the first or last occurrence.

For your use case you could also use one query key and divide the values, with this method you have a consistent result on both the server and client side. `/products?cat=shoes,shirts,bags`.

Since it's use-case depended and we want to keep Fiber's API as clean as possible and inline with [expressjs](https://expressjs.com/en/4x/api.html#req). I think that this would not have a valid purpose in this framework.

If you are stuck with `/products?cat=shoes?cat=shirts?cat=bags` my advice would be to use the `ctx.Fasthttp.QueryArgs().PeekMulti("cat")` or wrap a custom function for it,

I hope this answered your question, thanks for your contribution!

PR #291, opened 2020-04-24
add validator

Shareed2k:
you can register a custom validator using App#Validator and leverage third-party libraries.
like in [echo](https://echo.labstack.com/guide/request) 

Example below uses https://github.com/go-playground/validator framework for validation:

```go
type (
	User struct {
		Name  string `json:"name" validate:"required"`
		Email string `json:"email" validate:"required,email"`
	}

	CustomValidator struct {
		validator *validator.Validate
	}
)

func (v *CustomValidator) Validate(i interface{}) error {
	return v.validator.Struct(i)
}

app := fiber.New()
app.Validator = &CustomValidator{validator: validator.New()}

app.Post("/test", func(ctx *Ctx) {
		user:= new(User)
		if err := ctx.BodyParser(user); err != nil {
			log.Fatal(err)
		}

		if err := ctx.Validate(user); err != nil {
			log.Fatal(err)
		}
	})
```

### DIFF
--- a/app.go
+++ b/app.go
@@ -24,16 +24,22 @@ import (
 )
 
 // Version of current package
-const Version = "1.9.2"
+const Version = "1.9.3"
 
 // Map is a shortcut for map[string]interface{}
 type Map map[string]interface{}
 
+// Validator is the interface that wraps the Validate function.
+type Validator interface {
+	Validate(i interface{}) error
+}
+
 // App denotes the Fiber application.
 type App struct {
-	server   *fasthttp.Server // FastHTTP server
-	routes   []*Route         // Route stack
-	Settings *Settings        // Fiber settings
+	server    *fasthttp.Server // FastHTTP server
+	routes    []*Route         // Route stack
+	Settings  *Settings        // Fiber settings
+	Validator Validator        // Validator
 }
 
 // Settings holds is a struct holding the server settings
--- a/ctx.go
+++ b/ctx.go
@@ -6,7 +6,6 @@ package fiber
 
 import (
 	"bytes"
-
 	"encoding/json"
 	"encoding/xml"
 	"fmt"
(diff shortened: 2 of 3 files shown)


renanbastos93:
Interesting, Can you put more details about this approach?
What is the motivation?

Shareed2k:
> Interesting, Can you put more details about this approach?
> What is the motivation?

look in description

thomasvvugt:
Hi @Shareed2k ,

Thanks for creating your pull request on implementing a custom validator for structs.
However, I do think it would be a nice idea to be able to implement a validator per seperate struct.
Thus having multiple validators having a 1:1 relationship with the structs, creating an App.Validators array of Validator structs.
As an example; having a specific validator for a User struct and having a different validator the a Group struct.

Let me know what you think!

Shareed2k:
> Hi @Shareed2k ,
> 
> Thanks for creating your pull request on implementing a custom validator for structs.
> However, I do think it would be a nice idea to be able to implement a validator per seperate struct.
> Thus having multiple validators having a 1:1 relationship with the structs, creating an App.Validators array of Validator structs.
> As an example; having a specific validator for a User struct and having a different validator the a Group struct.
> 
> Let me know what you think!

not sure i'm understand you, you can use one validator with many structs, it caching structs inside for example in [go-playground/validator](https://github.com/go-playground/validator/blob/7e57ca0cf54664aeb738b2eb275e656362129cf9/validator_instance.go#L89), why do you need many validators? can you give me an example?

thomasvvugt:
Hi @Shareed2k , 
Right I was incorrect as I was thinking the other way around.
The provider example is clear to me, I'm awaiting the other maintainers their opinions.
Thanks!

Fenny:
Thanks @Shareed2k for your contribution. I thought about this PR for a while and came to the conclusion that it does not belong in Fiber. Let me try to explain why.

We try to keep the Fiber API in line with [expressjs API](https://expressjs.com/en/4x/api.html#req), this way we keep Fiber core clean and maintainable using almost no dependencies.

I think that using a Validator is up to the user on how to implement this. Fiber gives the tools to take care of the HTTP protocol / routing, so the user can build their app logic around it. 

I hope this makes sense, what are your thoughts @gofiber/maintainers

jozsefsallai:
I agree with @Fenny here, yeah. Perhaps something like this could work as a middleware?

koddr:
It's looks like middleware, agree with @jozsefsallai

Shareed2k:
as you wish... what can i say, and it can't work as middleware...

PR #308, opened 2020-04-26
Schema Decoder: Ignore unknown fields by default

khaibin:
Hi Guys,

Thanks for your great work! It seems like BodyParser is generating error when there are unknown fields in the form/query. For example:

```
type User struct {
	Email string `form:"email"`
	Password string `form:"password"`
}
```

If I send the request with these parameters:
email: abc@gmail.com
password: abc
token_type: jwt

It will generate error `schema: invalid path "token_type"`

I think it should be ignored peacefully. Let me know how you think :)

### DIFF
--- a/app.go
+++ b/app.go
@@ -82,7 +82,9 @@ type Group struct {
 // You can pass optional settings when creating a new instance.
 func New(settings ...*Settings) *App {
 	schemaDecoderForm.SetAliasTag("form")
+	schemaDecoderForm.IgnoreUnknownKeys(true)
 	schemaDecoderQuery.SetAliasTag("query")
+	schemaDecoderQuery.IgnoreUnknownKeys(true)
 	// Create app
 	app := new(App)
 	// Create settings


Fenny:
Welcome @khaibin, thanks for opening your first issue.

Fiber uses the [Schema]() package to decode the request body, this library does support [IgnoreUnknownKeys](https://pkg.go.dev/github.com/gorilla/schema?tab=doc#Decoder.IgnoreUnknownKeys) but it's not enabled by default.

I invite the @gofiber/maintainers to also give their feedback on whether we should enable `IgnoreUnknownKeys`(https://pkg.go.dev/github.com/gorilla/schema?tab=doc#Decoder.IgnoreUnknownKeys) by default in our [schema decoder](https://github.com/gofiber/fiber/blob/master/ctx.go#L64)

khaibin:
Hi @Fenny, consider this:

user := &User{}
err := c.BodyParser(user)
if err != nil {
  c.SendStatus(400)
}

By default, the server should not return with 400 (Bad request) when additional fields are present. There are cases where we need some extra fields to do further processing. 

Gorilla schema returns the error instance as MultiError and it will complicate the process of handling unknown fields.

Gin are ignoring the fields by default for all Binding methods:
https://github.com/gin-gonic/gin#model-binding-and-validation

Fenny:
@khaibin if you could add testing for this by adding another field to the struct in the [ctx_test.go#L156](https://github.com/gofiber/fiber/blob/master/ctx_test.go#L156) that would be great.

PR #312, opened 2020-04-26
use getString

bestgopher:
`getString` instead of `string`


bestgopher:
@Fenny add a `.gitignore`, please

Fenny:
@bestgopher we have a very clean root folder, there is nothing to ignore.
If you are talking about the `.github` folder, that is ignored by default 😃

PR #316, opened 2020-04-27
update: ctx.BodyParser

bestgopher:
If the query string is `id=1&name=tom&hobby=basketball&hobby=football`, then can be parsed and get all `hobby` param。
```
type User struct {
   Id  int
   Name string
  Hobby []string
}
```


Fenny:
I tested this and it works fine 👍 I will add a test after merging your PR

Issue #311, opened 2020-04-26
💡 ETag header / ctx.Fresh


Fenny:
This is a proposal to finish the [ctx.Fresh()](https://github.com/gofiber/fiber/blob/master/ctx.go#L411) method along with [ETag header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) support.

- The `Fresh` method originates from [expressjs/req.fresh](https://expressjs.com/en/4x/api.html#req.fresh) but we never finished it [context#fresh](https://docs.gofiber.io/context#fresh
)

- _When the response is still “fresh” in the client’s cache `true` is returned, otherwise `false` is returned to indicate that the client cache is now stale and the full response should be sent._
_When a client sends the Cache-Control: no-cache request header to indicate an end-to-end reload request, this module will return `false` to make handling these requests transparent [HTTP/1.1 Caching Specification](https://tools.ietf.org/html/rfc7234)._

- Express uses the [jshttp/fresh](https://github.com/jshttp/fresh) module, [source code](https://github.com/jshttp/fresh/blob/master/index.js) _MIT_. 
I found a Go net/http version from [go-http-utils](https://github.com/go-http-utils/fresh/blob/master/fresh.go)  _MIT_

- It would be a good idea to implement [ETag](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) support within the Fiber core before starting on the `Fresh` method.
[jshttp/etag](https://github.com/jshttp/etag/blob/master/index.js)  _MIT_
[go-http-utils/etag](https://github.com/go-http-utils/etag/blob/master/etag.go)  _MIT_
[expressjs/send-etag](https://github.com/expressjs/express/blob/3ffceff3ed356a711fce5fee6ed00a03f43affd8/lib/response.js#L145)
[expressjs/weak-etag](https://github.com/expressjs/express/blob/3ffceff3ed356a711fce5fee6ed00a03f43affd8/lib/utils.js#L43)
[expressjs/strong-etag](https://github.com/expressjs/express/blob/3ffceff3ed356a711fce5fee6ed00a03f43affd8/lib/utils.js#L21)

Let's add this to the project board 💪 @gofiber/maintainers

PR #317, opened 2020-04-27
Fresh(#311)

bestgopher:
#311, copied from [jshttp/fresh](https://github.com/jshttp/fresh/blob/10e0471669dbbfbfd8de65bc6efac2ddd0bfa057/index.js#L33)

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -156,6 +156,34 @@ func (c *testConn) SetDeadline(t time.Time) error      { return nil }
 func (c *testConn) SetReadDeadline(t time.Time) error  { return nil }
 func (c *testConn) SetWriteDeadline(t time.Time) error { return nil }
 
+// Copied from:
+// https://github.com/jshttp/fresh/blob/10e0471669dbbfbfd8de65bc6efac2ddd0bfa057/index.js#L110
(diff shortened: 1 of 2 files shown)


Fenny:
Awesome work @bestgopher, thanks for helping us out!

But since it's a new feature, it is required to be added to the `ctx_test.go` for testing.
Do you have time to add a test case?

utils.go
```diff
@@ -156,6 +156,34 @@ func (c *testConn) SetDeadline(t time.Time) error      { return nil }
 func (c *testConn) SetReadDeadline(t time.Time) error  { return nil }
 func (c *testConn) SetWriteDeadline(t time.Time) error { return nil }
 
+// Copied from:
```
jozsefsallai, line 159: "copied from" sounds kinda bad. I'd replace it with something like "based on" or "inspired by" or maybe even "adapted from"

bestgopher: I am sorry that I am not good at English, so there may be some words that are not appropriate.

And these codes are not tested.

Fenny: Could you pplease change it to:
`Adapted from https://github.com/jshttp/fresh`

PR #318, opened 2020-04-28
ctx_test: impl Test_SaveFile

tianhongw:
Impl function: `Test_SaveFile`


Fenny:
Thank you for your contribution!

Issue #277, opened 2020-04-19
💡 Template / View / Render enhancement


pharrisee:
Loving fiber so far, but I have one issue with the way that templating is done.  

The current templating functionality is passing the raw contents of the template. 

I generally use https://github.com/unrolled/render for template management, and it works really well in other Go web frameworks/routers.  However it requires the template name and not the contents to be passed to it.

I wrote a middleware to the templating interface (as the others already implemented do), but obviously it's not worked for the reason specified above.

Is there any way to get the templating engine to pass the name of the template rather than it's contents?

Fenny:
Hey @pharrisee, our templating middleware was a `quick-fix` solution. You could play with the [template](https://github.com/gofiber/template) repo and submit a PR.

Or leave a code suggestion regarding our template middleware here.

Fenny:
@pharrisee We are going to change the way we render templates in Fiber. I opened an discussion that relates to multiple issues including this one, if you have any ideas or feedback please post them here https://github.com/gofiber/fiber/issues/331

PR #320, opened 2020-04-28
Feature: Adds template caching

thlacroix:
Currently templates files are read and parsed for each call to Render, and are not cached.

This adds a `func (app *App) LoadHTMLTemplates() error` method that would load all templates specified by `TemplateFolder` and `TemplateExtension` into the template engine cache.

This currently only supports the default HTML cache, as currently template engines only receive the template file data. https://github.com/gofiber/fiber/issues/277 is needed before this could support other engines, but would probably break the lib API.

The Render call will then query the cache first if it's not empty (for the default engine), execute the template if it exists, or read it from the filesystem otherwise.

For example if we have (with `TemplateFolder = "views"` and `TemplateExtension = ".tmpl"`):
```
.
└── views
    ├── header.html.tmpl
    ├── index.html.tmpl
    └── shared
        └── index.hmtl.tmpl
```

It will have the following templates in cache: `header.html`, `index.html` and `shared/index.html`, names that can be used for nested templates.

Note: no unit tests for now, waiting for feedback first. Render doesn't have tests at the moment

### DIFF
--- a/app.go
+++ b/app.go
@@ -7,13 +7,17 @@ package fiber
 import (
 	"bufio"
 	"crypto/tls"
+	"errors"
 	"fmt"
+	"html/template"
+	"io/ioutil"
 	"log"
 	"net"
 	"net/http"
 	"net/http/httputil"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"reflect"
 	"runtime"
 	"strconv"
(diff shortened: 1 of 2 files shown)


Fenny:
Thank you for your PR @thlacroix, I would lie if this new feature would be used. We are currently planning a big change for our current template system. I created a issue regarding this matter, this PR is included (caching feature). Feel free to share your ideas https://github.com/gofiber/fiber/issues/331

PR #333, opened 2020-05-01
add media blog post

WLun001:
Is it I need to add to all `README.md`s as well?


jozsefsallai:
> Is it I need to add to all `README.md`s as well?

that would be nice :)

PR #350, opened 2020-05-08
activate race condition check in tests

ReneWerner87:
add mutex for app.server -> fix for TestListen && TestServe


Fenny:
Merged your PR https://github.com/gofiber/fiber/pull/351

Issue #382, opened 2020-05-15
🤔 Any way to set header instead of appending to existing value?


boolooper:
Is there any way to completely vanish previous values of a header and set a completely new value?

```
c.Append("foo","bar")
c.Append("foo","jar")
```

will result `Foo: bar, jar`

But I want the output as `Foo: jar`.

Any way?

boolooper:
`ctx.Set` is the solution. But it should be in documentation. Closing....

Fenny:
Welcome @boolooper! Sure thing, remember we almost have a 1:1 API like [expressjs](https://expressjs.com/en/4x/api.html)

Fiber uses the `ctx.Set(key, val)` method, which is documented under the context section https://docs.gofiber.io/context#set

boolooper:
Got it. Btw off-topic,  any work on implementing `context.Context` in `ctx`?

PR #383, opened 2020-05-16
Implement context.Context in App.Ctx.

boolooper:
`context.Context` - https://godoc.org/context#Context

Error string format -[ https://github.com/golang/go/wiki/CodeReviewComments#error-strings]( https://github.com/golang/go/wiki/CodeReviewComments#error-strings)

### DIFF
--- a/app.go
+++ b/app.go
@@ -391,7 +391,7 @@ func (app *App) Listen(address interface{}, tlsconfig ...*tls.Config) error {
 	if !ok {
 		port, ok := address.(int)
 		if !ok {
-			return fmt.Errorf("Listen: Host must be an INT port or STRING address")
+			return fmt.Errorf("fiber: Host must be an INT port or STRING address")
 		}
 		addr = strconv.Itoa(port)
 	}
@@ -436,7 +436,7 @@ func (app *App) Shutdown() error {
 	app.mutex.Lock()
 	defer app.mutex.Unlock()
 	if app.server == nil {
-		return fmt.Errorf("Server is not running")
+		return fmt.Errorf("fiber: server is not running")
 	}
 	return app.server.Shutdown()
 }
@@ -472,7 +472,7 @@ func (app *App) Test(request *http.Request, msTimeout ...int) (*http.Response, e
 		select {
 		case err = <-channel:
 		case <-time.After(time.Duration(timeout) * time.Millisecond):
-			return nil, fmt.Errorf("Timeout error %vms", timeout)
+			return nil, fmt.Errorf("fiber: timeout error %vms", timeout)
 		}
 	} else {
 		// Without timeout
--- a/ctx.go
+++ b/ctx.go
@@ -237,7 +237,7 @@ func (ctx *Ctx) BodyParser(out interface{}) error {
 		return schemaDecoderQuery.Decode(out, data)
 	}
 
-	return fmt.Errorf("BodyParser: cannot parse content-type: %v", ctype)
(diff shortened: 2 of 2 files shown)


boolooper:
Someone please fix the failed github workflow file.

Fenny:
@boolooper thank you for your first PR, Fiber already supports passing context values using [ctx.Locals()](https://docs.gofiber.io/context#locals).
Please take a minute to read our documentation https://docs.gofiber.io  or visit our examples https://github.com/gofiber/recipes. In these docs you can also find the `ctx.Get()` method you asked  about in https://github.com/gofiber/fiber/issues/382#issuecomment-629552382

boolooper:
This PR does not only adds "passing context values". It's about implementing the `context.Context`. Many libraries takes `context.Context` as parameter including standard `database/sql`. 

The goal of this PR is supporting App.Ctx for passing as a parameter to functions line [sql.Conn.BeginTx](https://godoc.org/database/sql#Conn.BeginTx).

`context.Context` is being used in lots of places in go ecosystem. Even `fasthttp.RequestCtx` implements this.

Fenny:
Thanks for your proposal, I will invite the @gofiber/maintainers to join the discussion.

Fiber almost implements an one on one [expressjs](https://expressjs.com/en/4x/api.html) like API. We are open for improvements/features to be added to the [Ctx](https://godoc.org/github.com/gofiber/fiber#pkg-index). I'm not sure if it's worth to add the [context.Context](https://godoc.org/context#Context) interface directly to the Fiber context.

In the end Fiber is a simplified layer build around [Fasthttp](https://github.com/valyala/fasthttp), so if you really need to pass a context interface you could use [ctx.Fasthttp (fasthttp.RequestCtx)](https://docs.gofiber.io/context#fasthttp).

boolooper:
Just saw, `fiber.Ctx` implements the `io.Writer` interface by adding `Ctx.Write` method ignoring express API design. I don't think adding `context.Context` will be a problem.

Fenny:
`write` is a method provided by the `OutgoingMessage` class from the node `http` module. Express.js inherits this class and therefore we implement the namespace in our API.

```js
// express
app.get('/user/:id', function (req, res) {
  res.write('<html>');
  res.write('<body>');
  res.write('<h1>Hello, World!</h1>');
  res.write('</body>');
  res.write('</html>');
  res.end()
});
```
```go
// fiber
app.Get("/user/:id", func(ctx *fiber.Ctx) {
  ctx.Write("<html>");
  ctx.Write("<body>");
  ctx.Write("<h1>Hello, World!</h1>");
  ctx.Write("</body>");
  ctx.Write("</html>");
});
```

boolooper:
But I would really request you to provide a function like `Ctx.Context` that returns an implementation of `context.Context`.
Passing `ctx.Fasthttp` as a `context.Context` really looks weird in code.
An `Context()` method in `fiber.Ctx` would be really great and self documented than `ctx.Fasthttp`. 

Moving from fasthttp to fiber is making my code looks weird for this silly reason. 
Please for the sake of readable code, provide a function like this or implement `context.Context` in `fiber.Ctx`.

Best regards.

Fenny:
You got a valid point, if the @gofiber/maintainers agree we could merge it if a test case is also provided.

thomasvvugt:
@gofiber/maintainers LGTM, I would like to see this merged in `1.10` as this would break all working code having the `ctx.Context` naming.

Fenny:
Added in [v1.10.0](https://github.com/gofiber/fiber/pull/399)

Issue #397, opened 2020-05-22
🔥 Docker-Image for static websites


Lol3rrr:
**Is your feature request related to a problem?**
Not really

**Describe the solution you'd like**
A simple docker image which is just a simple webserver, that serves static files located in some directory (maybe '/website/' but that is not the point right now). 

**Describe alternatives you've considered**
I first thought about creating it myself in a seperate repository, but this would just add more things to keep up with and it would be hard to automatically create a newer version with a new release.

**Additional context**
Because fiber is pretty fast and is already capable of serving static files, this would be a rather simple addition, but could be really useful when you just need to serve some static files and don't want to use something "heavier" like nginx.
The reason I think it would fit in the main project or a subdirectory/repo, like some middlewares, is simply that it is easier to keep up to date, by automatically creating it with every new release.

Lol3rrr:
When publishing the Docker-Image should it be published to Dockerhub or to Github's docker repository?
I would personally prefer Dockerhub, as it's kind of the default place to look for docker images

thomasvvugt:
Hi @Lol3rrr please see my comment at #398 
Feel free to post a PR in [gofiber/boilerplate](https://github.com/gofiber/boilerplate)!

Fenny:
Merged in [gofiber/boilerplate](https://github.com/gofiber/boilerplate!) 👍

PR #398, opened 2020-05-23
Docker-Image for static sites

Lol3rrr:
Related to Issue #397

This will run on every release and create a new docker image that is tagged with the latest version and is then published to the dockerhub repo.
When adding this there need to be 3 new secrets:
 * "DOCKER_USERNAME": The username for dockerhub
 * "DOCKER_PASSWORD": The password or access token to be able to log into the dockerhub account
 * "DOCKER_REPO": The name for the repository on dockerhub (most likely just gofiber/fiber)

The Docker-Image exposes port 80 and will serve all files that are in the "/app/website" directory


thomasvvugt:
Hi @Lol3rrr ,

Thanks for creating this PR regarding a `Dockerfile` example.
We like the idea, however we don't think this is the right place to merge these changes.

The [gofiber/boilerplate](https://github.com/gofiber/boilerplate) repository is rather a better place to post this PR.

Lastly, Fiber is framework on which applications can be built. The boilerplate is an example of an application structure when using Fiber.
Since Fiber is not application on itself, I don't think this should be listed in the Docker Hub.

Thanks for your contribution! 🎉

PR #404, opened 2020-05-23
Add interfaces and wrapper for ctx

azhai:
You must/can implement the request and response methods in another package or repo

### DIFF
--- a/contrib/response.go
+++ b/contrib/response.go
@@ -1,0 +1,104 @@
+// ⚡️ Fiber is an Express inspired web framework written in Go with ☕️
+// 🤖 Github Repository: https://github.com/gofiber/fiber
+// 📌 API Documentation: https://docs.gofiber.io
+
+package contrib
+
+import (
+	"fmt"
+	"net/http"
+	"unsafe"
+
+	"github.com/gofiber/fiber/base"
+)
+
+// Quickly to string
+func ToString(b []byte) string {
+	return *(*string)(unsafe.Pointer(&b))
+}
+
+func GetServiceCode(statusCode int) int {
+	if statusCode == 200 {
+		return 200
+	}
+	return statusCode
+}
+
+// Give the difference of reponse body or reponse data
+type ReplyBody base.Map
+
+type ResponseMixin struct {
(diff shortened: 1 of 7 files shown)


Fenny:
@gofiber/maintainers opinions

ReneWerner87:
interesting commits, abstraction is always good, to be able to evaluate the whole thing I have to take a closer look at everything first ... need some time

contrib/response.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/gofiber/fiber/base"
+)
+
+// Quickly to string
+func ToString(b []byte) string {
+	return *(*string)(unsafe.Pointer(&b))
+}
+
+func GetServiceCode(statusCode int) int {
+	if statusCode == 200 {
```
ReneWerner87, line 21: what is this job for ?
return statusCode 
should be enought

ctx.go
```diff
@@ -830,16 +865,27 @@ func (ctx *Ctx) Stale() bool {
...
 // This method is chainable.
 func (ctx *Ctx) Status(status int) *Ctx {
-	ctx.Fasthttp.Response.SetStatusCode(status)
+	ctx.SetStatus(status)
 	return ctx
 }
 
 // Type sets the Content-Type HTTP header to the MIME type specified by the file extension.
-func (ctx *Ctx) Type(extension string) *Ctx {
+func (ctx *Ctx) SetType(extension string) {
```
ReneWerner87, line 881: methods are based on express because our framework is designed to offer an easy changeover and the familiar handling of express
https://expressjs.com/de/4x/api.html#res.type

contrib/response.go
```diff
@@ -0,0 +1,104 @@
...
+	"fmt"
+	"net/http"
+	"unsafe"
+
+	"github.com/gofiber/fiber/base"
+)
+
+// Quickly to string
+func ToString(b []byte) string {
+	return *(*string)(unsafe.Pointer(&b))
```
ReneWerner87, line 17: please use the method which is set by the app for string conversion, depending on whether the user has set immutable to true or false in the app, a different method is used internally 

but I don't think we need this methodology here

PR #417, opened 2020-05-27
Use root Group instead of the App instance

azhai:
* Use root Group instead of the App instance
* Remove some methods of App when Group extends App

### DIFF
--- a/req_test.go
+++ b/req_test.go
@@ -1,0 +1,76 @@
+// ⚡️ Fiber is an Express inspired web framework written in Go with ☕️
+// 🤖 Github Repository: https://gitee.com/azhai/fiber-u8l
+// 📌 API Documentation: https://docs.gofiber.io
+
+package fiber
+
+import (
+	"net/http/httptest"
+	"strconv"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+var expectBodies = []string{
+	`{"code":200,"data":{"param":"john"}}`,
+	`{"code":200,"data":{"page":1,"size":20}}`,
+}
+
+func CreateTestGroup() *Group {
+	grp := New().Group("/test")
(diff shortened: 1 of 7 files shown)


renanbastos93:
> * Use root Group instead of the App instance
> * Remove some methods of App when Group extends App

Sorry, but I don't understand the gain of this change.
Can you explain this to us?

azhai:
Why Group should have the same methods as App, rather than extends App?

req_test.go
```diff
@@ -0,0 +1,76 @@
...
+		if sizeStr := c.FormValue("size"); sizeStr != "" {
+			size, _ = strconv.Atoi(sizeStr)
+		}
+		data := Map{"page": page, "size": size}
+		c.JSON(Map{"code": 200, "data": data})
+	})
+	return grp
+}
+
+//func CreateSeniorGroup() *Group {
```
renanbastos93, line 41: Why do you comment it?

renanbastos93:
> Why Group should have the same methods as App, rather than extends App?

looks promising so I liked it to remove code and keeping your API, but this change requires many tests and reviews from the community in specially `maintainers`

req_test.go
azhai: The test is used for another unofficial version  [fiber-u8l](https://gitee.com/azhai/fiber-u8l)
I has add some methods like ParamStr(), FetchInt() in Ctx class.

Fenny:
Hi @azhai, thank you for your PR. It's an interesting approach, could you share a use-case for this change some can see the benefit. Thank you!

Fenny:
I'm closing this issue because we have no plans in changing the return types for `v1`, feel free to re-open to continue the discussion with a good use-case.

Thank you for your PR!

PR #424, opened 2020-05-29
Create CODEOWNERS

renanbastos93:
This approach is better what the last approach to auto-assign now we can setting to authorize merge pull request when a maintainer approves the request. We need changes in settings from the repository.

Refer: https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners


renanbastos93:
> Good idea @renanbastos93, could you move the file to the `./github` folder?

It's ok, I moved it

PR #425, opened 2020-05-29
Feature: Add Settings StartupBanner & HideStartupBanner

soulchainer:
Current `Settings` when creating a new **App** allow to enable or disable the «startup message», which is shown when the server is started.

This message, like in most frameworks, displays a «banner» of the library (in this case «FIBER» in ASCII art), and a message with info about the address the server is currently listening.

The thing is the listening info is pretty informative and usually we want to show it (at least we know that way when the server is ready), but the «banner» before, most of the times is just decorative.
Or we want to replace it with another banner or info (as example, if we're developing an app that uses «Fiber» but we want to add instead a «banner» of the project using it, its version, some initial comments, etcetera).

So I added a pair of extra settings just for that. One to allow the `StartupBanner` customization (it defaults to the current Fiber ASCII Art if is not passed) and another for being able to `HideStartupBanner` if we want (and only show the «listening» message).

I read the [CONTRIBUTING](https://github.com/gofiber/fiber/blob/master/.github/CONTRIBUTING.md) and it says that unit tests are mandatory for new features... but I reviewed the unit tests and there are no testing about **App** `Settings`...At least I didn't find where the startup message content is being tested, if the test is there, so IDK. If tests about that should be written, I will try to do it :3. This is my first PR to a `golang` repository, so be easy on me :þ.

If this PR passes the cut, I could also update the docs about it (I saw that is in another repository).

**Note**: in fact, I tought it would be preferable something like a callback in `app.Listen` (just like Express does, if I'm not wrong), so when it's successful, you control exactly what you want to display (instead of several config settings or not configurable enough settings), but I just don't know if that is feasible with golang/Fiber or how (I'm pretty new with golang), so because of I went this way. It seems like Golang doesn't have optional arguments, just and array of extra arguments of the same type and `Listen` & `Serve` already has that occupied with a `tlsconfig` so >_<.

### DIFF
--- a/app.go
+++ b/app.go
@@ -28,6 +28,8 @@ import (
 // Version of current package
 const Version = "1.10.1"
 
+var fiberASCIIBanner = fmt.Sprintf("        _______ __\n  ____ / ____(_) /_  ___  _____\n_____ / /_  / / __ \\/ _ \\/ ___/\n  __ / __/ / / /_/ /  __/ /\n    /_/   /_/_.___/\\___/_/ v%s\n", Version)
+
 // Map is a shortcut for map[string]interface{}, usefull for JSON returns
 type Map map[string]interface{}
 
@@ -91,8 +93,14 @@ type Settings struct {
 	// By default all header names are normalized: conteNT-tYPE -> Content-Type
 	DisableHeaderNormalizing bool // default: false
 
-	// When set to true, it will not print out the fiber ASCII and "listening" on message
-	DisableStartupMessage bool
+	// Startup banner
+	StartupBanner string // default: fiberASCIIBanner
+
+	// When set to true, it will hide the startup banner
+	HideStartupBanner bool // default: false
+
+	// When set to true, it will not print out the StartupBanner and "listening" on message
+	DisableStartupMessage bool // default: false
 
 	// Templates is the interface that wraps the Render function.
 	Templates Templates
@@ -166,9 +174,10 @@ func New(settings ...*Settings) *App {
 		},
 		// Set default settings
 		Settings: &Settings{
-			Prefork:     utils.GetArgument("-prefork"),
-			BodyLimit:   4 * 1024 * 1024,
-			Concurrency: 256 * 1024,
+			Prefork:       utils.GetArgument("-prefork"),
+			BodyLimit:     4 * 1024 * 1024,
+			Concurrency:   256 * 1024,
+			StartupBanner: fiberASCIIBanner,
 		},
 	}
 	// Overwrite settings if provided
@@ -187,6 +196,9 @@ func New(settings ...*Settings) *App {
 		if app.Settings.Immutable {
 			getBytes = getBytesImmutable
 			getString = getStringImmutable
+		}
+		if app.Settings.StartupBanner == "" {
+			app.Settings.StartupBanner = fiberASCIIBanner
 		}
 	}
 	// Initialize app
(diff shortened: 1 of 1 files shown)


app.go
```diff
@@ -341,9 +355,13 @@ func (app *App) Listen(address interface{}, tlsconfig ...*tls.Config) error {
 	if len(tlsconfig) > 0 {
 		ln = tls.NewListener(ln, tlsconfig[0])
 	}
-	// Print listening message
+	// Print startup message
 	if !app.Settings.DisableStartupMessage && !utils.GetArgument("-child") {
-		fmt.Printf("        _______ __\n  ____ / ____(_) /_  ___  _____\n_____ / /_  / / __ \\/ _ \\/ ___/\n  __ / __/ / / /_/ /  __/ /\n    /_/   /_/_.___/\\___/_/ v%s\n", Version)
+		if !app.Settings.HideStartupBanner {
+			if !app.Settings.HideStartupBanner {
+				fmt.Print(app.Settings.StartupBanner)
```
renanbastos93, line 362:
```suggestion
				fmt.Println(app.Settings.StartupBanner)
```

soulchainer: @renanbastos93 The thing is that previously the StartupBanner was written directly with `fmt.Printf` (you can see it in removed line 346), because the newline is already included in the banner text, so use `fmt.Println` would result in an extra line that wasn't there before.
So, if this change is applied, we should then remove the final line in the `fiberASCIIBanner`, to keep it the same way it has been rendered since always it seems.

Fenny:
Thank you @soulchainer for your detailed PR. here are my two cents on this.

I agree that the `listening` information is helpful and that it would be better to only disable the ASCII art. Since I want to keep the settings minimal and don't change to much, it might be better to just change the behaviour by only disabling the ASCII art when setting the current `DisableStartupMessage` to `true`, I would accept a PR for this.

Regarding your `Listen` callback, that would be hard to implement and I don't really like callbacks in Go. Feel free to change my mind!

soulchainer:
@Fenny Yeah, that would simplify the settings stuff (I already commented that I really don't like so much to have several settings for that). I will change the PR in that direction.

But then, a question: it's there any way of printing some additional, custom message when the server starts listening on one address? Some «listener» from where, when the server succesfully starts listening, you can trigger some code (in this case, to print some extra info). Because the change in `DisableStartupMessage` allows to hide/unhide the Fiber banner, and that's all: you can't customize the startup message.

I guess there is some way of doing that and is just me who doesn't know how to do it?

Thanks!

soulchainer:
PR updated according to @Fenny suggestions.
The PR changes `DisableStartupMessage` to only hide the «Fiber» ASCII art, while preserving the «Listening to...» message, due to its usefulness.

I also tried to add some test to cover these settings, but after checking docs, stackoverflow, several blog entries... and trying everything, I can't get a way to test what the server is writting to the standard output which actually works here.

app.go
renanbastos93: sure

renanbastos93:
> PR updated according to @Fenny suggestions.
> The PR changes `DisableStartupMessage` to only hide the «Fiber» ASCII art, while preserving the «Listening to...» message, due to its usefulness.
> 
> I also tried to add some test to cover these settings, but after checking docs, stackoverflow, several blog entries... and trying everything, I can't get a way to test what the server is writting to the standard output which actually works here.

Thanks, you made a good job.

PR #435, opened 2020-06-03
Bump Fasthttp v1.14.0

Fenny:


router_test.go
```diff
@@ -4,7 +4,7 @@
 
 package fiber
 
-// go test -v ./... -run=^$ -bench=Benchmark_Router_Handler -benchmem -count=3
+// go test -v ./... -run=^$ -bench=Benchmark_Router -benchmem -count=2
```
renanbastos93, line 7: I believe that's interesting we show results about these changes but now we are using other parameters. I referer about add this info on the description of this pull request.

ReneWerner87: is only a comment, what was used internally, so here is no problem

renanbastos93: @ReneWerner87 I know, but I believe that it is always good to have rich descriptions to better tell the community about the changes, etc.

PR #476, opened 2020-06-17
Abstract Router interface for App and Group.

kiyonlin:
Router interface defines all router handle interface includes app and group router.When dealing with sub-routers, users do not need to care about the specific implementation.

### DIFF
--- a/app.go
+++ b/app.go
@@ -27,9 +27,6 @@ import (
 	fasthttp "github.com/valyala/fasthttp"
 )
 
-// Ensure App implement Router interface
-var _ Router = (*App)(nil)
-
 // Version of current package
 const Version = "1.12.0-alpha"
 
--- a/app.go
+++ b/app.go
@@ -28,7 +28,7 @@ import (
 )
 
 // Ensure App implement Router interface
-var _ Router = &App{}
+var _ Router = (*App)(nil)
 
 // Version of current package
 const Version = "1.12.0-alpha"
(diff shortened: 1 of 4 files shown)


kiyonlin:
> Could you fix the error thrown by the tests? Thanks!

All tests passed on my machine before I created the pull request. I don’t know why it failed in the CI.

Fenny:
```go
##[error]./app.go:31:16: cannot use &(App literal) (value of type *App) as Router value in variable declaration: wrong type for method Add
##[error]./group.go:13:16: cannot use &(Group literal) (value of type *Group) as Router value in variable declaration: wrong type for method Add
```

kiyonlin:
Sorry for using wrong way to ensure App and Group implement Router interface. I’ll add a test case for this.

renanbastos93:
Why do you add an interface that`s be not using?
So we must set this contract with the struct sure.
e.g.
```go
type Router interface{
    Get()
}

type Routing struct {
    Router
}

```

kiyonlin:
@renanbastos93 This Router interface is for users who use fiber with sub-router.

e.g.
```
func setupRoutes(router fiber.Router) {
  router.Get("/path1", handler1);
  router.Post("/path2", handler2);
}
```

The `router` maybe an `*App` instance or a `*Group` instance.

Fenny:
Lgtm, if we can add a test case we can merge and tag `v1.12.0`

Issue #479, opened 2020-06-18
🚀 Session Interface


hendratommy:
**Is your feature request related to a problem?**
No

**Describe the solution you'd like**
Right now the `gofiber/session` is more like a store than a middleware.
In every handlers that need session, we need to get the `sessions`, `session.Get(c)`, then save it.
It became very repetitive tasks, became dependent to the session store implementation, and if we have handlers in separate package then we need to pass it around. While we can abstract the store and let the handlers to use the session without referencing to the variable directly.

```go
type Session interface {
    ID() string
    Get(key string) interface{}
    Set(key string, value interface{})
    Destroy()
    Regenerate()
}
```
and in Ctx have function to get and set the Session
```go
func (c *Ctx) SetSession(session Session) {}
func (c *Ctx) Session() Session
```

finally, use it like:
```go
sessions := session.New()

app.Use(func(c *fiber.Ctx) {
    store := sessions.Get(c)
    defer store.Save()
    c.SetSession(store)
    c.Next()
})

app.Get("/web", func(c *fiber.Ctx) {
    c.Session().ID()                   // returns session id
    c.Session().Destroy()              // delete storage + cookie
    c.Session().Get("john")            // get from storage
    c.Session().Regenerate()           // generate new session id
    c.Session().Delete("john")         // delete from storage
    c.Session().Set("john", "doe")     // save to storage
})
```

**Additional context**
- This will not have breaking changes
- This will also enable third party middleware that requires or optionally use session to get the session more easily (ie. `csrf` store the token to session instead of cookies).
- Switch to any session store including custom one more doable, and can be done without breaking any middlewares and handlers.
- If by any chances in the future, templates can access `Ctx` they can access the session too

enBonnet:
This feature looks great! Can improve use of sessions

Fenny:
I'm closing this proposal after discussions with the @gofiber/maintainers, at this time we are not planning to extend the session me to the Fiber core. Maybe later in the future we could re-open this issue and discuss the possibilities.

I appreciate your proposal @hendratommy, thank you 👍

PR #480, opened 2020-06-18
add session interface

hendratommy:
PR for #479 if it's approved

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -43,6 +43,7 @@ type Ctx struct {
 	values       []string             // Route parameter values
 	err          error                // Contains error if passed to Next
 	Fasthttp     *fasthttp.RequestCtx // Reference to *fasthttp.RequestCtx
+	session      Session              // Reference to Session, default nil. Use `SetSession` to provide concrete Session struct
 }
 
 // Range data for ctx.Range
@@ -77,6 +78,15 @@ type Views interface {
 	Render(io.Writer, string, interface{}, ...string) error
 }
 
+// Session interface
+type Session interface {
+	ID() string
+	Get(key string) interface{}
+	Set(key string, value interface{})
+	Destroy()
+	Regenerate() error
+}
+
 // AcquireCtx from pool
 func (app *App) AcquireCtx(fctx *fasthttp.RequestCtx) *Ctx {
 	ctx := app.pool.Get().(*Ctx)
@@ -102,6 +112,7 @@ func (app *App) ReleaseCtx(ctx *Ctx) {
 	ctx.values = nil
 	ctx.Fasthttp = nil
 	ctx.err = nil
+	ctx.session = nil
 	app.pool.Put(ctx)
 }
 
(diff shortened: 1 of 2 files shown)


ReneWerner87:
shouldn't the session instance stay the same when setting ?(@hendratommy ) think using a reference would be good

hendratommy:
@ReneWerner87 
the user session lifetime is only per request, and we will to clear it upon releasing the `ctx` since the `ctx` might be reused by different user.
we only provide the session interface not the concrete session store (or the factory) itself, and we might not want to depend on `fiber/session` directly.
the reference to the store should be managed by the middleware or user land code, for example:
```go
sessions := session.New()
// below handler might be provided by the middleware
app.Use("/web", func(c *fiber.Ctx) {
    store := sessions.Get(c)
    defer store.Save()
    c.SetSession(store)
    c.Next()
})
```
different session factory, might have have their own mechanism about how to get the user's session and how to save them.
how about it?

ReneWerner87:
i mean actually only that a copy of the instance is made if it is set with the setter without reference.

with reference its the same instance
```go
type TestStr struct {
	testValue int
}
a := TestStr{};
a.testValue = 3;
fmt.Println(a.testValue);
func(b *TestStr) {
	b.testValue = 6
	fmt.Println(b.testValue);
	return
}(&a)
fmt.Println(a.testValue);
// output:
// 3
// 6
// 6 <-
```

without reference its a different instance
```go
type TestStr struct {
	testValue int
}
a := TestStr{};
a.testValue = 3;
fmt.Println(a.testValue);
func(b TestStr) {
	b.testValue = 6
	fmt.Println(b.testValue);
	return
}(a)
fmt.Println(a.testValue);
// output:
// 3
// 6
// 3 <-
```

would be cool when you change it to 
```go
func (ctx *Ctx) SetSession(session *Session) {
// code
}

func (ctx *Ctx) Session() *Session {
// code
}
```

ReneWerner87:
image
there can already be different instances between the requests, but you should not make a copy of them when you set it in the ctx

as you can see in the image, in all properties the object instances contain is a reference to it

for the fasthttp requestctx it is always different, but it is referenced so that it is the same as the one used in the fasthttp process and our process

thomasvvugt:
I've looked into this issue and proposal as well, but it seems to me this is a bit of over-engineering.
We specifically developed a middleware (library) for sessions since not every application built with Fiber would be using this feature.
Eventhough I like the idea of wrapping the Fiber application along with the session middleware in a struct with setters and getters, I don't think this should be a feature which should be implementer in Fiber itself but rather should be handler by a developer using Fiber as a framework to build their application on.

Let me explain further using a code example:
```go
package main

import (
	"github.com/gofiber/fiber"
	"github.com/gofiber/session"
)

type App struct {
	Fiber *fiber.App
	Sessions *session.Session
	Store *session.Store
}

func main() {
	app := App{
		Fiber: fiber.New(),
		Sessions: session.New(),
	}

	app.Fiber.Use(func(c *fiber.Ctx) {
		app.Store = app.Sessions.Get(c)
		defer app.Store.Save()
		c.Next()
	})

	app.Fiber.Get("/", func(c *fiber.Ctx) {
		app.Store.ID()                   // returns session id
		app.Store.Destroy()              // delete storage + cookie
		app.Store.Get("john")            // get from storage
		app.Store.Regenerate()           // generate new session id
		app.Store.Delete("john")         // delete from storage
		app.Store.Set("john", "doe")     // save to storage
	})
}
```

hendratommy:
@thomasvvugt 
agree, i like to see it as placeholder, we have the way but not the implementor. it's more or less the same with render, we provide the render function but the template itself is not implemented in the main package.

Fenny:
https://github.com/gofiber/fiber/issues/479#issuecomment-652799369

PR #482, opened 2020-06-18
ctx: do not use normalized path

humaidq:


jozsefsallai:
Hi!
As @ReneWerner87 mentioned, could you also add a test case that ensures `/////admin` doesn't match `/admin` and sends 404? Thank you so much 🙏

PR #499, opened 2020-06-21
improve timeout middleware

ReneWerner87:


middleware/timeout.md
```diff
@@ -1,9 +1,7 @@
-# Timeout wrapper
-
-Why use this middleware?
-
-...
+# Timeout
+Wrapper function which provides a handler with a timeout.
 
+If the handler takes longer than the given duration, the timeout error is set and forwarded to the next method.
```
Fenny, line 4: If the handler takes longer than the given duration to return a response, the timeout error is set and forwarded to the error handler.

PR #535, opened 2020-07-03
Localize reuseport logic

Fenny:


reuseport_windows.go
```diff
@@ -0,0 +1,22 @@
+package fiber
```
savsgio, line 1: Please, add the license disclaimer, because you are copied from fasthttp [reuseport/windows](https://github.com/valyala/fasthttp/blob/master/reuseport/reuseport_windows.go) code.

Thanks!

reuseport.go
```diff
@@ -0,0 +1,45 @@
+// +build !windows
```
savsgio, line 1: Please, add the license disclaimer, because you are copied from fasthttp [reuseport](https://github.com/valyala/fasthttp/blob/master/reuseport/reuseport.go) code.

Thanks!

PR #542, opened 2020-07-04
🩹 Improve flag handling

### LINKED ISSUE #522: 🐛 Fiber's flags are part of the printed usage
**Fiber version**

`v0.12.1`

**Issue description**

I use `flag.Parse` in my app to parse things like bind address, port and more app-specific settings. The app is offered for self-hosting to others, so it's not an app I just run myself on a server. With the switch to Fiber two flags show up when doing `app -h` that I never defined myself : `-prefork` and `-child`. _This would confuse the users of my app_.

I found a workaround for this, but it's kind of ugly and I was wondering if there's an "official" recommended approach to deal with this.

The code that probably causes this is here: https://github.com/gofiber/fiber/blob/ec514589405a7ab637f97e8ce410a76a9433f5f4/prefork.go#L28-L29

There was a GitHub issue where `flag.Parse` vs. scanning `os.Args` was discussed, which concluded with keeping to use `os.Args` scanning, but somehow that got changed later anyway: https://github.com/gofiber/fiber/issues/301

**Code snippet**

Reproducible with this:

```go
package main

import (
	"flag"

	"github.com/gofiber/fiber"
)

var (
	port = flag.Int("port", 8080, "Port")
)

func main() {
	flag.Parse()

	app := fiber.New()
		c.Send("Hello, World!")
	})
	app.Listen(*port)
}
```

Then do `go run . -h` or `go build -o fiber-flags && ./fiber-flags -h`.

This is printed:

> Usage of /path/to/fiber-flags:
>   -port int
>         Port (default 8080)
>   -prefork
>         is child proc
>   -prefork-child
>         use prefork

But of course I only want my app users to see the `-port` flag.

**Workaround**

I'd like to share my workaround.

- The `flag.FlagSet` struct has a `Usage` field, which is just a `func()` that's called to print the usage
- When an app defines flag like `var port = flag.Int(...)` and calls `flag.Parse`, the global, exported variable `flag.CommandLine` is used, which is such a `flag.FlagSet`
- => So we overwrite `flag.CommandLine.Usage` with our own `func()`!

We can copy everything in `flag`'s `func (f *FlagSet) PrintDefaults()`, but add these lines:

```go
		// Skip printing usage info for Fiber's flags
		if f.Name == "prefork" || f.Name == "child" {
			return
		}
```

Some other code has to be copied from the stdlib as well to make the function compile, like `stringValue`, but it's not too much and in the end it works fine.

Link to example commit that contains all required changes for this workaround: https://github.com/Deflix-tv/go-stremio/commit/a57d3d418f5adf01a8854121245066692e97ab59

Now of course this is far from optimal as for example if we do this in a *library* that uses Fiber and our library user defines his own flags, he could define `-prefork` and wouldn't see the usage being printed!

That's why this is just a workaround.

doingodswork:
1. Don't define the prefork flag
2. Don't expose the prefork-child flag in the flag usage message

Closes #522

### DIFF
--- a/app.go
+++ b/app.go
@@ -256,10 +256,6 @@ func New(settings ...*Settings) *App {
 	if app.Settings.ErrorHandler == nil {
 		app.Settings.ErrorHandler = defaultErrorHandler
 	}
-
-	if !app.Settings.Prefork { // Default to -prefork flag if false
-		app.Settings.Prefork = utils.GetArgument(flagPrefork)
-	}
 	// Replace unsafe conversion functions
 	if app.Settings.Immutable {
 		getBytes, getString = getBytesImmutable, getStringImmutable
--- a/prefork.go
+++ b/prefork.go
@@ -7,6 +7,7 @@ import (
 	"net"
 	"os"
 	"os/exec"
+	"reflect"
 	"runtime"
 	"strconv"
 	"strings"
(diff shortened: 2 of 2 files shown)


doingodswork:
> Please, re-check CI output and fix errors.

The issue is that `strings.ReplaceAll` was only added in Go 1.12, but the test is run with Go 1.11.

It looks like `strings.ReplaceAll` is only a convenience method, so I could just replace it by what's called inside it.

But a quick question before I do that: Do you really need to support Go 1.11? Fiber is a fairly new framework so I guess people starting to build on top of Fiber use a recent version of Go. Go 1.11 was released almost 2 years ago. In fact, it doesn't receive any updates by the Go maintainers anymore, see the official release policy [here](https://golang.org/doc/devel/release.html#policy) and [here](https://github.com/golang/go/wiki/Go-Release-Cycle#release-maintenance):

> Each major Go release is supported until there are two newer major releases.

So as there are Go 1.14 and Go 1.13, Go 1.12 and earlier are not supported anymore.

PR #545, opened 2020-07-04
🔥 Feature Add ParamList Method to context

zerbitx:
[Here](https://github.com/zerbitx/gnockgnock/blob/master/gnocker/gnocker.go#L199) is my rather specific use case.  This test helper app will be dynamically configured to mock 3rd party apis, so it would be convenient to see what params are available at run time.

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -650,6 +650,15 @@ func (ctx *Ctx) Params(key string, defaultValue ...string) string {
 	return defaultString("", defaultValue)
 }
 
+// ParamList returns a list of all the parameter names for the current context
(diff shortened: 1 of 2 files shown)


Fenny:
Hi @zerbitx, I recently had the same issue where I needed the full Param list for a middleware I was creating.

Since Fiber's Context API is almost identical to Expressjs API, I don't want to bloat the Ctx with more methods. So I have another proposition:
<https://github.com/gofiber/fiber/blob/master/router.go#L48> let's make routeParams public in `*Route` under the name `Params` so it could be accessed like `c.Route().Params`.

Let me know what ya think

Issue #469, opened 2020-06-15
🐛 flag.Parse not working when using Prefork on Fiber


itsursujit:
**Fiber version**
1.11.x
**Issue description**
flag.Parse not working when using Prefork on Fiber
**Code snippet**

This piece of code is throwing error:
```go
port := flag.Int("port", 8080, "Port number to serve app")
    flag.Parse()

    App = fiber.New(&fiber.Settings{
        ErrorHandler:              nil,
        Prefork:                   true,
        DisableStartupMessage:     true,
        Templates:                 handlebars.New("./resources/views", "html"),
    })

    err := App.Listen(*port)
    if err != nil {
        panic(err)
    }
```
Error:
```
main: error: unknown short flag '-c', try --help
```

Fenny:
@itsursujit this should be fixed, you can try the alpha version here
<https://github.com/gofiber/fiber/releases/tag/v1.12.0-alpha>

PR #580, opened 2020-07-11
use prefork child env

kiyonlin:
Currently, `Fiber` uses `utils.GetArgument(flagChild)` to determine whether the current process is a child process from prefork. According to #522 and #469, we know that users have some side effects when using `flag.Parse()`.

My current approach is to use environment variables to determine whether the current process is a child process. When spawning a child process, add `FIBER_PREFORK_CHILD=1` environment variable for inter-process communication. This has several advantages:
- No longer affects anything related to the flag, so there is no need for the hijacking code added in #542 (some are copied from the standard library)
- `FIBER_PREFORK_CHILD` this environment variable is basically impossible to conflict with the user’s environment variables

As for the global variable names that control environment variables, I use `envPreforkChildKey` and `envPreforkChildVal`, if you think there are better variable names, I will replace them.

Finally, I was thinking about whether I need to add an environment variable, which has the same effect as `settings.Prefork`, for .e.g
```go
	if app.Settings.Prefork || os.Getenv("FIBER_PREFORK") == "1"{
		return app.prefork(addr, tlsconfig...)
	}
```

If this is the case, I also need to pass the `FIBER_PREFORK` environment variable to the child process.

### DIFF
--- a/prefork.go
+++ b/prefork.go
@@ -2,36 +2,24 @@ package fiber
 
 import (
 	"crypto/tls"
-	"flag"
 	"fmt"
 	"net"
 	"os"
 	"os/exec"
-	"reflect"
 	"runtime"
 	"strconv"
 	"strings"
 	"time"
-
-	utils "github.com/gofiber/utils"
 )
 
 var (
-	flagChild = "-prefork-child"
-	isChild   bool
+	envPreforkChildKey = "FIBER_PREFORK_CHILD"
+	envPreforkChildVal = "1"
 )
-
-func init() { //nolint:gochecknoinits
-	// Prevent users from defining the same flag on their own.
-	flag.BoolVar(&isChild, flagChild[1:], false, "Child Process")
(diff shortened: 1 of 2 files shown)


prefork.go
```diff
@@ -2,36 +2,24 @@ package fiber
...
-	"reflect"
 	"runtime"
 	"strconv"
 	"strings"
 	"time"
-
-	utils "github.com/gofiber/utils"
 )
 
 var (
```
Fenny, line 15: Please change this into `const`

prefork.go
```diff
@@ -2,36 +2,24 @@ package fiber
...
-	flagChild = "-prefork-child"
-	isChild   bool
+	envPreforkChildKey = "FIBER_PREFORK_CHILD"
```
Fenny, line 16: Please change these global variables into `const` 👍

doingodswork:
Nice solution to get rid of the `prefork-child` flag!

Issue #565, opened 2020-07-08
🚀 Logger coloration


YuilTripathee:
**Is your feature request related to a problem?**

My feature request is related to logger. Could there be solution for coloration?

**Describe the solution you'd like**

Same solution is being provided in Gin. Using colored logger, debugging looks easier and fancier to be precise 😋. This feature is not desired with the cost of performance.

**Describe alternatives you've considered**

Apart from coloration, saving the logger to a log file could be fruitful than this features. Alternative to this by saving logs using external gateway.

**Additional context**

![image](https://user-images.githubusercontent.com/28005756/86934179-1cf19300-c15b-11ea-9eb7-08b387a10d63.png)

But, if both coloration and saving to file is attained without losing performance at micro-level,  the logger will be useful than ever.

Fenny:
Hey @YuilTripathee, awesome idea. We have to think about how to customize colors by implementing a new template syntax.

PS: Our current Logger accepts an `io.Writer` as input, writing to a file should be no problem.
```go
package main

import (
	"bufio"
	"log"
	"os"
	"time"

	"github.com/gofiber/fiber"
	"github.com/gofiber/fiber/middleware"
)

func main() {
	// Create file
	f, _ := os.Create("./logs.txt")
	defer f.Close()

	// Create io.Writer
	w := bufio.NewWriter(f)

	// Flush to file every second
	go func() {
		for {
			w.Flush()
			time.Sleep(1 * time.Second)
		}
	}()

	app := fiber.New()

	// Pass io.Writer to logger
	app.Use(middleware.Logger(w))

	app.Get("/", func(c *fiber.Ctx) {
		c.Send("Hello, World!")
	})

	log.Fatal(app.Listen(3000))
}

```

YuilTripathee:
Working on template now!!!

YuilTripathee:
Will there arise an issue while saving log with coloration?
For this we got two options:
1. Use coloration while printing on screen only.
2. Use some sort of slicing during Flush.

Fenny:
I think the best way would be to have a single `Color bool` option.

If this is enabled, we first check if a custom `io.Writer` is given. If not we use our own [colorable writer](https://godoc.org/github.com/mattn/go-colorable#NewColorableStdout) and provide colored variables for like `method`, `statuscode` and `path`.

YuilTripathee:
Going through the multiple requests, Flush() is giving inconsistent result and not appending data properly.

[...]

PR #586, opened 2020-07-11
Fix: #565 🖌 Add coloration in logger middleware

YuilTripathee:


YuilTripathee:
Next PR will get coloration in logger template format level. I think it'll be much better implementation over existing one.
Like `${red}...`.
Coming soon.

PR #590, opened 2020-07-12
replace `register` method to `Add` in App.Use 

songjiayang:
changing for the same behavior with `app.Get`.


kiyonlin:
Thanks for the PR.

But I think in the package, use `register` is better. The `Add` is for users outside the package.

Fenny:
Thanks for the PR @songjiayang, @kiyonlin is right regarding the use of the `app.Add` method. Its purpose is to register HTTP methods, where `USE` is a custom verb in the Fiber framework.

We appreciate your contribution by providing optimizations/alternatives ❤️

Issue #589, opened 2020-07-12
🚀  `methodInt` refactor to contain `USE` method


songjiayang:
**Is your feature request related to a problem?**

currently , the `methodInt` method in `utils.go`  looks like: 

```go
func methodInt(s string) int {
	switch s {
	case MethodGet:
		return 0
        .......
	default:
		return 0
	}
}
```

and we use it check a method is valid method or `USE` in `app.go`, it's too cumbersome:

```go
func (app *App) register(method, pathRaw string, handlers ...Handler) *Route {
         .....
	if method != "USE" && methodInt(method) == 0 && method != MethodGet {
		log.Fatalf("Add: Invalid HTTP method %s", method)
	}
       ....
}
```

**Describe the solution you'd like**

- change `methodInt` default value to -1.
- add MethodUSE to `methodInt` case.

code looks like:

```go
func methodInt(s string) int {
	switch s {
	case MethodUse:
		return 9
        .......
	default:
		return -1
	}
}
```

then the above check code can change to :

```
func (app *App) register(method, pathRaw string, handlers ...Handler) *Route {
         .....
	if methodInt(method) == -1 {
		log.Fatalf("Add: Invalid HTTP method %s", method)
	}
       ....
}
```

Fenny:
PR https://github.com/gofiber/fiber/pull/594 got merged, thanks guys 👍

thomasvvugt:
Nice work, let us know whether the issue is resolved! 👍🏼

songjiayang:
fixed so quickly，confirm  it resolved

PR #594, opened 2020-07-12
👷 Add use methodInt and Extract "USE" to a global const variable

kiyonlin:
Ref #589

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -241,8 +241,10 @@ func methodInt(s string) int {
 		return 7
 	case MethodPatch:
 		return 8
+	case methodUse:
+		return 9
 	default:
-		return 0
+		return -1
 	}
 }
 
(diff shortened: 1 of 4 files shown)


utils.go
```diff
@@ -270,6 +272,7 @@ const (
 	MethodConnect = "CONNECT" // RFC 7231, 4.3.6
 	MethodOptions = "OPTIONS" // RFC 7231, 4.3.7
 	MethodTrace   = "TRACE"   // RFC 7231, 4.3.8
+	methodUse     = "USE"
```
renanbastos93, line 275: Maybe we should change it.
```suggestion
	MethodUse     = "USE"
```

kiyonlin: Now it’s just used inside the package.

renanbastos93: it's okay, hehehe my mistake

Fenny:
Awesome work, thank you ❤️

PR #596, opened 2020-07-13
🚨 extract func quoteString() for ctx.Attachment and ctx.Download

kiyonlin:
co-authored-by @ReneWerner87 

And `func quoteString` now is put in `Fiber core`'s util file, maybe it should be moved to `gofiber/utils` package someday @Fenny .

Add a test assert for `ctx.Download` incidentally.

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -7,6 +7,8 @@ package fiber
 import (
 	"bytes"
 	"fmt"
+	"github.com/valyala/bytebufferpool"
+	"github.com/valyala/fasthttp"
 	"hash/crc32"
 	"net"
 	"strings"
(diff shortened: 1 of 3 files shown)


utils.go
```diff
@@ -7,6 +7,8 @@ package fiber
 import (
 	"bytes"
 	"fmt"
+	"github.com/valyala/bytebufferpool"
```
songjiayang, line 10: The third part libraries should be imported after standard library, how about changing to:

```
"hash/crc32"
"net"		"net"
"strings"		"strings"

"github.com/valyala/bytebufferpool"
"github.com/valyala/fasthttp"
```

kiyonlin: Fixed. Thanks to point this out.

Fenny:
We could add this to utils one day, thnx for the PR!

PR #601, opened 2020-07-14
👶 Remove child procs on PID kill + IPv6

Fenny:


app.go
```diff
@@ -484,10 +490,13 @@ func (app *App) Listen(address interface{}, tlsconfig ...*tls.Config) error {
 	app.init()
 	// Start prefork
 	if app.Settings.Prefork {
+		if app.Settings.Network == "ipv6" || isIPv6(addr) {
```
songjiayang, line 493: The defaultNetwork is `tcp`,  maybe `tcp6` is better than `ipv6`.

kiyonlin: `tcp6` makes sense

Fenny: Made a typo, should be `tcp6` 👍

PR #602, opened 2020-07-14
🚀 refactor: improve logic method BodyParser

renanbastos93:
* Changed many `if` to a `switch case`
* Removed unused variable
* Defined alias tag when needed

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -222,47 +222,45 @@ func (ctx *Ctx) Body() string {
 // application/json, application/xml, application/x-www-form-urlencoded, multipart/form-data
 func (ctx *Ctx) BodyParser(out interface{}) error {
 	// TODO: Create benchmark ( Probably need a sync pool )
-	var schemaDecoderForm = schema.NewDecoder()
-	var schemaDecoderQuery = schema.NewDecoder()
-	schemaDecoderForm.SetAliasTag("form")
-	schemaDecoderForm.IgnoreUnknownKeys(true)
-	schemaDecoderQuery.SetAliasTag("query")
-	schemaDecoderQuery.IgnoreUnknownKeys(true)
+	var (
+		schemaDecoder = schema.NewDecoder()
+		ctype         = getString(ctx.Fasthttp.Request.Header.ContentType())
+	)
+	schemaDecoder.IgnoreUnknownKeys(true)
 
-	// get content type
-	ctype := getString(ctx.Fasthttp.Request.Header.ContentType())
-	// application/json
-	if strings.HasPrefix(ctype, MIMEApplicationJSON) {
+	switch ctype {
+	case MIMEApplicationJSON, MIMEApplicationJSONCharsetUTF8:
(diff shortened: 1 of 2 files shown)


jozsefsallai:
This looks fine **but** I'll have to check the behavior of this when there are other tokens appended to the content type (iirc, multipart/form-data does that a lot when appending the boundary).

ctx.go
```diff
@@ -222,47 +222,44 @@ func (ctx *Ctx) Body() string {
...
-	// application/json
-	if strings.HasPrefix(ctype, MIMEApplicationJSON) {
+	var (
+		schemaDecoder = schema.NewDecoder()
+		ctype         = getString(ctx.Fasthttp.Request.Header.ContentType())
+	)
+	schemaDecoder.IgnoreUnknownKeys(true)
+
+	switch ctype {
+	case MIMEApplicationJSON: // application/json
```
Fenny, line 232: Please add the `utf-8` variant
```go
case MIMEApplicationJSON, MIMEApplicationJSONUTF8:
```

ctx.go
```diff
@@ -222,47 +222,44 @@ func (ctx *Ctx) Body() string {
...
 		return json.Unmarshal(ctx.Fasthttp.Request.Body(), out)
-	}
-	// application/xml text/xml
-	if strings.HasPrefix(ctype, MIMEApplicationXML) || strings.HasPrefix(ctype, MIMETextXML) {
+	case MIMEApplicationXML: // application/xml
```
Fenny, line 234: Please add the `utf-8` variant
```go
case MIMEApplicationXML, MIMEApplicationXMLUTF8:
```

ctx.go
```diff
@@ -222,47 +222,44 @@ func (ctx *Ctx) Body() string {
...
 		return xml.Unmarshal(ctx.Fasthttp.Request.Body(), out)
-	}
-	// application/x-www-form-urlencoded
-	if strings.HasPrefix(ctype, MIMEApplicationForm) {
+	case MIMETextXML: // text/xml
```
Fenny, line 236: Please add the `utf-8` variant
```go
case MIMETextXML, MIMETextXMLUTF8:
```

Fenny:
> multipart/form-data

This is correct, the `multipart/form-data` header could be unique on every request i.e. 
`Content-Type: multipart/form-data; boundary=something`

We should move this one outside the switch statement and perform a `Contains` check.

renanbastos93:
Alright I will improve it, thanks for feedbacks

renanbastos93:
> This looks fine **but** I'll have to check the behavior of this when there are other tokens appended to the content type (iirc, multipart/form-data does that a lot when appending the boundary).
> 
> The original code checks if the supplied content type starts with the XML, JSON, etc. content types, not that they're equal.

Nice catch, thanks I improved. Please review guys

Fenny:
Awesome work, thnx!

PR #604, opened 2020-07-14
Improve coverage

### LINKED ISSUE #598: 🚀 Improve unit test coverage > 80%
**Question description**

run `go test -count=1 -cover` got fibber code coverage is `77.8%`, how about to improve it to `>80%` or higher.

kiyonlin:
```
go test -count=1 -cover
PASS
coverage: 87.0% of statements
```

close #598

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -388,7 +388,7 @@ func (ctx *Ctx) Format(body interface{}) {
 			ctx.Send(body) // Fallback
 			log.Println("Format: error serializing json ", err)
 		}
-	case "text":
+	case "txt":
 		ctx.SendString(b)
 	case "xml":
 		raw, err := xml.Marshal(body)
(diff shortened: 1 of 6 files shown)


koddr:
LGTM, great job, @kiyonlin 👍

ctx.go
```diff
@@ -753,9 +763,11 @@ func (ctx *Ctx) Range(size int) (rangeData Range, err error) {
 		})
 	}
 	if len(rangeData.Ranges) < 1 {
-		return rangeData, fmt.Errorf("range: unsatisfiable range")
+		err = ErrRangeUnsatisfiable
```
songjiayang, line 766: Why not return `rangeData, ErrRangeUnsatisfiable` here

kiyonlin: > Why not `return rangeData, ErrRangeUnsatisfiable` here 

@songjiayang I think this will reassign rangeData again and I like unifying the code style :smile:

songjiayang: If you return the error directly at above code it looks unifying yet 😄 , for example  `return nil,  ErrRangeMalformed`

`No Need Change`  it's personal behavior

kiyonlin: When using `named return vars`, I prefer the `empty return` way.

Fenny:
Awesome work 👌

PR #608, opened 2020-07-15
Improve coverage

kiyonlin:
```bash
go test -count=1 -cover
PASS
coverage: 93.9% of statements
```

I did some trick on `prefork.go` test
```go
cmd := exec.Command(os.Args[0], os.Args[1:]...)
if testPreforkMaster {
	// When test prefork master,
	// just start the child process
	// a cmd on all os is best
	cmd = exec.Command("date")
}
```

I'm wondering if this will be fine.

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http/httptest"
+	"reflect"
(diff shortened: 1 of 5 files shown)


Fenny:
Awesome work like always 👏

PR #615, opened 2020-07-15
🚽 remove invalid benchmark

kiyonlin:


renanbastos93:
Why are invalid?

kiyonlin:
> Why are invalid?

I should test `ctx.Send`and `ctx.SendBytes` directly

PR #638, opened 2020-07-18
✏️  Fix typo for app and ctx warnings

DarthBenro008:
Replacing use instead of us


kiyonlin:
Nice found :smile:

DarthBenro008:
Had used the wrong emoji before , just replaced the correct one :smile:

PR #651, opened 2020-07-20
🚀 compress same route's handler and sort routes when registering

kiyonlin:
- compress route handers when it is identical with pre-one
- use routes slice to store all registered routes
- Get route shares handlers with Head route which is spawned by itself

### DIFF
--- a/router_test.go
+++ b/router_test.go
@@ -364,29 +364,6 @@ func Benchmark_Router_WithCompression(b *testing.B) {
 	}
 }
 
-// go test -v ./... -run=^$ -bench=Benchmark_Router_Compress -benchmem -count=4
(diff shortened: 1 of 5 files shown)


router_test.go
```diff
@@ -329,7 +329,12 @@ func Benchmark_Router_Chain(b *testing.B) {
 	handler := func(c *Ctx) {
 		c.Next()
 	}
-	app.Get("/", handler, handler, handler, handler, handler, handler)
+	app.Get("/", handler)
```
ReneWerner87, line 332: please don´t change the benchmark, in this bench the handlers should be in one line

to measure the compression of the route we should create an independent benchmark, if that was the idea behind the change

kiyonlin: Ok I’ll add another one

PR #652, opened 2020-07-21
🔥 Add logger middleware time zone support

kallydev:
The default value of TimeZone is "Local", using the system time zone.
You can also specify "UTC" and "America / New_York" and "Asia / Chongqing", etc.

**Example:**

```go
app := fiber.New()
app.Use(middleware.Logger(middleware.LoggerConfig{
	TimeFormat: "2006-01-02 15:04:05",
	TimeZone:   "UTC",
}))
app.Get("/", func(ctx *fiber.Ctx) {
	ctx.Write("hello world")
})
if err := app.Listen(3000); err != nil {
	log.Fatalln(err)
}
```

**Output:**

```text
#29499 - 2020-07-21 05:15:29 200 - 0s         GET /
```

### DIFF
--- a/middleware/logger.go
+++ b/middleware/logger.go
@@ -193,6 +193,9 @@ func logger(config LoggerConfig) fiber.Handler {
 	if config.Format == "" {
 		config.Format = LoggerConfigDefault.Format
 	}
+	if config.TimeZone == "" {
+		config.TimeZone = LoggerConfigDefault.TimeZone
+	}
 	if config.TimeFormat == "" {
 		config.TimeFormat = LoggerConfigDefault.TimeFormat
 	}
@@ -212,17 +215,13 @@ func logger(config LoggerConfig) fiber.Handler {
 
 	var tmpl loggerTemplate
 	tmpl.new(config.Format, "${", "}")
-	var location, err = time.LoadLocation(config.TimeZone)
-	if err != nil {
-		log.Fatalf("Logger: failed to load time zone: %e\n", err)
-	}
-	timestamp := time.Now().In(location).Format(config.TimeFormat)
+	timestamp := nowTimeString(config.TimeZone, config.TimeFormat)
 	// Update date/time every second in a separate go routine
 	if strings.Contains(config.Format, "${time}") {
 		go func() {
 			for {
 				mutex.Lock()
-				timestamp = time.Now().In(location).Format(config.TimeFormat)
+				timestamp = nowTimeString(config.TimeZone, config.TimeFormat)
 				mutex.Unlock()
 				time.Sleep(500 * time.Millisecond)
 			}
@@ -364,6 +363,18 @@ func logger(config LoggerConfig) fiber.Handler {
 		}
 		bytebufferpool.Put(buf)
 	}
+}
+
+func nowTimeString(timeZone, layout string) string {
+	// This is different from Golang's time package which returns UTC, and Local is better than it
+	if timeZone == "" {
+		timeZone = "UTC"
+	}
+	location, err := time.LoadLocation(timeZone)
+	if err != nil {
+		log.Fatalf("Logger: failed to load time zone: %e\n", err)
+	}
+	return time.Now().In(location).Format(layout)
 }
 
 // Use Golang's time package to determine whether the TimeZone is available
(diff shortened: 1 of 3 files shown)


kiyonlin:
Please consider that `Logger(options...interface())` supports `TimeZone` and add some necessary test cases.

kallydev:
> Please consider that `Logger(options...interface())` supports `TimeZone` and add some necessary test cases.

OK, I have found a bug and I will fix it. Thanks for your suggestion.

https://github.com/gofiber/fiber/blob/a097820313910409e38aa012c6b118c7183e6657/middleware/logger.go#L163-L168

kiyonlin:
LGTM! Thanks for your contribution!

middleware/logger.go
```diff
@@ -203,14 +212,17 @@ func logger(config LoggerConfig) fiber.Handler {
 
 	var tmpl loggerTemplate
 	tmpl.new(config.Format, "${", "}")
-
-	timestamp := time.Now().Format(config.TimeFormat)
+	var location, err = time.LoadLocation(config.TimeZone)
+	if err != nil {
```
thomasvvugt, line 216: I have my doubts about this integration.
It would be nice to have `Local` as default timezone and maybe as fallback.
@gofiber/maintainers, let me know what you think.
Thanks for your contribution! 🎉

kiyonlin: Agreed. `Local` is better.

kallydev: Yes, `Local` is already defined in `LoggerConfigDefault`.

https://github.com/gofiber/fiber/blob/1fa775fa29aeab066823a8914ba83c84747a0e6a/middleware/logger.go#L137-L143

kiyonlin: What if `LoggerConfig.TimeZone` is empty string?

You should set a default value after:
https://github.com/gofiber/fiber/blob/1fa775fa29aeab066823a8914ba83c84747a0e6a/middleware/logger.go#L198

kallydev: > What if `LoggerConfig.TimeZone` is empty string?
> 
> You should set a default value after:
> https://github.com/gofiber/fiber/blob/1fa775fa29aeab066823a8914ba83c84747a0e6a/middleware/logger.go#L198-199

1. If the name is "", `LoadLocation` returns “UTC”.
2. This is an oversight, I will solve it, sorry.

kallydev: Now there is a problem. The empty string in the Golang time package should be "UTC". But our default value is empty will conflict with it, do we want to change it to "Local"'?

kiyonlin:
```go
if config.TimeZone == "" {
	config.TimeZone = LoggerConfigDefault.TimeZone
}
```

middleware/logger.go
```diff
@@ -354,6 +365,24 @@ func logger(config LoggerConfig) fiber.Handler {
...
 }
 
+func nowTimeString(timeZone, layout string) string {
+	// This is different from Golang's time package which returns UTC, and Local is better than it
+	if timeZone == "" {
+		timeZone = "UTC"
+	}
+	location, err := time.LoadLocation(timeZone)
+	if err != nil {
+		log.Fatalf("Logger: failed to load time zone: %e\n", err)
```
Fenny, line 375: I think we should print the error and default to "UTC" instead of crashing the application.

kallydev: Because `middleware.LoggerConfig` does not use pointer, I did not find a way to change the `config` value from within the function. If I force the use of pointer, it may break the design of fibir.Could you give me some suggestions, please?

thomasvvugt: That was indeed also my opinion. Crashing the application is not needed here imho

kiyonlin: > Because `middleware.LoggerConfig` does not use pointer, I did not find a way to change the `config` value from within the function. If I force the use of pointer, it may break the design of fibir.Could you give me some suggestions, please?

I think you can check TimeZone first. If it is invalid, set the value to `Local`

Fenny: Use the following logic after line 374
```go
log.Printf("logger: failed to load time zone: %e\n", err)
return time.Now().Format(layout)
```

Fenny:
I will merge this PR, and make a small change 👍 thanks for your contribution!

Fenny:
https://github.com/gofiber/fiber/pull/654

PR #669, opened 2020-07-24
Use regexp to find next segment end

larrylv:

This PR changes the implementation of `findNextRouteSegmentEnd` from iterating the delimiters and find their first index, to using regexp to find the first match of either `/`, `-`, or `.`. Existing tests should cover testing this change.

I also plan to add `(` and `)` support in the route path in the next couple of days.

**Commit formatting** 

### DIFF
--- a/path.go
+++ b/path.go
@@ -7,6 +7,7 @@
 package fiber
 
 import (
+	"regexp"
 	"strings"
 	"sync/atomic"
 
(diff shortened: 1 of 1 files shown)


kiyonlin:
Thanks for your contribution! Could you please provide benchmark? We are sensitive to performance.

ReneWerner87:
thanks for your effort

we have tried in version 1.9.6 to avoid regex usage for parsing and registering the route, because regex is much slower when analyzing the strings

please create a benchmark, if it is faster than the pre-behavior we will merge it but if it is slower we will not adopt it

https://github.com/gofiber/fiber/releases/tag/v1.9.6
> Improved router by ditching regexp and increased performance by 817% without allocations.

larrylv:
@kiyonlin @ReneWerner87 thanks for the quick feedback. yeah, this is definitely slower than the `for` + `strings.IndexByte` implementation. I'll keep running benchmark in mind before opening next PR.

ReneWerner87:
thanks anyway for trying to bring the fiber framework forward

PR #673, opened 2020-07-25
📈 Add regular expression support in route path

larrylv:

Express supports regular expression with `?`, `+`, `*`, and `()` in their route path. So this PR adds the support for those too.
Unfortunately I have to use `regexp` since part of the route path will be a regular expression. If you have better suggestion than this, happy to make the change.

Note that even with the `regexp` introduced, it won't be used for all previous route path cases. This PR only uses it when the segment has a regular expression char (`?`, `+`, `*`, and `()`) in it, so the old code won't get the performance hit.

Goal of this PR is to provide the same route functionality as Express does https://expressjs.com/en/guide/routing.html
Tests in this PR are also found in their document page.

**Commit formatting** 

### DIFF
--- a/path.go
+++ b/path.go
@@ -146,7 +146,7 @@ func isSegRegexp(seg string) bool {
 func buildSegRegexp(seg string) (*regexp.Regexp, error) {
 	// `\A` is added because we need the match from the beginning
 	//  `*` is replaced with `.*` because we need to turn the wildcard to a regular expression.
-	return regexp.Compile("\\A" + strings.ReplaceAll(seg, "*", ".*"))
+	return regexp.Compile("\\A" + strings.Replace(seg, "*", ".*", -1))
 }
 
 // getMatch parses the passed url and tries to match it against the route segments and determine the parameter positions
--- a/path.go
+++ b/path.go
@@ -7,6 +7,8 @@
 package fiber
 
 import (
+	"fmt"
+	"regexp"
 	"strings"
 	"sync/atomic"
 
@@ -21,18 +23,20 @@ type routeParser struct {
 
 // paramsSeg holds the segment metadata
 type paramSeg struct {
-	Param      string
-	Const      string
-	IsParam    bool
-	IsOptional bool
-	IsLast     bool
-	EndChar    byte
+	Param          string
+	Const          string
+	IsParam        bool
+	IsOptional     bool
+	IsLast         bool
+	IsRegexp       bool           // for `Const` segment, it might be a regular expression
+	CompiledRegexp *regexp.Regexp // compiled regular expression if `IsRegexp = true`
+	EndChar        byte
 }
 
 // list of possible parameter and segment delimiter
 // slash has a special role, unlike the other parameters it must not be interpreted as a parameter
-// TODO '(' ')' delimiters for regex patterns
 var routeDelimiter = []byte{'/', '-', '.'}
+var regexpCharacters = []byte{'?', '+', '*', '(', ')'}
 
 const wildcardParam string = "*"
 
@@ -79,6 +83,23 @@ func parseRoute(pattern string) (p routeParser) {
 			lastSeg = len(out) - 1
 		}
 
+		// only check if the segment is regexp for non param segments
+		if !out[lastSeg].IsParam && !out[lastSeg].IsRegexp {
+			if isSegRegexp(out[lastSeg].Const) {
+				re, err := buildSegRegexp(out[lastSeg].Const)
+				if err != nil {
+					// If there is an error when compiling the regexp, we log an
+					// error and treat it like a normal Const segment instead of
+					// panicing.
+					fmt.Printf("invalid regexp route path: %v, %v\n", out[lastSeg].Const, err)
+				} else {
+					out[lastSeg].IsRegexp = true
+					out[lastSeg].CompiledRegexp = re
+				}
+
+			}
+		}
+
 		if delimiterPos != -1 && len(pattern) >= delimiterPos+1 {
 			out[lastSeg].EndChar = pattern[delimiterPos]
 			pattern = pattern[delimiterPos+1:]
@@ -105,6 +126,27 @@ func findNextRouteSegmentEnd(search string) int {
 	}
 
 	return nextPosition
+}
+
+// isSegRegexp checks if the segment contains regexp chars in `regexpCharacters`.
+// Note that `regexpCharacters` also includes `wildcardParam`, but only `Constant`
+// segment (isParam != true) uses this method. So as long as it has `*`, it will
+// always be part of the regular expression but not the wildcard.
+func isSegRegexp(seg string) bool {
+	for _, regexpChar := range regexpCharacters {
+		if pos := strings.IndexByte(seg, regexpChar); pos != -1 {
+			return true
+		}
+	}
+
+	return false
+}
+
+/// buildSegRegexp builds the Regexp based on the segment string.
+func buildSegRegexp(seg string) (*regexp.Regexp, error) {
+	// `\A` is added because we need the match from the beginning
+	//  `*` is replaced with `.*` because we need to turn the wildcard to a regular expression.
+	return regexp.Compile("\\A" + strings.ReplaceAll(seg, "*", ".*"))
 }
 
 // getMatch parses the passed url and tries to match it against the route segments and determine the parameter positions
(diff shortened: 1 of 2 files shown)


ReneWerner87:
Okay thanks the pull request, I will check the performance and if it does not affect the other cases, agree.  The logic of the given characters should even be quite easy to convert to string analysis without regex, maybe we will do that in the next step.

Hope the consumers of fiber don't get confused about the wildcard characters, because once it returns parameters and once it is in the part of a segment it doesn't.

larrylv:
> The logic of the given characters should even be quite easy to convert to string analysis without regex
can you say more? It might require us building a subset of the regular expression engine, but I can't think of a way without regexp (not the package, but just regular expression) in general.

Here is the benchmark on master:
```
$ go test -v ./... -run=^$ -bench=Benchmark_Router_Github_API -benchmem -count=4
goos: darwin
goarch: amd64
pkg: github.com/gofiber/fiber
Benchmark_Router_Github_API
Benchmark_Router_Github_API-12              1755            647651 ns/op             137 B/op          1 allocs/op
Benchmark_Router_Github_API-12              1789            643144 ns/op             134 B/op          1 allocs/op
Benchmark_Router_Github_API-12              1755            644639 ns/op             137 B/op          1 allocs/op
Benchmark_Router_Github_API-12              1796            642922 ns/op             134 B/op          1 allocs/op
PASS
ok      github.com/gofiber/fiber        4.985s
PASS
ok      github.com/gofiber/fiber/middleware     0.113s
```

and here is with this PR
```
➜  ~/code/gopath/src/github.com/larrylv/fiber ruby:(2.7.0) go:(1.14.5) git:(larrylv/question_mark_in_route_path) ✔
$ go test -v ./... -run=^$ -bench=Benchmark_Router_Github_API -benchmem -count=4
goos: darwin
goarch: amd64
pkg: github.com/gofiber/fiber
Benchmark_Router_Github_API
Benchmark_Router_Github_API-12              1664            707483 ns/op             181 B/op          2 allocs/op
Benchmark_Router_Github_API-12              1614            686752 ns/op             187 B/op          2 allocs/op
Benchmark_Router_Github_API-12              1718            696610 ns/op             176 B/op          1 allocs/op
Benchmark_Router_Github_API-12              1729            680646 ns/op             175 B/op          1 allocs/op
PASS
ok      github.com/gofiber/fiber        5.131s
PASS
ok      github.com/gofiber/fiber/middleware     0.117s
```

The performance is degraded. I think it's probably because we will check if a Constant segment contains those regular expression characters https://github.com/gofiber/fiber/pull/673/commits/5d8cd24f0a4b46b25560e6d59f027070e4fe7f23#diff-167fa3abc5af405dd53a00a09a35b800R88

So I updated this PR a little bit in the last commit by only checking if the incremental Const part is regexp, and here is the _latest_ benchmark:
```
$ go test -v ./... -run=^$ -bench=Benchmark_Router_Github_API -benchmem -count=4
goos: darwin
goarch: amd64
pkg: github.com/gofiber/fiber
Benchmark_Router_Github_API
Benchmark_Router_Github_API-12              1666            668311 ns/op             181 B/op          2 allocs/op
Benchmark_Router_Github_API-12              1731            660890 ns/op             174 B/op          1 allocs/op
Benchmark_Router_Github_API-12              1808            695656 ns/op             167 B/op          1 allocs/op
Benchmark_Router_Github_API-12              1737            661348 ns/op             174 B/op          1 allocs/op
PASS
ok      github.com/gofiber/fiber        5.086s
PASS
ok      github.com/gofiber/fiber/middleware     0.205s
```

it is still degraded,  but that as much as before. Let me know if this is acceptable. I'll also think more about this to optimize this.

ReneWerner87:
Okay, it looks pretty much the same as before, but today I don't have time 

We will see tomorrow if we can do this without affecting the normal routes 

Regarding the regex characters, this should also be possible with string analysis , or almost 90%....  The logics of the individual characters are well known, 

On Wildcard you just look until you find the substring after the characters, with plus the same, only that here at least one character must be present and question mark says only that the substring is optional, we have already implemented these logics in the parameters (wildcard, question mark)

larrylv:
I would agree with you if we only allow those characters appearing once in one segment. Otherwise with the combination of various of them, it would be impossible to deal with only with string analysis. You would end up with a simplified regular expression engine.
The implementation of current wildcard and question mark is simple, yes, but it could only handle simpler cases:
1. if the whole segment is `*`
2. if the last char is `?`.
So it supports those two chars only in those scenarios.

Fenny:
Hey @larrylv, thanks for the idea and the PR 👍 Regex has been a big topic the last months, @ReneWerner87 discussed to add support for it in `v1.12.0`. But we moved the idea because our current router is heavily optimized and we want to make sure it does not conflict with the current implementation.

That being said, I will make some time this week to dive deep into this PR/idea again regarding performance and compatibility. We also have to think about specific regex patterns that return param values.

To be continued....

larrylv:
Thank you both for the quick review and feedback.
> We also have to think about specific regex patterns that return param values.

yeah, this PR only supports regexp with const segment, but doesn't support regexp with param. If that's a think gofiber also wants to support. I could give it a try after this PR.

kiyonlin:
Hey @larrylv ! Thank you for your outstanding contribution.  For regular expression matching of routing paths, I have some ideas of my own.  

First of all, by using regular expressions, routing matching will become very flexible, but it will definitely bring performance problems.  Further more, attackers can construct very performance-consuming strings to attack(ddos) servers.  

Secondly, it is precisely because of flexibility that it will be difficult for us to extend route matching in the future—-no backward compatible.  

So, like @ReneWerner87 and @Fenny mentioned above, now we start with simple string parsing and slowly add more matching symbols and matching algorithms.  

In short, it is going to implement a simple regular expression matcher by ourselves.  If you are interested, you can participate in this work and improve it together!

larrylv:
With the last two commits, I used the standard `strings` API `IndexAny` to find the index of the first appeared delimiter in `findNextRouteSegmentEnd`, and `ContainsAny` to check if it has any regexp character. Performance seems very close to master.
This PR:
```
$ go test -v ./... -run=^$ -bench=Benchmark_Router_Github_API -benchmem -count=4
goos: darwin
goarch: amd64
pkg: github.com/gofiber/fiber
Benchmark_Router_Github_API
Benchmark_Router_Github_API-12              1665            669391 ns/op             181 B/op          2 allocs/op
Benchmark_Router_Github_API-12              1677            676372 ns/op             180 B/op          2 allocs/op
Benchmark_Router_Github_API-12              1718            668879 ns/op             176 B/op          1 allocs/op
Benchmark_Router_Github_API-12              1766            663753 ns/op             171 B/op          1 allocs/op
PASS
ok      github.com/gofiber/fiber        4.995s
PASS
ok      github.com/gofiber/fiber/middleware     0.119s
```

master:
```
$ go test -v ./... -run=^$ -bench=Benchmark_Router_Github_API -benchmem -count=4
goos: darwin
goarch: amd64
pkg: github.com/gofiber/fiber
Benchmark_Router_Github_API
Benchmark_Router_Github_API-12              1617            669397 ns/op             149 B/op          2 allocs/op
Benchmark_Router_Github_API-12              1712            668729 ns/op             140 B/op          1 allocs/op
Benchmark_Router_Github_API-12              1662            665205 ns/op             145 B/op          2 allocs/op
Benchmark_Router_Github_API-12              1686            661555 ns/op             143 B/op          1 allocs/op
PASS
ok      github.com/gofiber/fiber        4.988s
PASS
ok      github.com/gofiber/fiber/middleware     0.163s
```

kiyonlin:
Found an interesting [article](https://checkmarx.gitbooks.io/go-scp/content/general-coding-practices/regular-expressions.html) about ReDos, and maybe we shouldn't worry about ReDos in golang.

thomasvvugt:
The benchmarks look promising, also I don't think we really have to worry for ReDos when using the default `regexp` package.
The default `regexp` does lack fundamental features [as compared to dlclark/regexp2](https://github.com/dlclark/regexp2#compare-regexp-and-regexp2) but therefore does have a linear-time performance as proven by the benchmarks.

larrylv:
gentle ping @ReneWerner87 @Fenny , anything I can do to make things move forward? Thanks!

Fenny:
Hey @larrylv, sorry this PR is staled for a while. The reason for this is because we are also working on a radix tree based router behind the scenes. If you don't mind I want to keep this PR open until we have a clear vision where we want to go regarding our routing system.

It would be a bummer if I merge it now and it would get overwritten before it gets tagged, I will keep you updated on the progress. Feel free to join us on [Discord](https://gofiber.io/discord) to join the discussion.

ReneWerner87:
support for the plus sign will come in the next fiber version, without regular expressions (means faster) and the optional groups come later

PR #675, opened 2020-07-26
🚀 Don't use regexp for no-cache value check

larrylv:

While we are on the `regexp` topic, I found another place that's using it for `no-cache` value check. The regular expression is pretty straightforward, so I implemented a function to do the same thing -- if the string has `no-cache`, and its previous non-space char is either nothing or comma, and same with the char after.

I also added a benchmark to make sure this improves the performance:
AFTER:
```
$ go test -v -run=^$ -bench=Benchmark_Ctx_Fresh_WithNoCache -benchmem -count=4
goos: darwin
goarch: amd64
pkg: github.com/gofiber/fiber
Benchmark_Ctx_Fresh_WithNoCache
Benchmark_Ctx_Fresh_WithNoCache-12      10466191               117 ns/op               0 B/op          0 allocs/op
Benchmark_Ctx_Fresh_WithNoCache-12      10317930               118 ns/op               0 B/op          0 allocs/op
Benchmark_Ctx_Fresh_WithNoCache-12       9983041               115 ns/op               0 B/op          0 allocs/op
Benchmark_Ctx_Fresh_WithNoCache-12      10603621               113 ns/op               0 B/op          0 allocs/op
PASS
ok      github.com/gofiber/fiber        5.397s
```

BEFORE:
```
➜  ~/code/gopath/src/github.com/larrylv/fiber ruby:(2.7.0) go:(1.14.5) git:(master) ✗
$ go test -v -run=^$ -bench=Benchmark_Ctx_Fresh_WithNoCache -benchmem -count=4
goos: darwin
goarch: amd64
pkg: github.com/gofiber/fiber
Benchmark_Ctx_Fresh_WithNoCache
Benchmark_Ctx_Fresh_WithNoCache-12       4557762               260 ns/op               0 B/op          0 allocs/op
Benchmark_Ctx_Fresh_WithNoCache-12       4668902               260 ns/op               0 B/op          0 allocs/op
Benchmark_Ctx_Fresh_WithNoCache-12       4625556               264 ns/op               0 B/op          0 allocs/op
Benchmark_Ctx_Fresh_WithNoCache-12       4619721               259 ns/op               0 B/op          0 allocs/op
PASS
ok      github.com/gofiber/fiber        6.133s
```

So it cuts the the per op time by more than 50%.

**Commit formatting** 

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -17,7 +17,6 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
-	"regexp"
 	"strconv"
 	"strings"
 	"sync"
@@ -417,8 +416,6 @@ func (ctx *Ctx) FormValue(key string) (value string) {
 	return getString(ctx.Fasthttp.FormValue(key))
 }
 
-var cacheControlNoCacheRegexp, _ = regexp.Compile(`(?:^|,)\s*?no-cache\s*?(?:,|$)`)
-
 // Fresh returns true when the response is still “fresh” in the client's cache,
 // otherwise false is returned to indicate that the client cache is now stale
 // and the full response should be sent.
@@ -439,7 +436,7 @@ func (ctx *Ctx) Fresh() bool {
 	// to support end-to-end reload requests
 	// https://tools.ietf.org/html/rfc2616#section-14.9.4
 	cacheControl := ctx.Get(HeaderCacheControl)
-	if cacheControl != "" && cacheControlNoCacheRegexp.MatchString(cacheControl) {
+	if cacheControl != "" && isNoCache(cacheControl) {
 		return false
 	}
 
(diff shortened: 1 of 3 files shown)


ctx.go
```diff
@@ -479,6 +476,38 @@ func (ctx *Ctx) Fresh() bool {
 	return true
 }
 
+// isNoCache checks if the cacheControl header value is a `no-cache`.
+func isNoCache(cacheControl string) bool {
```
larrylv, line 480: to be clear, this function is implemented to replicate the previous regexp `(?:^|,)\s*?no-cache\s*?(?:,|$)`. If we think having random chars before `,no-cache,` is invalid too, happy to change it too.

kiyonlin:
Nice work @larrylv . It is worth mentioning that we just consider standard http header values until now.

ctx.go
Fenny: LGTM

PR #678, opened 2020-07-27
🧹 Refactor etag stale checking

larrylv:
Move logics of checking if the etag is stale to `utils` and replace the existing `parseTokenList` function. And also, we don't need to allocate a string slice to store all etags. Instead, we could just check in place to see if it matches the etag.

**Commit formatting** 

### DIFF
--- a/ctx_test.go
+++ b/ctx_test.go
@@ -632,6 +632,23 @@ func Test_Ctx_FormValue(t *testing.T) {
 	utils.AssertEqual(t, StatusOK, resp.StatusCode, "Status code")
 }
 
+// go test -v -run=^$ -bench=Benchmark_Ctx_Fresh_StaleEtag -benchmem -count=4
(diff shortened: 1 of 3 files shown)


ctx_test.go
```diff
@@ -632,6 +632,23 @@ func Test_Ctx_FormValue(t *testing.T) {
 	utils.AssertEqual(t, StatusOK, resp.StatusCode, "Status code")
 }
 
+// go test -v -run=^$ -bench=Benchmark_Ctx_Fresh_StaleEtag -benchmem -count=4
+func Benchmark_Ctx_Fresh_StaleEtag(b *testing.B) {
```
larrylv, line 636: Benchmark shows no improvements or degradation. It's probably because the allocated string slice doesn't matter too much to the performance over all. But I think it's still worthing removing the unnecessary slice allocation and then iterate it.

PR #682, opened 2020-07-29
Improve test

kiyonlin:
Add more coverage test cases

`coverage: 97.5% of statements`

**Commit formatting** 


utils.go
```diff
@@ -248,7 +264,6 @@ func (a testAddr) String() string {
 }
 
 type testConn struct {
-	net.Conn
```
ReneWerner87, line 251: not 100% sure about this change, but since we're running tests, it shouldn't be a problem

ReneWerner87: okay had a talk with kiyonlin, testConn already implement all the needed stuff, so net.Conn should be not necassary

thomasvvugt:
Awesome work, LGTM! 🎉

Issue #691, opened 2020-07-30
🚀 Extending Ctx using an interface


thomasvvugt:
**Is your feature request related to a problem?**
Fiber uses a `struct` type for it's Context whereas other frameworks like Echo use an `interface` type. Using the `interface` type allows extending the current Ctx struct.

**Describe the solution you'd like**
A guide on how to do this in Echo can be found here: https://echo.labstack.com/guide/context

**Describe alternatives you've considered**
As can be seen from the following example, interfaces are the best way to extend both the fields and the methods of an object in GoLang: https://gist.github.com/josephspurrier/55d76c2a10026e085226#file-structs_interface-go-L160

**Additional context**
The main difference between Fiber's Ctx and Echo's Context can be compared here:
https://github.com/gofiber/fiber/blob/master/ctx.go#L33
https://github.com/labstack/echo/blob/master/context.go#L19

Fenny:
Will be shipped in `v1.14.0` 👍

gulien:
Perfect, thank you! 🚀

PR #694, opened 2020-07-30
🚀 Adding interface for Ctx

thomasvvugt:
Instead of using a `struct` type, I suggest we implement an `interface` type for the context. By using an interface, we provide more expandability to our end-users for them to modify or embed our Ctx with custom logic.

This PR is related to `🚀 Extending Ctx using an interface`, see #691

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -29,8 +29,8 @@ import (
 )
 
 type (
-	// ICtx represents the interface for the Context
-	ICtx interface {
+	// Ctx represents the interface for the Context
+	Ctx interface {
 		// Accepts checks if the specified extensions or content types are acceptable.
 		Accepts(offers ...string) string
 
@@ -95,6 +95,9 @@ type (
 
 		// Error contains the error information passed via the Next(err) method.
 		Error() error
+
+		// Fasthttp return the Fasthttp Request Context
+		Fasthttp() *fasthttp.RequestCtx
 
 		// Format performs content-negotiation on the Accept HTTP header.
 		// It uses Accepts to select a proper format.
@@ -248,11 +251,10 @@ type (
 		Stale() bool
 
 		// Status sets the HTTP status for the response.
-		// This method is chainable.
-		Status(status int) *Ctx
+		Status(status int)
 
 		// Type sets the Content-Type HTTP header to the MIME type specified by the file extension.
-		Type(extension string, charset ...string) *Ctx
+		Type(extension string, charset ...string)
 
 		// Vary adds the given header field to the Vary response header.
 		// This will append the header, if not already listed, otherwise leaves it listed in the current location.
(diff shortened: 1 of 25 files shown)


thomasvvugt:
Once merged, we might want to move to 1.14 as this PR has some impacting changes, such as:
- Removing the chainability of `ctx.Status()` and `ctx.Type()`
- Changing the Handler type from `func(*Ctx)` to `func(Ctx)`, which would require a lot of re-writing for our users

ctx.go
```diff
@@ -28,22 +28,263 @@ import (
...
+		Write(bodies ...interface{})
+
+		// XHR returns a Boolean property, that is true, if the request's X-Requested-With header field is XMLHttpRequest,
+		// indicating that the request was issued by a client library (such as jQuery).
+		XHR() bool
+	}
+
+	// Context represents the Context which hold the HTTP request and response.
+	// It has methods for the request query string, parameters, body, HTTP headers and so on.
+	Context struct {
```
kiyonlin, line 273: What about making Context struct private?

ctx.go
```diff
@@ -960,62 +1211,59 @@ func (ctx *Ctx) Subdomains(offset ...int) []string {
...
-func (ctx *Ctx) Status(status int) *Ctx {
-	ctx.Fasthttp.Response.SetStatusCode(status)
-	return ctx
+func (ctx *Context) Status(status int) {
+	ctx.Fasthttp().Response.SetStatusCode(status)
 }
 
 // Type sets the Content-Type HTTP header to the MIME type specified by the file extension.
-func (ctx *Ctx) Type(extension string, charset ...string) *Ctx {
+func (ctx *Context) Type(extension string, charset ...string) {
```
ReneWerner87, line 1224: why did you remove this return value here ? is this not implementable in the interface ?

ctx.go
```diff
@@ -960,62 +1211,59 @@ func (ctx *Ctx) Subdomains(offset ...int) []string {
...
+func (ctx *Context) Stale() bool {
 	return !ctx.Fresh()
 }
 
 // Status sets the HTTP status for the response.
-// This method is chainable.
-func (ctx *Ctx) Status(status int) *Ctx {
-	ctx.Fasthttp.Response.SetStatusCode(status)
-	return ctx
+func (ctx *Context) Status(status int) {
```
ReneWerner87, line 1219: why did you remove this return value here ? is this not implementable in the interface ?

ctx.go
thomasvvugt: I made it private, however AcquireCtx() now return the unexported type of this struct.

ctx.go
thomasvvugt: I re-added the chainability of these methods by using the interface

thomasvvugt:
Interface types can store the copy of a value, or a value can be shared with the interface by storing a pointer to the value’s address. An example of using a working struct pointer to extend upon the `Ctx` struct in Fiber's current version would be:

```go
type Custom struct {
    *fiber.Ctx
}

func (c *Custom) SendString(body string) {
    c.Send("Overriden, sorry 👋\n\n")
}

func main() {
    app := fiber.New()

        cc := &Custom{c}
        cc.SendString("Hello, World!")
    })

}
```

Some arguments to use/not use the interface approach in comparison to the pointer struct approach;
- The interface implementation decides the type of parameter (a pointer or a value) to pass to a method/function that accepts the interface as a parameter.
- Since both type of receivers are allowed, there could be varying implementations of the same interface, which could lead to possible runtime errors during type conversion.

Let me know what you guys think about both approaches.

PR #742, opened 2020-08-27
Add fiber-cache middleware to README(s)

codemicro:

Added my fiber-cache middleware to the 3rd party middlewares section of the readmes ([codemicro/fiber-cache](https://github.com/codemicro/fiber-cache)). I also sorted the list alphabetically.

~~**Commit formatting** 
Use emojis on commit messages so it provides an easy way of identifying the purpose or intention of a commit. Check out the emoji cheatsheet here: https://gitmoji.carloscuesta.me/~~ I sort of attempted to do this and then failed massively.

### DIFF
--- a/.github/README.md
+++ b/.github/README.md
@@ -498,16 +498,17 @@ The Fiber middleware modules listed here are maintained by the [Fiber team](http
 ## 🌱 Third Party Middlewares
 
 This is a list of middlewares that are created by the Fiber community, please create a PR if you want to see yours!
-- [arsmn/fiber-swagger](https://github.com/arsmn/fiber-swagger)
 - [arsmn/fiber-casbin](https://github.com/arsmn/fiber-casbin)
 - [arsmn/fiber-introspect](https://github.com/arsmn/fiber-introspect)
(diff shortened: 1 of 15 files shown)


kiyonlin:
Hey, @codemicro  thanks for the awesome work!

I'v looked at your code, and have some questions:
- There should be data competition in the global `statusCodes` map
- And how about the response headers? I can see you only cache the response body :smile:

codemicro:
Morning! Funnily enough, I thought of the data race in the `statusCodes` map while I was drifting off to sleep last night. I've sorted that in v1.0.1.

In regards to the headers, I don't see an especially easy way of handling them. Some headers (like `Date`) have to be removed and others need to stay. Determining that is a bit of a pain, hence I decided to forgo headers. If someone really needs specific headers to be added to that endpoint, they'd have to add another middleware before the cache middleware to add it. I've added a note to the readme about this.

Fenny:
Awesome work, I had the same issue regarding the `Date` header. I even got to a point where I stored the raw HTTP request ( headers + body ) in memory and override the date header. Many possibilities, but I think this is a great start!

PR #751, opened 2020-09-07
🐛 Fix filesystem tests, make compatible with v15

hi019:
* Updated filesystem middleware to be compatible with v15 by using `c.Request()` instead of `c.FastHTTP()`
* Fixed filesystem middleware tests. It seems the previous syntax was using `utils.AssertEqual(test, actual, expected)` when it should be `utils.AssertEqual(test, expected, actual)`


Fenny:
Awesome work, didn't had time to adjust yet ^^

PR #793, opened 2020-09-16
Improve test

kiyonlin:

**Commit formatting** 

Co-authored-by: Rene rene@gofiber.io

### DIFF
--- a/middleware/cors/cors.go
+++ b/middleware/cors/cors.go
@@ -84,6 +84,9 @@ func New(config ...Config) fiber.Handler {
 		if cfg.AllowMethods == "" {
 			cfg.AllowMethods = ConfigDefault.AllowMethods
 		}
+		if cfg.AllowOrigins == "" {
+			cfg.AllowOrigins = ConfigDefault.AllowOrigins
+		}
 	}
 
 	// Convert string to slice
--- a/middleware/cors/cors.go
+++ b/middleware/cors/cors.go
@@ -81,12 +81,6 @@ func New(config ...Config) fiber.Handler {
 		cfg = config[0]
 
 		// Set default values
-		if cfg.Next == nil {
-			cfg.Next = ConfigDefault.Next
-		}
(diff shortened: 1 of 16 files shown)


middleware/cors/cors.go
```diff
@@ -81,12 +81,6 @@ func New(config ...Config) fiber.Handler {
 		cfg = config[0]
 
 		// Set default values
-		if cfg.Next == nil {
-			cfg.Next = ConfigDefault.Next
-		}
-		if cfg.AllowOrigins == "" {
```
ReneWerner87, line 87: why you remove that ?
default allowOrigins is "*" and should be used, when nothing is added in the custom user config

middleware/cors/cors_test.go
```diff
@@ -41,7 +42,13 @@ func Test_CORS_Wildcard(t *testing.T) {
 	handler := app.Handler()
 
 	// OPTIONS (preflight) response headers when AllowOrigins is *
-	app.Use(New(Config{AllowOrigins: "*", AllowCredentials: true, MaxAge: 3600}))
+	app.Use(New(Config{
```
ReneWerner87, line 45: can you add an test with the default configuration ? and check then the expected headers/behavior

middleware/cors/cors.go
kiyonlin: Yeah you are right. I made a mistake :smile:

middleware/cors/cors_test.go
```diff
@@ -41,7 +42,13 @@ func Test_CORS_Wildcard(t *testing.T) {
...
+		AllowOrigins:     "*",
```
ReneWerner87, line 46: can you please add another test, where you leave an empty configuration

ReneWerner87 (approved):

middleware/cors/cors_test.go
ReneWerner87: thx for the additional test case

PR #823, opened 2020-09-23
update readme prometheusmiddleware

ansrivas:
#### Purpose:

Added prometheus middleware implementation to 3rd party middlewares list.
Will help users integrate a metrics end point easily in their applications.


Fenny:
Thanks for the mw 🆗, could you also change the other 14 README files?

PR #855, opened 2020-09-29
🩹 Fix: revise coloration in logger middleware

YuilTripathee:
🎨Coloration for METHOD is added.

This is as far I can get!

Coloration present to v1 is being restored.


YuilTripathee:
Issue is being fixed with #856

PR #863, opened 2020-09-30
Fix minor issue on cache middleware and add tests

klipitkas:
This PR addresses the following:
- Fixes an issue with a magic number on cache middleware.
- Adds a basic test for cache middleware.
- Fixes an issue with cache expiration.
- Adds more tests thus increasing coverage for the cache middleware.

### DIFF
--- a/middleware/cache/cache.go
+++ b/middleware/cache/cache.go
@@ -16,7 +16,7 @@ type Config struct {
 	// Optional. Default: nil
 	Next func(c *fiber.Ctx) bool
 
-	// Expiration is the time that an cached response will live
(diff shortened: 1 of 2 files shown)


klipitkas:
Actually I am going to increase coverage on this, so please wait a bit if you were planning on merging.

klipitkas:
@Fenny I have an issue that I cannot troubleshoot, it appears that one of my tests causes a race condition. This is the test that checks the cache expiration. If you can have a look let me know.

Fenny:
Thanks for the unit tests, changed some small things in https://github.com/gofiber/fiber/pull/864 💯

PR #881, opened 2020-10-04
Enable timeout tests

lcd1232:

Enable timeout tests.

**Commit formatting** 

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -276,18 +276,26 @@ func Test_App_Use_Params(t *testing.T) {
 	app := New()
 
 	app.Use("/prefix/:param", func(c *Ctx) error {
-		utils.AssertEqual(t, "john", c.Params("param"))
+		param := c.Params("param")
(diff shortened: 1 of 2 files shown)


lcd1232:
@koddr it looks like Params has data race. Is it ok if I update it by creating copies c.route.Params each call?

Fenny:
The data race seems is caused by the timeout middleware, we use the same logic as fasthttp https://github.com/valyala/fasthttp/blob/master/server.go#L418.
@kiyonlin do you have any idea's left?

lcd1232:
I added time.Sleep for Params test and it has data race too

Fenny:
I'm afraid this is currently not possible to fix with the current design. I added a warning message regarding this issue https://github.com/gofiber/fiber/pull/885/commits/251a2e46e24c6c2148a0770d52e3e385a517bfb5

I'm closing this for now until we make some internal changes, thanks for trying to add unit tests 👍

PR #889, opened 2020-10-07
✨c.GetAll

napei:

This PR introduces `c.GetAll()` to accompany `c.Get()` by returning a `map[string]string` of all the request headers.  

Several people have asked me how to get all the headers of a request at once. Fiber doesn't have a method for this out of the box, instead relying on fasthttp.

This may or may not be relevant to everyone, but it covers an edge-case more than core functionality and promotes use of fiber instead of just going to fasthttp for that need.

In some cases, this may be more performant than running `c.Get()` for multiple headers. However, `GetAll` would not have default values like `Get` does.

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -455,6 +455,14 @@ func (c *Ctx) Get(key string, defaultValue ...string) string {
 	return defaultString(getString(c.fasthttp.Request.Header.Peek(key)), defaultValue)
 }
 
+// GetAll returns a map of all HTTP request headers
+func (c *Ctx) GetAll() (out map[string]string) {
+	c.fasthttp.Request.Header.VisitAll(func(key, value []byte) {
+		out[getString(key)] = getString(value)
+	})
+	return out
+}
+
 // Hostname contains the hostname derived from the Host HTTP header.
 // Returned value is only valid within the handler. Do not store any references.
 // Make copies or use the Immutable setting instead.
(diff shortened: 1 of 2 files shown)


renanbastos93:
@napei Please, adjust your commits on pull request using emoji.
Well, about this proposal I don't see great advantages use it then I believe that you need to set more examples to this case from ensuring performance.
About code, I liked your commitment. Good job, thanks a lot for the contribution.

napei:
Hey @renanbastos93, I had some discussion with @Fenny regarding this in discord.

My viewpoint is that through last week, 6 people messaged me wanting help with fiber and all were non-express users and non-go users that wanted helpers like this as implementing `VisitAll` was confusing. After some discussion, it appears that this kind of addition probably doesn't follow the core principles of Fiber as `GetAll` isn't present in the express API and that this probably shouldn't go in.

Reasoning was just simplification and keeping ease of development in mind. Understandable maps are expensive also which is why this probably isn't a good route to go down with performance in mind. 

Happy to abandon this as it's just an idea. :)

Fenny:
After some discussions, we decided not to add `GetAll` to the Fiber API because it's not an efficient method.
An alternative would be using `VisitAll` that does no copies:
```go
c.Request().Header.VisitAll(func(k, v []byte) {
      fmt.Printf("%s: %s", k, v)
})
```

PR #894, opened 2020-10-07
use string library to split the string

neil85:

Use string library to split credential to []string. Add checking ensure the result must have only 2 data which is username & password. 

**Commit formatting** 

### DIFF
--- a/middleware/basicauth/basicauth.go
+++ b/middleware/basicauth/basicauth.go
@@ -118,20 +118,19 @@ func New(config Config) fiber.Handler {
 
 				// Convert to string
 				cred := utils.UnsafeString(raw)
+				credArray := strings.Split(cred, ":")
 
-				// Find semicolumn
-				for i := 0; i < len(cred); i++ {
-					if cred[i] == ':' {
-						// Split into user & pass
-						user := cred[:i]
-						pass := cred[i+1:]
+				// Check array size must be 2
+				if len(credArray) == 2 {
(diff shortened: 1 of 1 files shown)


Fenny:
@neil85 thanks for your PR. `strings.Split` will cause an extra allocation which results in slower results.
I added a quick benchmark test https://github.com/gofiber/fiber/pull/895 ( was not included yet ) to make sure this is correct.

```go
// before
Benchmark_Middleware_BasicAuth
Benchmark_Middleware_BasicAuth-16        3809520               311 ns/op              48 B/op          3 allocs/op
Benchmark_Middleware_BasicAuth-16        3870937               311 ns/op              48 B/op          3 allocs/op
Benchmark_Middleware_BasicAuth-16        3870939               316 ns/op              48 B/op          3 allocs/op
Benchmark_Middleware_BasicAuth-16        3857341               310 ns/op              48 B/op          3 allocs/op

// after
Benchmark_Middleware_BasicAuth
Benchmark_Middleware_BasicAuth-16        3100776               385 ns/op              80 B/op          4 allocs/op
Benchmark_Middleware_BasicAuth-16        3108799               385 ns/op              80 B/op          4 allocs/op
Benchmark_Middleware_BasicAuth-16        3116878               386 ns/op              80 B/op          4 allocs/op
Benchmark_Middleware_BasicAuth-16        3116877               384 ns/op              80 B/op          4 allocs/op
```

PR #896, opened 2020-10-07
🔥set a `Cache Control` header for static ressources

r-52:

Right now, each static request returns the file without any `Cache-Control` header. If you are using `app.Static` to serve your assets, the browser won't cache it. This PR introduces a way to set the `cache control` header (if the file was found) on the file response to set a custom `max-age` to e.g. cache it 1 year.

A general use case for 

```go
app.Static("/public", "./public")
```

could be to serve your CSS and JS assets. Right now, there is no `Cache-Control` header present on the file-response. With this change, the `Cache-Control`-header can be configured through something like this

```go
app.Static("/public", "./ public", Static{MaxAge: 100})
```

Which improves the performance for your webpage and reduces the bandwidth.

**Commit formatting** 

✅ done

### DIFF
--- a/app.go
+++ b/app.go
@@ -261,6 +261,12 @@ type Static struct {
 	// The name of the index file for serving a directory.
 	// Optional. Default value "index.html".
 	Index string `json:"index"`
+
+	// The value for the Cache-Control HTTP-header
+	// that is set on the file response. MaxAge is defined in seconds.
+	//
+	// Optional. Default value 0.
+	MaxAge int `json:"max_age"`
 }
 
 // Default Config values
--- a/router.go
+++ b/router.go
@@ -7,6 +7,7 @@ package fiber
 import (
 	"fmt"
 	"sort"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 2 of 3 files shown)


app.go
```diff
@@ -261,6 +261,10 @@ type Static struct {
 	// The name of the index file for serving a directory.
 	// Optional. Default value "index.html".
 	Index string `json:"index"`
+
+	// The value for the Cache-Control that is set on the file response.
```
Fenny, line 265: You might want to clarify that the `int` argument defines as `seconds`

router.go
```diff
@@ -354,6 +358,7 @@ func (app *App) registerStatic(prefix, root string, config ...Static) Router {
 		// Return request if found and not forbidden
 		status := c.fasthttp.Response.StatusCode()
 		if status != StatusNotFound && status != StatusForbidden {
+			c.fasthttp.Response.Header.Set(HeaderCacheControl, "public, max-age="+strconv.Itoa(maxAge))
```
Fenny, line 361: We should build the `CacheControl` value once outside of the handler value.
```go
maxAgeVal := "public, max-age="+strconv.Itoa(maxAge)
```

And check if the value is higher than 1 before setting it
```go
if maxAge > 0 {
  c.fasthttp.Response.Header.Set(HeaderCacheControl, maxAgeVal)
}
```

Fenny (changes requested):
Nice PR, could you please address my comments above 👍

router.go
r-52: Yea, that was my pain point at this PR. I've applied your hints

app.go
r-52: Good catch! I've added a better explanation. Thinking out loud: What about an external link to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control?

r-52:
> Nice PR, could you please address my comments above 👍

Thanks for the review 👍 I've (hopefully) applied the changes

PR #897, opened 2020-10-07
Add custom types to compress middleware

klipitkas:
@Fenny this is more of an idea than a PR.

I believe it would help a lot if we used custom types for specific configurations:
- It enforces people to use a value that matches the type.
- We can limit only to the proper or "right" values.
- It works nicely with auto completion tools.
- It helps maintain a stricter interface and gives us more control.

If you agree I can make similar changes to the rest of the middleware and then take it from there.

### DIFF
--- a/middleware/compress/compress_test.go
+++ b/middleware/compress/compress_test.go
@@ -40,7 +40,7 @@ func Test_Compress_Gzip(t *testing.T) {
 	utils.AssertEqual(t, 200, resp.StatusCode, "Status code")
 	utils.AssertEqual(t, "gzip", resp.Header.Get(fiber.HeaderContentEncoding))
 
-	// Validate the file size is shrinked
+	// Validate the file size is shrunk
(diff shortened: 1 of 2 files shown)


middleware/compress/compress_test.go
```diff
@@ -40,15 +40,15 @@ func Test_Compress_Gzip(t *testing.T) {
 	utils.AssertEqual(t, 200, resp.StatusCode, "Status code")
 	utils.AssertEqual(t, "gzip", resp.Header.Get(fiber.HeaderContentEncoding))
 
-	// Validate the file size is shrinked
+	// Validate the file size is shrunk
```
napei, line 43: `has shrunk`

past-tense

klipitkas: Indeed, changed. I only fixed what golangci-lint pointed out there.

Fenny:
Sounds good 👍

PR #900, opened 2020-10-08
Add custom store to limiter middleware

codemicro:

This PR adds the ability to use a custom store for the `limiter` middleware, which can be useful when running in `prefork` mode, for example.

**Details**

This PR adds the following storage interface to the `limiter` middleware (as defined [here](https://github.com/hi019/fiber-sessions/blob/master/store.go#L14)).

Mode switching is used in order to keep performance as good as possible when not using a custom store.

Benchmark before any changes:

Benchmark after changes (with and without the default store):

WindowsTerminal_2020-10-08_20-07-22

In the above example with a custom store enabled, an in memory store is still being used. It's considerably slower than default because of extra encoding/decoding that needs to occur in order to store data as bytes in the custom store.

When a custom store is disabled, an approach almost identical to the original middleware is used, with a single `map[string]trackedSession`.


codemicro:
Wooo, 900th thing 😂

Aa12345667 (approved):

Aa12345667 (approved):

WLun001:
@codemicro i like your commit message 😂

renanbastos93 (approved):
Good job, thanks for your contributions to Fiber.

codemicro:
I seriously have no idea what I've done with Git here to make this mess of a set of commits BUT this now has `msgp` and `fiber.Handler` integrated. Performance is also really good compared to the previos implementation with a sample in memory custom store.

image

PR #908, opened 2020-10-10
🐛 Make SendFile return 404 if file not found

hi019:
fixes #907

PR #911, opened 2020-10-10
Add test for sendfile in logger

klipitkas:
Adds test for #908 in order to cover regressions.

### DIFF
--- a/middleware/logger/logger_test.go
+++ b/middleware/logger/logger_test.go
@@ -125,6 +125,19 @@ func Test_Logger_AppendUint(t *testing.T) {
 	utils.AssertEqual(t, "0 5 200", buf.String())
 }
 
+func TestLogger_FileNotFound(t *testing.T) {
+	app := fiber.New()
+
+	app.Use(New())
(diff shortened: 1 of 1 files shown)


kiyonlin:
Hi @klipitkas thanks for the PR. But I don't think it is the right place to put this test case.

klipitkas:
> Hi @klipitkas thanks for the PR. But I don't think it is the right place to put this test case.

Where would the best place be to add this? I think we need to test all cases where bugs have been found but not sure of the proper way to do it. Any help would be appreciated.

kiyonlin:
The bug was found in `Ctx.SendFile`, so it is better to add a test case in `ctx_test.go`.

klipitkas:
> The bug was found in `Ctx.SendFile`, so it is better to add a test case in `ctx_test.go`.

I think this is already covered at https://github.com/gofiber/fiber/blob/master/ctx_test.go#L1306

kiyonlin:
Hey @klipitkas I reviewed the test code and improved it ☝ . Thanks for your PR again!

PR #922, opened 2020-10-12
Refactor basicauth middleware

klipitkas:
- Split nested ifs to simpler statements
- Make the tests run in parallel (perhaps we should do this in all test cases under the middlewares)


Fenny:
Nice PR, they are run in parallel using `-race` in our workflow cmd 👍

PR #924, opened 2020-10-13
Refactor proxy middleware and increase coverage

klipitkas:
- Refactor the proxy middleware.
- Update coverage from 0% to 92%.

### DIFF
--- a/middleware/proxy/proxy.go
+++ b/middleware/proxy/proxy.go
@@ -44,7 +44,7 @@ var ConfigDefault = Config{
 
 // New is deprecated
 func New(config Config) fiber.Handler {
-	fmt.Println("proxy.New is deprecated, please us proxy.Balancer instead")
+	fmt.Println("proxy.New is deprecated, please use proxy.Balancer instead")
 	return Balancer(config)
 }
 
@@ -58,7 +58,9 @@ func Balancer(config Config) fiber.Handler {
 		cfg.Next = ConfigDefault.Next
 	}
 	if len(cfg.Servers) == 0 {
-		panic("Servers cannot be empty")
+		return func(c *fiber.Ctx) (err error) {
+			panic("Servers cannot be empty")
+		}
 	}
 
 	client := fasthttp.Client{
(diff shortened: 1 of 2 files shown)


middleware/proxy/proxy.go
```diff
@@ -58,7 +58,9 @@ func Balancer(config Config) fiber.Handler {
 		cfg.Next = ConfigDefault.Next
 	}
 	if len(cfg.Servers) == 0 {
-		panic("Servers cannot be empty")
+		return func(c *fiber.Ctx) (err error) {
+			panic("Servers cannot be empty")
+		}
```
hi019, line 63: Wouldn't it be better to keep it as it is and panic on MW creation?

middleware/proxy/proxy_test.go
```diff
@@ -1,148 +1,161 @@
...
+func Test_Proxy_Empty_Upstream_Servers(t *testing.T) {
+	app := fiber.New()
+	app.Use(recover.New(), Balancer(Config{Servers: []string{}}))
+
+	resp, err := app.Test(httptest.NewRequest("GET", "/", nil))
+	utils.AssertEqual(t, nil, err)
+	utils.AssertEqual(t, fiber.StatusInternalServerError, resp.StatusCode)
+}
 
 // // go test -run Test_Proxy_Next
```
hi019, line 25: Should be ` // go test -run Test_Proxy_Next`

hi019:
Hi :wave:

Tests were broken because localhost was used in `app.Listen`. I've committed to your repository and fixed the problem. Otherwise, nice PR!

middleware/proxy/proxy.go
hi019, line 63: Shouldn't we keep the code as it is and panic on MW creation?

middleware/proxy/proxy_test.go
hi019, line 25: Remove two forward slashes. Line 30 as well :)

middleware/proxy/proxy.go
klipitkas: This has an issue where it cannot be caught by the `recover` middleware.

middleware/proxy/proxy_test.go
klipitkas: we should be good here, thanks

PR #936, opened 2020-10-19
improve base uri for ctx

imxyb:
bench compare result:
```
benchmark                    old ns/op     new ns/op     delta
Benchmark_Ctx_BaseURL-12     58.0          2.24          -96.14%

benchmark                    old allocs     new allocs     delta
Benchmark_Ctx_BaseURL-12     1              0              -100.00%

benchmark                    old bytes     new bytes     delta
Benchmark_Ctx_BaseURL-12     32            0             -100.00%
```

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -39,6 +39,7 @@ type Ctx struct {
 	indexHandler int                  // Index of the current handler
 	method       string               // HTTP method
 	methodINT    int                  // HTTP method INT equivalent
+	baseURI      string               // HTTP base uri
 	path         string               // Prettified HTTP path -> string copy from pathBuffer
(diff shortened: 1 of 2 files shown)


ctx.go
```diff
@@ -20,11 +20,12 @@ import (
 	"text/template"
 	"time"
 
+	"github.com/valyala/fasthttp"
```
kiyonlin, line 23: Please merge all third libs together.

kiyonlin:
Looks good to me @imxyb

ctx.go
imxyb: done

kiyonlin (approved):

PR #962, opened 2020-10-25
🚨 Add missing header and compression result assertion in different compression levels test

sbielenica:
Adjusted one of the compression tests. Previously it didn't compress results at all. It just checked if 200 is returned without verifying compression results.


kiyonlin (approved):
Thanks for the improvement

PR #981, opened 2020-10-29
Implementing storage for CSRF Middleware

malusev998:
CSRF token middleware now complies with the ```fiber.Storage``` interface, which makes it extendable for other services like Redis in a production environment. Token generation is also made modular with the ```Generator``` interface, by default it uses cryptographically secure random bytes, which makes it not guessable.


codemicro (approved):
Hi @malusev998 - thanks for the PR! 

Everything looks good to me! At some point soon, we'll swap out the `storage` package to an in memory store from another repo, but this is perfect in the mean time.

I'm not going to merge this just yet, just because I'd like someone else from the team to take a look at it beforehand.

malusev998:
@codemicro Thanks for the review.
I was also thinking about extracting ```storage``` into its own repository, there we can implement all kinds of providers (Redis, SQL, etc...)

codemicro:
It's already under construction!

middleware/requestid/requestid.go
```diff
@@ -31,9 +31,9 @@ type Config struct {
 
 // ConfigDefault is the default config
 var ConfigDefault = Config{
-	Next:      nil,
-	Header:    fiber.HeaderXRequestID,
-	Generator: utils.UUID,
+	Next:       nil,
```
kiyonlin, line 34: Please remove the requestid mw related stuff from this PR. You can make another one of course:smile:

middleware/requestid/requestid_test.go
```diff
@@ -64,12 +64,12 @@ func Test_RequestID_Locals(t *testing.T) {
 
 	var ctxVal string
 
-	app.Use(func (c *fiber.Ctx) error {
+	app.Use(func(c *fiber.Ctx) error {
```
kiyonlin, line 67: Please remove the requestid mw related stuff from this PR. You can make another one of course:smile:

middleware/csrf/storage/memory.go
```diff
@@ -0,0 +1,77 @@
...
+	return nil
+}
+
+func (m *memoryStorage) gc() {
+	for {
+		// GC the tokens every 10 seconds to avoid
+		time.Sleep(10 * time.Second)
+		m.Lock()
+		for t := range m.tokens {
+			if time.Now().Unix() >= m.tokens[t] {
```
kiyonlin, line 71: Why not extracting a `now` variable for `time.Now()`

middleware/csrf/storage/memory.go
```diff
@@ -0,0 +1,77 @@
+package storage
```
kiyonlin, line 1: I think it is unnecessary to separate memory storage into another folder/package

middleware/csrf/csrf.go
```diff
@@ -100,8 +104,16 @@ func New(config ...Config) fiber.Handler {
...
 		}
+
+	}
+
+	if cfg.Storage == nil {
+		cfg.Storage = storage.NewMemoryStorage()
+	}
+
+	if cfg.Generator == nil {
+		cfg.Generator = cryptoRandomGenerator{length: 32}
```
kiyonlin, line 115: If you want to make generator customized, copy the [option](https://github.com/gofiber/fiber/blob/master/middleware/requestid/requestid.go#L20-L23) and [init stuff](https://github.com/gofiber/fiber/blob/master/middleware/requestid/requestid.go#L36) from requestid mw is fine.

middleware/csrf/csrf.go
```diff
@@ -271,3 +266,10 @@ func csrfFromCookie(param string) func(c *fiber.Ctx) (string, error) {
 		return token, nil
 	}
 }
+
+func shouldCheckCSRFToken(httpMethod string) bool {
```
kiyonlin, line 270: We can remove CSRF stuff because we are in csrf package imo

middleware/csrf/storage/memory.go
```diff
@@ -0,0 +1,77 @@
...
+	return storage
+}
+
+func (m *memoryStorage) Get(id string) ([]byte, error) {
+	m.RLock()
+	t, ok := m.tokens[id]
+	m.RUnlock()
+	// Check if token exist or expired
+	if !ok || time.Now().Unix() >= t {
+		return nil, errors.New("invalid key")
```
kiyonlin, line 32: It's better to add `csrf: ` prefix here to make the error clearer.

middleware/csrf/memory.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+func (m *memoryStorage) gc() {
+	for {
+		// GC the tokens every 10 seconds to avoid
+		time.Sleep(10 * time.Second)
+		now := time.Now().Unix()
+		for t := range m.tokens {
+			if now >= m.tokens[t] {
+				m.Lock()
```
kiyonlin, line 72: Why not putting this outside the for-range?

middleware/csrf/memory.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/gofiber/fiber/v2/utils"
+)
+
+type memoryStorage struct {
+	sync.RWMutex
+	tokens map[string]int64
+}
+
+// NewMemoryStorage - Creates new in-memory storage for CSRF tokens
+func NewMemoryStorage() fiber.Storage {
```
kiyonlin, line 18: I think it's better to make this factory method private

middleware/csrf/memory.go
kiyonlin, line 72: I'm wondering if we don't put this outside the for-range, will we have a race condition? @Fenny

malusev998: i think not, reading from map with multiple goroutines should be safe, write lock will not allow reads from map, so as long as writes are safe, this will not be race condition

middleware/csrf/memory.go
malusev998: what is the upside of making it private?

middleware/csrf/memory.go
malusev998: It's really bad for performance to acquire a lock every 10 seconds, maybe a Read lock, but i dont think this is important, memory provider should not be used in production anyway, scaling is impossible

middleware/csrf/memory.go
kiyonlin: I usually make stuff public only when it is necessary. We can change private to public but it's hard to do the opposite.

malusev998: i will make the change

middleware/csrf/csrf.go
```diff
@@ -1,10 +1,10 @@
 package csrf
 
 import (
+	"context"
```
renanbastos93, line 4: Maybe, Can we use context from Fiber?

malusev998: I don't think it's necessary, std context should work fine, besides many users will use std context.
If in the future ctx is exposed through config for other reasons that will provide compatibility between libraries

middleware/csrf/csrf.go
```diff
@@ -47,6 +47,16 @@ type Config struct {
 	//
 	// Optional. Default value "csrf".
 	ContextKey string
+
+	// CSRF Storage
+	//
+	// Optional. Default value MemoryStorage
+	Storage fiber.Storage
```
Fenny, line 54: Can you call it `Store`

middleware/csrf/csrf.go
```diff
@@ -47,6 +47,16 @@ type Config struct {
...
+
+	// Generator defines a function to generate the unique identifier.
+	//
+	// Optional. Default: utils.UUID
+	Generator func() string
```
Fenny, line 59: Could you name this `Key`

middleware/csrf/csrf.go
```diff
@@ -100,8 +106,15 @@ func New(config ...Config) fiber.Handler {
 		} else {
 			cfg.Cookie = ConfigDefault.Cookie
 		}
+
+		if cfg.Generator == nil {
```
Fenny, line 110: `cfg.Key`

middleware/csrf/csrf.go
```diff
@@ -100,8 +106,15 @@ func New(config ...Config) fiber.Handler {
...
+			cfg.Generator = utils.UUID
+		}
+	}
+
+	if cfg.Storage == nil {
```
Fenny, line 115: `cfg.Store`

Fenny (changes requested):
Please see comments

PR #983, opened 2020-10-31
Add custom key to cache middleware

### LINKED ISSUE #979: 🚀 add configuration option for key in cache mw
Currently the cache middleware uses a hard coded `c.Path()` as cache key.

Sometimes a more complicated cache key is required, so please add a configuration option to support a custom cache key function.

https://github.com/gofiber/fiber/blob/master/middleware/cache/cache.go#L126

codemicro:
This PR adds support for a custom key to be used in the cache middleware, akin to that found in the limiter middleware.

Fixes #979.

```go
app.Use(cache.New(cache.Config{
    Key: func (c *fiber.Ctx) string {
        return c.OriginalURL()
    },
})
```


marcusti:
Thanks for implementing this. Looks like exactly what I need. :+1:

PR #986, opened 2020-11-01
Add note about go mod to installation instructions

hi019:


hi019:
Need to add GOPROXY to CN README.

PR #988, opened 2020-11-02
🩹 Fix: Allow Request timestamp to be more accurate by using Time() from fasthttp.RequestCtx

Zate:
Also fixed formatting to use TimeFormat.

- I commended out the timestamp code, was not sure on the rules of just straight up removing it, as it may still serve a need?  I couldnt see that it did, so I shut it off and my testing didnt show any difference.

- Added code to take the output from the *fasthttp.RequestCtx context for the request, and use the Time() function to putll the time.Time object it had from starting to process the Request. 

- Used the Logger.Config options for TimeFormat and Timezone to output a string to go into the 'time' tag/variable.

When working with HTTP requests it is desirable to have at least milliseconds fidelity on the timestamp.  There can be thousands of requests per second happening and if we want to trace a particular one for security or operational reasons, it can be difficult if the timestamps are not setup correctly.

It looks like fasthttp already calculates these time stamps when it processes a Request and so instead of computing these, I think we could use the ones provided.  fasthttp put these in the *fasthttp.RequestCtx struct and I see that is inside the *fiber.Ctx struct.  This also removes the need for the timestamp creating go-routine that was sleeping for 750ms for performance reasons.  I think this might even improve performance.

⚡️🔥💩🔊

### DIFF
--- a/middleware/logger/logger.go
+++ b/middleware/logger/logger.go
@@ -159,15 +159,15 @@ func New(config ...Config) fiber.Handler {
 	var timestamp atomic.Value
 	timestamp.Store(time.Now().In(cfg.timeZoneLocation).Format(cfg.TimeFormat))
 
-	// Update date/time every 750 milliseconds in a separate go routine
-	if strings.Contains(cfg.Format, "${time}") {
-		go func() {
-			for {
-				time.Sleep(750 * time.Millisecond)
-				timestamp.Store(time.Now().In(cfg.timeZoneLocation).Format(cfg.TimeFormat))
-			}
-		}()
-	}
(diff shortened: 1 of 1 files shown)


ReneWerner87 (changes requested):
okay sorry, your change will affect the performance for everyone, we made benchmarks that showed that the creation of the timestamp is very time consuming and therefore we created this timestamp in a go routine like in the fasthttp library

a good fix would be to extend the configuration with the time intervals, then you can leave the default for all at 750ms and define a smaller interval for yourself

Fenny:
https://github.com/gofiber/fiber/pull/993 👍

PR #1006, opened 2020-11-10
Code refractor to avoid duplicate lines

JonanOribe:
Just a few constants to avoid duplicate code


Fenny:
Please use https://github.com/gofiber/fiber/blob/master/helpers.go#L450 for textplain const

JonanOribe:
> Please use https://github.com/gofiber/fiber/blob/master/helpers.go#L450 for textplain const

Thanks! now it´s done

PR #1009, opened 2020-11-11
🧹 housekeeping

Fenny:


router.go
```diff
@@ -349,7 +349,9 @@ func (app *App) registerStatic(prefix, root string, config ...Static) Router {
 		if maxAge > 0 {
 			cacheControlValue = "public, max-age=" + strconv.Itoa(maxAge)
 		}
-
+		if config[0].CacheDuration != 0 {
```
ReneWerner87, line 352: what happens if someone wants to set it to 0 ? is the 0 value possible

ReneWerner87 (approved):

router.go
Fenny: If you want to disable it, you need to provide a negative number i.e. `-1`

ReneWerner87: okay, not needed 
https://github.com/valyala/fasthttp/blob/805af0ee73bd914c4ee3ee8eccf400edaa0ccfb7/fs.go#L369
0 is not possible in fs.go

green light for the PR

codemicro (approved):
Sure, whatever, approved, I was too lazy to read all 2k lines - you're welcome :P

middleware/session/store.go
```diff
@@ -0,0 +1,60 @@
...
+func New(config ...Config) *Store {
+	// Set default config
+	cfg := configDefault(config...)
+
+	return &Store{
+		cfg,
+	}
+}
+
+func (s *Store) Get(c *fiber.Ctx) *Session {
```
kirillDanshin, line 22: looking at this signature, it doesn't feel right to me, as there's no way to know if there was an error or not, right? so you could rely on some security-related stuff to be in the session, but you could get get a fresh session instead of old one, which could end up in lost session-wide parameters, "ghost sessions", failed aug PAKE auth etc.

Fenny: Fixed in <https://github.com/gofiber/fiber/pull/1009/commits/2cafd486ef3b821adb64385b3a0d95f6f8488e7f> thanks

PR #1024, opened 2020-11-16
update readme, Key is deprecade

ancogamer:


middleware/limiter/README.md
```diff
@@ -36,9 +36,7 @@ app.Use(limiter.New(limiter.Config{
 	},
 	Max:          20,
 	Duration:     30 * time.Second,
-	Key:          func(c *fiber.Ctx) string {
-		return c.Get("x-forwarded-for")
-	},
+	KeyGenerator: func(c *fiber.Ctx) string,
```
hi019, line 39: We'll need a valid function, this is just a signature. Maybe

```go
KeyGenerator: func(c *fiber.Ctx) string {
  return "key"
},
```

hi019 (changes requested):

middleware/limiter/README.md
ancogamer: i will keep the default, return c.IP(), what you think ?

PR #1028, opened 2020-11-19
Fix: App Error Handler.

melbahja:
Fix to allow app `ErrorHandler` to handle 404 errors.

App config `ErrorHandler` not working with 404 errors.

### DIFF
--- a/router.go
+++ b/router.go
@@ -127,15 +127,14 @@ func (app *App) next(c *Ctx) (match bool, err error) {
 		return match, err // Stop scanning the stack
 	}
 
-	// If c.Next() does not match, return 404
-	_ = c.SendStatus(StatusNotFound)
-	_ = c.SendString("Cannot " + c.method + " " + c.pathOriginal)
-
 	// If no match, scan stack again if other methods match the request
 	// Moved from app.handler because middleware may break the route chain
 	if !c.matched && methodExist(c) {
 		err = ErrMethodNotAllowed
+		return
 	}
+
+	err = fmt.Errorf("Cannot %s %s", c.method, c.pathOriginal)
 	return
 }
 


codemicro:
Hi @melbahja, thanks for the PR!

**Unfortunately, this PR is not going to be merged.** This change would be a breaking change that changes fundamental behaviour of the error handler in Fiber, which is also intended behaviour.

The pre-existing error handling logic in Fiber is designed to deal with errors that are returned by a handler somewhere in the stack of middlewares and handlers. It is not suitable for handling errors that have been caused by an invalid request (for example, a 404) as the error handler for these errors returns a HTTP 500 responses by default, which is the incorrect response code.

If you need to setup a custom 404 handler, you can find information on how to do this here, in our docs: https://docs.gofiber.io/extra/faq#how-do-i-handle-custom-404-responses

kiyonlin:
> If you need to setup a custom 404 handler, you can find information on how to do this here, in our docs: https://docs.gofiber.io/extra/faq#how-do-i-handle-custom-404-responses

Beside @codemicro mentioned above, you can also write your own `ErrorHandler` to handle all kinds of errors.

melbahja:
On line 137 if we return `app.Error` instead that, this will fix the tests!
```go
// this
err := Error{Code: StatusNotFound, Message: fmt.Sprintf("Cannot %s %s", c.method, c.pathOriginal)}

// instead of
err = fmt.Errorf("Cannot %s %s", c.method, c.pathOriginal)
```
 I was gonna pass the tests, but you closed the PR quickly.

kiyonlin:
Hey @melbahja , what does `App config ErrorHandler not working with 404 errors.` mean exactly? Do you have any examples?

melbahja:
@kinbiko 

```go
package main

import (
	"github.com/gofiber/fiber/v2"
)

func main() {

		// Override default error handler
		ErrorHandler: func(ctx *fiber.Ctx, err error) error {
			// Statuscode defaults to 500
			code := fiber.StatusInternalServerError

			// Retreive the custom statuscode if it's an fiber.*Error
			if e, ok := err.(*fiber.Error); ok {
				code = e.Code
			}

			// Send custom error page
			return ctx.Status(code).SendString("My ErrorHandler Message")
		},
	})

		return c.SendString("Hello, World!")
	})

	app.Listen(":3000")

}
```

```bash
$ curl -i http://localhost:3000/not_found

HTTP/1.1 404 Not Found
Date: Thu, 26 Nov 2020 15:10:49 GMT
Content-Length: 21

Cannot GET /not_found
```
Expecting: `My ErrorHandler Message`

kiyonlin:
I think you misunderstood what `ErrorHandler` does. In your example, no error happens. It's correct to return 404 because you don't have a route with path `not_found`.

melbahja:
Maybe fiber needs `NotFoundHandler`?

kiyonlin:
> Maybe fiber needs `NotFoundHandler`?

Maybe you should make a feature issue.

codemicro:
> I was gonna pass the tests, but you closed the PR quickly.

The problem with this is that if you need to change the test cases, you're introducing changes that will also break any apps using Fiber when they're upgraded to a new minor or patch version of the framework. This is the sort of change that would go in the v3 release of Fiber, which isn't planned for some time yet.

@melbahja

Fenny:
> Maybe fiber needs `NotFoundHandler`?

<https://docs.gofiber.io/extra/faq#how-do-i-handle-custom-404-responses>

PR #1035, opened 2020-11-26
:seedling: Add fiber_vhost middleware

K0enM:

I wrote a small middleware to enable the use of virtual hosts by comparing the Host header with the Hostname specified in the Config. I plan on extending the middleware to match Host's by regex as well.


kiyonlin (approved):
LGTM

kiyonlin:
After a quick glance at the code, what do you think if we return [handler result](https://github.com/K0enM/fiber_vhost/blob/master/main.go#L34) directly?

K0enM:
Whoops, that should indeed be `return cfg.Handler(c)`. That's what you get by coding at 2 AM 😴 I'll fix it tomorrow.

PR #1059, opened 2020-12-07
Update timeout.go

JuniaWonter:

**Commit formatting** 


renanbastos93 (changes requested):
seems not good to me, can you explain more about your commit and about your idea?

PR #1076, opened 2020-12-14
🔥 Feature:  Add CustomTypeRegister for form

rockcreation7:
Add "CustomTypeRegister" for form data bodyparser, allow you to input any custom type, provide flexibility for form, so you can register you own custom type in form.

I added "CustomTypeRegister" struct just above the bodyparser, and optional parameter for form bodyparser.
The comment above "CustomTypeRegister" provide example and how to use.

You can do custom type in JSON format, Create custom type and related "UnmarshalJSON" method.
But in form you need to register converter in decoder, I added option for supporting this feature. In order to solve problem when you need custom type for form.

For example: 
It is mostly happen when people will sent you "2020-12-15" in browser form instead of RFC3339 format.
You can register this custom format, instead of force to use RFC3339. Other use case like, You can register own member ID type, etc.

**Commit formatting** 

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -77,6 +77,13 @@ type Views interface {
 	Render(io.Writer, string, interface{}, ...string) error
 }
 
+// BodyParserType require two element, type and converter for register.
+// Use BodyParserType with BodyParser for parsing custom type in form data.
+type BodyParserType struct {
+	Customtype interface{}
+	Converter  func(string) reflect.Value
+}
+
 // AcquireCtx retrieves a new Ctx from the pool.
 func (app *App) AcquireCtx(fctx *fasthttp.RequestCtx) *Ctx {
 	c := app.pool.Get().(*Ctx)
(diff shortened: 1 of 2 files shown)


ctx.go
```diff
@@ -244,13 +244,42 @@ var decoderPool = &sync.Pool{New: func() interface{} {
...
 // BodyParser binds the request body to a struct.
 // It supports decoding the following content types based on the Content-Type header:
 // application/json, application/xml, application/x-www-form-urlencoded, multipart/form-data
 // If none of the content types above are matched, it will return a ErrUnprocessableEntity error
-func (c *Ctx) BodyParser(out interface{}) error {
+func (c *Ctx) BodyParser(out interface{}, customTypeRegister ...CustomTypeRegister) error {
 	// Get decoder from pool
 	schemaDecoder := decoderPool.Get().(*schema.Decoder)
+
+	if len(customTypeRegister) > 0 {
```
kiyonlin, line 277: I think this sentence is redundant.

rockcreation7: Thanks for your time to review, updated pull request.

rockcreation7:
For your reference : https://github.com/gorilla/schema/blob/c1a93f601268c10220a683cd59d08e2fcc55ad3e/decoder.go#L57

I have provided a unit test with parameter, just under Test_Ctx_BodyParser, it could be doing in Test_Ctx_BodyParser, but it require extra condition and ReleaseCtx.

I could provide example on usage if needed. Will help working on doc.

Fenny:
PR looks good, I'm just not sure about struct naming. Since it's a public method attached to Fiber, we won't be able to change it in the future.

What do you think about `BodyParserType` or `BodyParserRegister`?

rockcreation7:
Thanks, it is about adding type and eventually it is adding type to BodyParser, "BodyParserType" is related and descriptive. one consideration is bodyparser is also working on json, dont know if user will confuse/think BodyParserType will work on json too.

But I believe doc will explain, I plan to work on the doc on this too. 

Btw the doc of fiber and recipe section is clear and good, it helps a lot to me.

Fenny:
After I saw https://github.com/gofiber/fiber/issues/1084, maybe we should provide an option to override the schema decoder globally?

rockcreation7:
Will be update

rockcreation7:
More elaboration on the code:

Added Following function for override the schema decoder globally :

```
customType := BodyParserType{
		Customtype: customType{},
		Converter:  timeConverter,
}

SetBodyParserDecoder(
         BodyParserConfig{
		IgnoreUnknownKeys: true,
		BodyParserType:    []BodyParserType{customType},
		ZeroEmpty:         true,
		SetAliasTag:       "form",
	}
)
```

The test is also update in last commit.

Please let me know if there any change/idea should be update for merging.
Enjoy Holidays guys

rockcreation7:
Closing this branch, duplication, new branch with bigger scope

Issue #1084, opened 2020-12-21
🐛  BodyParser does not save empty string to field


sandro:
**Fiber version**
 Fiber v2.1.0

**Issue description**
I'm trying to use BodyParser to save an empty string to a field in a struct, but the empty string will not overwrite the existing data in the field.

The bug disappears when I'm able to set the decoder's zeroEmpty to true, i.e. `decoder.ZeroEmpty(true)`.

I wonder if there's a way to allow users to change the decoder when necessary. Perhaps by allowing the user to pass a custom decoder into the BodyParser function?

**Code Sample**
```go
package main

import (
	"fmt"
	"log"
	"net/http"

	"github.com/gofiber/fiber/v2"
	"github.com/gofiber/fiber/v2/utils"
)

type Post struct {
	Title string `form:"title"`
	Body  string `form:"body"`
}


func main() {
	app := fiber.New()

        // curl -d "title=&body=new_body" localhost:3000/posts
	app.Post("/posts", func(c *fiber.Ctx) error {
		p := Post{Title: "Existing title", Body: "Existing Body"}
		c.BodyParser(&p)
		fmt.Printf("form value of title %q\n", c.FormValue("title"))
		fmt.Printf("form value of title %q\n", c.FormValue("body"))
		fmt.Printf("%+v\n", p)
		return c.SendStatus(http.StatusOK)
	})

        // curl -d "title=&body=new_body" localhost:3000/posts_use_empty_string
	app.Post("/posts_use_empty_string", func(c *fiber.Ctx) error {
		p := Post{Title: "Existing title", Body: "Existing Body"}
		var decoder = NewDecoder()
		decoder.IgnoreUnknownKeys(true)
		decoder.ZeroEmpty(true)
		decoder.SetAliasTag("form")
		data := make(map[string][]string)
		c.Context().PostArgs().VisitAll(func(key []byte, val []byte) {
			data[utils.UnsafeString(key)] = append(data[utils.UnsafeString(key)], utils.UnsafeString(val))
		})
		decoder.Decode(&p, data)
		fmt.Printf("form value of title %q\n", c.FormValue("title"))
		fmt.Printf("form value of title %q\n", c.FormValue("body"))
		fmt.Printf("%+v\n", p)
		return c.SendStatus(http.StatusOK)
	})

	log.Fatal(app.Listen(":3000"))
}
```

Sending an empty string to BodyParser does not overwrite the existing value

[...]

PR #1100, opened 2021-01-03
🔥 Add function to override form decoder setting

rockcreation7:

- SetBodyParserDecoder function map all option to form decoder, which allow you to change decoder option globally, use with type BodyParserConfig, BodyParserType.

- Added Test_Ctx_BodyParser_WithSetBodyParserDecoder for testing

- You can do custom type in JSON format, Create custom type and related "UnmarshalJSON" method, But in form you need to register converter in decoder, I added option for supporting this feature. In order to solve problem when you need custom type for form.

- Allow set the decoder's zeroEmpty to true, allow not skipping empty input as it will overwrite existing value.  #1084

```
customType := BodyParserType{
		Customtype: CustomType{},
		Converter:  timeConverter,
}

SetBodyParserDecoder(
         BodyParserConfig{
		IgnoreUnknownKeys: true,
		BodyParserType:    []BodyParserType{customType},
		ZeroEmpty:         true,
		SetAliasTag:       "form",
	}
)
```

**Commit formatting** 


rockcreation7:
Any suggestion for moving forward on this PR will be appreciated, or any idea should be further investigate, wish to provide help, Thanks.

ctx.go
```diff
@@ -244,6 +259,22 @@ var decoderPool = &sync.Pool{New: func() interface{} {
 	return decoder
 }}
 
+// SetBodyParserDecoder allow globally change the option of form decoder, update decoderPool
+func SetBodyParserDecoder(bodyParserConfig BodyParserConfig) {
```
ReneWerner87, line 263: can we use the method for the initial write of the decoder pool ?
https://github.com/gofiber/fiber/blob/master/ctx.go#L274
@rockcreation7

rockcreation7: Sure, I can help update the doc on this, thanks.

Please feel free to let me know anything that can help.

rockcreation7: or you mean I put a change to the initial write?

ReneWerner87: > or you mean I put a change to the initial write?
@rockcreation7 
yes exactly, with the initial set you can also use your method or ?

will merge it directly after the adjustment

ReneWerner87:
@rockcreation7 can you rename it a bit, because the decoder is used by the query parser and body parser

maybe something just called parser config or type

rockcreation7:
Thanks, will update.

rockcreation7:
https://github.com/gofiber/fiber/pull/1555/

Renamed function:
SetBodyParserDecoder to SetParserDecoder

Type:
BodyParserType > ParserType
bodyParserConfig > parserConfig
BodyParserConfig > ParserConfig

Issue #1108, opened 2021-01-09
c.JSON() driver 🚀 


pjebs:
It would be nice to be able to pick different json marshalling drivers instead of stdlib's "encoding/json".

There is so many faster ones:

1. https://github.com/json-iterator/go (drop in replacement for encoding/json)
2. https://github.com/valyala/fastjson
3. https://github.com/francoispqt/gojay
4. https://github.com/buger/jsonparser

hi019:
We use https://github.com/segmentio/encoding/blob/master/json/README.md for json operations

Fenny:
Fiber currently uses segmentio which is currently the fastest json encoder as a drop in replacement for STF (faster than json iterator https://github.com/json-iterator/go/issues/455 ) without generating files like fastjson.

JSON benchmark results: https://www.techempower.com/benchmarks/#section=test&runid=706be841-9ca9-4930-a8dd-c238592e3767&hw=ph&test=json&l=zijxtr-73&a=2

I would accept an PR where you could override the JSON engine in `Config` 👍

yquansah:
I seek to address this issue here: https://github.com/gofiber/fiber/pull/1121 if you all want to take a look.

Fenny:
https://github.com/gofiber/fiber/releases/tag/v2.4.0

PR #1121, opened 2021-01-20
feat: add definition for overridable json engine

yquansah:
The intention of this PR is to address the issue raised here: https://github.com/gofiber/fiber/issues/1108

It allows for configuration of the underlying JSON marshaller/unmarshaller by clients of fiber. It will just default to the `encoding/json` implementation if one isn't provided a client.

**Commit formatting** 


yquansah:
Would appreciate suggestions on naming throughout this PR. Any and all feedback is more than welcomed.

Fenny:
Thanks for the PR, `JSONEncoder` would be better. What ya think @kiyonlin @ReneWerner87 @hi019 @yquansah

renanbastos93 (approved):
LGTM

yquansah:
@Fenny I agree. I can make that change 👍

Fenny:
@yquansah, I do agree that your solution is more fitting for a more flexible approach if we add a json decoder in the future. But I highly doubt we will add that because we think that is not the responsibility of a web framework.

I still suggest simplifying it by using the STD json marshal signatures ( and add support for `Unmarshal` for BodyParser )

```go
JSONMarshal func(interface{}) ([]byte, error)
JSONUnmarshal func([]byte, interface{}) error
```

yquansah:
@Fenny Yeah I was thinking about that too, I do not think there will be a use case of doing json decoding, and if there is then it shouldn't be too hard to add. Let me know if my changes address what you are talking about.

ReneWerner87 (approved):

yquansah:
@ReneWerner87 @Fenny Hey y'all just wanted to check in and see if there is anything left to do on this PR?

ReneWerner87 (approved):
code looks good

can you add some words in 
https://github.com/gofiber/docs/blob/master/api/fiber.md
-> PR in the docs repository

Fenny:
LGTM! Thanks!

Issue #1083, opened 2020-12-20
Custom Error Func for CSRF Middleware🚀 


twhit-mrwebapps:
**Is your feature request related to a problem?**
Not so much a problem, more about improving usability and functionality.

**Describe the solution you'd like**
I'd like the option to be able to add a custom error response (in the way of a func) for the CSRF token as I wish to have a specific response when the CSRF token has expired / failed which would allow me to ensure the practices being used in the project we're developing are followed.

**Describe alternatives you've considered**
None.

**Additional context**
I don't know if I could expand on this request.

hi019:
Agreed this would be a good feature. Should be straightforward, just add the relevant config and replace the two `return fiber.ErrForbidden` with `config.ErrorHandler(c). Would you like to make a PR? We are always available in [the discord](https://gofiber.io/discord) for help.

twhit-mrwebapps:
When I get a chance, I'll look into it, thanks 👍

xenowits:
is this issue resolved?? @hi019

Fenny:
https://github.com/gofiber/fiber/releases/tag/v2.4.0

PR #1129, opened 2021-01-22
🚀Add Custom Error Func for CSRF Middleware

hinoguma:
#1083

### DIFF
--- a/middleware/csrf/config.go
+++ b/middleware/csrf/config.go
@@ -80,6 +80,11 @@ type Config struct {
 
 	// Deprecated, please use KeyLookup
 	TokenLookup string
+
+	// ErrorHandler is executed when an error is returned from fiber.Handler.
+	//
+	// Optional. Default: DefaultErrorHandler
+	ErrorHandler fiber.ErrorHandler
 }
 
 // ConfigDefault is the default config
@@ -89,6 +94,12 @@ var ConfigDefault = Config{
 	CookieSameSite: "Strict",
 	Expiration:     1 * time.Hour,
 	KeyGenerator:   utils.UUID,
+	ErrorHandler:   DefaultErrorHandler,
+}
+
+// Default ErrorHandler that process return error from fiber.Handler
+var DefaultErrorHandler = func(c *fiber.Ctx, err error) error {
+	return fiber.ErrForbidden
 }
 
 // Helper function to set default values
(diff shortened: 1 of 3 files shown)


middleware/csrf/config.go
```diff
@@ -89,6 +94,12 @@ var ConfigDefault = Config{
 	CookieSameSite: "Strict",
 	Expiration:     1 * time.Hour,
 	KeyGenerator:   utils.UUID,
+	ErrorHandler:   DefaultErrorHandler,
+}
+
+// Default ErrorHandler that process return error from fiber.Handler
+var DefaultErrorHandler = func(c *fiber.Ctx, err error) error {
```
kiyonlin, line 101: It's better to make `DefaultErrorHandler` private.

middleware/csrf/csrf_test.go
```diff
@@ -233,3 +233,54 @@ func Test_CSRF_From_Cookie(t *testing.T) {
...
+	})
+
+	h := app.Handler()
+	ctx := &fasthttp.RequestCtx{}
+
+	// Empty/invalid CSRF token
+	ctx.Request.Reset()
+	ctx.Response.Reset()
+	ctx.Request.Header.SetMethod("POST")
+	ctx.Request.Header.Set("X-CSRF-Token", "johndoe")
```
kiyonlin, line 282: This case does not set a invalid cookie. And `empty` also means `without cookie` right?

kiyonlin (changes requested):

middleware/csrf/config.go
hinoguma: @kiyonlin 
Thank you for review.
I fixed. 1ad5a618cb0da239c41a218d3b06029556e7e571

middleware/csrf/csrf_test.go
hinoguma: > This case does not set a invalid cookie.

I fixed Test_CSRF_ErrorHandler_InvalidToken for comparison between generated uuid token and invalid token.

> And empty also means without cookie right?

Empty means that token is not exists in request.
And I add Test_CSRF_ErrorHandler_EmptyToken in 1ad5a618cb0da239c41a218d3b06029556e7e571.


Test_CSRF_ErrorHandler_InvalidToken for
https://github.com/gofiber/fiber/blob/19e6a4429d1f9b6ae5f818c4b66a7357c78e58e4/middleware/csrf/csrf.go#L88-L103

Test_CSRF_ErrorHandler_EmptyToken for
https://github.com/gofiber/fiber/blob/19e6a4429d1f9b6ae5f818c4b66a7357c78e58e4/middleware/csrf/csrf.go#L82-L86

kiyonlin (approved):
LGTM, Thanks!

PR #1130, opened 2021-01-22
✨ add Next function to app.Static config

mlukasik-dev:

I've added option to to skip app.Static middleware. Mechanism is identical with proxy.Config.Next function from `proxy` middleware.

This can be feature is crucial for example when you want to serve different static files based on whether the user is authenticated.
Or when you just want to protect static file from unauthenticated users.

**Commit formatting** 

### DIFF
--- a/app.go
+++ b/app.go
@@ -302,6 +302,11 @@ type Static struct {
 	//
 	// Optional. Default value 0.
 	MaxAge int `json:"max_age"`
+
+	// Next defines a function to skip this middleware when returned true.
+	//
+	// Optional. Default: nil
+	Next func(c *Ctx) bool
 }
 
 // Default Config values
--- a/app_test.go
+++ b/app_test.go
@@ -758,6 +758,48 @@ func Test_App_Static_Trailing_Slash(t *testing.T) {
 	utils.AssertEqual(t, MIMETextPlainCharsetUTF8, resp.Header.Get(HeaderContentType))
 }
 
+func Test_App_Static_Next(t *testing.T) {
+	app := New()
+	app.Static("/", ".github", Static{
+		Next: func(c *Ctx) bool {
+			// If value of the header is any other from "skip"
+			// c.Next() will be invoked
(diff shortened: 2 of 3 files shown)


kiyonlin (commented):
Could you please add some test cases?

mlukasik-dev:
> Could you please add some test cases?

Sure, and thanks for reply!)

kiyonlin (approved):
LGTM, thanks!

ReneWerner87 (approved):

ReneWerner87 (commented):
Could you add the new options also in the gofiber docs repository
https://github.com/gofiber/docs/blob/master/api/app.md

mlukasik-dev:
@ReneWerner87 
I'll do that today.

app.go
```diff
@@ -302,6 +302,11 @@ type Static struct {
 	//
 	// Optional. Default value 0.
 	MaxAge int `json:"max_age"`
+
+	// Next defines a function to skip this middleware when returned true.
+	//
+	// Optional. Default: nil
+	Next func(c *Ctx) bool
```
Fenny, line 309: Could you change this to the `Handler` type?

Fenny (changes requested):

app.go
mlukasik-dev: Assuming that you're talking about this `Handler` type:
```go
// Handler defines a function to serve HTTP requests.
type Handler = func(*Ctx) error
```
I'm not sure under what circumstances the middleware should be skipped.
If `Handler` returns `nil` or `error`? 

Here is current login of `app.Static`:
```go
if config != nil && config[0].Next != nil && config[0].Next(c) {
    return c.Next()
}
// Serve file
// ...
```

mlukasik-dev:
@Fenny
So what's the status of this pull request?
Should I really change type to `Handler`

Fenny:
@mlukasik-dev totally missed the `bool` mb!

Issue #1119, opened 2021-01-18
🚀 Recover Middleware Debug Stacktrace


OrenBen-Meir:
I think it would be a good idea that the recover middleware has an option to show the stack trace of your panic. It should be off by default but for development and staging environments, it is very helpful when debugging.

ReneWerner87:
thanks @tianjipeng for the pull request #1131

Fenny:
https://github.com/gofiber/fiber/releases/tag/v2.4.0

PR #1131, opened 2021-01-23
🔥 Feature: recover middleware support print stack trace

tianjipeng:
The intention of this PR is to address the issue raised here: [1119](https://github.com/gofiber/fiber/issues/1119)
It allows for configuration of printing debug stacktrace.

Add variable to control whether print debug stacktrace  and can define custom print stack trace function.

**Commit formatting** 

### DIFF
--- a/middleware/recover/config.go
+++ b/middleware/recover/config.go
@@ -10,11 +10,25 @@ type Config struct {
 	//
 	// Optional. Default: nil
 	Next func(c *fiber.Ctx) bool
+
+	// EnableStackTrace enables handling stack trace
+	//
+	// Optional. Default: false
+	EnableStackTrace bool
+
+	// StackTraceHandler defines a function to handle stack trace
+	//
+	// Optional. Default: defaultStackTraceHandler
+	StackTraceHandler func(e interface{})
 }
+
+var defaultStackTraceBufLen = 1024
 
 // ConfigDefault is the default config
 var ConfigDefault = Config{
-	Next: nil,
+	Next:              nil,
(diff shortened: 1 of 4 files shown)


middleware/recover/config.go
```diff
@@ -10,11 +10,25 @@ type Config struct {
...
 }
 
+var defaultStackTraceBufLen = 1024
+
 // ConfigDefault is the default config
 var ConfigDefault = Config{
-	Next: nil,
+	Next:             nil,
+	EnableStackTrace: false,
+	PrintHandler:     defaultPrintHandler,
```
kiyonlin, line 31: I'm wondering if `StackTraceHandler` is more appropriate? It maybe does not just print the stack trace, but also can store it or send it to remote hosts.

kiyonlin (changes requested):

middleware/recover/config.go
tianjipeng: Got it. ``StackTraceHandler`` is better than ``PrintHandler ``

kiyonlin (approved):
LGTM 😄

ReneWerner87 (approved):

ReneWerner87:
Could you also add the new options in the gofiber docs repository
https://github.com/gofiber/docs/blob/master/api/middleware/recover.md#config

tianjipeng:
@ReneWerner87 docs pr is here: https://github.com/gofiber/docs/pull/147

Fenny:
LGTM

Issue #1116, opened 2021-01-16
 Cache Middleware: cache only responses where status == 200


GarryGaller:
How to make sure that responses with a status not equal to 200 are not cached?

hi019:
You can use the Next config param to check the status code. Check the documentation in the middleware folder. I'm on mobile right now so let me know if you need a sample later.

GarryGaller:
The Next function is not applicable for the response status, because the status is assigned to the response only before it is sent.
Accordingly, you cannot check the status in the Next function.

xenowits:
Hi @Fenny . I want to solve this issue. Can u plz point me to the correct direction??
Also @GarryGaller How did u know that nothing was cached? Plz help me understand the issue.

GarryGaller:
xenowits, "How did u know that nothing was cached?"
My logger (not the fiber logger) outputs intermediate information about the execution of the work.
If the data is cached, there is no output. There is only the final fiber logger.

kiyonlin:
Please try v2.4.1, and feel free to re-open this if you still have problems.

PR #1133, opened 2021-01-24
add NeedCache to control cache behavior

tianjipeng:
The intention of this PR is to address the issue raised here: [1116](https://github.com/gofiber/fiber/issues/1116)
It allows for configuration of controlling whether to cache or not

Provide NeedCache function to control whether to cache

**Commit formatting** 

### DIFF
--- a/middleware/cache/README.md
+++ b/middleware/cache/README.md
@@ -81,6 +81,13 @@ type Config struct {
 	//
 	// Default: an in memory store for this process only
 	Storage fiber.Storage
+
+	// NeedCache allows you to control whether to cache or not
+	//
+	// Default: func(c *fiber.Ctx) bool {
+	//   return true
+	// }
+	NeedCache func(*fiber.Ctx) bool
 }
 ```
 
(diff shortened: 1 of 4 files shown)


kiyonlin:
There's no need to add an extra option to solve that issue. The `Next` option is enough, but it is in the wrong place. Cache middleware is a post-run handler, so we do cache stuff after `Ctx.Next`, so should `config.Next`.

tianjipeng:
Cache Middleware has two part: get stuff from cache and put stuff to cache, ``config.Next`` now can only control both or not. So I think adding a ``NeedCache`` is reasonable to control whether to cache the response stuff. Do I understand it wrong?

kiyonlin:
What do you think is the meaning of config.Next?

tianjipeng:
I know the meaning of ``config.Next``.  But cache middleware is not only a post-run handler (put stuff to cache) but also a pre-run handler(get stuff from cache). I can't figure out how to implement the feature by ``config.Next``.  Could you give me some hint?

kiyonlin:
Only if response is cached(post run), then pre run handler will be hit, right? So if you don't want a cached response(in pre run), you shouldn't cache it. Do I make it clear?

tianjipeng:
Oh, I got it. Just put ``config.Next`` after ``ctx.Next``:
```go

if err := c.Next(); err != nil {
	return err
}

if cfg.Next != nil && cfg.Next(c) {
	return nil
}
```
Is tha right? If true, I will close this pr and reopen one.

kiyonlin:
Yes, that is what I mean. When you make another pull request, could you please add some related test cases?

tianjipeng:
ok, I will do that.

PR #1135, opened 2021-01-24
ajust config.Next position

tianjipeng:
The intention of this PR is to address the issue raised here: [1116](https://github.com/gofiber/fiber/issues/1116)

Move ``config.Next`` after ``ctx.Next``, so we can determine whether to cache or not after handling biz logic.

### DIFF
--- a/middleware/cache/cache_test.go
+++ b/middleware/cache/cache_test.go
@@ -231,6 +231,44 @@ func Test_Cache_NothingToCache(t *testing.T) {
 	}
 }
 
+func Test_Cache_CustomNext(t *testing.T) {
+	app := fiber.New()
+
+	app.Use(New(Config{
+		Next: func(c *fiber.Ctx) bool {
+			return !(c.Response().StatusCode() == fiber.StatusOK)
+		},
+		CacheControl: true,
+	}))
+
+	app.Get("/", func(c *fiber.Ctx) error {
+		return c.SendString(time.Now().String())
+	})
+
+	app.Get("/error", func(c *fiber.Ctx) error {
+		return c.Status(fiber.StatusInternalServerError).SendString(time.Now().String())
+	})
+
+	resp, err := app.Test(httptest.NewRequest("GET", "/", nil))
+	utils.AssertEqual(t, nil, err)
+	body, err := ioutil.ReadAll(resp.Body)
+	utils.AssertEqual(t, nil, err)
+
+	respCached, err := app.Test(httptest.NewRequest("GET", "/", nil))
+	utils.AssertEqual(t, nil, err)
+	bodyCached, err := ioutil.ReadAll(respCached.Body)
+	utils.AssertEqual(t, nil, err)
+	utils.AssertEqual(t, true, bytes.Equal(body, bodyCached))
+	utils.AssertEqual(t, true, respCached.Header.Get(fiber.HeaderCacheControl) != "")
+
+	_, err = app.Test(httptest.NewRequest("GET", "/error", nil))
+	utils.AssertEqual(t, nil, err)
+
+	errRespCached, err := app.Test(httptest.NewRequest("GET", "/error", nil))
+	utils.AssertEqual(t, nil, err)
+	utils.AssertEqual(t, true, errRespCached.Header.Get(fiber.HeaderCacheControl) == "")
+}
+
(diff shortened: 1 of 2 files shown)


middleware/cache/cache_test.go
```diff
@@ -231,6 +231,53 @@ func Test_Cache_NothingToCache(t *testing.T) {
...
+			return true
+		},
+	}))
+
+	app.Get("/", func(c *fiber.Ctx) error {
+		return c.SendString(time.Now().String())
+	})
+
+	app.Get("/error", func(c *fiber.Ctx) error {
+		return c.Status(fiber.StatusInternalServerError).Send([]byte(time.Now().String()))
```
kiyonlin, line 251: Why do you use Send here but not SendString?

middleware/cache/cache_test.go
```diff
@@ -231,6 +231,53 @@ func Test_Cache_NothingToCache(t *testing.T) {
...
+	})
+
+	resp, err := app.Test(httptest.NewRequest("GET", "/", nil))
+	utils.AssertEqual(t, nil, err)
+	body, err := ioutil.ReadAll(resp.Body)
+	utils.AssertEqual(t, nil, err)
+
+	time.Sleep(500 * time.Millisecond)
```
kiyonlin, line 259: What's the meaning of Sleep here?

middleware/cache/cache_test.go
```diff
@@ -231,6 +231,53 @@ func Test_Cache_NothingToCache(t *testing.T) {
...
+	bodyCached, err := ioutil.ReadAll(respCached.Body)
+	utils.AssertEqual(t, nil, err)
+	utils.AssertEqual(t, true, bytes.Equal(body, bodyCached))
+
+	resp, err = app.Test(httptest.NewRequest("GET", "/error", nil))
+	utils.AssertEqual(t, nil, err)
+	body, err = ioutil.ReadAll(resp.Body)
+	utils.AssertEqual(t, nil, err)
+
+	time.Sleep(500 * time.Millisecond)
```
kiyonlin, line 272: And here

middleware/cache/cache_test.go
```diff
@@ -231,6 +231,53 @@ func Test_Cache_NothingToCache(t *testing.T) {
 	}
 }
 
+func Test_Cache_CustomNext(t *testing.T) {
+	app := fiber.New()
+
+	app.Use(New(Config{
+		Next: func(c *fiber.Ctx) bool {
+			if c.Response().StatusCode() == fiber.StatusOK {
```
kiyonlin, line 239: Please simplified these code 😃

kiyonlin (changes requested):

middleware/cache/cache_test.go
tianjipeng: Oh,  I will change that.

middleware/cache/cache_test.go
tianjipeng: At the begining, I just want to make sure to return different string. Now I found it doesn't need  to  do that. I will fix it.

kiyonlin: Maybe we can take advantage of CacheControl option? If it is true, we'll get extra response header if it's cached.

tianjipeng: I add a new commit, plz have a look.

kiyonlin:
Please check error in CI

l3uddz:
Will the Custom Config readme example still be relevant with this change?

https://github.com/gofiber/fiber/tree/master/middleware/cache#custom-config

tianjipeng:
@kiyonlin I check the errorr, but it seems like proxy middleware test cases have data race:
image

Fenny:
Re-running tests

tianjipeng:
@Fenny @kiyonlin I check CI error and review the test code, I can't figure out why there is an error on windows platform. Could you have a look on my test case?

kiyonlin:
> @Fenny @kiyonlin I check CI error and review the test code, I can't figure out why there is an error on windows platform. Could you have a look on my test case?

How about removing body check?

tianjipeng:
@kiyonlin I have removed body check, plz have a look.

kiyonlin (approved):

PR #1149, opened 2021-02-04
✨ Add fiber-inertia to README.md

theArtechnology:
It is an inertia.js server-side adapter which helps creating single-page apps without building an API, using classic server-side routing and controllers.


kiyonlin:
Could you add it to other language reamdes?

theArtechnology:
> Could you add it to other language reamdes?

Yes

theArtechnology:
There you go !

kiyonlin (approved):
Thanks a lot!

PR #1152, opened 2021-02-06
🩹 Fix: add missing utf-8 charset

herla97:

* While testing an API, I detected that the response using the JSON function was returning strange characters, so I checked the Content-type header and saw that the **charset** was missing.

Based on the standard [RFC 3629](https://tools.ietf.org/html/rfc3629). I modified the [MIME Type](https://pkg.go.dev/github.com/gofiber/fiber#pkg-constants) of the JSON function in the ctx as follows:

**Before:**
```go
c.fasthttp.Response.Header.SetContentType(MIMEApplicationJSON)
```
**After:**
```go
c.fasthttp.Response.Header.SetContentType(MIMEApplicationJSONCharsetUTF8)
```
🩹🩹🩹🩹

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -541,7 +541,7 @@ func (c *Ctx) JSON(data interface{}) error {
 		return err
 	}
 	c.fasthttp.Response.SetBodyRaw(raw)
-	c.fasthttp.Response.Header.SetContentType(MIMEApplicationJSON)
+	c.fasthttp.Response.Header.SetContentType(MIMEApplicationJSONCharsetUTF8)
 	return nil
 }
 
--- a/ctx_test.go
+++ b/ctx_test.go
@@ -1411,7 +1411,7 @@ func Test_Ctx_JSON(t *testing.T) {
 		"Age":  20,
 	})
 	utils.AssertEqual(t, `{"Age":20,"Name":"Grame"}`, string(c.Response().Body()))
-	utils.AssertEqual(t, "application/json", string(c.Response().Header.Peek("content-type")))
+	utils.AssertEqual(t, "application/json; charset=utf-8", string(c.Response().Header.Peek("content-type")))
 
 	testEmpty := func(v interface{}, r string) {
 		err := c.JSON(v)


kiyonlin:
Sorry, I don't think it is mandatory to set charset to `utf-8` imo(correct me if I was wrong). You can change it in a middleware to fit your requirement.

Thanks for your PR anyway.

herla97:
> Sorry, I don't think it is mandatory to set charset to `utf-8` imo(correct me if I was wrong). You can change it in a middleware to fit your requirement.
> 
> Thanks for your PR anyway.

I have used other frameworks (buffalo, beego, echo) and they bring it by default, I suppose they are based on the mentioned standard.

I also made a middleware as follows:
```go
		c.Set("Content-type", "application/json; charset=utf-8")
		return c.Next()
	})
```

But it didn't work for me using the JSON function, I think the JSON function overwrites my Content-type. 

If I'm wrong about something, tell me. Thank for your time!

kiyonlin:
You should do it in post run
```go
	err := c.Next()
	c.Set("Content-type", "application/json; charset=utf-8")
	return err
})
```

kiyonlin:
> I have used other frameworks (buffalo, beego, echo) and they bring it by default, I suppose they are based on the mentioned standard.

But not all http server will response json with charset, such as laravel and [gin](https://github.com/gin-gonic/gin/blob/e899771392ecf35de8ce10a030ed8fed2207e9cb/context.go#L905-L909). 

We also can't just change the behavior as it would break things for people currently using it.

herla97:
> > I have used other frameworks (buffalo, beego, echo) and they bring it by default, I suppose they are based on the mentioned standard.
> 
> But not all http servers response json with charset, such as laravel and [gin](https://github.com/gin-gonic/gin/blob/e899771392ecf35de8ce10a030ed8fed2207e9cb/context.go#L905-L909).
> 
> We also can't just change the behavior as it would break things for people currently using it.

You are right. Thank you very much for your support.

Issue #1154, opened 2021-02-06
🤗 IPv6 support


suntong:
**Question description**

In #601, IPv6 support was added

https://github.com/gofiber/fiber/blob/b8fbaca1ea0b1a80c34257baa0a9c12e49c11de4/app.go#L111-L113

and tcp4 + tcp6 was default.

Now, 

https://github.com/gofiber/fiber/blob/3b956ef3a81de3228afb76bc5f56e1db5f797e39/app.go#L559

It seems only tcp4 is supported. 

Why is that, and how can I enable IPv6 again?

Thx

**Code snippet** _Optional_

```go
package main

import "github.com/gofiber/fiber/v2"

func main() {
  app := fiber.New()
  // ..
}
```

kiyonlin:
The primary reason is because of [prefork](https://github.com/gofiber/fiber/blob/master/prefork.go#L38-L39), we can't use `tcp` in prefork mode. And in additional, most cases are using `tcp4`, so we removed `tcp6` and only used `tcp4` by default in v2.

If you indeed need to start a `tcp6` server, we can discuss here.

suntong:
Ok, understand now. 

I do need to start a tcp6 server besides tcp4, what options I have?

kiyonlin:
Made a PR(#1155), we can move the discussion there.

suntong:
Now that the PR has been closed, I'll continue the discussion here...

> we can't use tcp in prefork mode.

I'd like to know more about that, because I can see that both tcp6 and tcp4 work just fine with prefork mode. So what was the problem?

The reason I'm asking is that, previously, besides _" "tcp4" (IPv4-only), "tcp6" (IPv6-only) "_, there is also a "`tcp`" setting that allows starting a tcp6 server _besides_ tcp4, which is the mode I was hoping to get...

kiyonlin:
That PR hasn't been merged yet. 

`tcp` doesn't work for linux and darwin in prefork mode. But it may works for windows which I don't have test environment.

suntong:
Oh, I don't care about Windows, 

Hmm, in that case, OK, I can live with only one, not both. 

thx for the PR, @kiyonlin

suntong:
Hmm,

> WARNING: When prefork is set to true, only "tcp4" and "tcp6" can be chose.

I don't use prefork, can I use `NetworkTCP` to have both tcp4 & tcp6 server then?

kiyonlin:
> Hmm,
> 
[...]

PR #1155, opened 2021-02-07
🔥 Add network option

kiyonlin:
Solve #1154 

```go
package main

import (
    "log"

    "github.com/gofiber/fiber/v2"
)

func main() {
        Prefork: true,
        Network: fiber.NetworkTCP6,
    })

        return c.SendString("Hello IPv6 👋")
    })

    log.Println(app.Listen("[::]:3000"))
}
```
By adding `Network` option, users can decide which network stack(`tcp`, `tcp4`, `tcp6`) to use.

**Commit formatting** 

### DIFF
--- a/app.go
+++ b/app.go
@@ -284,7 +284,7 @@ type Config struct {
 	// Known networks are "tcp", "tcp4" (IPv4-only), "tcp6" (IPv6-only)
 	// WARNING: When prefork is set to true, only "tcp4" and "tcp6" can be chose.
 	//
-	// Default: "tcp4"
+	// Default: NetworkTCP4
 	Network string
 }
 
@@ -403,7 +403,7 @@ func New(config ...Config) *App {
 		app.config.JSONEncoder = json.Marshal
 	}
 	if app.config.Network == "" {
-		app.config.Network = "tcp4"
+		app.config.Network = NetworkTCP4
 	}
 
 	// Init app
--- a/app.go
+++ b/app.go
@@ -280,6 +280,12 @@ type Config struct {
 	// Allowing for flexibility in using another json library for encoding
 	// Default: json.Marshal
 	JSONEncoder utils.JSONMarshal `json:"-"`
+
+	// Known networks are "tcp", "tcp4" (IPv4-only), "tcp6" (IPv6-only)
+	// WARNING: When prefork is set to true, only "tcp4" and "tcp6" can be chose.
+	//
+	// Default: "tcp4"
+	Network string
 }
 
 // Static defines configuration options when defining static assets.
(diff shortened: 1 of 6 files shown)


ReneWerner87 (approved):

prefork_test.go
```diff
@@ -23,15 +23,15 @@ func Test_App_Prefork_Child_Process(t *testing.T) {
...
+	err := app.prefork("tcp4", "invalid", nil)
 	utils.AssertEqual(t, false, err == nil)
 
 	go func() {
 		time.Sleep(1000 * time.Millisecond)
 		utils.AssertEqual(t, nil, app.Shutdown())
 	}()
 
-	utils.AssertEqual(t, nil, app.prefork("[::]:", nil))
+	utils.AssertEqual(t, nil, app.prefork("tcp6", "[::]:", nil))
```
hsblhsn, line 34: Wouldn't it be better if you keep this strings (`tcp`, `tcp4`, `tcp6`) as constant and use the constant everywhere?
Also make those constants exported, so that when I use the network option, i can use it like `Network: fiber.NetworkTCP4` instead of `Network: "tcp4"`. And typing hardcoded strings can be misspelled.

hsblhsn (changes requested):

app.go
```diff
@@ -396,6 +402,9 @@ func New(config ...Config) *App {
 	if app.config.JSONEncoder == nil {
 		app.config.JSONEncoder = json.Marshal
 	}
+	if app.config.Network == "" {
+		app.config.Network = "tcp4"
+	}
```
hsblhsn, line 407: Wouldn't it be better if you keep this strings (`tcp`, `tcp4`, `tcp6`) as constant and use the constant everywhere?
Also make those constants exported, so that when I use the network option, i can use it like `Network: fiber.NetworkTCP4` instead of `Network: "tcp4"`. And typing hardcoded strings can be misspelled.

hsblhsn (changes requested):

app.go
suntong: I like the idea, but now that the PR has been approved, would you send in another to fix it pls @hsblhsn?

hsblhsn: @suntong This PR is approved but not merged yet, I think @kiyonlin can add this little change in this PR. Or I have to wait until this PR merges. I don't want to open a PR with duplicated code from another PR with just a little bit of change. Whatever, I will try to open a PR if @kiyonlin don't add it in this PR.
Thank you!

suntong: gotya

kiyonlin: > Wouldn't it be better if you keep this strings (`tcp`, `tcp4`, `tcp6`) as constant and use the constant everywhere?
> 
> Also make those constants exported, so that when I use the network option, i can use it like `Network: fiber.NetworkTCP4` instead of `Network: "tcp4"`. And typing hardcoded strings can be misspelled.

Good idea! Will improve this.

suntong:
:+1: 

Just that the internal usages of "`tcp4`" might better be replaced as well.

kiyonlin:
> 👍
> 
> Just that the internal usages of "`tcp4`" might better be replaced as well.

Okey

Issue #1110, opened 2021-01-13
Can i custom session_id position?🤗 


xuyunan:
Like "Lookup" in gofiber/session, I want  save it in header:session_id

```
// Lookup is a string in the form of "<source>:<name>" that is used
// to extract session id from the request.
// Possible values: "header:<name>", "query:<name>" or "cookie:<name>"
// Optional. Default value "cookie:session_id".
Lookup string
```

leozhantw:
Hi, may I try to make a PR for this issue?

hi019:
> Hi, may I try to make a PR for this issue?

Sure

hi019:
No

PR #1159, opened 2021-02-09
🔥 Customize the source of session_id

leozhantw:

Try to support the need of https://github.com/gofiber/fiber/issues/1110.

I refer to [gofiber/session](https://github.com/gofiber/session) and [fasthttp/session](https://github.com/fasthttp/session) to implement this requirement.

**Commit formatting** 

### DIFF
--- a/middleware/session/config.go
+++ b/middleware/session/config.go
@@ -1,6 +1,8 @@
 package session
 
 import (
+	"fmt"
+	"strings"
 	"time"
 
 	"github.com/gofiber/fiber/v2"
@@ -17,9 +19,11 @@ type Config struct {
 	// Optional. Default value memory.New()
 	Storage fiber.Storage
 
-	// Name of the session cookie. This cookie will store session key.
-	// Optional. Default value "session_id".
-	CookieName string
+	// KeyLookup is a string in the form of "<source>:<name>" that is used
+	// to extract session id from the request.
+	// Possible values: "header:<name>", "query:<name>" or "cookie:<name>"
+	// Optional. Default value "cookie:session_id".
+	KeyLookup string
 
 	// Domain of the CSRF cookie.
 	// Optional. Default value "".
@@ -44,13 +48,32 @@ type Config struct {
 	// KeyGenerator generates the session key.
 	// Optional. Default value utils.UUIDv4
 	KeyGenerator func() string
+
+	// Deprecated, please use KeyLookup
+	CookieName string
+
+	// Source defines where to obtain the session id
+	source Source
+
+	// The session name
+	sessionName string
 }
+
+type Source string
+
+const (
+	SourceCookie   Source = "cookie"
+	SourceHeader   Source = "header"
+	SourceURLQuery Source = "query"
+)
 
 // ConfigDefault is the default config
 var ConfigDefault = Config{
 	Expiration:   24 * time.Hour,
-	CookieName:   "session_id",
+	KeyLookup:    "cookie:session_id",
 	KeyGenerator: utils.UUIDv4,
+	source:       "cookie",
+	sessionName:  "session_id",
 }
 
 // Helper function to set default values
@@ -67,11 +90,32 @@ func configDefault(config ...Config) Config {
 	if int(cfg.Expiration.Seconds()) <= 0 {
 		cfg.Expiration = ConfigDefault.Expiration
 	}
-	if cfg.CookieName == "" {
-		cfg.CookieName = ConfigDefault.CookieName
+	if cfg.CookieName != "" {
+		fmt.Println("[session] CookieName is deprecated, please use KeyLookup")
+		cfg.KeyLookup = fmt.Sprintf("cookie:%s", cfg.CookieName)
+	}
+	if cfg.KeyLookup == "" {
+		cfg.KeyLookup = ConfigDefault.KeyLookup
 	}
 	if cfg.KeyGenerator == nil {
 		cfg.KeyGenerator = ConfigDefault.KeyGenerator
 	}
+
+	selectors := strings.Split(cfg.KeyLookup, ":")
+	if len(selectors) != 2 {
+		panic("[session] KeyLookup must in the form of <source>:<name>")
+	}
+	switch Source(selectors[0]) {
+	case SourceCookie:
+		cfg.source = SourceCookie
+	case SourceHeader:
+		cfg.source = SourceHeader
+	case SourceURLQuery:
+		cfg.source = SourceURLQuery
+	default:
+		panic("[session] source is not supported")
+	}
+	cfg.sessionName = selectors[1]
+
 	return cfg
 }
--- a/middleware/session/session.go
+++ b/middleware/session/session.go
@@ -103,8 +103,8 @@ func (s *Session) Destroy() error {
 		return err
 	}
 
-	// Expire cookie
-	s.delCookie()
+	// Expire session
+	s.delSession()
 	return nil
 }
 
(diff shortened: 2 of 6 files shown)


middleware/session/session.go
```diff
@@ -171,6 +171,11 @@ func (s *Session) setCookie() {
...
 	s.ctx.Response().Header.SetCookie(fcookie)
 	fasthttp.ReleaseCookie(fcookie)
+
+	if s.config.Source == SourceHeader {
+		s.ctx.Request().Header.SetBytesV(s.config.CookieName, []byte(s.id))
+		s.ctx.Response().Header.SetBytesV(s.config.CookieName, []byte(s.id))
+	}
 }
 
 func (s *Session) delCookie() {
```
hi019, line 181: I think we should have separate functions for extracting a session ID from a cookie and from a header. You can find an example of this in our [csrf middleware](https://github.com/gofiber/fiber/blob/master/middleware/csrf/csrf.go#L30-L39).

middleware/session/session.go
```diff
@@ -171,6 +171,11 @@ func (s *Session) setCookie() {
 
 	s.ctx.Response().Header.SetCookie(fcookie)
 	fasthttp.ReleaseCookie(fcookie)
+
+	if s.config.Source == SourceHeader {
+		s.ctx.Request().Header.SetBytesV(s.config.CookieName, []byte(s.id))
+		s.ctx.Response().Header.SetBytesV(s.config.CookieName, []byte(s.id))
+	}
```
hi019, line 178: The code above this if statement will still execute even if the source is a header. It should be in an if statement that checks the source.

middleware/session/session.go
```diff
@@ -197,4 +202,9 @@ func (s *Session) delCookie() {
 
 	s.ctx.Response().Header.SetCookie(fcookie)
 	fasthttp.ReleaseCookie(fcookie)
+
+	if s.config.Source == SourceHeader {
+		s.ctx.Request().Header.Del(s.config.CookieName)
+		s.ctx.Response().Header.Del(s.config.CookieName)
+	}
```
hi019, line 209: Same here

middleware/session/session.go
hi019, line 178: A cookie will still be set even if the source is Header. We should put the above code in an if statement to prevent this. We should also rename the function to `setSession` or similar

middleware/session/session.go
hi019, line 209: Same here

middleware/session/session_test.go
```diff
@@ -203,7 +203,9 @@ func Test_Session_Save(t *testing.T) {
 	t.Parallel()
 
 	// session store
-	store := New()
+	store := New(Config{
+		Source: SourceHeader,
+	})
```
hi019, line 208: We should probably have separate test cases for saving a session to a header and saving a session to a cookie

middleware/session/store.go
```diff
@@ -72,6 +72,33 @@ func (s *Store) Get(c *fiber.Ctx) (*Session, error) {
 	return sess, nil
 }
 
+// getSessionID will returns the session id from:
```
hi019, line 75:
```suggestion
// getSessionID will return the session id from:
```

middleware/session/store.go
```diff
@@ -72,6 +72,33 @@ func (s *Store) Get(c *fiber.Ctx) (*Session, error) {
...
+// 2. http headers
+// 3. query string
+func (s *Store) getSessionID(c *fiber.Ctx) string {
+	id := c.Cookies(s.CookieName)
+	if len(id) > 0 {
+		return id
+	}
+
+	if s.Source == SourceHeader {
+		id = string(c.Request().Header.Peek(s.CookieName))
```
hi019, line 86: We shouldn't use the cookie name as the header name, this will be confusing to users. Instead we should use a key-value syntax, see our csrf middleware for an example: https://github.com/gofiber/fiber/blob/master/middleware/csrf/csrf.go#L30-L40

hi019 (changes requested):
Thanks for the PR!

leozhantw:
Hi,

I'm a bit busy recently, sorry to reply to you so late.

And I have made some adjustments, plz help me to confirm it again, thanks!

middleware/session/session.go
leozhantw: Thanks! it has been adjusted to `setSession`

middleware/session/session.go
leozhantw: Thanks, adjusted.

middleware/session/session_test.go
leozhantw: Thanks!

Done for the separation of the test case.

middleware/session/store.go
leozhantw: Fixed, thanks~!

middleware/session/store.go
leozhantw: I have been added `KeyLookUp` to instead it.

hi019:
Apologies for the delay, will review this tomorrow

ReneWerner87:
@leozhantw could you refresh the pull request and solve the change request ?

then we can also merge this slowly^^

ReneWerner87:
will do the review tonight

middleware/session/store.go
```diff
@@ -87,9 +86,36 @@ func (s *Store) Get(c *fiber.Ctx) (*Session, error) {
...
+
+	if s.source == SourceHeader {
+		id = string(c.Request().Header.Peek(s.sessionName))
+		if len(id) > 0 {
+			return id
+		}
+	}
+
+	if s.source == SourceURLQuery {
+		id = c.FormValue(s.sessionName)
```
ReneWerner87, line 107: this should be the wrong function or ? if we want to get it from the query parameters it should be c.Query(s.sessionName)

ReneWerner87 (approved):
not sure about one part i have marked in the review, otherwise the whole thing looks very good

although i think you could still optimize some things by adding benchmarks and look that you reduce the allocations

like for example when splitting the string, you could look instead of split with index checks, but that should be insignificant, because it is before the process of the requests and only affects the configuration

good work and thx

middleware/session/store.go
leozhantw: Although `c.FormValue()` will search the value from query string first, 

but you are right, it should only search from query string.

Changed here to use `c.Query()`. Thanks you!

#### Ref
https://github.com/valyala/fasthttp/blob/v1.23.0/server.go#L1004-L1036

middleware/session/config.go
```diff
@@ -17,9 +19,11 @@ type Config struct {
 	// Optional. Default value memory.New()
 	Storage fiber.Storage
 
-	// Name of the session cookie. This cookie will store session key.
-	// Optional. Default value "session_id".
-	CookieName string
+	// KeyLookup is a string in the form of "<source>:<name>" that is used
+	// to extract session id from the request.
+	// Possible values: "header:<name>", "query:<name>" or "cookie:<name>"
+	// Optional. Default value "cookie:session_id".
```
hi019, line 25:
```suggestion
	// to extract a session id from the request.
	// Possible values: "header:<name>", "query:<name>" or "cookie:<name>"
	// Optional. Default value: "cookie:session_id".
```

hi019 (approved):

hi019:
Thank you for the PR! Other then the small grammatical thing I pointed out, looks good 👍

ReneWerner87:
@leozhantw can you update the pull request again with the master, i had unfortunately produced an incorrect test in the master in the meantime

after that i will start the tests from PR again and merge them directly if they were successful

leozhantw:
> @leozhantw can you update the pull request again with the master, i had unfortunately produced an incorrect test in the master in the meantime
> 
> after that i will start the tests from PR again and merge them directly if they were successful

Done~!

PR #1162, opened 2021-02-09
New feature: ListenConfigTLS

DevDevious:
The function accepts a custom tls.Config parameter, supporting more advanced TLS configuration scenarios.

### DIFF
--- a/app.go
+++ b/app.go
@@ -638,6 +638,29 @@ func (app *App) ListenTLS(addr, certFile, keyFile string) error {
 	return app.server.ServeTLS(ln, certFile, keyFile)
 }
 
+func (app *App) ListenConfigTLS(addr string, config *tls.Config) error {
+	if app.config.Prefork {
+		return app.prefork(app.config.Network, addr, config)
+	}
+
+	// Setup listener
+	ln, err := net.Listen(app.config.Network, addr)
+	if err != nil {
+		return err
+	}
+	// prepare the server for the start
+	app.startupProcess()
+	// Print startup message
+	if !app.config.DisableStartupMessage {
+		app.startupMessage(addr, true, "")
+	}
+
+	// Start listening
+	return app.Server().Serve(
+		tls.NewListener(ln, config),
+	)
+}
+
 // Config returns the app config as value ( read-only ).
 func (app *App) Config() Config {
 	return app.config
--- a/app.go
+++ b/app.go
@@ -638,9 +638,12 @@ func (app *App) ListenTLS(addr, certFile, keyFile string) error {
 	return app.server.ServeTLS(ln, certFile, keyFile)
 }
 
-func (app *App) ListenConfigTLS(addr string, config *tls.Config) error {
+// ListenConfigTLS serves HTTPs requests from the given addr.
+//  addr - address to listen on
+//  conf - custom TLS configuration
+func (app *App) ListenConfigTLS(addr string, conf *tls.Config) error {
 	if app.config.Prefork {
-		return app.prefork(app.config.Network, addr, config)
+		return app.prefork(app.config.Network, addr, conf)
 	}
 
 	// Setup listener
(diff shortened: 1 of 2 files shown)


ReneWerner87:
Thank you for the effort,

Think we need at least one more test for the new feature to maintain the quality of the framework and a new pull request in the documentation repository with information on how to use it.

@Fenny @kiyonlin What do you think about the new function? Offer this additionally or replace the old one?

DevDevious:
@ReneWerner87  : my pleasure. 

A bit of background on `ListenConfigTLS`'s purpose. 

**1 -** Currently, if a client certificate is absent or otherwise invalid, the connection is closed at transport layer. This is fine for basic client authentication scenarios, however if one needs to take further action, such as redirecting users to custom error pages, one would need to implement custom client certificate validation and error handling at application layer. 

**2 -** This is currently achievable by passing in a custom TLS listener. However starting `Fiber` with a custom TLS listener will lead it to display `http://addr:port` in its startup message instead of `https://addr:port`. This happens because Fiber can only tell it's a `HTTPS` endpoint if started with `ListenTLS`. Anything else and it will display `HTTP`.

 `ListenConfigTLS` solves and **(2)** and opens up **(1)**.

ReneWerner87:
@DevDevious thanks for the background, can you add a test case(unittest) for the new function and a little documentation(PR) in this repository https://github.com/gofiber/docs/blob/master/api/app.md#listentls like the listentls

DevDevious:
Docs @ https://github.com/gofiber/docs/pull/152

ReneWerner87:
We need a unittest for the new function.

ReneWerner87 (approved):

Fenny:
@DevDevious can we rename it to `ListenTLSConfig` ?

DevDevious:
@Fenny I named it `ListenConfigTLS` as opposed to `ListenTLSConfig` in order avoid the merger of  the 3 letter acronym **TLS** with the upper case **C** in Config, resulting in **TLSC**. It is something I tend to do in my code, but I'm down to rename it if you want.

Another possibility would be to merge `ListenTLS` and `ListenConfigTLS`, keeping the **name of the first** and the **signature of the second**, like @ReneWerner87 mentioned earlier.

hi019 (approved):

kiyonlin:
> @Fenny @kiyonlin What do you think about the new function? Offer this additionally or replace the old one?

I haven't seen such Listen signature in fasthttp or gin. And I think `app.Listener` is enough, for example:
```go
package main

import (
	"crypto/tls"

	"github.com/gofiber/fiber/v2"
)

func main() {
	app := fiber.New()

	cer, _ := tls.LoadX509KeyPair("crt", "key")

	config := &tls.Config{
		Certificates: []tls.Certificate{cer},
		// Other configurations
	}

	ln, _ := tls.Listen("tcp", ":3000", config)

	app.Listener(ln)
}
```

And the new feature will be 
```go
package main

import (
	"crypto/tls"

	"github.com/gofiber/fiber/v2"
)

func main() {
	app := fiber.New()

	cer, _ := tls.LoadX509KeyPair("crt", "key")

	config := &tls.Config{
		Certificates: []tls.Certificate{cer},
		// Other configurations
	}

	app.ListenConfigTLS(":3000", config)
}
```

There is no need to add an api with such little benefit.

ReneWerner87:
think @kiyonlin has a good point

DevDevious:
```go
ln, _ := tls.Listen("tcp", ":3000", config)
```

messes-up the start-up message which will continue to show HTPP as protocol. It's the little things...

kiyonlin:
> ```go
> ln, _ := tls.Listen("tcp", ":3000", config)
> ```
> 
> messes-up the start-up message which will continue to show HTPP as protocol. It's the little things...

This is a bug and can be fixed. Thanks for reporting!

DevDevious:
Suit yourself.

kiyonlin:
> Suit yourself.

Sorry we can't accept this PR, but thanks for your efforts anyway!

PR #1170, opened 2021-02-15
📚 fix wrong session code example

ReneWerner87:


middleware/session/README.md
```diff
@@ -76,7 +76,8 @@ app.Get("/", func(c *fiber.Ctx) error {
 		panic(err)
 	}
 
-	return fmt.Fprintf(ctx, "Welcome %v", name)
+	_, err = fmt.Fprintf(c, "Welcome %v", name)
```
hi019, line 79: Maybe we should use c.SendString? Just to make it a bit simpler and use Fiber functions

ReneWerner87: Jo, i will change it tomorrow

PR #1175, opened 2021-02-18
add embed example for go 1.16

circa10a:
[Go embedding](https://golang.org/pkg/embed/) was released in Go 1.16 yesterday. This PR is to add an example for how to use it with the filesystem middleware

### DIFF
--- a/middleware/filesystem/README.md
+++ b/middleware/filesystem/README.md
@@ -5,11 +5,13 @@ Filesystem middleware for [Fiber](https://github.com/gofiber/fiber) that enables
 ⚠️ **`:params` & `:optionals?` within the prefix path are not supported!**
 
 ## Table of Contents
+
 - [Filesystem Middleware](#filesystem-middleware)
 	- [Table of Contents](#table-of-contents)
 	- [Signatures](#signatures)
 	- [Examples](#examples)
 		- [Config](#config)
+		- [embed](#embed)
 		- [pkger](#pkger)
 		- [packr](#packr)
 		- [go.rice](#gorice)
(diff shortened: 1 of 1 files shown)


hi019:
That was fast, thank you! Can we also put a notice about recommending this over the other solutions listed? 

> If your environment (Go 1.16+) supports it, we recommend using Go Embed instead of the other solutions listed as this one is native to Go and the easiest to use.

circa10a:
> That was fast, thank you! Can we also put a notice about recommending this over the other solutions listed?
> 
> > If your environment (Go 1.16+) supports it, we recommend using Go Embed instead of the other solutions listed as this one is native to Go and the easiest to use.

Added 👍

middleware/filesystem/README.md
```diff
@@ -4,12 +4,16 @@ Filesystem middleware for [Fiber](https://github.com/gofiber/fiber) that enables
 
 ⚠️ **`:params` & `:optionals?` within the prefix path are not supported!**
 
+> If your environment (Go 1.16+) supports it, we recommend using Go Embed instead of the other solutions listed as this one is native to Go and the easiest to use.
```
hi019, line 7: Lets move this to above `### embed`

jorgefuertes:
Good idea, but seems needed to make an example to embed a static tree, not only one file, I'm stuck in that right now. Thanks!

jorgefuertes:
Please, note this in order to serve a directory tree:

```
//go:embed public
var public embed.FS

subFS, _ := fs.Sub(public, "public")
app.Use("/", filesystem.New(filesystem.Config{
	Root:         http.FS(subFS),
}))
```

chiroro-jr:
Thank you so much for this merge. I was looking for something exactly like this to serve static assets from an embedded filesystem.

Issue #1114, opened 2021-01-15
🤗 How to use fasthttp client in fiber?


gxdvip:
**Question description**

I don’t know how to get fasthttp client in fiber.
I want to request an open api.

kiyonlin:
Fiber doesn't wrap anything about fasthttp client, you can use it directly.

Fenny:
Fiber does not wrap the fh client like @kiyonlin mentioned. But you can easily import the fh package to use the client along with the `Request` and `Response` properties directly.

```go
package main

import (
	"log"

	"github.com/gofiber/fiber/v2"
	"github.com/valyala/fasthttp"
)

func main() {
    app := fiber.New()
    app.Get("/", func(c *fiber.Ctx) error {
        req := c.Request()
        res := c.Response()
        req.SetRequestURI("https://api.ipify.org?format=json")
        return client.Do(req, res)
    })
    log.Fatal(app.Listen(":3000"))
}
```

gxdvip:
@Fenny Why not wrap it up? This is more convenient to use.

PR #1177, opened 2021-02-19
🔥 Implement Fiber http Client

kiyonlin:
Implement #1114, it is on top of fasthttp http client 

# Fiber Client and Agent
Fiber http client has two struct:
```go
// Client implements http client.
//
// It is safe calling Client methods from concurrently running goroutines.
type Client struct {
	// UserAgent is used in User-Agent request header.
	UserAgent string

	// NoDefaultUserAgentHeader when set to true, causes the default
	// User-Agent header to be excluded from the Request.
	NoDefaultUserAgentHeader bool

	// When set by an external client of Fiber it will use the provided implementation of a
	// JSONMarshal
	//
	// Allowing for flexibility in using another json library for encoding
	JSONEncoder utils.JSONMarshal

	// When set by an external client of Fiber it will use the provided implementation of a
	// JSONUnmarshal
	//
	// Allowing for flexibility in using another json library for decoding
	JSONDecoder utils.JSONUnmarshal
}
```
And 
```go
// Agent is an object storing all request data for client.
// Agent instance MUST NOT be used from concurrently running goroutines.
type Agent struct {
	// Name is used in User-Agent request header.
	Name string

	// NoDefaultUserAgentHeader when set to true, causes the default
	// User-Agent header to be excluded from the Request.
	NoDefaultUserAgentHeader bool

	// HostClient is an embedded fasthttp HostClient
	*fasthttp.HostClient
}
```

`Client` has methods like `Get`, `Post`, etc, they'll create a new `Agent` instance and parse request url automatically. And all field values in `Client` will be passed to `Agent` instance.

You can also get an `Agent` instance via `AcquireAgent` and parse request url manually:
```go
a := AcquireAgent()
req := a.Request()
req.Header.SetMethod(MethodGet)
req.SetRequestURI("http://example.com")

if err := a.Parse(); err != nil {
	panic(err)
}

code, body, errs := a.Bytes()
// ...
```

`Agent` is on top of [fasthttp `HostClient`](https://github.com/valyala/fasthttp/blob/master/client.go#L603) which has lots of convenient helper methods and several `send request` methods(will show them later).

For an http request, it will be divided into two phases:
1. prepare request data
2. send request

Each phase may accumulates errors in the `Agent` instance. So you may get multiple errors(`[]error`) after sending a request.

**NOTE** that `Agent` instances MUST NOT be used from concurrently running goroutines.

# Prepare request data
## Set Header
`Agent` instance has handy methods about commonly used http header:
- `ContentType` sets Content-Type header value.
- `Referer` sets Referer header value.
- `UserAgent` sets UserAgent header value.
- `ConnectionClose` sets 'Connection: close' header.
- `Cookie` sets one 'key: value' cookie.
- `Set` sets the given 'key: value' header.
- `Add` adds the given 'key: value' header.

Some of them have corresponding `Bytes` methods.

## Set URI
`Agent` instance also has handy methods about commonly used URI attributes:
- `Host` sets host for the URI.
- `QueryString` sets URI query string.
- `BasicAuth` sets URI username and password.

All of them have corresponding `Bytes` methods.

## Set Request body
`Agent` instance has several ways to set request body:
- `BodyString` sets request body.
- `BodyStream` sets request body stream and, optionally body size.
- `JSON` sends a JSON request.
- `XML` sends a XML request.
- `Form` sends form request with body if args is non-nil.
- `MultipartForm` sends multipart form request with k-v and files.

Some of them have corresponding `Bytes` methods.

`MultipartForm` has some helper methods to attach files and change the boundary:
- `FileData` appends files for multipart form request.
- `SendFile` reads file and appends it to multipart form request.
- `SendFiles`  reads files and appends them to multipart form request.
- `Boundary` sets boundary for multipart form request.

### Debug
`Debug` mode enables logging request and response detail:
```go
fiber.Get("http://example.com").Debug().Bytes()
```
```hs
Connected to example.com(1.1.1.1:80)

GET / HTTP/1.1
User-Agent: fiber
Host: example.com

HTTP/1.1 200 OK
Date: Fri, 26 Feb 2021 02:11:27 GMT
Content-Length: 5

debug
```

`Debug` can also accept an optional `io.Writer` to redirect the output, and it uses `os.Stdout` by default.

### Reuse
`Reuse` enables the `Agent` instance to be used again after one request. By default, after a request, `Agent` instance will be released automatically. So if an `Agent` instance is reusable, then it should be released manually when it is no longer used. This is helping for sending same requests.

### Timeout
`Timeout` sets request timeout duration.

### MaxRedirectsCount
`MaxRedirectsCount` sets max redirect count for method `GET` and method `HEAD` and specific redirectable response status code: `301 Moved Permanently`, `302 Found`, `303 See Other`, `307 Temporary Redirect` and `308 Permanent Redirect`.

### InsecureSkipVerify
`InsecureSkipVerify` controls whether the `Agent` instance verifies the server certificate chain and host name.

### TLSConfig
`TLSConfig` sets tls config.

### SetResponse
`SetResponse` sets custom response for the `Agent` instance.

It is recommended obtaining custom response via `AcquireResponse` and release it manually in performance-critical code.

### Dest
`Dest` sets custom dest.

The contents of dest will be replaced by the response body, if the dest is too small a new slice will be allocated.

# Send request
Currently there are three methods for sending request:
- `Bytes` returns the status code, bytes body and errors of url.
- `String` returns the status code, string body and errors of url.
- `Struct` returns the status code, bytes body and errors of url. And bytes body will be unmarshalled to given v.


sujit-baniya:
@kiyonlin I think it's better to include following features as well:

- Throttle per host
- Retry Mechanism

ReneWerner87:
@sujit-baniya 
good idea, but not for the first step, it is important to first make progress and then add features, feel free to add this later as a feature request, i also have a few things in mind that can be added as features when the first step is finished

ReneWerner87 (approved):
really awesome work

ReneWerner87:
maybe we should add something in the docs repository @kiyonlin 

kiyonlin:
> maybe we should add something in the docs repository @kiyonlin
> https://github.com/gofiber/docs

Yeah, we should add an extra section for client.

PR #1180, opened 2021-02-20
🩹 Fix: CSRF middleware cookie<>storage bug squashed and other improvements

sixcolors:
Bug fixes and improvements to csrf middleware:

Fixed bug where token was deleted from manager storage but cookie was not removed so it was not re-issued.
Re-structured handler switch statement to assume that anything not defined as 'safe' by RFC7231 needs protection.
Removed code deleting token from storage.
Add (re)generate token if not set on all methods.

### DIFF
--- a/middleware/csrf/csrf.go
+++ b/middleware/csrf/csrf.go
@@ -48,18 +48,12 @@ func New(config ...Config) fiber.Handler {
 		}
 
 		var token string
-		generateToken := false
 
 		// Action depends on the HTTP method
 		switch c.Method() {
 		case fiber.MethodGet, fiber.MethodHead, fiber.MethodOptions, fiber.MethodTrace:
 			// Declare empty token and try to get existing CSRF from cookie
 			token = c.Cookies(cfg.CookieName)
-
-			// Generate CSRF token if not exist
-			if token == "" {
-				generateToken = true
-			}
 		default:
 			// Assume that anything not defined as 'safe' by RFC7231 needs protection
 
@@ -83,32 +77,30 @@ func New(config ...Config) fiber.Handler {
 				})
 				return cfg.ErrorHandler(c, err)
 			}
-
-			// The token is validated, time to delete it
-			manager.delete(token)
-			generateToken = true
 		}
 
-		if generateToken {
+		// Generate CSRF token if not exist
+		if token == "" {
 			// And generate a new token
 			token = cfg.KeyGenerator()
-			// Add token to Storage
-			manager.setRaw(token, dummyValue, cfg.Expiration)
-
-			// Create cookie to pass token to client
-			cookie := &fiber.Cookie{
-				Name:     cfg.CookieName,
-				Value:    token,
-				Domain:   cfg.CookieDomain,
-				Path:     cfg.CookiePath,
-				Expires:  time.Now().Add(cfg.Expiration),
-				Secure:   cfg.CookieSecure,
-				HTTPOnly: cfg.CookieHTTPOnly,
-				SameSite: cfg.CookieSameSite,
-			}
-			// Set cookie to response
-			c.Cookie(cookie)
 		}
+
+		// Add/update token to Storage
+		manager.setRaw(token, dummyValue, cfg.Expiration)
+
+		// Create cookie to pass token to client
+		cookie := &fiber.Cookie{
+			Name:     cfg.CookieName,
+			Value:    token,
+			Domain:   cfg.CookieDomain,
+			Path:     cfg.CookiePath,
+			Expires:  time.Now().Add(cfg.Expiration),
+			Secure:   cfg.CookieSecure,
+			HTTPOnly: cfg.CookieHTTPOnly,
+			SameSite: cfg.CookieSameSite,
+		}
+		// Set cookie to response
+		c.Cookie(cookie)
 
 		// Protect clients from caching the response by telling the browser
 		// a new header value is generated
--- a/middleware/csrf/csrf.go
+++ b/middleware/csrf/csrf.go
@@ -84,21 +84,20 @@ func New(config ...Config) fiber.Handler {
 			if err != nil {
 				return cfg.ErrorHandler(c, err)
 			}
+			
+			// Expire cookie
+			c.Cookie(&fiber.Cookie{
+				Name:     cfg.CookieName,
+				Domain:   cfg.CookieDomain,
+				Path:     cfg.CookiePath,
+				Expires:  time.Now().Add(-1 * time.Minute),
+				Secure:   cfg.CookieSecure,
+				HTTPOnly: cfg.CookieHTTPOnly,
+				SameSite: cfg.CookieSameSite,
+			})
 
 			// if token does not exist in Storage
 			if manager.getRaw(token) == nil {
-
-				// Expire cookie
-				c.Cookie(&fiber.Cookie{
-					Name:     cfg.CookieName,
-					Domain:   cfg.CookieDomain,
-					Path:     cfg.CookiePath,
-					Expires:  time.Now().Add(-1 * time.Minute),
-					Secure:   cfg.CookieSecure,
-					HTTPOnly: cfg.CookieHTTPOnly,
-					SameSite: cfg.CookieSameSite,
-				})
-
 				return cfg.ErrorHandler(c, err)
 			}
 
--- a/middleware/csrf/csrf.go
+++ b/middleware/csrf/csrf.go
@@ -48,21 +48,52 @@ func New(config ...Config) fiber.Handler {
 		}
 
 		var token string
+		generateToken := false
 
 		// Action depends on the HTTP method
 		switch c.Method() {
-		case fiber.MethodGet:
+		case fiber.MethodGet, fiber.MethodHead, fiber.MethodOptions, fiber.MethodTrace:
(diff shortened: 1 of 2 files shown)


hi019 (approved):

hi019 (changes requested):
Thanks for the PR! We'll need some tests for making sure a token is generated on HEAD, OPTIONS & TRACE?

sixcolors:
Why would you want to test all of the 'safe' methods got getting the csrf_ token instead or any of them (ie GET)?

hi019 (approved):

sixcolors:
Fiat enough. Tests added.

sixcolors:
docs updated with PR https://github.com/gofiber/docs/pull/154

middleware/csrf/csrf.go
```diff
@@ -98,12 +81,33 @@ func New(config ...Config) fiber.Handler {
...
-
 				return cfg.ErrorHandler(c, err)
 			}
 
 			// The token is validated, time to delete it
 			manager.delete(token)
+			generateToken = true
+		}
+
+		if generateToken {
```
kiyonlin, line 92: I think you change the original logic of this middleware. When using safe methods, and we already get token, we should get a new cookie with new expiration, And then we call `c.Vary(fiber.HeaderCookie)`.

kiyonlin:
Our strategy is [per-session token](https://stackoverflow.com/questions/8655817/csrf-protection-do-we-have-to-generate-a-token-for-every-form)

sixcolors:
It wasn’t per session before. And it should regenerate after safe methods if it deletes it (like it was).

sixcolors:
If that was the strategy that’s not what it was doing. It was deleting when validated before.

middleware/csrf/csrf.go
sixcolors: If you update the cookie here it will fail to validate later as you weren’t updating the manager storage ttl. The token is compared to stored value and not just the cookie in your code.

sixcolors:
```
case fiber.MethodGet:
			// Declare empty token and try to get existing CSRF from cookie
			token = c.Cookies(cfg.CookieName)

			// Generate CSRF token if not exist
			if token == "" {
				// Generate new CSRF token
				token = cfg.KeyGenerator()

				// Add token to Storage
				manager.setRaw(token, dummyValue, cfg.Expiration)
			}

			// Create cookie to pass token to client
			cookie := &fiber.Cookie{
				Name:     cfg.CookieName,
				Value:    token,
				Domain:   cfg.CookieDomain,
				Path:     cfg.CookiePath,
				Expires:  time.Now().Add(cfg.Expiration),
				Secure:   cfg.CookieSecure,
				HTTPOnly: cfg.CookieHTTPOnly,
				SameSite: cfg.CookieSameSite,
			}
			// Set cookie to response
			c.Cookie(cookie)

		case fiber.MethodPost, fiber.MethodDelete, fiber.MethodPatch, fiber.MethodPut:
			// Extract token from client request i.e. header, query, param, form or cookie
			token, err = extractor(c)
			if err != nil {
				return cfg.ErrorHandler(c, err)
			}

			// if token does not exist in Storage
			if manager.getRaw(token) == nil {

				// Expire cookie
				c.Cookie(&fiber.Cookie{
					Name:     cfg.CookieName,
					Domain:   cfg.CookieDomain,
					Path:     cfg.CookiePath,
					Expires:  time.Now().Add(-1 * time.Minute),
					Secure:   cfg.CookieSecure,
					HTTPOnly: cfg.CookieHTTPOnly,
					SameSite: cfg.CookieSameSite,
				})

				return cfg.ErrorHandler(c, err)
			}

			// The token is validated, time to delete it
			manager.delete(token)
		}
```

main branch logic flow:

GET method only:

If no token in cookie present: generate token->store in manager with ttl now + config.Expiration->set csrf cookie with now + config.Expiration

If token cookie present: set csrf cookie with now + config.Expiration (but don't extent ttl in manager).

POST, DELETE, PATCH, PUT methods:

If token from extractor '' or token does not match manager.getRaw(token) delete cookie, feturn cfg.ErrorHandler(c, err)

If token valid: manager.delete(token)

If token valid and manager.delete(token): Next GET request will not generate a new token as the cookie is set, however it will fail the next validation as the token was deleted from storage.

middleware/csrf/csrf.go
sixcolors: ie you're extending the expiry of one element for the compare but not the other.

sixcolors:
> Our strategy is [per-session token](https://stackoverflow.com/questions/8655817/csrf-protection-do-we-have-to-generate-a-token-for-every-form)

We should have a token rotation strategy. I used rotate on expire or consume in the PR. This strategy requires that the cookie be the source of token not a from value from statically generated html for parallel requests so AJAX for example on the client.

If we want to just keep setting the csrf cookie, extending the expiry, and only have one csrf token per session: We would have to also manager.setRaw(token, dummyValue, cfg.Expiration) each time.

Finally the csrf token should be generated by all methods not must GET as per the main branch. A post method for example can return a body with a HTML from that can then be submitted.

sixcolors:
[](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html)

sixcolors:
[Cross-Site_Request_Forgery_Prevention_Cheat_Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html)

middleware/csrf/csrf.go
sixcolors: It would be easy enough to extend the token in manager.setRaw(token, dummyValue, cfg.Expiration) by calling it on every request along with setting the cookie. That way both get expirations extended each request.

That would have performance implications, do we accept those?

middleware/csrf/csrf.go
```diff
@@ -98,13 +75,32 @@ func New(config ...Config) fiber.Handler {
...
+		}
+
+		// Generate CSRF token if not exist
+		if token == "" {
+			// And generate a new token
+			token = cfg.KeyGenerator()
+		}
 
-			// The token is validated, time to delete it
-			manager.delete(token)
```
sixcolors, line 106: if doing csrf per session don't remove token from manager when validated.

middleware/csrf/csrf.go
```diff
@@ -98,13 +75,32 @@ func New(config ...Config) fiber.Handler {
...
+			Value:    token,
+			Domain:   cfg.CookieDomain,
+			Path:     cfg.CookiePath,
+			Expires:  time.Now().Add(cfg.Expiration),
+			Secure:   cfg.CookieSecure,
+			HTTPOnly: cfg.CookieHTTPOnly,
+			SameSite: cfg.CookieSameSite,
 		}
+		// Set cookie to response
+		c.Cookie(cookie)
```
sixcolors, line 103: generate csrf token and set token for each request if not exists.

middleware/csrf/csrf.go
```diff
@@ -98,13 +75,32 @@ func New(config ...Config) fiber.Handler {
...
-
 				return cfg.ErrorHandler(c, err)
 			}
+		}
+
+		// Generate CSRF token if not exist
+		if token == "" {
+			// And generate a new token
+			token = cfg.KeyGenerator()
+		}
```
sixcolors, line 86: All requests should set/extend the token

middleware/csrf/csrf.go
```diff
@@ -51,34 +51,12 @@ func New(config ...Config) fiber.Handler {
...
-				Value:    token,
-				Domain:   cfg.CookieDomain,
-				Path:     cfg.CookiePath,
-				Expires:  time.Now().Add(cfg.Expiration),
-				Secure:   cfg.CookieSecure,
-				HTTPOnly: cfg.CookieHTTPOnly,
-				SameSite: cfg.CookieSameSite,
-			}
-			// Set cookie to response
-			c.Cookie(cookie)
```
sixcolors, line 79: moved to all methods

middleware/csrf/csrf.go
```diff
@@ -51,34 +51,12 @@ func New(config ...Config) fiber.Handler {
 
 		// Action depends on the HTTP method
 		switch c.Method() {
-		case fiber.MethodGet:
+		case fiber.MethodGet, fiber.MethodHead, fiber.MethodOptions, fiber.MethodTrace:
```
sixcolors, line 54: RFC7231 defines GET HEAD OPTIONS and TRACE as safe, no need to validate the token, but check to see if the cookie is set, we generate token and set cookie later if not.

middleware/csrf/csrf.go
sixcolors: assume yes, commit made

sixcolors:
> Our strategy is [per-session token](https://stackoverflow.com/questions/8655817/csrf-protection-do-we-have-to-generate-a-token-for-every-form)

Seems valid. I ref this: [Cross-Site_Request_Forgery_Prevention_Cheat_Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html)

kiyonlin (approved):
LGTM

PR #1189, opened 2021-02-24
🔥 Add optional FileSystem config to middleware/favicon

dechristopher:
This change exists to support go 1.16 embedded filesystems and other network filesystems that can't simply be read as a file from a system path.

Since go 1.16 just dropped with embedded filesystem support, I thought fiber would benefit from fully supporting alternate filesystems. The filesystem middleware already supports any filesystem but I noticed the favicon middleware was still built to simply read from a system path.

### DIFF
--- a/middleware/favicon/favicon.go
+++ b/middleware/favicon/favicon.go
@@ -2,6 +2,7 @@ package favicon
 
 import (
 	"io/ioutil"
+	"net/http"
 	"strconv"
 
 	"github.com/gofiber/fiber/v2"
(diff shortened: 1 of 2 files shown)


dechristopher:
Looks like my test uses go 1.16 functions. I'll fix with a more suitable 1.14+ capable calls.

dechristopher:
I have mimicked the `http.FileSystem.Open` function in a wrapper `mockFS` type to simulate an alternate `http.FileSystem` for the purposes of the test since `os.DirFS` is not available <1.16. If there is a more suitable way to mock `http.FileSystem`, please let me know and I'll update the test accordingly.

dechristopher:
This is now ready for review.

kiyonlin (approved):
LGTM. Thank you for the contributing!

Issue #1184, opened 2021-02-22
🐛 The middleware has session operation, and there is no setting for the first visit 


da4z:
**Fiber version**
v2.5.0
**Issue description**
The session middleware has session operation, and there is no setting for the first visit 
**Code snippet**

```go
package main

import (
	"github.com/gofiber/fiber/v2"
	"github.com/gofiber/fiber/v2/middleware/session"
)

func main() {
	store := session.New()

	app := fiber.New()

	app.Use(func(c *fiber.Ctx) error {
		sess, err := store.Get(c)
		if err != nil {
			return err
		}
		defer sess.Save()

		sess.Set("id", "1")

		return c.Next()
	})

	app.Get("/", func(c *fiber.Ctx) error {
		sess, err := store.Get(c)
		if err != nil {
			return err
		}
		defer sess.Save()

		//EXEC SET

		sess.Set("name", "john")

		return c.JSON(fiber.Map{
			"id":   sess.Get("id"),
			"name": sess.Get("name"),
		})
	})

	app.Get("/id", func(c *fiber.Ctx) error {
		sess, err := store.Get(c)
		if err != nil {
			return err
		}
		defer sess.Save()

		//NO SET

		return c.JSON(fiber.Map{
			"id": sess.Get("id"),
		})
	})

	app.Listen("127.0.0.1:3008")
}
```

![x](https://user-images.githubusercontent.com/46473093/108671043-e2d2d700-751a-11eb-8b69-55b061de8f5b.gif)

ReneWerner87:
the reason should be that you execute the "save" function at the end, i.e. after the "next" and therefore after the next 
handler.

![image](https://user-images.githubusercontent.com/7063188/108675640-5f4bc280-74e7-11eb-8097-6b4e7ea91570.png)

probably everything works as you expect it, if you take away the defer for the "save" call

i will test it again, but the assumption looks like this for now

chinaapus:
@ReneWerner87 
If the middleware "Save" is removed, it will not be saved if it encounters a route without session 
```go
package main

import (
	"github.com/gofiber/fiber/v2"
	"github.com/gofiber/fiber/v2/middleware/session"
)

func main() {
	store := session.New()

	app := fiber.New()

	app.Use(func(c *fiber.Ctx) error {
		sess, err := store.Get(c)
		if err != nil {
			return err
		}
		//defer sess.Save()

		sess.Set("id", "1")

		return c.Next()
	})

[...]

PR #1191, opened 2021-02-27
🐛fix: lookup cookie in response header

tianjipeng:

This pr try to fix this issue: https://github.com/gofiber/fiber/issues/1184

``ctx.Cookies`` will look up coookie in response header if cookie is not found in request header

**Commit formatting** 

### DIFF
--- a/middleware/session/session_test.go
+++ b/middleware/session/session_test.go
@@ -274,6 +274,30 @@ func Test_Session_Cookie(t *testing.T) {
 	utils.AssertEqual(t, 84, len(ctx.Response().Header.PeekCookie(store.CookieName)))
 }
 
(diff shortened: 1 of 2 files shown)


ReneWerner87 (changes requested):
Please don't change the cookie method, but the content of the session manager and please consider the fresh status like in my comments in the issue, otherwise the session will be marked as not new in the second handler, although it didn't exist before the request start

tianjipeng:
@ReneWerner87 Sorry, my mistake.  Now it is updated, please review it again, thanks.

ReneWerner87:
Look better now, will check it again on my pc in a few hours.

ReneWerner87 (approved):

Issue #1207, opened 2021-03-08
binding to "0.0.0.0" startupMessage reports "127.0.0.1" 🐛 


gregchalmers:
**Fiber version**
v2.5.0

**Issue description**
If I bind Listen to "0.0.0.0" startupMessage reports "127.0.0.1", but OS says it's bound to "0.0.0.0".

**Code snippet**

```go
package main
import "github.com/gofiber/fiber/v2"
func main() {
  app := fiber.New()
  log.Fatal(app.Listen("0.0.0.0:8080"))
}
```
netstat 
```
$ netstat -lnp | grep LISTEN | grep 8080
tcp        0      0 0.0.0.0:8080            0.0.0.0:*               LISTEN      189950/./main
```
output
```
 ┌───────────────────────────────────────────────────┐ 
 │                    Fiber v2.5.0                   │ 
 │               http://127.0.0.1:8080               │ 
 │                                                   │ 
 │ Handlers ............. 6  Processes ........... 1 │ 
 │ Prefork ....... Disabled  PID ............ 189363 │ 
 └───────────────────────────────────────────────────┘ 
 ```
code responsible in fiber/app.go
```
	host, port := parseAddr(addr)
	if host == "" || host == "0.0.0.0" {
		host = "127.0.0.1"
	}
```

kiyonlin:
You want to see `0.0.0.0` in banner like #1198 do?

gregchalmers:
ah right sorry yes same as [#1198](https://github.com/gofiber/fiber/issues/1198), sorry I didn't find the other one I should have searched the closed bugs as well.

I cannot really see how printing `127.0.0.1` when it's bound to `0.0.0.0` is valid. 
`127.0.0.1` == bind to the loopback device, i.e. gofiber is only accessible by processes on the same machine.
`0.0.0.0` == bind to all network cards on that machine, i.e. the gofiber process is likely accessible to remote machines.

If someone uses `0.0.0.0` by mistake and start banner says `127.0.0.1` this seems a security issue to me.
Also as in my case when I bind to `0.0.0.0` and I read the startup banner say `127.0.0.1` its confusing. Hence I checked with `netstat` to make sure I wasn't going crazy.

kiyonlin:
The reason of why we use `127.0.0.1`, is that `0.0.0.0` can't be clicked on the terminal to open the address on a browser. I think we should treat them as two address (bound address and visit address) if necessary.

gregchalmers:
[...]

PR #1210, opened 2021-03-09
🔥 adds and additional line to starting banner "bound: http://0.0.0.0:8080" 

gregchalmers:
I submitted an issue #1207 saying when I bound to socket "0.0.0.0" the banner said 127.0.0.1.
i.e.
```
 ┌───────────────────────────────────────────────────┐ 
 │                    Fiber v2.5.0                   │ 
 │               http://127.0.0.1:3000               │ 
 │                                                   │ 
 │ Handlers ............. 0  Processes ........... 1 │ 
 │ Prefork ....... Disabled  PID ............ 382963 │ 
 └───────────────────────────────────────────────────┘ 
```
with this PR you will now see the following when host == "0.0.0.0" or ""
```
 ┌───────────────────────────────────────────────────┐ 
 │                    Fiber v2.5.0                   │ 
 │              https://127.0.0.1:38171              │ 
 │            bound https://0.0.0.0:38171            │ 
 │                                                   │ 
 │ Handlers ............. 0  Processes ........... 1 │ 
 │ Prefork ....... Disabled  PID ............ 384749 │ 
 └───────────────────────────────────────────────────┘ 
```
I'm just a newbie at go so it might be worth a review, or a rejection. I tested manually and `go test`


koddr:
@gregchalmers and @kiyonlin

I think it would be a good idea to color-code the clickable address.

kiyonlin:
I love @koddr's idea

gregchalmers:
thanks @koddr I updated the message, yeah its way clearer.

kiyonlin (approved):
LGTM

koddr:
By the way, @kiyonlin 

Maybe it is better to use the `switch` operator in this case than `if...else` ? I mean, do it like this:

```go
	// ...
	switch host {
	case "0.0.0.0":
		logoFmt += " │ %s │\n"
		logoArgs = append(logoArgs,
			center(fmt.Sprintf("%s://127.0.0.1:%s", scheme, port), 49),
			center(fmt.Sprintf("bound %s://0.0.0.0:%s", scheme, port), 49),
		)
	default:
		logoArgs = append(logoArgs, center(fmt.Sprintf("%s://%s:%s", scheme, host, port), 49))
	}
```

We may still have other cases and it will look better this approach (in the future). What do you think?

kiyonlin:
I think we only have one special case `0.0.0.0`. So it is not very necessary imo. We could use `switch` if more cases happen.

koddr:
> I think we only have one special case 0.0.0.0. So it is not very necessary imo. We could use switch if more cases happen.

Okay! So, LGTM now.

ReneWerner87 (approved):

PR #1213, opened 2021-03-10
🐛  fix: prefork is working on macOS

renanbastos93:
based on `fasthttp` our prefork doesn't work to macOS so we need to link the file descriptor with the listener/proc, however, we are implementing it. whatever consideration tells me. I am open to feedback or criticals.

I realized that the net.Listener cannot get the file descriptor, I thought to use net.TCPListener. What do you think?

### DIFF
--- a/prefork.go
+++ b/prefork.go
@@ -10,8 +10,6 @@ import (
 	"strconv"
 	"strings"
 	"time"
-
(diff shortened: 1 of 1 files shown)


renanbastos93:
I did tests in macOS Big Sur but doesn't work according to scenery. My laptop is chip a new M1. Someone help us?

PR #1229, opened 2021-03-17
🐛 Cache Middleware - Race condition in the cache expiration logic

defshift:
This PR is for investigating and fixing the problem with a race condition that is shown by flaky test:

```
=== RUN   Test_Cache_Expired
Error:     cache_test.go:76: Cache should not have expired: 1615964974123514694, 1615964974123857398
--- FAIL: Test_Cache_Expired (2.00s)
```

**Upd**: details in comments

### DIFF
--- a/middleware/cache/cache.go
+++ b/middleware/cache/cache.go
@@ -61,7 +61,7 @@ func New(config ...Config) fiber.Handler {
 		// Get timestamp
 		ts := atomic.LoadUint64(&timestamp)
 
-		if e.exp != 0 && ts >= e.exp  {
+		if e.exp != 0 && ts >= e.exp {
 			// Check if entry is expired
 			manager.delete(key)
 			// External storage saves body data with different key
--- a/middleware/cache/cache_test.go
+++ b/middleware/cache/cache_test.go
@@ -39,7 +39,8 @@ func Test_Cache_CacheControl(t *testing.T) {
 func Test_Cache_Expired(t *testing.T) {
 	app := fiber.New()
 
-	expiration := 1 * time.Second
+	// To avoid test failure we should set expiration time > 1s since the timer error in the cache middleware is 1s
+	expiration := 1*time.Second + 500*time.Millisecond
 
 	app.Use(New(Config{
 		Expiration: expiration,


defshift:
Will dive into it this evening

da4z:
Can fmt.Printf be optional?

defshift:
@da4z it's just for debugging purposes. I'll remove them after finding the root cause of the problem

defshift:
Finally have found the root cause.

There is a timer for updating current timestamp https://github.com/gofiber/fiber/blob/b517de85ec80e7873d9a1c3e61a463fbf084a5e7/middleware/cache/cache.go#L37-L42 that is used for performance optimization purposes to avoid extra sys calls.
 
Sometimes we catch a situation when the timer increases the current timestamp value right after a new cache entry is saved.

Now imagine that if the cache entry is saved with exp time `16......0001` and then timer increases current timestamp by 1 second (+ some time error) and the timestamp becomes `16......0002`. We've got a situation when entry is expired right after the saving.

Also due the timer error, the loop saves the same timestamp twice - `16......0001`, `16......0001`, because of the timer precision.

As a solution we could decrease sleep time or make it odd, but of course it does not rule out an error.
So here we could define a TTL error for this cache middleware - 1s. For example for TTL=6s cache entry could expire in 5 or 7 seconds.

Here to fix the flaky test I increased expiration time from 1s to 1.5s.

@ReneWerner87 thoughts?

ReneWerner87:
Okay, don't understand this 100% yet, but can you also set the timer to 0,5s, would the error then also go?

defshift:
@ReneWerner87 yep, it's possible and will fix test as well. But unfortunately the timer error will remain. It will be 0.5s in this case. For TTL=5s cache entry could expire in a range within 4.5s...5.5s, for TTL=1s - 0.5s...1.5s, etc.

We could reduce timer to 0.1s, to have timer error 0.1s, but not sure if we want having this timer ticking too often.

The only one way to solve this completely - to remove this timer and get current timestamp using `time.Now()` every time middleware is getting request.

ReneWerner87:
okay, will try to look at it over the weekend, need to understand the error completely, currently still can't get through the cause, because we use atomic and so there should be no race condition here, wonder if this is a real world problem or rather a problem that only occurs in the test world

ReneWerner87:
okay, now i've looked through it, you only made the change in the test,

had just another idea, maybe we should make the times of the barbarian collector configurable at some point and give over the configuration of the caching middleware in the configuration of the storages, but for this all storages would have to have an option to set the time for the garbage collector intervals

normally we would have to do that too, because people could also set a duration for the item which is smaller than the gc interval

what do you think ? @hi019 @kiyonlin 

I merge the PR now, of course, because it becomes more stable

ReneWerner87:
image
https://github.com/gofiber/fiber/runs/2321097511

@defshift the problem unfortunately still exists, can you look again if you get it corrected

defshift:
@ReneWerner87 yeah, I'll look into it over the weekend

PR #1234, opened 2021-03-20
Fix incorrect comment lines.

iredmail:
Fix incorrect comment lines in 2 files:

- middleware/csrf/config.go
- middleware/session/config.go

One question left: in csrf middleware, default value is `Strict`, but in session, its `Lax` (see [line 163-170](https://github.com/iredmail/fiber/blob/master/middleware/session/session.go#L163)). Should we set it to `Strict` too in session by default?

### DIFF
--- a/middleware/csrf/config.go
+++ b/middleware/csrf/config.go
@@ -49,7 +49,7 @@ type Config struct {
 	// Optional. Default value false.
 	CookieHTTPOnly bool
 
-	// Indicates if CSRF cookie is HTTP only.
+	// Value of SameSite cookie.
 	// Optional. Default value "Strict".
 	CookieSameSite string
 
--- a/middleware/session/config.go
+++ b/middleware/session/config.go
@@ -37,8 +37,8 @@ type Config struct {
 	// Optional. Default value false.
 	CookieHTTPOnly bool
 
-	// Indicates if CSRF cookie is HTTP only.
-	// Optional. Default value false.
+	// Value of SameSite cookie.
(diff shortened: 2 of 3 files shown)


kiyonlin:
Thanks!

> One question left: in csrf middleware, default value is Strict, but in session, its Lax (see line 163-170). Should we set it to Strict too in session by default?

We should not cuz it'll cause a breaking change.

iredmail:
Reasonable. But better set the default value to “strict” and mention this breaking change in release notes, so that all users get better security. :)

kiyonlin:
> Reasonable. But better set the default value to “strict” and mention this breaking change in release notes, so that all users get better security. :)

Then we should release fiber v3 according to [Semantic versioning ](https://semver.org/) 😂 

User can override the value if they wish. So don't worry about the different default values.

iredmail:
Understand. Let’s fix the comment lines first. :)

where should we put a TODO to remind fiber developers to change it to “strict” when preparing v3?

kiyonlin:
> where should we put a TODO to remind fiber developers to change it to “strict” when preparing v3?

This is really a good point. Could you please add this?

iredmail:
I’d like to, but where should I put this remind? A new issue? Or add a comment line in source file?

kiyonlin:
Just push another commit about the TODO stuff in this PR.

iredmail:
Added a `TODO` item in comment line, hope it's ok for you. :)

kiyonlin:
thanks

PR #1235, opened 2021-03-21
Fix session middleware

ReneWerner87:
First tests with encoding/gob instead of gotiny library


ReneWerner87:
image

Hmm, the benchmark looks really bad, we should find another package

PR #1243, opened 2021-03-26
✨ Add method to get an integer directly from the Ctx params (route params)

eacp:

I added a method to Ctx and its tests. The method is `ParamsInt`. It helps developers get an integer directly from the request

It is very common to send integers like IDs directly in the route as parameters. I added an utility function, `ParamsInt` to directly get an integer from the Ctx that is passed on each request. It returns zero and an error if the param cannot be found OR if the param is not a valid integer

**Commit formatting** 

:sparkles:

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -672,6 +672,14 @@ func (c *Ctx) Params(key string, defaultValue ...string) string {
 	return defaultString("", defaultValue)
 }
 
+// ParamsInt is used to get an integer from the route parameters
+// it defaults to zero if the parameter is not found or if the
+// parameter cannot be converted to an integer
+func (c *Ctx) ParamsInt(key string) (int, error) {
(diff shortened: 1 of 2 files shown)


ReneWerner87:
Thx for the contribution and effort

Could you add an pull request here
https://github.com/gofiber/docs/pulls

ReneWerner87 (approved):

ReneWerner87:
thanks for the implementation, 

we have internally consulted a bit and merge the pull request, but it may be that we still make optimizations and add another method for floats.

ReneWerner87:
@eacp 
Please don´t forget to add it to our documentation repository

eacp:
@ReneWerner87 I have added the documentation in [this pr](https://github.com/gofiber/docs/pull/161)

PR #1247, opened 2021-03-30
Use sliding window instead of fixed window algorithm in limiter middleware

conorbros:

Refactored the rate limiter to use a sliding window algorithm instead of fixed window. Added a new test that can demonstrate how to "cheat" the old rate limiting algorithm.  

Sliding window algorithm offers significant improvement over the fixed window algorithm and has very little overhead for this gain. Fixed window makes it possible for an API consumer to wait for a reset window and then send more requests, cheating the allowed limit.

**Commit formatting** 


ReneWerner87:
Thanks for the implemtation,
Very interesting the other more complex algorithm 

Unfortunately I am not an expert for this algorithm, I have to check it myself. 

Would it be possible that you prevent the new import by using the internal package and check if the failure of the test has to do with your adjustments.

With the additional data you have collected, would it actually be possible to make the algorithm configurable and depending on which method is used, let the user decide what kind of window he wants to use? As I have read, both offer advantages and disadvantages that the user of the middleware may want to decide for themselves.

ReneWerner87:
Thanks for the customization based on my comments, 

what do you think of the idea of configuring the windows behavior?

conorbros:
Thanks for the prompt feedback. I've fixed those failing tests by switching to the internal package.

It might be good to add configurable rate limiting algorithms but I think that the sliding window should be the default one because there is no advantage the fixed window has over it except for a negligible performance difference. 

Benchmarks with fixed window algorithm:
image

Benchmarks with sliding window algorithm:
image

Here are some resources detailing why the sliding window algorithm is the best choice for most cases:  
https://blog.cloudflare.com/counting-things-a-lot-of-different-things/
https://konghq.com/blog/how-to-design-a-scalable-rate-limiting-algorithm/
https://cloud.google.com/solutions/rate-limiting-strategies-techniques#techniques-enforcing-rate-limits

middleware/limiter/limiter.go
```diff
@@ -68,18 +67,24 @@ func New(config ...Config) fiber.Handler {
 
 		} else if ts >= e.exp {
 			// Check if entry is expired
-			e.hits = 0
+			e.prevHits = e.currHits
```
kiyonlin, line 70: `e.prevHits = e.currHits` only happens when `ts == e.exp`, otherwise `e` was GCed.

middleware/limiter/limiter.go
```diff
@@ -68,18 +67,24 @@ func New(config ...Config) fiber.Handler {
...
-		e.hits++
+		e.currHits++
 
 		// Calculate when it resets in seconds
 		expire := e.exp - ts
 
-		// Set how many hits we have left
-		remaining := cfg.Max - e.hits
+		// Calculate current rate
+		windowEnd := expiration - uint64(expire)
```
kiyonlin, line 83: windowEnd is always less then expiration

middleware/limiter/limiter.go
```diff
@@ -68,18 +67,24 @@ func New(config ...Config) fiber.Handler {
...
+		rate := uint64(e.prevHits)*(expiration-windowEnd/expiration) + uint64(e.currHits)
```
kiyonlin, line 84: rate always equals to `e.currHits`

kiyonlin (commented):
I think you should check your implementation again.

kiyonlin:
And you should take a look at [golang/time/rate](https://github.com/golang/time) if you have interesting.

middleware/limiter/limiter.go
conorbros: Fixed this by setting the garbage collection expiry to last 2 windows. There was also an issue with the expiry not being set correctly. It was being set to `time of first request in new window + expiration` when it should be 
`time window started + expiration`.

middleware/limiter/limiter.go
```diff
@@ -55,34 +55,48 @@ func New(config ...Config) fiber.Handler {
...
+		// rate = request count in previous window - weight + request count in current window
+		rate := int(float64(e.prevHits)*weight) + e.currHits
+
+		// Calculate how many hits can be made based on the current rate
+		remaining := cfg.Max - rate
 
-		// Update storage
-		manager.set(key, e, cfg.Expiration)
+		// Update storage. Garbage collect after 2 windows
+		manager.set(key, e, cfg.Expiration*2)
```
kiyonlin, line 99: I don't think it is a good idea to use 2 windows. Can't you implement this with 1 window?

conorbros: We need to store the key for the previous window and the current window at minimum.

kiyonlin:
Here is my idea about slide window algorithm: 
In order to simplify the problem, we assume max=3 and Exp=3, which means that we can request at most 3 times in 3 seconds, and we can resume a request every second. When ts=100.1, the first request came, at this time exp=ts+Exp = 103.1, hits = 1 Then we can show the results we want through the table below:
|no	|ts	|	exp		|hits|
---|---|---|---
|1	|100.1|	103.1| 	1|
|2	|100.2|	103.2| 	2|
|3	|100.3|	103.3| 	3|
|4	|100.4|	103.4| 	4(exceed max, back to 3)|
|5	|102.4|	105.1| 	2|

And explination for no.5:
1. resume 2 requests: Exp - (103.4-102.4) = 2
2. hits = 3 - 2 = 1
3. hits++

I hope my explanation is clear enough. Let me know what you think.

conorbros:
@kiyonlin I think that would be a simpler and elegant. I will have a go at implementing that this weekend. We could also get rid of the mutex and asynchronously update the memory storage which should increase throughput of the limiter.

kiyonlin:
> @kiyonlin I think that would be a simpler and elegant. I will have a go at implementing that this weekend. We could also get rid of the mutex and asynchronously update the memory storage which should increase throughput of the limiter.

This is only a POC. Please be aware of boundary of `resuming`.

And what's more, if we use slide window, should we change Retry-After Header?
 ```go
if remaining < 0 {
	// Return response with Retry-After header
	// https://tools.ietf.org/html/rfc6584
	c.Set(fiber.HeaderRetryAfter, strconv.FormatUint(expire, 10))

	// Call LimitReached handler
	return cfg.LimitReached(c)
}
```
Because in the next second we'll release one request opportunity/chance/token whatever.

conorbros:
Can you elaborate on the boundary of `resuming`? Yes we should change the retry after header. 

We should also inform the client of their current rate per second (RPS) and and the allowed RPS. I've seen numerous APIs do this so the client can rate limit themselves.

kiyonlin:
> Can you elaborate on the boundary of resuming?

I haven't thought about this problem in depth, mainly when Exp cannot divide max, like `max=30` and `Exp=100`, but it's not a big issue.

ReneWerner87:
@conorbros @kiyonlin 
what about these pull requests ? what is the latest status, what is still open ?

is actually a very cool feature, which improves this middleware enormously

conorbros:
@ReneWerner87 Thanks for the reminder, been a bit busy lately. I will try and get to it tonight.

ReneWerner87:
@conorbros any progress ?

Gusted:
Hi!

I'm gusted, a developer for userstyles.world[1][2].

We are currently happy with the current limiter middleware but noticed some people trying to attack the server by "bypassing"[6] the limiter as after each window you can simply just send the whole maximum amount of requests which was noticeable in CPU spikes[6] and performance.

I found about this PR a few weeks ago and hoped that it could be merged soon. However as it seems to be 'dead', we've decided to fork fiber[3] and implement this patch[4]. And going to deploy a version with this patch to production soon.

I'd like to report back in a few days/week to tell how this PR is working in production against real life data so it can hopefully be merged and we can remove our fork of this fiber.

I'd like to note I found 1 improvement and 1 implementation bug in this PR, please check the patches[5][6] I unfortunately don't have much time to make a comment about them and add a review to this PR for that.

Thank you for reading this and hopefully considering this to be merged soon.

1. https://userstyles.world/
2. https://github.com/userstyles-world/userstyles.world
3. https://github.com/userstyles-world/fiber/
4. https://github.com/userstyles-world/fiber/blob/master/patches/window.patch
5. https://github.com/userstyles-world/fiber/blob/master/patches/gctime.patch
6. https://github.com/userstyles-world/fiber/blob/master/patches/windowbug.patch
image
(We have set the window to have a duration of 1 minute and if it wasn't to rate-limit the spike would be higher than 30% which is already a lot to cause by 1 person, this person was targeting every 1.10 minute to take into account that the rate limiter window was not yet GC'ed from the cache)

ReneWerner87:
Thanks, we will check it next week

conorbros:
Thanks @Gusted! Glad to see this patch worked for you. I can put your bug fixes in this PR aswell.

ReneWerner87:
@conorbros yes please transfer the fixes into the pull request and refresh it, can you please provide a setting to switch between the old and the new method? find it better to give the consumer a choice

if everything fits then(documentation, tests) i will merge it quite fast

conorbros:
@ReneWerner87 What should be default?

ReneWerner87:
normally i would say the old one, but since the new implementation brings more security, i would use it in this case as default and give the consumers the possibility to switch back to the old variant if there are reasons to do so.

ReneWerner87:
image
https://en.wikipedia.org/wiki/Rate_limiting

https://nlogn.in/design-a-scalable-rate-limiting-algorithm-system-design/

since there are several possible algorithms, such a configuration is quite reasonable, maybe someone will implement another variant someday, then the middleware users can decide

Gusted:
Okay, so hi again!
Sorry for the extra delay of a week. 

I have luckily good news, as the implementation seems to have a good effect onto countering abusers.
Actually funny to see after doing a manual review of the server logs, that people who are rate limiting and where trying to get high CPU spikes actually gave up after finding out that their "method" didn't work anymore.

As for normal users the implementation doesn't seem to have any effect on them. The performance and memory usages of this new implementation seems to be in pair with the "old" implementation.

I do agree with @ReneWerner87 that having a option to switch back to the old implementation is good as users that uses some kind of standard rate limiting on public API's will most likely not have a sliding window.

Regards,
Gusted

ReneWerner87:
Thx for the feedback.

Would love to adopt the code, please just make the final adjustments and then I will merge it.

ReneWerner87:
@conorbros any updates?

Gusted:
> Would love to adopt the code, please just make the final adjustments and then I will merge it.

#1580

PR #1248, opened 2021-03-30
🐛 fix: deadlock on session

si3nloong:
The mutex lock never unlock if the subsequent code return error

This PR mainly to fix deadlock and prevent memory leak issue on `acquireCookie`, allow cookie to release back to pool if `cookie.ParseBytes` return error

### DIFF
--- a/middleware/session/session.go
+++ b/middleware/session/session.go
@@ -142,12 +142,12 @@ func (s *Session) Save() error {
 
 	// Convert data to bytes
 	mux.Lock()
+	defer mux.Unlock()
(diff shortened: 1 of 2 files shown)


ReneWerner87:
Thanks for the fix.

ReneWerner87:
We had then tried everywhere to get the maximum performance out, since the use of defer is a bit slower, that probably would not be used here and over time the code has grown, creating several cases that are of course not covered, here we should use the defer method

because the readability and the error vulnerability is then no longer in the right proportion to the performance gain.
Thx

PR #1257, opened 2021-04-01
✨ Create method to get a net.IP from the context

eacp:

Created a method in ctx to get the IP address of a request but returns a `net.IP` object instead of a string.

Currently the only way to get an IP Address from the context is to use the `ctx.IP()` method, which returns a string. The
`fasthttp`reference is NOT public, so it cannot be used outside the package. I created a method that returns the IP directly, so
the user doesn't need to parse it.

**Commit formatting** 

✨


ReneWerner87:
Thanks 

Please don't forget to add some documentation to our fiber docs repository.

kiyonlin:
> The fasthttp reference is NOT public

Do you mean `ctx.Context().RemoteIP()`?

ReneWerner87:
since the method is externally accessible and it is just a redirect of the fasthttp methodic, we(maintainers) decided not to merge this extra method

thanks for your work anyway, unfortunately we sometimes have to sort out some stuff so that our wrapper is not packed with helper functions with no added value

PR #1280, opened 2021-04-11
🚀 new possibility to escape special routing parameters

ReneWerner87:
🚀 new possibility to escape special routing parameters,
which gives the possibility to follow the google api design guide https://cloud.google.com/apis/design/custom_methods

### DIFF
--- a/path.go
+++ b/path.go
@@ -41,11 +41,12 @@ type routeSegment struct {
 
 // different special routing signs
 const (
-	wildcardParam    byte = '*' // indicates a optional greedy parameter
-	plusParam        byte = '+' // indicates a required greedy parameter
-	optionalParam    byte = '?' // concludes a parameter by name and makes it optional
-	paramStarterChar byte = ':' // start character for a parameter with name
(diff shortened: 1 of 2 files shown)


mlukasik-dev:
Oh, that's really cool! Thank you very much this feature will be very helpful :)

path.go
```diff
@@ -339,3 +360,11 @@ func GetTrimmedParam(param string) string {
 
 	return param[start:end]
 }
+
+// RemoveEscapeChar remove escape charaters
+func RemoveEscapeChar(word string) string {
+	if -1 != strings.IndexByte(word, escapeChar) {
```
mlukasik-dev, line 366: @ReneWerner87 , just asking: don't you think that `strings.IndexByte(word, escapeChar) != -1` is a bit clearer and in sync with "the best practices"?)

ReneWerner87:
gofiber/docs#165

path.go
ReneWerner87: Yes, I can do that too, I am used to the yoda notation from other programming languages, should also bring only advantages here

Issue #1284, opened 2021-04-17
🚀 ETag Middleware


fernandobandeira:
**Is your feature request related to a problem?**
Currently, there's no way to set a custom Etag header on a response, the middleware just ignores it and overrides with it's own hash.

**Why is this even a problem?**
Etags can be used for more things than just saving response bandwidth to your CDN, for example,

Assuming that you have an endpoint that returns a list of products ordered by the creation date, if this list is returning, let's say, 10k products (it's just an example), you could use the id field of the most recent product in this list as the Etag.

When the CDN sends the next request you can pick this Etag up and query in the DB to check if the most recent item has the same id, instead of querying and returning 10k products, the DB would just return 1 product, you can then check if it matches with the Etag and then return the response as Not modified without even querying the DB with the expensive query, saving up a lot of computing resources in the process.

**Describe the solution you'd like**
I think that the only thing that we have to do is to check if the response has an Etag header set and forward it to the client, we could also optionally use it to check if we should send the Not modified status, although I believe that this would already be set by the developer on the handler in this use case.

**Describe alternatives you've considered**
Currently, you can bypass the middleware if you don't have a response body or if the status is different than OK but in this case, the uncached response would have an OK status with a body, skipping the custom header that we set.

Cloning the middleware and patching it myself is always an option but this behavior would fit nicely with the framework by default so I think we could change it.

**Additional context**
If this is an interesting thing we could add this as a tip in the docs as well so people are aware of this and can use on their apps.

https://medium.com/build-succeeded/how-to-properly-use-cloudfront-to-cache-an-api-with-cache-control-and-http-304-to-provide-cache-b893f6822475

I can / would make PR's for this in this repo and on the docs but I'm not familiar with the community here yet, but I certainly want to contribute to this amazing project 🤗

kiyonlin:
Sounds interesting. I'm looking forward to your PR 😃

PR #1288, opened 2021-04-19
✨ Etag middleware support to custom etags

fernandobandeira:

Skip the middleware if an Etag response header already exists.
This change should be compatible with ```ctx.Fresh()```, you can set the custom header and call it to do the verification.

Also added a test to make sure that the If-Match and 412 responses for Put requests aren't messed up by the middleware to avoid problems with future changes.

Some things to consider:

We could export the ETag generation as a function so users could call it inside their handlers and compare it to return the 412 on their own, currently, there's no way to achieve that easily without using custom tags. However IMO returning the updatedAt as a custom Etag and using that is way more practical so we could keep it like this.

Currently, we're missing some behavior to conform with the specs:
- If-Match header doesn't do anything, it should have the reverse behavior of the If-None-Match header, however honestly I have no idea why someone would use this apart from the PUT use case that I've added in the tests.
- Both If-Match and If-None-Match headers can also accept a * (just ignore the header, ctx.Fresh() does this but the middleware doesn't) and a list of etags with the format "etag", "etag2", "etag3", again never seen this in practice but good to keep in mind

Specs links: 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Match
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match

Solves #1284

### DIFF
--- a/middleware/etag/etag.go
+++ b/middleware/etag/etag.go
@@ -39,6 +39,10 @@ func New(config ...Config) fiber.Handler {
 		if len(body) <= 0 {
 			return
 		}
+		// Skip ETag if header is already present
+		if c.Response().Header.PeekBytes(normalizedHeaderETag) != nil {
+			return
+		}
 
 		// Generate ETag for response
 		bb := bytebufferpool.Get()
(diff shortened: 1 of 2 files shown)


kiyonlin (approved):
LGTM, thanks!

PR #1290, opened 2021-04-19
Revert "🚀 new possibility to escape special routing parameters"

mlukasik-dev:
Reverts gofiber/fiber#1280

Doesn't work for such use cases:
- `/v1/users/:id\\:undelete` - very crucial use case.
- `/v1/\\:customVerb` - not really needed, but it's good to support.

Sure I can help with implementation, can start with writing tests for these 2 endpoints.

### DIFF
--- a/path.go
+++ b/path.go
@@ -41,12 +41,11 @@ type routeSegment struct {
 
 // different special routing signs
 const (
-	wildcardParam    byte = '*'  // indicates a optional greedy parameter
-	plusParam        byte = '+'  // indicates a required greedy parameter
-	optionalParam    byte = '?'  // concludes a parameter by name and makes it optional
-	paramStarterChar byte = ':'  // start character for a parameter with name
-	slashDelimiter   byte = '/'  // separator for the route, unlike the other delimiters this character at the end can be optional
-	escapeChar       byte = '\\' // escape character
+	wildcardParam    byte = '*' // indicates a optional greedy parameter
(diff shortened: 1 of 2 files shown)


mlukasik-dev:
Also I was wondering if we can support both `:param` and `{param}` syntaxes, is that possible?
That will be really great!

ReneWerner87:
we have not yet released since the last merge  ?

ReneWerner87:
> Also I was wondering if we can support both :param and {param} syntaxes, is that possible?
That will be really great!

where does this syntax come from ? it is not included in the expressjs syntax

mlukasik-dev:
> > Also I was wondering if we can support both :param and {param} syntaxes, is that possible?
> > That will be really great!
> 
> where does this syntax come from ? it is not included in the expressjs syntax

From go standard `net/http` library. People might migrate not only from expressjs, but also from other golang frameworks.

ReneWerner87:
currently it is not planned to allow several different syntaxes there on the other hand, maybe later

currently you can and should orientate to expressjs, because all other functionalities, orientate to it too

ReneWerner87:
> From go standard net/http library.

can you provide me with a link for this, can't find it in the documentation

mlukasik-dev:
> can you provide me with a link for this, can't find it in the documentation
My fault, didn't know it doesn't support path params.

ReneWerner87:
okay, i thought, you must have been watching the gorilla mux :D

PR #1299, opened 2021-04-24
🐛 Print logger's error to stderr if given io.writer failed

cep-ter:

Currently, loggger middleware doesn't do anything if it fails to write to the given io.writer. It might be a good idea to at least log something to inform the user there is a problem.

I think it might be a good idea to print the error to stderr to inform the user there is  a problem with logger (rather than doing nothing) also some logging package like logrus do the same thing(you can see the code here   [link to logrus code](https://github.com/sirupsen/logrus/blob/b50299cfaaa1bca85be76c8984070e846c7abfd2/entry.go#L295) )

### DIFF
--- a/middleware/logger/logger.go
+++ b/middleware/logger/logger.go
@@ -298,7 +298,7 @@ func New(config ...Config) fiber.Handler {
 			// Write error to output
 			if _, err := cfg.Output.Write([]byte(err.Error())); err != nil {
 				// There is something wrong with the given io.Writer
-				// TODO: What should we do here?
+				fmt.Fprintf(os.Stderr, "Failed to write to log, %v\n", err)
 			}
 		}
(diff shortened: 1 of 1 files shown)


ReneWerner87 (approved):

ReneWerner87:
thanks, looks good

cep-ter:
@ReneWerner87,  Is there any rough estimate of when this would be released(Is there any releases plan)?

ReneWerner87:
> @ReneWerner87 thank you for review, Is there any rough estimate of when this would be released(Is there any releases plan)?

i just merged it, so it should be in the next version

since we have just released, this can take a while, we usually collect some pull requests until we make the next release

maybe if we get some other important features in the next time we will release earlier

PR #1302, opened 2021-04-28
🩹Fix: Startup message doesn't show correct amount of Handlers after mounting

capthiron:
After mounting an `App` instance or `Group` to an existing instance (root), the handlerCount of the root is not being updated. I came across this issue when I noticed that the startup message showed `Handlers=0` even though my app had mounted a router with multiple handlers.

For example...
```go
router := New()
router.Get("/doe", func(c *Ctx) error {
    return c.SendStatus(StatusOK)
})

app := New()
app.Mount("/john", router)
```
... currently results in the following startup message.

image

You get the same result when mounting the `router` to a Group.

I fixed this by simply adding the handlerCount of the mounted instance to the root App's own handlerCount in the corresponding Mount-function of the App and Group struct.

**Side note**
When writing the tests I noticed that when registering a Get-Route, not *1* but **2** Handlers are being registered. This is not the case for all the other HTTP-Methods. After investigating I noticed that App.Get() not only adds the Get-Method but also the Head-Method which results in the second handler. I'm wondering whether that is this the desired behaviour or not?

Best regard, Dario ⚡

### DIFF
--- a/app.go
+++ b/app.go
@@ -423,6 +423,11 @@ func (app *App) Mount(prefix string, fiber *App) Router {
 			app.addRoute(route.Method, app.addPrefixToRoute(prefix, route))
 		}
 	}
+
+	app.mutex.Lock()
+	app.handlerCount += fiber.handlerCount
+	app.mutex.Unlock()
+
 	return app
 }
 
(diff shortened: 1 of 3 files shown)


ReneWerner87 (approved):
thanks, its looking good

PR #1310, opened 2021-05-03
🐛 Cache Middleware - Race condition in the cache expiration logic. Round 2

defshift:
After a not quite successful attempt to fix flaky tests for the cache middleware in #1229  I'd want to propose to get current timestamp every time when the middleware is executed, since it's the most stable way to expire cache in time.

Alternative way would be to not check the expiration at all and leave it up to caching the backend.

### DIFF
--- a/middleware/cache/cache.go
+++ b/middleware/cache/cache.go
@@ -5,11 +5,17 @@ package cache
 import (
 	"strconv"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/gofiber/fiber/v2"
 	"github.com/gofiber/fiber/v2/utils"
 )
+
+// timestampUpdatePeriod is the period which is used to check the cache expiration.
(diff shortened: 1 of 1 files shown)


ReneWerner87 (changes requested):
we had tested this performance boost and it has a very big impact, so building back would not be good

defshift:
got it. Ok, let me think what we can do here

defshift:
@ReneWerner87 Updated the PR. Decreased timestamp update period from 1s to 300ms. It should be enough to provide acceptable cache expiration error and it's not too overwhelming. What do you think?

Btw. Checked benchmarks. You'r right, the impact is significant.

ReneWerner87 (approved):
looks good

ReneWerner87:
hopefully the problems will disappear, thought we had to adjust the test, but a third of the time for the loop is also okay

PR #1314, opened 2021-05-05
Use atomic instead of mutex when counting the route handler.

bestgopher:
Use atomic instead of mutex when counting.

**Commit formatting** 


ReneWerner87 (approved):
Thanks

Issue #1313, opened 2021-05-03
🐛 Data race when running tests in parallel


diegocsandrim:
**Fiber version**
v2.8.0

**Issue description**

I got a data race from fiber app when I was trying to make my tests run in parallel. This only happens if the app is configured to use `Immutable: true`.

It looks like the problem happens here:

https://github.com/gofiber/fiber/blob/a48cb806f39f4717cb1326eee7f9fee96f68e2d5/app.go#L395-L397

`getBytes` and `getString` are package level function variables being changed when the `app.New` is called.

I may be able to fix that if you provide some guidance: where should we put `getBytes` and `getString`? `app.getBytes` and `app.getString` is a good place?

**Code snippet**

```go
package main

import (
	"net/http/httptest"
	"testing"

	"github.com/gofiber/fiber/v2"
)

func Test1(t *testing.T) {
	t.Parallel()
	app := fiber.New(fiber.Config{Immutable: true})
	app.Test(httptest.NewRequest("GET", "/", nil))
}

func Test2(t *testing.T) {
	t.Parallel()
	app := fiber.New(fiber.Config{Immutable: true})
	app.Test(httptest.NewRequest("GET", "/", nil))
}

```

After run `go test -race ./...` i got this:

```
==================
WARNING: DATA RACE
Write at 0x000000b11b68 by goroutine 8:
  github.com/gofiber/fiber/v2.New()
      /home/diego/go/pkg/mod/github.com/gofiber/fiber/v2@v2.8.0/app.go:396 +0x513
  bug.Test2()
      /home/diego/src/bug/bug_test.go:18 +0xa4
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1194 +0x202

Previous write at 0x000000b11b68 by goroutine 7:
  github.com/gofiber/fiber/v2.New()
      /home/diego/go/pkg/mod/github.com/gofiber/fiber/v2@v2.8.0/app.go:396 +0x513
  bug.Test1()
      /home/diego/src/bug/bug_test.go:12 +0xa4
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1194 +0x202

Goroutine 8 (running) created at:
  testing.(*T).Run()
      /usr/local/go/src/testing/testing.go:1239 +0x5d7
  testing.runTests.func1()
      /usr/local/go/src/testing/testing.go:1512 +0xa6
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1194 +0x202
[...]

PR #1316, opened 2021-05-05
Fix app test parallel

diegocsandrim:
## The problem

`fiber.New()` changes the package variables `getBytes` and `getString`, and these are used a lot across the app. This may cause data race problems, as mentioned in the issue #1313, as well as other problems like: when using 2 apps, a second of fiber may override these functions, causing the first app to use incorrect configurations.

## The solution
In order to avoid sharing app configuration between two created apps, I have moved the `getBytes` and `getString` from the package to the app struct, and fixed the usages of these functions.

## Detailed changes

- Create the test to show the problem
- Deletes `getBytes` and `getString` function: they should not be package variables.
- Adds `getBytes` and `getString` to the `App` struct
- Default `getBytes` and `getString` uses the same defaults as before: `utils.GetBytes` and `utils.GetString`
- When config Immutable is true, it use `getBytesImmutable` and `getStringImmutable`
- Changed a lot of usages of `getBytes` and `getString` to `app.getBytes` and `app.getString` in files `ctx.go`, `router.go`, `ctx_test.go`, `router_test.go` and`app_test.go` .
- `client.go` was changed to always use the `utils.UnsafeBytes`. I think this change is safe, because the function the client uses should not be change by a call to `app.New()`.
- Add the App as a receiver for `quoteString` and `isEtagStale`: `func (app *App) ...`, so they can use the functions defined in the App configuration
- `Benchmark_Utils_Unescape` now uses the default functions: `utils.UnsafeBytes` and `utils.UnsafeString`

## Commits

I made two commits: one adding a test to show the bug happening, and one to fix it.

- 24f55ef - The test: You can run it using `go test -race -run Test_App_New_Test_Parallel` to see the error happening.
- 0dc67e8 - The fix.

### DIFF
--- a/app.go
+++ b/app.go
@@ -105,6 +105,10 @@ type App struct {
 	server *fasthttp.Server
 	// App config
 	config Config
+	// Converts string to a byte slice
+	getBytes func(s string) (b []byte)
+	// Converts byte slice to a string
+	getString func(b []byte) string
 }
 
 // Config is a struct holding the server settings.
@@ -364,7 +368,9 @@ func New(config ...Config) *App {
 			},
 		},
 		// Create config
-		config: Config{},
+		config:    Config{},
+		getBytes:  utils.GetBytes,
+		getString: utils.GetString,
 	}
 	// Override config if provided
 	if len(config) > 0 {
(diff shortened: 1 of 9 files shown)


Bader5110:
F

kiyonlin:
Thanks a lot for your contributing!

kiyonlin (approved):

ReneWerner87 (approved):

PR #1318, opened 2021-05-06
🐞 fix: reduce coupling by removing auto registering Head method on App.Get() call

capthiron:
Hey Pals 🤓 

The declaration of a Head-endpoint should be explicit to achieve a lower coupling and leaner performance for Fiber. Currently Fiber registers an additional Head handler when setting up a Get-endpoint (see code snippet).

```go
// Get registers a route for GET methods that requests a representation
// of the specified resource. Requests using GET should only retrieve data.
func (app *App) Get(path string, handlers ...Handler) Router {
	return app.Add(MethodHead, path, handlers...).Add(MethodGet, path, handlers...)
}
```

The current state can be misleading and redundant since it registers extra handlers unknowingly. The implicit handler therefore ought to be removed. When a Head-handler is needed it should be defined via the designated Head-method. See `Side note` in [PR-1302](https://github.com/gofiber/fiber/pull/1302) for further information.

I committed and tested the changes to be applied.

Best regard, Dario ⚡

### DIFF
--- a/app.go
+++ b/app.go
@@ -465,7 +465,7 @@ func (app *App) Use(args ...interface{}) Router {
 // Get registers a route for GET methods that requests a representation
 // of the specified resource. Requests using GET should only retrieve data.
 func (app *App) Get(path string, handlers ...Handler) Router {
-	return app.Add(MethodHead, path, handlers...).Add(MethodGet, path, handlers...)
+	return app.Add(MethodGet, path, handlers...)
 }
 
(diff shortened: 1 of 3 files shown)


kiyonlin:
Hey, thanks for your work! But this is an expected behavior. Some of the web frameworks add HEAD route while adding GET route, and Fiber also does so.

If you really don't need the HEAD route, we may add a configuration option to control it. But your current implementation will cause a breaking change.

PR #1341, opened 2021-05-19
🔥 Adding user context

### LINKED ISSUE #1249: 🤗 Context Propagation
We are currently implementing OpenTelemetry in a huge application. Fiber is our last missing part :)

OpenTelemetry and similar frameworks make use of go's Context propagation. Basically you do something like this:

```
func(ctx context.Context) {
		var span trace.Span
		ctx, span = tracer.Start(ctx, "operation")
		defer span.End()
		
		// continue with this ctx
		// callMyDb(ctx, xxx)
```

invoked sub-systems (think about databases, gprc clients etc), do the same over again and extend the Context with more and more info - or even serialize it to other microservces where this continues.

Now with Fiber v2, I do not know how to overwrite the Context in a middleware. It seems this was somehow possible in version v1 . Maybe also the adapter middleware is doing something like that.

I am are that you want to keep fiber simple - and that's great.

Maybe it's an idea to use your locals or likewise feature:
-> setting a context as local with a special key
-> c.Context() would return this instead the fasthttpctx

Next to tracing, this is of course also an essential feature for distributed logging.

Certainly, we could make a middleware to store the context in a local; and then use a small helper to extract it again and continue with that in functions and sub-systems.

But maybe... you also like the idea or have suggestions in a similar direction.

patil-kshitij:
Resolved #1249 

This adds an additional field called `userContext` which is of type` context.Context`.
It could be set by user.

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -320,7 +320,7 @@ func (c *Ctx) UserContext() context.Context {
 	return c.userContext
 }
 
-//SetUserContext sets a context implementation by user.
+// SetUserContext sets a context implementation by user.
 func (c *Ctx) SetUserContext(ctx context.Context) {
 	c.userContext = ctx
 }
--- a/ctx.go
+++ b/ctx.go
@@ -311,8 +311,8 @@ func (c *Ctx) Context() *fasthttp.RequestCtx {
 	return c.fasthttp
 }
 
-//UserContext returns a context implementation that was set by
-//user earlier or returns a non-nil, empty context, it was not set earlier.
+// UserContext returns a context implementation that was set by
+// user earlier or returns a non-nil, empty context, it was not set earlier.
 func (c *Ctx) UserContext() context.Context {
 	if c.userContext == nil {
 		c.userContext = context.Background()
--- a/ctx.go
+++ b/ctx.go
@@ -312,7 +312,7 @@ func (c *Ctx) Context() *fasthttp.RequestCtx {
 }
 
 // UserContext returns a context implementation that was set by
-// user earlier or returns a non-nil, empty context, it was not set earlier.
+// user earlier or returns a non-nil, empty context,if it was not set earlier.
 func (c *Ctx) UserContext() context.Context {
 	if c.userContext == nil {
 		c.userContext = context.Background()
--- a/ctx.go
+++ b/ctx.go
@@ -6,6 +6,7 @@ package fiber
 
 import (
 	"bytes"
+	"context"
 	"encoding/xml"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


liaohongxing:
1. expressjs styles

```
c.UserContext(context.Background)   // set Context

c.UserContext()   // get Context

```
2. why

```
c.Locals("c", context.Background)   // setContext
c.Locals("c")  // getContext
```

patil-kshitij:
1. Should I write setting and getting context in expressjs styles ?

2. We need context propagation, where Go defined context interface's implementation could be set by user.

I could implement point 1.
Please let me know if you are okay with point2.

liaohongxing:
This is just my suggestion, listen to the suggestions of other mainline maintainers

ReneWerner87:
@patil-kshitij please add something in the README for this or/and the docs repository

only one known feature is used 😄

ctx.go
```diff
@@ -309,6 +311,17 @@ func (c *Ctx) Context() *fasthttp.RequestCtx {
 	return c.fasthttp
 }
 
+func (c *Ctx) UserContext() context.Context {
```
ReneWerner87, line 314: add a method docblock with description 

we receive a lot of support requests and i think this could eliminate some requests from the start

ctx.go
```diff
@@ -309,6 +311,17 @@ func (c *Ctx) Context() *fasthttp.RequestCtx {
...
+	if c.userContext == nil {
+		c.userContext = context.Background()
+	}
+	return c.userContext
+}
+
+func (c *Ctx) SetUserContext(ctx context.Context) {
```
ReneWerner87, line 321: add a method docblock with description

we receive a lot of support requests and i think this could eliminate some requests from the start

ctx.go
patil-kshitij: Done, 
Kindly review

ctx.go
patil-kshitij: Done, 
Kindly review

hi019: Need a space after the last `/`

patil-kshitij: Done
Kindly review

patil-kshitij:
@ReneWerner87, should I add a parallel PR on [https://github.com/gofiber/docs] to add documentation regarding setting and getting user-context, which could be used for observability purposes?

ReneWerner87:
Yes please add a PR in the docs repository.

patil-kshitij:
Hi @ReneWerner87 ,
Have raised PR for documentation as well.
Kindly review, https://github.com/gofiber/docs/pull/173

ReneWerner87:
Will check it in a few hours

ReneWerner87 (approved):

ctx.go
patil-kshitij: Hi @hi019,
Added space after last /.
Please review

patil-kshitij:
Hi @ReneWerner87, @hi019 
Can we please merge this PR is review is completed ?

hi019 (approved):

ReneWerner87:
@patil-kshitij could you check https://github.com/gofiber/fiber/pull/1522/files

PR #1343, opened 2021-05-20
🔥 Feature: Add encrypt cookies middleware

amir9480:
## Description
This PR adds a new middleware to encrypt cookie values to increase security
([Similar middleware in Laravel](https://github.com/laravel/framework/blob/8.x/src/Illuminate/Cookie/Middleware/EncryptCookies.php))

Middleware decrypts request cookies and encrypts response cookies.
If a user manually changes the cookie value to something invalid cookie value will convert to `""`.

## Usage
```go
app := fiber.New()

app.Use(encryptcookie.New())

	c.Cookie(&fiber.Cookie{
		Name:  "SomeThing",
		Value: "A Secret Value Here",
	})
	// ...
})
```

The middleware uses `AES-256-GCM` by default, and generates a new encryption key on each run but we should always use a secret encryption key like this:
```go
app.Use(encryptcookie.New(
    encryptcookie.Config{
        Key: "QjEyRTBWcU1NTGpzT0xCRXd2SUN3VFZBOUVoRzJzWUs=" // A secret base64 encoded key
    }
))
```

Also, it is possible to use a custom function to encrypt and decrypt cookies, So developers can use their own encryption algorithms:
```go
app.Use(encryptcookie.New(encryptcookie.Config{
	Encryptor: func(decryptedString, _ string) (string, error) {
		return base64.StdEncoding.EncodeToString([]byte(decryptedString)), nil // Not safe, just a sample
	},
	Decryptor: func(encryptedString, _ string) (string, error) {
		decodedBytes, err := base64.StdEncoding.DecodeString(encryptedString)  // Not safe, just a sample
		return string(decodedBytes), err
	},
}))
```

We can also exclude specific cookies from being encrypted like cookies using for Google Auth:
```go
app.Use(encryptcookie.New(encryptcookie.Config{
	Except: []string{
		"g_csrf_token", // This cookie using for Google One Tap and should not be encrypted
	},
}))
```

(Sorry for not providing documentation because I'm not fluent in English and writing documents.)

### DIFF
--- a/middleware/encryptcookie/README.md
+++ b/middleware/encryptcookie/README.md
@@ -1,5 +1,66 @@
 # Encrypt Cookie Middleware
 
-Encrypt middleware for [Fiber](https://github.com/gofiber/fiber)
+Encrypt middleware for [Fiber](https://github.com/gofiber/fiber) which encrypts cookie values. Note: this middleware does not encrypt cookie names.
 
-TODO
+
+## Signaures
+
+```go
+func New(config ...Config) fiber.Handler
+
+func GenerateKey(length int) string
+```
+
+## Setup
+
+First import the middleware from Fiber,
+
+import (
+  "github.com/gofiber/fiber/v2"
+  "github.com/gofiber/fiber/v2/middleware/cache"
+)
+
+Then create a Fiber app with `app := fiber.New()`.
+
+## Minimum Config
+
+```go
+// `Key` must be a 32 character string. It's used to encrpyt the values, so make sure it is random and keep it secret. 
+// You can call `encryptcookie.GenerateKey()` to create a random key for you. 
+// Make sure not to set `Key` to `encryptcookie.GenerateKey()` because that will create a new key every run.
+app.Use(encryptcookie.New(encryptcookie.Config{
+    Key: "secret-thirty-2-character-string",
+}))
+```
+
+## Config
+
+```go
+type Config struct {
+	// Next defines a function to skip this middleware when returned true.
+	//
+	// Optional. Default: nil
+	Next func(c *fiber.Ctx) bool
+
+	// Array of cookie keys that should not be encrypted.
+	//
+	// Optional. Default: []
+	Except []string
+
+	// Base64 encoded unique key to encode & decode cookies.
+	//
+	// Required. Key length should be 32 characters.
+	// You may use `encryptcookie.GenerateKey(32)` to generate a new key.
+	Key string
+
+	// Custom function to encrypt cookies.
+	//
+	// Optional. Default: EncryptCookie
+	Encryptor func(decryptedString, key string) (string, error)
+
+	// Custom function to decrypt cookies.
+	//
+	// Optional. Default: DecryptCookie
+	Decryptor func(encryptedString, key string) (string, error)
+}
+```
--- a/middleware/encryptcookie/README.md
+++ b/middleware/encryptcookie/README.md
@@ -25,8 +25,8 @@ Then create a Fiber app with `app := fiber.New()`.
 ## Minimum Config
 
 ```go
-// `Key` must be a 32 character string. It's used to encrpyt the values, so make sure it is random and keep it secret. 
-// You can call `encryptcookie.GenerateKey()` to create a random key for you. 
+// `Key` must be a 32 character string. It's used to encrpyt the values, so make sure it is random and keep it secret.
+// You can call `encryptcookie.GenerateKey()` to create a random key for you.
 // Make sure not to set `Key` to `encryptcookie.GenerateKey()` because that will create a new key every run.
 app.Use(encryptcookie.New(encryptcookie.Config{
     Key: "secret-thirty-2-character-string",
@@ -50,7 +50,7 @@ type Config struct {
 	// Base64 encoded unique key to encode & decode cookies.
 	//
 	// Required. Key length should be 32 characters.
-	// You may use `encryptcookie.GenerateKey(32)` to generate a new key.
+	// You may use `encryptcookie.GenerateKey()` to generate a new key.
 	Key string
 
 	// Custom function to encrypt cookies.
--- a/middleware/encryptcookie/README.md
+++ b/middleware/encryptcookie/README.md
@@ -6,9 +6,11 @@ Encrypt middleware for [Fiber](https://github.com/gofiber/fiber) which encrypts 
 ## Signaures
 
 ```go
+// Intitializes the middleware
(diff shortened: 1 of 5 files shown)


Fenny:
LGTM, I might have one suggestion
> The middleware uses AES-256-GCM by default, and generates a new encryption key on each run but we should always use a secret encryption key like this:

Maybe we should default to a unique identifier related to the machine, so it won't change the encryption key on restarts. This will cause previous set cookie values to be invalid on server restarts/crashes. Thoughts?

amir9480:
> LGTM @amir9480, I might have one suggestion
> 
> > The middleware uses AES-256-GCM by default, and generates a new encryption key on each run but we should always use a secret encryption key like this:
> 
> Maybe we should default to a unique identifier related to the machine, so it won't change the encryption key on restarts. This will cause previous set cookie values to be invalid on server restarts/crashes. (Only if no config is provided) Thoughts?

Sure, But which parameter should be used which is unique & also private & no accessable by others?

amir9480:
@Fenny

Any idea?
What about making key a required parameter like `gofiber/jwt`

https://github.com/gofiber/jwt/blob/master/jwt.go#L92

hi019:
> What about making key a required parameter like `gofiber/jwt`

I think this is our best option. If we autogenerate the key statelessly, users will be confused because cookies from the previous application version (especially apparent during development where application restarts are frequent) will no longer be valid.

middleware/encryptcookie/config.go
```diff
@@ -0,0 +1,75 @@
...
+import "github.com/gofiber/fiber/v2"
+
+// Config defines the config for middleware.
+type Config struct {
+	// Next defines a function to skip this middleware when returned true.
+	//
+	// Optional. Default: nil
+	Next func(c *fiber.Ctx) bool
+
+	// Array of cookies that should not encrypt
```
hi019, line 12:
```suggestion
	// Array of cookie keys that should not be encrypted
```

amir9480:
@hi019 
`Key` is a required parameter now.
I also improved the key generation function

hi019:
Will review and test soon

middleware/encryptcookie/config.go
```diff
@@ -0,0 +1,75 @@
...
+	Next func(c *fiber.Ctx) bool
+
+	// Array of cookie keys that should not be encrypted
+	//
+	// Optional. Default: []
+	Except []string
+
+	// Base64 unique key to encode & decode cookies
+	//
+	// Optional. Default: Generating new key on every run
```
hi019, line 19:
```suggestion
```

middleware/encryptcookie/config.go
```diff
@@ -0,0 +1,75 @@
...
+	Key string
```
hi019, line 20: We should add a doc block and maybe a runtime check regarding how this value needs to be at least 32 chars. Otherwise the code will panic on cookie creation

middleware/encryptcookie/config.go
```diff
@@ -0,0 +1,75 @@
...
+			cfg.Encryptor = ConfigDefault.Encryptor
+		}
+
+		if cfg.Decryptor == nil {
+			cfg.Decryptor = ConfigDefault.Decryptor
+		}
+	}
+
+	if cfg.Key == "" {
+		panic("Fiber: Encrypt cookie middleware requires key")
```
hi019, line 71:
```suggestion
		panic("fiber: encrypt cookie middleware requires key")
```

middleware/encryptcookie/encryptcookie.go
```diff
@@ -0,0 +1,56 @@
...
+			return c.Next()
+		}
+
+		// Decrypt request cookies
+		c.Request().Header.VisitAllCookie(func(key, value []byte) {
+			keyString := string(key)
+			if !isDisabled(keyString, cfg.Except) {
+				decryptedValue, err := cfg.Decryptor(string(value), cfg.Key)
+				if err != nil {
+					c.Request().Header.SetCookie(string(key), "")
```
hi019, line 26: Using `c.Request().Header.SetCookieBytesKV(key, nil)` here is probably faster

middleware/encryptcookie/README.md
```diff
@@ -0,0 +1,5 @@
+# Encrypt Cookie Middleware
+
+Encrypt middleware for [Fiber](https://github.com/gofiber/fiber)
+
+TODO
```
hi019, line 5: TODO for when I write these docs: cookie name is not encrypted

hi019 (changes requested):

middleware/encryptcookie/config.go
amir9480: I'm not sure about runtime checks because of `aes.NewCipher` checks for key lengths.
Also, it's possible to use a key with 16/24 characters to use `AES-128`/`AES-192`.

middleware/encryptcookie/README.md
tomaswarynyca: @amir9480 please remove TODO

hi019:
Thanks for the changes, I'll look at them and write the docs later today.

middleware/encryptcookie/config.go
```diff
@@ -0,0 +1,76 @@
...
+
+	// Array of cookie keys that should not be encrypted.
+	//
+	// Optional. Default: []
+	Except []string
+
+	// Base64 encoded unique key to encode & decode cookies.
+	//
+	// Required. Key length should be 32 characters.
+	// you may use `encryptcookie.GenerateKey(32)` to generate a new key.
```
hi019, line 20:
```suggestion
	// You may use `encryptcookie.GenerateKey(32)` to generate a new key.
```

middleware/encryptcookie/utils.go
```diff
@@ -0,0 +1,88 @@
...
+	plaintext, err := gcm.Open(nil, nonce, ciphertext, nil)
+	if err != nil {
+		return "", err
+	}
+
+	return string(plaintext), nil
+}
+
+// GenerateKey Generates an encryption key
+func GenerateKey(length int) string {
```
hi019, line 69: This probably doesn't need to take a parameter, since the length should always be 32

hi019 (approved):

ReneWerner87:
I will check it tonight

efectn:
Nice feature

ReneWerner87 (approved):

ReneWerner87:
@amir9480 thanks for your work, sorry for the long wait, can you add some documentation in the docs https://github.com/gofiber/docs

PR #1368, opened 2021-06-06
🔥 Feature: add cache header to show cache status

WLun001:
Cache middleware to returns `S-Cache` header to show cache status, `hit`, `miss` or `unreachable`

Not using `X-Cache` because I think some CDN uses this value?

### DIFF
--- a/middleware/cache/README.md
+++ b/middleware/cache/README.md
@@ -65,6 +65,13 @@ type Config struct {
 	// Optional. Default: 1 * time.Minute
 	Expiration time.Duration
 
+	// CacheHeader header on response header, indicate cache status, with the following possible return value
+	//
+	// hit, miss, unreachable
+	//
+	// Optional. Default: S-Cache
+	CacheHeader string
+
 	// CacheControl enables client side caching if set to true
 	//
 	// Optional. Default: false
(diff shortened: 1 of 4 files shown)


middleware/cache/README.md
```diff
@@ -91,6 +98,7 @@ type Config struct {
 var ConfigDefault = Config{
 	Next:         nil,
 	Expiration:   1 * time.Minute,
+	CacheHeader:  "S-Cache",
```
hi019, line 101: The default header should be X-Cache, it's used by many other CDN's and X indicates a non standard header

hi019 (changes requested):

hi019:
Thanks for the PR, can you also add tests?

ReneWerner87 (approved):

middleware/cache/cache_test.go
```diff
@@ -299,6 +299,61 @@ func Test_CustomKey(t *testing.T) {
...
+		return c.SendString(c.Query("cache"))
+	})
+
+	app.Get("/error", func(c *fiber.Ctx) error {
+		return c.Status(fiber.StatusInternalServerError).SendString(time.Now().String())
+	})
+
+	resp, err := app.Test(httptest.NewRequest("GET", "/", nil))
+	utils.AssertEqual(t, nil, err)
+	utils.AssertEqual(t, "miss", resp.Header.Get("X-Cache"))
```
hi019, line 326: Should be able to use the cache status constants in these tests

PR #1381, opened 2021-06-12
Improve errors handling

Elias506:
**Hi!**

I changed all places, where we using `fmt.Errorf` to `errors.New`, because it work much faster without any allocations.
Method `Errorf` needed, when we, for example, pull out the type of variable, or use `Unwrap` method 

Снимок экрана 2021-06-12 в 01 49 13

Снимок экрана 2021-06-12 в 01 49 31

So the same situation with `fmt.Sprintf` and `strings.Join`. The last one is better to use: 

Снимок экрана 2021-06-12 в 13 28 19

Here `d` is a global int variable 

Снимок экрана 2021-06-12 в 13 29 09

We have 1 allocation with method `Join` because of using int to string conversation 

There are actually many places where we can do this, but it will impair the readability of the code like this:
```Go
   - return uuid, fmt.Errorf("invalid urn prefix: %q", s[:9])		
   
   + return uuid, errors.New(`invalid urn prefix: "` + s[:9] + `"`)
```

 Perhaps error handling is not worth the quickest solution.

### DIFF
--- a/internal/uuid/dce.go
+++ b/internal/uuid/dce.go
@@ -6,8 +6,9 @@ package uuid
 
 import (
 	"encoding/binary"
-	"fmt"
 	"os"
+	"strconv"
+	"strings"
 )
 
 // A Domain represents a Version 2 domain
@@ -76,5 +77,5 @@ func (d Domain) String() string {
 	case Org:
 		return "Org"
 	}
-	return fmt.Sprintf("Domain%d", int(d))
+	return strings.Join([]string{"Domain", strconv.Itoa(int(d))}, "")
 }
--- a/internal/uuid/marshal.go
+++ b/internal/uuid/marshal.go
@@ -4,7 +4,10 @@
 
 package uuid
 
-import "fmt"
+import (
+	"errors"
+	"strconv"
+)
 
 // MarshalText implements encoding.TextMarshaler.
 func (uuid UUID) MarshalText() ([]byte, error) {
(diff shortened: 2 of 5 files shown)


internal/uuid/dce.go
```diff
@@ -76,5 +77,5 @@ func (d Domain) String() string {
 	case Org:
 		return "Org"
 	}
-	return fmt.Sprintf("Domain%d", int(d))
+	return strings.Join([]string{"Domain", strconv.Itoa(int(d))}, "")
```
ReneWerner87, line 80: Benchmark_Error/fmt.Sprintf("Domain%d",_int(d))-12         	
13134072	        
88.10 ns/op

Benchmark_Error/strings.Join([]string{"Domain",_strconv.Itoa(int(d))},_"")-12         	
35117198	        
43.97 ns/op

internal/uuid/marshal.go
```diff
@@ -31,7 +34,7 @@ func (uuid UUID) MarshalBinary() ([]byte, error) {
 // UnmarshalBinary implements encoding.BinaryUnmarshaler.
 func (uuid *UUID) UnmarshalBinary(data []byte) error {
 	if len(data) != 16 {
-		return fmt.Errorf("invalid UUID (got %d bytes)", len(data))
+		return errors.New("invalid UUID (got " + strconv.Itoa(len(data)) + " bytes)")
```
ReneWerner87, line 37: Benchmark_Error/fmt.Errorf("invalid_UUID_(got_%d_bytes)",_len(data))-12         	 
7709733	       
132.9 ns/op

Benchmark_Error/errors.New("invalid_UUID_(got_"_+_strconv.Itoa(len(data))_+_"_bytes)")-12         	
16772258	        
75.60 ns/op

internal/uuid/sql.go
```diff
@@ -26,7 +27,7 @@ func (uuid *UUID) Scan(src interface{}) error {
 		// see Parse for required string format
 		u, err := Parse(src)
 		if err != nil {
-			return fmt.Errorf("Scan: %v", err)
+			return errors.New("Scan: " + err.Error())
```
ReneWerner87, line 30: Benchmark_Error/fmt.Errorf("Scan:_%v",_err)-12         	 
5615924	       
188.9 ns/op

Benchmark_Error/errors.New("Scan:_"_+_err.Error())-12  	
18481023	        
64.12 ns/op

ReneWerner87 (approved):

PR #1386, opened 2021-06-16
Improved some conditions 

Ja7ad:


utils/http.go
```diff
@@ -8,15 +8,15 @@ const MIMEOctetStream = "application/octet-stream"
 
 // GetMIME returns the content-type of a file extension
 func GetMIME(extension string) (mime string) {
-	if len(extension) == 0 {
+	if extension == "" {
```
ReneWerner87, line 11:
```go
Benchmark_Compare/""_==_""-12         	1000000000	         0.4273 ns/op
Benchmark_Compare/"test"_==_""-12     	1000000000	         0.3923 ns/op
Benchmark_Compare/len("")_==_0-12     	1000000000	         0.3785 ns/op
Benchmark_Compare/len("test")_==_0-12 	1000000000	         0.3814 ns/op
```
len(x) == 0 is faster than x == ""

Ja7ad: Commit https://github.com/gofiber/fiber/pull/1386/commits/1652a85638632ad3733f72b2e7d11b7bcae80677 added for back to condition len(x) == 0

ReneWerner87 (approved):

Issue #1375, opened 2021-06-09
🤗  How to handle errors from internal package?


jorgii:
**Question description**
The internal package declares `schema.MultiError` and `schema.ConversionError`. These are returned from ctx.QueryParser or ctx.BodyParser when the input is not in the expected type (for example providing `foo` when expecting an ISO 8601 datetime string). I would like to customise the error message for such errors. Normally I would import the error type and based on it would decide the error message. Since these error types are in the internal package I cannot import them. Is there another way of customising the error messages?

**Code snippet**

```go
package main

import (
    "log"
    "time"

    "github.com/gofiber/fiber/v2"
)

type MyFilter struct {
    CreatedAt time.Time `query:"created_at"`
}

func hello(ctx *fiber.Ctx) error {
    var filter MyFilter
    if err := ctx.QueryParser(&filter); err != nil {
        // handle err
        panic(err)
    }
    return ctx.SendString("Hello, World!")
}

func main() {
    app := fiber.New()
    app.Get("/hello", hello)
    log.Fatalln(app.Listen(":8085"))
}
```
**Request**

```bash
curl --location --request GET 'http://localhost:8085/hello?created_at=foo' 
```

ReneWerner87:
thank you for the report,
you are absolutely right, we should move the errors from the internal package and make them available

can you do that and create a pull request for us, then i will review this?

GibranHL0:
Is @jorgii working on this issue? 🤔 If not, could I give it a try?

jorgii:
@GibranHL0 I was planning to do it around the end of the week. If you can start on it sooner please go ahead.

GibranHL0:
@jorgii I got assigned another issue, I'll focus on the other one.

PR #1387, opened 2021-06-16
Expose internal errors.

jorgii:

This pull request solves exposes the internal errors so that they can be used publicly.

The problem is described in https://github.com/gofiber/fiber/issues/1375.

**Commit formatting** 

### DIFF
--- a/error.go
+++ b/error.go
@@ -1,0 +1,14 @@
+package fiber
+
+import "github.com/gofiber/fiber/v2/internal/schema"
+
+type (
+	// Conversion error exposes the internal schema.ConversionError for public use.
+	ConversionError = schema.ConversionError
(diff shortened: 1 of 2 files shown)


ReneWerner87 (approved):

jorgii:
This is the easy version of the fix, aliasing the internal errors. This brings the potential problem where someone working on the internal error might not realise that they are exposed to the public.
I am pushing this as a start.
This could be improved by moving all the error definitions outside so that it is explicit they are public.

Does that make sense to do?

ReneWerner87:
yes it makes sense, but because this internal code is a copy, it will be difficult to do so

ReneWerner87:
but when we refresh the code we copied, this customization flies out again, so i think the current solution is good

jorgii:
Alright, making the PR ready for merge.

Issue #1401, opened 2021-06-23
🚀 Add automatic decompression for gzip, deflate, brotli


sneakykiwi:
**Is your feature request related to a problem?**
Partly yes, I have an API that I rent out to my clients, the api's core function requires uploading a really big json, one of my clients came up with the idea of compressing the payload to reduce upload latency. He used the `Content-Encoding: gzip` headers, to my surprise, the payload didn't get uncompressed in the default fiber context body though.
**Describe the solution you'd like**
Have automatic compression detection and use that to decompress the body of the request
**Describe alternatives you've considered**
Turns out fasthttp has in fact an implementation for that, and for other compressions as well. I was able to write a simple switch statement and just do `ctx.Request().BodyGunzip()` and retrieve the uncompressed body
**Additional context**
Would make it a really cool feature!

PR #1402, opened 2021-06-23
✨ Added decompression cases for Body function

sneakykiwi:

According to this issue I created a bit earlier #1401, I have implemented a filter to return the correct body each time the `Body` function of the fiber.Ctx struct is called

Fasthttp provides by default methods for decompressing different request payloads, currently, gofiber directly returns the raw body, if the body is compressed using gzip/deflate/brotli then it gets returned like that and the users has to decompress it manually. This PR fixes this issue

- Added a switch for the encoding to be able to determine what type it is and how to decompress it. Usage remains the same nonetheless.
- Updated tests and added new benchmark for updates
	
I have also updated the tests accordingly


ReneWerner87 (approved):

ReneWerner87 (approved):

tomaswarynyca (approved):

hi019 (approved):

navossoc:
I saw this update yesterday when I updated my project....

> // Body contains the **raw body** submitted in a POST request.
// Returned value is only valid within the handler. Do not store any references.
// Make copies or use the Immutable setting instead.

I think this is no longer true then...
Also the len(c.Body()) is not the `Content-Length` anymore.

This also broken for who doesn't want to deal with the uncompressed data.
Let's say I just want do receive it and store it.

Anyway, just mentioning a few points that come to my mind.

[]'s

PR #1407, opened 2021-06-24
Fix: Session middleware issues

### LINKED ISSUE #1395: sess.Regenerate() issue 🐛
**Fiber version**
v2.13.0

**Issue description**
I want to write a `refresh` api for regenerating a session but it doesn't regenerate session every second call.
When call `refresh` in first time the `fresh` of sess is true and I have a session. In second time the `fresh` of sess is false and `sess.Save()` doesn't save a new session. In third time the `fresh` of sess is true and I have a session.
I think issue is `sess.Regenerate()` does't set `s.fresh = true`.

**Code snippet**

```go
package main

import (
	"fmt"
	"github.com/gofiber/fiber/v2"
	"github.com/gofiber/fiber/v2/middleware/session"
	"log"
)

var store = session.New()

func main() {
	app := fiber.New()

	app.Get("/", func(ctx *fiber.Ctx) error {
		sess, err := store.Get(ctx)
		if err != nil {
			return err
		}
		name := sess.Get("name")
		str := fmt.Sprintf("%v", name)
		return ctx.SendString(str)
	})
	app.Get("/refresh", func(ctx *fiber.Ctx) error {
		sess, err := store.Get(ctx)
		if err != nil {
			return err
		}

		fmt.Println(sess.ID(), sess.Fresh())
		if err = sess.Regenerate(); err != nil {
			return err
		}

		sess.Set("name", "bob")

		fmt.Println(sess.ID(), sess.Fresh())
		if err = sess.Save(); err != nil {
			return err
		}

		fmt.Println("The session is refreshed")
		return ctx.SendString("The session is refreshed")
	})
	log.Fatal(app.Listen(":8080"))
}

```

Spedoske:
Fix: Session.Regenerate does not set Session.fresh to be true.
Fix: Session should be regenerated if the session can not be found in the storage.
This PR is intended to close #1395 issue and #1408 issue.

Tests have been added.
Some old tests have been modified.
If there are some questions, I will answer them.

### DIFF
--- a/middleware/session/session.go
+++ b/middleware/session/session.go
@@ -120,7 +120,10 @@ func (s *Session) Regenerate() error {
 
 	// Create new ID
 	s.id = s.config.KeyGenerator()
-
+	
+	// We assign a new ID to the session, so the session must be fresh
+	s.fresh = true
+	
 	return nil
 }
 
--- a/middleware/session/session.go
+++ b/middleware/session/session.go
@@ -118,13 +118,19 @@ func (s *Session) Regenerate() error {
 		return err
 	}
 
-	// Create new ID
-	s.id = s.config.KeyGenerator()
-	
-	// We assign a new ID to the session, so the session must be fresh
-	s.fresh = true
-	
+	// generate a new session, and set session.fresh to be true
+	s.refresh()
+
 	return nil
+}
+
+// refresh generates a new session, and set session.fresh to be true
+func (s *Session) refresh() {
+	// Create a new id
+	s.id = s.config.KeyGenerator()
+
+	// We assign a new id to the session, so the session must be fresh
+	s.fresh = true
 }
 
 // Save will update the storage and client cookie
--- a/middleware/session/session.go
+++ b/middleware/session/session.go
@@ -118,7 +118,7 @@ func (s *Session) Regenerate() error {
 		return err
 	}
 
-	// generate a new session, and set session.fresh to be true
+	// Generate a new session, and set session.fresh to true
 	s.refresh()
 
 	return nil
--- a/middleware/session/store.go
+++ b/middleware/session/store.go
@@ -79,7 +79,11 @@ func (s *Store) Get(c *fiber.Ctx) (*Session, error) {
 		} else if err != nil {
 			return nil, err
 		} else {
+			// raw is nil, which means id is not in the storage
+			// so it means that id is not valid (mainly because of id is expired or user provides an invalid id)
+			// therefore, we regenerate a id
 			sess.fresh = true
+			sess.id = s.KeyGenerator()
 		}
 	}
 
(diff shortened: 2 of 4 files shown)


ReneWerner87:
Since these bugs were not found by the tests, can you please add some to ensure that future code changes do not cause this problem again?

middleware/session/store.go
```diff
@@ -79,7 +79,11 @@ func (s *Store) Get(c *fiber.Ctx) (*Session, error) {
 		} else if err != nil {
 			return nil, err
 		} else {
+			// raw is nil, which means id is not in the storage
+			// so it means that id is not valid (mainly because of id is expired or user provides an invalid id)
+			// therefore, we regenerate a id
 			sess.fresh = true
```
ReneWerner87, line 85: can you create a method that combines the two commands and use it in sess.Regenerate and store, think that would be even cleaner

```go
func (s *Session) refresh() {
    s.id = s.config.KeyGenerator()
    s.fresh = true
}
```

Spedoske: Done.

hi019:
Hey, thanks for the PR! Some comments:

> Fix: Session.Regenerate does not set Session.fresh to be true.

I think the current behaviour is sufficent. `Session.Regenerate` only resets the session ID. The data inside the sessions still persists, so technically the session is not new.

I was working on a similar problem a few days ago, where `Session.SetExpiry` only sets the session expiry if the session is fresh. Instead of setting the session to fresh, I added a new private field to `Session` called `shouldSetSession`. Here is my commit (it's on a testing branch, I haven't created a PR yet): https://github.com/gofiber/fiber/commit/3d821f4a3a05a679414d8450cdb550bb031be498. Do you think we could do a similar thing here?

Spedoske:
It’s true that we need to set session for user after  `Regenerate` and `SetExpiry` since the session id  and the expiration time changes, which are stored in the user’s browser.

I agree that you should set `shouldSetSession` to be true when `SetExpiry` for a session, because the session is not changed, only the expiration time changes.

I am curious which is a more appropriate definition of regenerate ID, here are two versions:
1. Remove all data associating with the old session, and assign a new session to the user.
2. Assign a new session to the user, copy the data for the new session from the old session, and finally destroy the old session.

For version 1, the new session do not have any data. So the session is fresh.
For version 2, the data are transferred from the old session to the new session. So the session is not fresh, but session id changes.

Both of these two versions requires fiber to set session for user.

According to the main branch of fiber, the `Regenerate` implements the version 1, the function destroys all data in the storage for the old session, and assigns a new ID for the session with no data. So I think we should set `fresh` to be true.


If you think that version 2 is the more appropriate implementation, then we should set `shouldSetSession` to be true.

hi019:
My mistake. I looked at the `Regenerate` code again, and it does delete the session data in addition to regenerating the ID. The PR looks fine as it is, and I'll test it out tomorrow!

ReneWerner87 (approved):

middleware/session/session.go
```diff
@@ -118,12 +118,21 @@ func (s *Session) Regenerate() error {
 		return err
 	}
 
-	// Create new ID
-	s.id = s.config.KeyGenerator()
+	// generate a new session, and set session.fresh to be true
```
hi019, line 121:
```suggestion
	// generate a new session, and set session.fresh to true
```

hi019 (approved):

PR #1418, opened 2021-07-01
🔥 FEATURE: Allow adding App Name to customizing the startup message

sujit-baniya:
With this feature, user could add business app and build details on startup message alongside with other details

With this change, extra field is added in the config to allow adding App Name


ReneWerner87:
thx,

can you add the expectation to the test, unfortunately it is also missing in the initial test
would be good if we check what comes out and not just run it and see that no error happens

and please also add a pull request to our doc repository

sujit-baniya:
@ReneWerner87 updated the test. The doc was already added and made PR here https://github.com/gofiber/docs/pull/183

ReneWerner87:
Thx

ReneWerner87 (approved):

Issue #1414, opened 2021-06-29
🐛 QueryParser returns no error when required field is empty


tuhao1020:
```go
type QueryParams struct {
	Name string `query:"name,required"`
	Addr string `query:"addr,required"`
}

func ParseHandler(ctx *fiber.Ctx) error {
	param := new(QueryParams)
	err := ctx.QueryParser(param)
	if err != nil {
		ctx.Write([]byte(err.Error()))
	} else {
		err = ctx.JSON(param)
	}
	return err
}
```

http://localhost:8080/test?nameaaa=hello&addrbbb=world

```json
{
  "Name": "",
  "Addr": ""
}
```

http://localhost:8080/test?n=hello&addrbbb=world

```
name is empty
```

The above two cases are expected to return the following contents：

```
name is empty (and 1 other error)
```

tuhao1020:
L160: `strings.HasPrefix(key, path)` should be `key == path` ?

https://github.com/gofiber/fiber/blob/28ba42a6fd89b0a94de241931ba2ce1503d78fef/internal/schema/decoder.go#L151-L167

ReneWerner87:
think it was solved with hasPrefix, because in deeper structures, they wanted to check the previous parts


https://github.com/gorilla/schema/blob/master/decoder.go#L160
https://github.com/gorilla/schema/pull/126
https://github.com/gorilla/schema/commit/61751c9687439bf8cd824ad7eec0d21db1a1b804

here is also an issue about it
-> https://github.com/gorilla/schema/issues/176

maybe you can support it and give some information, as soon as it is adjusted in the library, i can try to take it over

tuhao1020:
@ReneWerner87 do you mean nested struct like this?

ReneWerner87:
![image](https://user-images.githubusercontent.com/7063188/123915923-4cf40080-d981-11eb-94f9-5350f33f6ad4.png)

tuhao1020:
@ReneWerner87 
```go
type R2 struct {
	A struct {
		B int `query:"b"`
	} `query:"a,required"`
}
```

I think if A has no required fields , it does't need to keep a `required` option.

ReneWerner87:
i know, only the test was filed with the adjustments you noted
and I think that's where the check with hasPrefix comes from

he wanted to check if a is prepended in the string

tuhao1020:
[...]

PR #1424, opened 2021-07-05
Schema parser: use full match instead of strings.HasPrefix

tuhao1020:
#1414 

https://github.com/gorilla/schema/issues/176

### DIFF
--- a/internal/schema/decoder.go
+++ b/internal/schema/decoder.go
@@ -157,7 +157,10 @@ func isEmptyFields(fields []fieldWithPrefix, src map[string][]string) bool {
 				return false
 			}
 			for key := range src {
-				if !isEmpty(f.typ, src[key]) && strings.HasPrefix(key, path) {
+				// issue references:
+				// https://github.com/gofiber/fiber/issues/1414
+				// https://github.com/gorilla/schema/issues/176
+				if !isEmpty(f.typ, src[key]) && key == path {
 					return false
 				}
 			}
--- a/internal/schema/decoder.go
+++ b/internal/schema/decoder.go
@@ -160,7 +160,7 @@ func isEmptyFields(fields []fieldWithPrefix, src map[string][]string) bool {
 				// issue references:
 				// https://github.com/gofiber/fiber/issues/1414
 				// https://github.com/gorilla/schema/issues/176
-				nested := strings.Contains(key, ".")
+				nested := strings.IndexByte(key, '.') != -1
 
 				// for non required nested structs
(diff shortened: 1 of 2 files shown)


tuhao1020:
@ReneWerner87 
Test_Cache_Expired failed:
```
=== RUN   Test_Cache_Expired
Error:     cache_test.go:77: Cache should not have expired: 1625478442551438654, 1625478442551768379
--- FAIL: Test_Cache_Expired (2.00s)
```

ReneWerner87:
thank you, in the cache test there is a not stable test

if you have time you can analyze and stabilize it, i'll start the tests again for now

ReneWerner87:
https://github.com/gorilla/schema/pull/178#issuecomment-873103020

i think he is right and you should include this in the solution

tuhao1020:
@ReneWerner87 
I'm not very familiar with gorilla/schema, so I don't understand which case he was talking about,  can you provide a test case to me?

ReneWerner87:
```go
	A struct {
		B int `schema:"b"`
	} `schema:"a,required"`
	
	C struct {
		D int `schema:"d,required"`
	} `schema:"c"`
```
I think he wants both to work

with your change you have destroyed the first case which checks that the object lying on "a" is not empty

in your example you only check if in "c" the property "d" is set


i.e. you would have to check if there is a dot in the current path and then do the partial check with it or if it is a path of the deepest element, then do the check for equality

i think the check with the HasPrefix is important, in my opinion there is still a dot missing in the string and as second condition the equality check

to understand the problem, i would recommend you to debug the unittests for this testcase, then you will see what is matched at this point and you should be able to look at the problem also from the second point of view

tuhao1020:
@ReneWerner87  Thank you! I'll have a try

internal/schema/decoder.go
```diff
@@ -160,7 +160,17 @@ func isEmptyFields(fields []fieldWithPrefix, src map[string][]string) bool {
 				// issue references:
 				// https://github.com/gofiber/fiber/issues/1414
 				// https://github.com/gorilla/schema/issues/176
-				if !isEmpty(f.typ, src[key]) && key == path {
+				nested := strings.Contains(key, ".")
```
ReneWerner87, line 163: use strings.IndexByte() ... it should be faster

tuhao1020: done

tuhao1020 (commented):
viewed

tuhao1020 (commented):
done

internal/schema/decoder.go
ReneWerner87: thx

Issue #1419, opened 2021-07-01
🐛 Memory Session Storage error with string of random length


chrigeeel:
**Fiber version**
2.12.0

**Issue description**
When using the default memory storage as a session storage, and using sess.Set() with a key and a string of different lengths, the session handling stops working with errors such as `Unexpected EOF`.
In the below example, the key "randomStr" is set to a string with a random length on page refresh.
If you do that from two sessions (browser), you can reproduce the error after a few refreshes.
I attached a video for clarification.
This bug doesn't happen when using a DB as a storage, e.g. Redis or Mongo.

https://user-images.githubusercontent.com/76595857/124153801-1e674a00-da95-11eb-9b8c-c8e24e06e617.mp4

**Code snippet**

```go
package main

import (
	"crypto/rand"
	"fmt"
	"math/big"

	"github.com/gofiber/fiber/v2"
	"github.com/gofiber/fiber/v2/middleware/session"
)

var Store = session.New()

func main() {
	app := fiber.New()
	app.Get("/", handleExample)
	app.Listen(":3000")
}

func handleExample(c *fiber.Ctx) error {
	sess, err := Store.Get(c)
	if err != nil {
		return c.Status(fiber.StatusInternalServerError).JSON(fiber.Map{
			"error": err.Error(),
		})
	}
	randomStr, err := GenerateRandomString(GenerateRandomInt(10,20))
	if err != nil {
		return c.Status(fiber.StatusInternalServerError).JSON(fiber.Map{
			"error": err.Error(),
		})
	}
	fmt.Println("random string: " + randomStr)
	defer sess.Save()
	sess.Set("randomStr", randomStr)
	return c.Status(fiber.StatusOK).JSON(fiber.Map{
		"randomStr": randomStr,
	})
}

func GenerateRandomInt(min, max int) int {
	max++
	bg := big.NewInt(int64(max - min))

	n, err := rand.Int(rand.Reader, bg)
	if err != nil {
		panic(err)
	}

	return int(n.Int64() + int64(min))
}

func GenerateRandomString(n int) (string, error) {
	const letters = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"
	ret := make([]byte, n)
	for i := 0; i < n; i++ {
		num, err := rand.Int(rand.Reader, big.NewInt(int64(len(letters))))
		if err != nil {
			return "", err
		}
		ret[i] = letters[num.Int64()]
	}

[...]

Issue #1420, opened 2021-07-01
🤗 How to manage/log template errors in fiber


Koredeoluwafemi:
I am getting these template errors on production app

1. gob: decoding into local type *map[string]interface {}, received remote type unknown type,

2. extra data in buffer,

3. gob: encoded unsigned integer out of range

I use fiber.Map interface type in passing data from the handler to the template. 

I am unable to replicate the issue, it just shows up once in a while.

ReneWerner87:
Gob is used in the session middleware, please share the code for your usage with this here.

Koredeoluwafemi:
Here is an example of my handler

    func Users(c *fiber.Ctx) error {

	db := database.DB
	session, _ := sess.Store.Get(c) // get/create new session

	var users []models.User

	notify := controllers.GetNotify(session)

	db.Find(&users)
	
	output := fiber.Map{
		"Title":     config.App.Name + ": Manage Users",
		"Notify":    notify,
		//"Admin":     admin,
		"Users":     users,
	}

	_ = session.Save()
	return c.Render("users/users", output)
}

Koredeoluwafemi:
@ReneWerner87

Koredeoluwafemi:
Or can you share a pointer to how I can correct or fix it @ReneWerner87

ReneWerner87:
no unfortunately not yet, have the cause of the problem unfortunately not yet found

the report #1419 should have the same cause

not sure where the problem with the gob encoder/decode is at the moment

ReneWerner87:
```go
	"Notify":    notify,
	//"Admin":     admin,
	"Users":     users,
```
Just see that you use some structs, have you registered them ?
![image](https://user-images.githubusercontent.com/7063188/124481661-d1db8180-dda8-11eb-9ec3-01a0bdd6293d.png)
https://github.com/gofiber/fiber/tree/master/middleware/session

f.e.: https://github.com/gofiber/fiber/blob/master/middleware/session/session_test.go#L126
so that the session middleware can pass this information to gob and gob knows how to encode and decode it ?

ReneWerner87:
the question is also what you save in the session

Koredeoluwafemi:
So I run a series of tests and discovered I am getting error while trying to get or set a new session


    func Login(c *fiber.Ctx) error {

	session, err := sess.Store.Get(c) // get/create new session
	if err != nil {
		log.Println("login error")
		log.Println(err.Error())
		return err
	}
	
[...]

PR #1425, opened 2021-07-07
Fix: copy the data in buffer instead of using original data.

Spedoske:
This PR is intended to close the #1419 issue.
The explanation of the bug is in the following comment.
https://github.com/gofiber/fiber/issues/1419#issuecomment-875422869

### DIFF
--- a/middleware/session/session.go
+++ b/middleware/session/session.go
@@ -160,8 +160,12 @@ func (s *Session) Save() error {
 		return err
 	}
 
-	// pass raw bytes with session id to provider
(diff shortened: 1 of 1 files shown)


ReneWerner87:
thanks for the fix, will review it after i understand the great commentary

ReneWerner87 (approved):

ReneWerner87:
thank you, i hope this solves the problems with 
https://github.com/gofiber/fiber/issues/1420 and https://github.com/gofiber/fiber/issues/1419

will try that in the next few days to test

of course you can do that too, you just have to get the commit in the master via dependencie
https://jfrog.com/blog/go-big-with-pseudo-versions-and-gocenter/

PR #1428, opened 2021-07-09
cookie: add possibility to disable same site attribute

wja513:
fix some old browsers cann't set the cookie if it contains "SameSite" attribute.

```go
ctx.Cookie(c, &fiber.Cookie{
	Name:     cookie.Name,
	Value:    cookie.Value,
        ...
	//SameSite: "",// this attribute should be constant.（why not use fasthttp constant directly?）
})
```


ReneWerner87 (approved):

ReneWerner87:
@wja513 please check the tests
image

ctx.go
```diff
@@ -363,13 +362,17 @@ func (c *Ctx) Cookie(cookie *Cookie) {
...
 	fcookie.SetHTTPOnly(cookie.HTTPOnly)
 
-	switch utils.ToLower(cookie.SameSite) {
-	case "strict":
-		fcookie.SetSameSite(fasthttp.CookieSameSiteStrictMode)
-	case "none":
-		fcookie.SetSameSite(fasthttp.CookieSameSiteNoneMode)
-	default:
-		fcookie.SetSameSite(fasthttp.CookieSameSiteLaxMode)
+	if cookie.SameSite == "" {
```
ReneWerner87, line 365: think that this change will cause a breaking change, since the disabled and the lax share the same case

Please add more identifiers for sameSite

like for example "lax" or "disabled"

but please make sure that no breaking change occurs and the identifiers are documented to the cookie struct

furthermore additional unittests for the new cases would be good

ReneWerner87 (changes requested):

ctx.go
wja513: please check code again~

ReneWerner87 (approved):

Issue #1422, opened 2021-07-02
🚀 Could add a requestId tag in logger middleware? like TagRequestId?


goldwind-ting:
```
package main
import (
     "github.com/natefinch/lumberjack"
     "github.com/gofiber/fiber/v2"
     "github.com/gofiber/fiber/v2/middleware/logger"
)
app := fiber.New()
app.Use(requestid.New())
app.Use(logger.New(
logger.Config{
        Output: &lumberjack.Logger{
        Filename:   "./foo.log", 
        MaxSize:    1, 
        MaxBackups: 3, 
        MaxAge:     7, 
        Compress:   true,
        },
        Format: "[${time}] ${status}  ${latency} ${path} ${requestId} ${queryParams} ${body} ${resBody}\n",
}))
```

I use a RequestId middleware and that set a response header, but how could I get this header (default X-Request-Id) and record it in log file?

ReneWerner87:
Please add a description

ReneWerner87:
you could use 

${header:x-request-id} in the log message

https://github.com/gofiber/fiber/tree/master/middleware/logger
![image](https://user-images.githubusercontent.com/7063188/124480901-0864cc80-dda8-11eb-9f7c-3b9b7958e268.png)

goldwind-ting:
it does not work. and when using ${header:x-request-id} in the log message, then c.Get is called,  it means that the x-request-id header is getted from request header, but x-request-id is a response header.
```
func (c *Ctx) Get(key string, defaultValue ...string) string {
	return defaultString(c.app.getString(c.fasthttp.Request.Header.Peek(key)), defaultValue)
}
```

ReneWerner87:
okay, then we should either allow more prefixes like response or request

or add a possibility to define in its own placeholder with function 


i.e. your request is not possible with the current implementation

pull requests are welcome

goldwind-ting:
It can get x-request-id by `c.Response().Header.Peek("X-Request-Id")`,but there is no tag that could get response header

goldwind-ting:
> okay, then we should either allow more prefixes like response or request
> 
> or add a possibility to define in its own placeholder with function
> 
> i.e. your request is not possible with the current implementation
> 
> pull requests are welcome

Thank you very much!

dpcks666:
How about this format?
${locals:requestid}

goldwind-ting:
> How about this format?
> ${locals:requestid}
[...]

PR #1446, opened 2021-07-21
[Logger]: add tag for response header

goldwind-ting:
fix issue #1422

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -33,7 +33,6 @@ const maxParams = 30
 
 const queryTag = "query"
 
-
 // Ctx represents the Context which hold the HTTP request and response.
 // It has methods for the request query string, parameters, body, HTTP headers and so on.
 type Ctx struct {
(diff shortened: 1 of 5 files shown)


ctx.go
```diff
@@ -526,6 +525,15 @@ func (c *Ctx) Get(key string, defaultValue ...string) string {
 	return defaultString(c.app.getString(c.fasthttp.Request.Header.Peek(key)), defaultValue)
 }
 
+// GetRespHeader returns the HTTP response header specified by field.
+// Field names are case-insensitive
+// Returned value is only valid within the handler. Do not store any references.
+// Make copies or use the Immutable setting instead.
+func (c *Ctx) GetRespHeader(key string, defaultValue ...string) string {
```
ReneWerner87, line 532: we need a part in the docs repository for this new method

middleware/logger/logger.go
```diff
@@ -40,7 +40,8 @@ const (
 	TagBytesReceived     = "bytesReceived"
 	TagRoute             = "route"
 	TagError             = "error"
-	TagHeader            = "header:"
+	TagReqHeader         = "reqHeader:"
```
ReneWerner87, line 43: we should not remove the old key, otherwise we would no longer be downward compatible with the version

you can add the old tag here and in the switch, mark it as deprecated and direct it to the TagReqHeader in the switch

ReneWerner87 (changes requested):

ctx.go
goldwind-ting: How about this?

goldwind-ting: Sorry, I am not clear, do you mean I should add the annotation for this new method in https://github.com/gofiber/docs/blob/master/api/ctx.md?

ReneWerner87: yes and please add a unitttest for the method

goldwind-ting: 👌

ReneWerner87:
@goldwind-ting can you still implement the note from me and make downward compatibility

ReneWerner87:
@goldwind-ting could you also update https://github.com/gofiber/docs/blob/master/api/middleware/logger.md in a PR

ReneWerner87 (approved):

PR #1458, opened 2021-07-29
🐛 fix parse gzipped body

faceair:
Now when we use `filber.Ctx.BodyParser` to deserialize JSON, it uses `filber.Ctx.fasthttp.Request.Body` method internally to take the raw body data, which works fine in most cases.

But when the request body is gzip-compressed, `filber.Ctx.fasthttp.Request.Body` takes the compressed data and there is no way to deserialize it directly.
image

So in this PR I changed the method inside `filber.Ctx.BodyParser` that takes the request body to `filber.Ctx.Body` method, which automatically detects the encoding of the request body and decompresses it itself. I have also added test cases for compression requests.

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -291,7 +291,7 @@ func (c *Ctx) BodyParser(out interface{}) error {
 	// Parse body accordingly
 	if strings.HasPrefix(ctype, MIMEApplicationJSON) {
 		schemaDecoder.SetAliasTag("json")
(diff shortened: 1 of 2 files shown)


ReneWerner87:
Thanks for the pull request, understand the code also, but not quite the description, can you adjust this again a bit, so that everyone knows what should be achieved with the pull request, if someone goes through the link in the release on the pull request

faceair:
Thanks to review, I have added more complete description information.

Issue #1452, opened 2021-07-25
🚀  JSONDecoder to config


supernova0730:
Hello everyone, I think it will be better if you will add JSONDecoder to config as JSONEncoder  
![image](https://user-images.githubusercontent.com/46263611/126909486-9e9410ac-2d9d-4984-a571-ead8aa8a1095.png)

FlameMida:
yes,it is important to custom json package

pd-pranay:
@ReneWerner87  Can I start working on it  ?

ReneWerner87:
Sure

pd-pranay:
To summarize things I need to do,
create JSONEncoder and JSONDecoder functions  ,
which will marshal or unmarshal data depending, upon the set up used at the time of App initialization ?

PR #1489, opened 2021-08-15
📝 Update: Add JSONDecoder to config

pd-pranay:

Add json decoder to config 

This pull request is in reference to [#1452](https://github.com/gofiber/fiber/issues/1452)
**Commit formatting** 

### DIFF
--- a/app.go
+++ b/app.go
@@ -308,6 +308,13 @@ type Config struct {
 	// Default: json.Marshal
 	JSONEncoder utils.JSONMarshal `json:"-"`
 
+	// When set by an external client of Fiber it will use the provided implementation of a
+	// JSONUnmarshal
+	//
+	// Allowing for flexibility in using another json library for encoding
+	// Default: json.Unmarshal
+	JSONDecoder utils.JSONUnmarshal `json:"-"`
+
 	// Known networks are "tcp", "tcp4" (IPv4-only), "tcp6" (IPv6-only)
 	// WARNING: When prefork is set to true, only "tcp4" and "tcp6" can be chose.
 	//
@@ -458,6 +465,9 @@ func New(config ...Config) *App {
 	}
 	if app.config.JSONEncoder == nil {
 		app.config.JSONEncoder = json.Marshal
+	}
+	if app.config.JSONDecoder == nil {
+		app.config.JSONDecoder = json.Unmarshal
 	}
 	if app.config.Network == "" {
 		app.config.Network = NetworkTCP4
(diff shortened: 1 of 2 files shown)


pd-pranay:
Let me know if everything is fine.

ReneWerner87:
Could you please add a test for the configuration possibility and a part in the docs repository.

pd-pranay:
> Could you please add a test for the configuration possibility and a part in the docs repository.

Sure 👍

ReneWerner87 (commented):
See the possibility to define the decoder, but no use within the framework/library, then shouldn't we use somewhere the method which was configured?

pd-pranay:
> See the possibility to define the decoder, but no use within the framework/library, then shouldn't we use somewhere the method which was configured?

Yes, I was thinking to add this to methods in ctx.go with using app.config's  Encoder and Decoder.
Will this do the work ? @ReneWerner87 

```
func (c *Ctx)JSONDecoder(raw []byte, data interface{}) error {
	err := c.app.config.JSONDecoder(raw, data)
	if err != nil {
		return err
	}
	return nil
}
```

```
func (c *Ctx)JSONEncoder(data interface{}) ([]byte, error) {
	raw, err := c.app.config.JSONEncoder(data)
	if err != nil {
		return nil, err
	}
	return raw, nil
}
```

ReneWerner87:
thought rather that the decoder should be used in the bodyParser or everywhere where we directly use the json unmarshal

i wouldn't provide an extra method in the context because it doesn't really have any reference to it and the one who sets the decoders and encoders from the outside in the config can use them from the outside capsule too

the only benefit i can think of is that you exchange the internal usages

ReneWerner87:
@pd-pranay could you change all the usages for the unmarshall method

pd-pranay:
@ReneWerner87   can you help me with references where I should be changing usages of unmarshal method  ?
I found one is this correct one  ? 
https://github.com/gofiber/fiber/blob/master/ctx.go#L294

ReneWerner87:
@pd-pranay Yes, that should be the only

pd-pranay:
For [docs](https://docs.gofiber.io/api/fiber#:~:text=//%20...-,Config%20fields,-Property) what type should I be defining for JSONDecoder ?

ReneWerner87:
https://docs.gofiber.io/api/fiber#config

it should just show up here so users know they can customize it, maybe you can add the doc for JSONEncode as well

ReneWerner87 (approved):

ReneWerner87:
> https://docs.gofiber.io/api/fiber#config
> 
> it should just show up here so users know they can customize it, maybe you can add the doc for JSONEncode as well

@pd-pranay can you add JSONEncode and JSONDecode to the documentation ?
https://github.com/gofiber/docs/blob/master/api/fiber.md#config

Issue #1487, opened 2021-08-13
🐛 panic: unaligned 64-bit atomic operation [32 bit machines]


Hexiro:
**Fiber version**
v2.17.0

**Issue description**
On 32 bit machines, fiber sometimes crashes on startup when using certain middleware.
I believe this is something to do with `sync/atomic`

https://pkg.go.dev/sync/atomic#pkg-note-BUG

```
On 386, the 64-bit functions use instructions unavailable before the Pentium MMX.

On non-Linux ARM, the 64-bit functions use instructions unavailable before the ARMv6k core.

On ARM, 386, and 32-bit MIPS, it is the caller's responsibility to arrange for 64-bit alignment of 64-bit words accessed atomically. The first word in a variable or in an allocated struct, array, or slice can be relied upon to be 64-bit aligned.
```

**Code snippet**

```go
package main

import (
        "github.com/gofiber/fiber/v2"
        "github.com/gofiber/fiber/v2/middleware/limiter"
        "log"
)

func main()  {
        app := fiber.New()
        app.Use(limiter.New())
        log.Fatal(app.Listen(":3000"))
}
```
```
panic: unaligned 64-bit atomic operation

goroutine 20 [running]:
runtime/internal/atomic.panicUnaligned()
        /usr/local/go/src/runtime/internal/atomic/unaligned.go:8 +0x24
runtime/internal/atomic.Store64(0x1d0016c, 0x6116b072, 0x0)
        /usr/local/go/src/runtime/internal/atomic/asm_arm.s:278 +0x14
github.com/gofiber/fiber/v2/internal/memory.(*Storage).updater(0x1d00150, 0x3b9aca00, 0x0)
        /home/pi/go/pkg/mod/github.com/gofiber/fiber/v2@v2.17.0/internal/memory/memory.go:69 +0x40
created by github.com/gofiber/fiber/v2/internal/memory.New
        /home/pi/go/pkg/mod/github.com/gofiber/fiber/v2@v2.17.0/internal/memory/memory.go:26 +0x128
```

PR #1502, opened 2021-08-22
FIX: panic: unaligned 64-bit atomic operation [32 bit machines] #1487

ReneWerner87:
FIX: panic: unaligned 64-bit atomic operation [32 bit machines] #1487


ReneWerner87:
Steps to reproduce:
32Bit docker engine
`docker run -it --rm golang@sha256:633dff80ab7628865ee68d090ccc2fe52c7b6b9c9e34dcb7250f4efcfbc98dd0 bash`
`git clone https://github.com/ReneWerner87/fiber.git`
`cd fiber`
`git checkout 32Bit-machines`
`go mod download && go mod vendor && go mod tidy`
`go test ./... -v`


image

PR #1513, opened 2021-08-31
✨ support logging color for custom logging formats

### LINKED ISSUE #1467: 🚀 I want to use color in the custom logger middleware
**Is your feature request related to a problem?**
no
**Describe the solution you'd like**
make it export in config

amalshaji:
**Support logger coloring for custom logging formats**

Add a function to check for a valid custom format. A valid format is one having both `${status}` and `${method}`. Fixed #1467

### DIFF
--- a/middleware/logger/config.go
+++ b/middleware/logger/config.go
@@ -3,6 +3,7 @@ package logger
 import (
 	"io"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/gofiber/fiber/v2"
@@ -56,6 +57,20 @@ var ConfigDefault = Config{
 	enableColors: true,
 }
 
+// Function to check if the logger format is compatible for coloring
+func validCustomFormat(format string) bool {
+	var validTemplates = []string{"${status}", "${method}"}
+	if format == "" {
+		return true
+	}
+	for _, template := range validTemplates {
+		if !strings.Contains(format, template) {
+			return false
+		}
+	}
+	return true
+}
+
 // Helper function to set default values
 func configDefault(config ...Config) Config {
 	// Return default config if nothing provided
@@ -67,7 +82,7 @@ func configDefault(config ...Config) Config {
 	cfg := config[0]
 
 	// Enable colors if no custom format or output is given
-	if cfg.Format == "" && cfg.Output == nil {
+	if validCustomFormat(cfg.Format) && cfg.Output == nil {
 		cfg.enableColors = true
 	}
 


ReneWerner87:
can you please provide unittests that test the new code in use and make sure what we expect happens

amalshaji:
how do test for colors in output? or just test the newly added function?

ReneWerner87:
Use a buffer as weitter in the configuration and check the generated output.

ReneWerner87:
@amalshaji any progress ? 
you could test the feature in the test and define a buffer as the writer in the config and then check the created output
-> color output has a special prefix and suffix

amalshaji:
working on it now

amalshaji:
```go
if validCustomFormat(cfg.Format) && cfg.Output == nil {
    cfg.enableColors = true
}
```

If I use `buf` as Output, the colors won't be enabled!

ReneWerner87:
ok then use my function to capture the output
https://github.com/gofiber/fiber/blob/master/app_test.go#L1263-L1291

ReneWerner87:
@amalshaji can you please create the tests

amalshaji:
I'm using your function to capture the output. The following code is used for testing the output:

```go
func Test_Logger_Custom_Format_Color(t *testing.T) {
	output := captureOutput(func() {
		app := fiber.New()
		app.Use(New(Config{
			Format: "",
		}))
		app.Test(httptest.NewRequest("GET", "/test", nil))
	})
	fmt.Println("captured Output: ", output)
	utils.AssertEqual(t, 5, 5)
}
```

When the format is `""`, I get:

```bash
❯ go test -timeout 30s -run ^Test_Logger_Custom_Format_Color
captured Output:  18:32:03 | 404 |      0s |         0.0.0.0 | GET     | /test

PASS
ok      github.com/gofiber/fiber/v2/middleware/logger   0.004s
```

When I use a custom template like `${status}`, I get:

```bash
❯ go test -timeout 30s -run ^Test_Logger_Custom_Format_Color
404captured Output:
PASS
ok      github.com/gofiber/fiber/v2/middleware/logger   0.005s
```

No output is being captured. Note the 404 printed to stdout.

So I decided to disable the validCustomTemplate function and use the previous check, i.e., `if cfg.Format == "" && ...`

 Got the same output as the empty format one. Unfortunately, the output does not capture any color tags. Also, no color tests were done previously, so I can't refer.

amalshaji:
`go run main.go` prints color to output. But `fiber dev` doesn't. Is this related to my problem?

image

ReneWerner87:
Yes should be, you can open a issue in  https://github.com/gofiber/cli

ReneWerner87:
@amalshaji that means we do not need the pull request ? the normally come the colors yes or ?

amalshaji:
Only when the format is "" (Empty). The issue was to support colors for custom format. I can't figure out how to test for colors.

ReneWerner87 (approved):

ReneWerner87:
@amalshaji  thx for the contribution and effort

have unfortunately also found no way

from the code it looks good

Issue #1500, opened 2021-08-21
🐛 ipv6 loopback


hotrungnhan:
**Fiber version**
v2.17.0
**Issue description**
i findout a little UI isuee with ipv6, that the console not giving me rightaway loopback ip. It must be [::1]:port not [::]
 ┌───────────────────────────────────────────────────┐
 │                     trueIPV6                      │
 │                   Fiber v2.17.0                   │
 │                 http://[::]:8000                  │
 │                                                   │
 │ Handlers ............. 2  Processes ........... 1 │
 │ Prefork ....... Disabled  PID .............. 9972 │
 └───────────────────────────────────────────────────┘
**Code snippet**
```go
package main

import "github.com/gofiber/fiber/v2"

func main() {

  var set = fiber.Config{
	  Network:                 fiber.NetworkTCP6,
  }
  app := fiber.New(set)

  // Steps to reproduce

  log.Fatal(app.Listen(":3000"))
}
```

techytoes:
@ReneWerner87 can I work on this issue? Thanks!

ReneWerner87:
Sure

techytoes:
@ReneWerner87 can you please help me with documentation for testing locally? 🙇🏼

hi019:
@techytoes I can help. What issue are you facing?

techytoes:
@hi019 I have made the changes. Just wanted to test these locally. Couldn't find any supporting documentation. Can you please help me with this?

hi019:
@techytoes you should be able to just run `go test`

PR #1517, opened 2021-09-03
🩹  Fix: corrects ipv6 loopback ip shown on UI

techytoes:
Fixes: #1500

### DIFF
--- a/app.go
+++ b/app.go
@@ -954,8 +954,13 @@ func (app *App) startupMessage(addr string, tls bool, pids string) {
 
 	host, port := parseAddr(addr)
 	if host == "" {
-		host = "0.0.0.0"
+		if app.config.Network == NetworkTCP6 {
+			host = "[::1]"
+		} else {
+			host = "0.0.0.0"
+		}
 	}
+
 
 	scheme := "http"
 	if tls {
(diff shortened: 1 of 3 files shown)


ReneWerner87:
you have only adjusted the tests in the pull request, where is the fix of the problem ?

techytoes:
@ReneWerner87 I have pushed the fix for the problem as well. Let me know if any other changes are also required.

ReneWerner87:
ok, had looked too early, there were only the test data in it

I check now

app.go
```diff
@@ -957,6 +957,11 @@ func (app *App) startupMessage(addr string, tls bool, pids string) {
 		host = "0.0.0.0"
 	}
 
+	if app.config.Network == NetworkTCP6 {
```
ReneWerner87, line 960: i think we should only set this if the host is not explicitly included in the address used for the listener

please move it in the `if host == "" {` body

ReneWerner87 (changes requested):

app.go
techytoes: okay. will do this. Thanks!

ReneWerner87 (approved):

PR #1522, opened 2021-09-04
:bug: Fix Mutability issue of UserContext in fiber.Ctx

### LINKED ISSUE #1518: 🐛 UserContext & SetUserContext suffer from Mutability of fiber.Ctx
**Fiber version**
v2.18.0

**Issue description**
Method `UserContext()` and `SetUserContext()` of `fiber.Ctx` suffer from mutability (regardless of `Immutable` config flag),
meaning that `context.Context` stored in `fiber.Ctx` via those methods is, most of the time, changing below our feet or being reused between across requests

**Code snippet**

```go
package main

import (
	"context"
	"fmt"
	"log"

	"github.com/gofiber/fiber/v2"
	"github.com/gofiber/fiber/v2/utils"
)

type contextKeyType string

const contextKey contextKeyType = "foobar-key"

// This is the first and foremost middleware, so we expect `c.UserContext()` to always return empty context.Context
func editUserContext(c *fiber.Ctx) error {
	ctx := c.UserContext()

	// In just a couple of consecutive requests, this condition below will be triggered
	if ctx.Value(contextKey) != nil {
		log.Fatalln("context.Context from UserContext() not empty in early request lifecycle, forcefully exiting...")
	}

	ctx = context.WithValue(ctx, contextKey, "foobar-value")
	c.SetUserContext(ctx)

	return c.Next()
}

func showUserContext(c *fiber.Ctx) error {
	ctx := c.UserContext()
	log.Printf("show %v\n", ctx)

	return c.Next()
}

func main() {
	// Immutable flag doesn't help in this issue
	app := fiber.New(fiber.Config{Immutable: true})

	app.Use(editUserContext, showUserContext)

	// I'm making sure that fiber.Ctx will not be the same each time, by making requests with unique query param
		input := utils.CopyString(c.Query("input", "NO_VALUE"))
		return c.SendString(fmt.Sprintf("hello-%s", input))
	})

	app.Listen(":3000")
}
```
One-line shell script to help squeeze out the issue
```shell
for ((i=1;i<=100;i++)); do curl "http://localhost:3000/?input=${i}" && echo; done
```
**Workaround**
From what I've investigated and tried, the workarounds are to use one of the following
- `c.Locals()`
- Underlying `*fasthttp.RequestCtx` from `c.Context()`, using methods `.SetUserValue()` and `.UserValue()`

```go
package main

import (
	"context"
	"fmt"
	"log"

	"github.com/gofiber/fiber/v2"
	"github.com/gofiber/fiber/v2/utils"
)

type contextKeyType string

const localContext = "__local_context__"
const contextKey contextKeyType = "foobar-key"

func editUserContext(c *fiber.Ctx) error {
	ctx, ok := c.Locals(localContext).(context.Context)
	if !ok {
		ctx = context.Background()
	}

	// With `c.Locals(...)`, the condition below never ever hit
	if ctx.Value(contextKey) != nil {
		log.Fatalln("Context from c.Locals(...) not empty in early request lifecycle, forcefully exiting...")
	}

	ctx = context.WithValue(ctx, contextKey, "foobar-value")
	c.Locals(localContext, ctx)

	return c.Next()
}

func showUserContext(c *fiber.Ctx) error {
	ctx, ok := c.Locals(localContext).(context.Context)
	if !ok {
		log.Fatalln("context.Context should not be nil in this middleware")
	}
	log.Printf("show %v\n", ctx)

	return c.Next()
}

func main() {
	app := fiber.New()

	app.Use(editUserContext, showUserContext)

		input := utils.CopyString(c.Query("input", "NO_VALUE"))
		return c.SendString(fmt.Sprintf("hello-%s", input))
	})

	app.Listen(":3000")
}
```

Thammachart:

Fix #1518

I've not much experience in FastHTTP, especially in Mutability nature, so this fix put user-defined `context.Context` into `*fasthttp.RequestCtx` with constant key `__local_user_context__`, instead of `fiber.Ctx` struct field, and keeping the `fiber.Ctx` API intact

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -53,7 +53,7 @@ type Ctx struct {
 	values              [maxParams]string    // Route parameter values
 	fasthttp            *fasthttp.RequestCtx // Reference to *fasthttp.RequestCtx
 	matched             bool                 // Non use route matched
-	userContext         context.Context
+	userContext         context.Context      // User-defined context.Context
 }
 
 // Range data for c.Range
--- a/ctx.go
+++ b/ctx.go
@@ -34,6 +34,9 @@ const maxParams = 30
 
 const queryTag = "query"
 
+// userContextKey define the key name for storing context.Context in *fasthttp.RequestCtx
+const userContextKey = "__local_user_context__"
+
 // Ctx represents the Context which hold the HTTP request and response.
 // It has methods for the request query string, parameters, body, HTTP headers and so on.
 type Ctx struct {
@@ -53,7 +56,6 @@ type Ctx struct {
 	values              [maxParams]string    // Route parameter values
 	fasthttp            *fasthttp.RequestCtx // Reference to *fasthttp.RequestCtx
 	matched             bool                 // Non use route matched
-	userContext         context.Context      // User-defined context.Context
 }
 
 // Range data for c.Range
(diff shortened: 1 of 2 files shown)


ReneWerner87:
@patil-kshitij could you check the pull request 

not sure since i have never used this feature before

the code looks good

patil-kshitij (approved):

patil-kshitij:
Looks good

ReneWerner87:
Ok, thanks for the review

ReneWerner87 (approved):

PR #1530, opened 2021-09-14
Add one minute load avg for monitor middleware

efectn:
Now we have same features as [express-status-monitor](https://github.com/RafalWilinski/express-status-monitor).


efectn:
Golint error is interesting. We use invoke in somewheres like [load_darwin.go#L54](https://github.com/efectn/fiber/blob/add-load-avg/internal/gopsutil/load/load_darwin.go#L54).

ReneWerner87 (approved):

PR #1555, opened 2021-10-01
♻️ Option to overide form decoder setting - Rename function

rockcreation7:

Rename function - SetBodyParserDecoder to SetParserDecoder

@ReneWerner87 :  rename it a bit, because the decoder is used by the query parser and body parser

https://github.com/gofiber/fiber/pull/1100

**Commit formatting** 


ReneWerner87:
Thx, there are merged conflicts, could you please update the branch

And maybe provide also a new test case where we use the new possibilities for the query parser.

ReneWerner87:

Can you also add a block for the new possibilities in the documentation repository?

rockcreation7:
I will first add the docs, and follow up a new test case for the query parser.

rockcreation7:
https://github.com/gofiber/docs/pull/203

Added block of doc on use with BodyPauser, may update again after QueryPauser test case

rockcreation7:
I think it can be integrate with app config, what you guys think? or just let it as stand alone function.

ReneWerner87:
Standalone is okay for now

ReneWerner87 (approved):

PR #1557, opened 2021-10-01
🚀 SaveFile to default External Storage

Aliqyan:
Allow client users to create a Storage interface for a default cloud based storage (like s3), and automatically save the file to the configured external file storage system if one exists.

S3 storage Addition (in GoFiber/Storage):  https://github.com/gofiber/storage/pull/227

Feature was suggested here: https://discordapp.com/channels/704680098577514527/848707314180423750/869831718876643388


ReneWerner87:
not quite sure if we should extend this method and the config with it, rather in favor of a new method which is used for this purpose 

e.g. SaveFileToStorage

ReneWerner87 (approved):

ReneWerner87:
please add a PR for the documentation repository

ReneWerner87:
https://github.com/gofiber/docs/pull/235

PR #1559, opened 2021-10-02
🐛 fix: actual error not forwarded if request is not timed out.

### LINKED ISSUE #1496: 🐛 Timeout middleware is ignoring errors returned by the handler
**Fiber version**

**Issue description**

**Code snippet**

```go
package main

import (
	"github.com/gofiber/fiber/v2"
	"github.com/gofiber/fiber/v2/middleware/timeout"
	"time"
)

func main() {
	app := fiber.New()

	// Steps to reproduce
	app.Get("/", timeout.New(func(c *fiber.Ctx) error {
		return fiber.NewError(fiber.StatusServiceUnavailable, "On vacation!")
	}, 1 * time.Minute))

	app.Listen(":3000")
}
```

The timeout middleware for timeout has this code snippet, and as shown it is ignoring errors returned by the handler
```
// logic is from fasthttp.TimeoutWithCodeHandler https://github.com/valyala/fasthttp/blob/master/server.go#L418
	return func(ctx *fiber.Ctx) error {
		ch := make(chan struct{}, 1)

		go func() {
			defer func() {
				_ = recover()
			}()
			_ = handler(ctx) /// <- Error being ignored
			ch <- struct{}{}
		}()

		select {
		case <-ch:
		case <-time.After(timeout):
			return fiber.ErrRequestTimeout
		}

		return nil
	}
```

akshaybharambe14:
Forwarded the actual error if request succeeds/fails in given timeout window.

fixes #1496

### DIFF
--- a/middleware/timeout/timeout.go
+++ b/middleware/timeout/timeout.go
@@ -23,12 +23,12 @@ func New(handler fiber.Handler, timeout time.Duration) fiber.Handler {
 	// logic is from fasthttp.TimeoutWithCodeHandler https://github.com/valyala/fasthttp/blob/master/server.go#L418
 	return func(ctx *fiber.Ctx) error {
 		ch := make(chan struct{}, 1)
-
+		var err error
 		go func() {
 			defer func() {
 				_ = recover()
 			}()
-			_ = handler(ctx)
+			err = handler(ctx)
 			ch <- struct{}{}
 		}()
 
@@ -38,6 +38,6 @@ func New(handler fiber.Handler, timeout time.Duration) fiber.Handler {
 			return fiber.ErrRequestTimeout
 		}
 
-		return nil
+		return err
 	}
 }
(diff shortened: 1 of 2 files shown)


ReneWerner87:
@akshaybharambe14 can you provide a test for this customization so we can make sure the workflow works as intended?

akshaybharambe14:
I checked existing tests and all were commented out. Anyways, I will add tests.

ReneWerner87:
ok thx
you can also try to reactivate the tests, I think someone had deactivated them because of the stability, we would have to try to get them back, best as stable as possible

akshaybharambe14:
Yup, I will work on that this weekend.

akshaybharambe14:
@ReneWerner87, is there any way to copy ctx?

Seems like, we need to copy Ctx fields.

ReneWerner87:
> @ReneWerner87, is there any way to copy ctx?
> 
> Seems like, we need to copy Ctx fields.
@akshaybharambe14 
not really, but you are welcome to provide a function that can do that, I think something like that would be helpful

ReneWerner87:
@akshaybharambe14

ReneWerner87:
@akshaybharambe14 do you have time to look at the problem again?

kevin-you2:
When will this issue be resolved and released?

Issue #1567, opened 2021-10-07
🐛 Limiter Middleware behaves wrongly after latest changes


DomiM94:
**Fiber version**
v 2.20+ (Limiter Middleware Commit #1542 )

**Issue description**
Since the new attributes for skip options were implemented the limiter does not work as expected anymore. 
When I don't set any of the two new attributes the request will be forwarded to the upstream server even in case if this request will hit the limit. Therefore I end up with executing a successful request and returning a 429 (since after this request's execution the limiter notices that the limit was hit/overdone). 

The solution would be 
(1):
Don't forward the request when both of the attributes are set to false.

(2):
Don't return 429 on requests which were in fact executed successfully on the upstream (this affects the lastest request which is executed before the limit was hit (or overdone with the new implementation)). This will hide the actual statuscode.

ReneWerner87:
@aliereno could you check this

aliereno:
Hi, I tried to solve it with this pr https://github.com/gofiber/fiber/pull/1568. 
This solution
1-  doesn't forward the request before the limit reached calculations. 
2- doesn't returns 429 on successful requests and doesn't overwrite the actual statuscode.

But there is a problem that I can't solve. I need some help.
When remaining number is 1 and next request should be skipped, this code will return 429 because it can't see the actual status code before checking limit reached.

PR #1568, opened 2021-10-07
Fix limit middleware skip options

aliereno:
Related: https://github.com/gofiber/fiber/issues/1567

This solution
1-  doesn't forward the request before the limit reached calculations. 
2- doesn't returns 429 on successful requests and doesn't overwrite the actual statuscode.

But there is a problem that I can't solve. I need some help.
When remaining number is 1 and next request should be skipped, this code will return 429 because it can't see the actual status code before checking limit reached.

### DIFF
--- a/middleware/limiter/limiter.go
+++ b/middleware/limiter/limiter.go
@@ -102,8 +102,8 @@ func New(config ...Config) fiber.Handler {
 		err := c.Next()
 
 		// Check for SkipFailedRequests and SkipSuccessfulRequests
-		if (cfg.SkipSuccessfulRequests && c.Response().StatusCode() < 400) ||
-			(cfg.SkipFailedRequests && c.Response().StatusCode() >= 400) {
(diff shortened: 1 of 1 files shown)


aliereno:
@ReneWerner87 

I checked the codes. They decreased the hit counter with a callback at the end of response (`res.on("finish",  ...)`). So basically my code does same thing. I think they have same problem. But I didn't test it for express package. 
I tried to find different libraries that has feature skip by response's status code. But I couldn't find any.

Also in this express library, they put some info about possible problem of this implementation.

https://github.com/nfriedly/express-rate-limit#skipsuccessfulrequests
```
# skipSuccessfulRequests
  When set to true successful requests (response status < 400) won't be counted. 
  (Technically they are counted and then un-counted, so a large number of slow 
  requests all at once could still trigger a rate-limit. This may be fixed in a future 
  release.)
```

I will think about optimal solution. But I can't promise anything.

Issue #1406, opened 2021-06-24
🤗 Documentation inconsistent in fiber.Storage, fiber memory storage and session middleware


Spedoske:
**Question description**

In the session middleware, the default storage is the fiber memory storage (`internal/storage/memory`), which implements the interface `fiber.Storage`. According the comments in `app.go`, `Storage.Get` should return `ErrNotFound` if the storage does not contain the key. However, when I check out the implementation in `internal/storage/memory/memory.go`, it returns `nil` even if the key is not in the map.

I want to ask the following questions.
+ What is the appropriate behavior for the function `Storage.Get` ?
+ What does the returned error mean when error is not nil ?
+ What does the returned data (`[]byte`) mean when data is nil ?

**Code snippet** _Optional_

```go
// middleware/session/store.go
type Store struct {
	Config
}
...
func (s *Store) Get(c *fiber.Ctx) (*Session, error) {
        ...
	// Fetch existing data
	if loadData {
		raw, err := s.Storage.Get(id)
		...
	}
        ...
}
```

```go
// middleware/session/config.go
// Config defines the config for middleware.
type Config struct {
        ...
	// Storage interface to store the session data
	// Optional. Default value memory.New()
	Storage fiber.Storage
        ...
}
```

```go
// app.go
// Storage interface for communicating with different database/key-value
// providers
type Storage interface {
	// Get gets the value for the given key.
	// It returns ErrNotFound if the storage does not contain the key.
	Get(key string) ([]byte, error)
        ...
}
```

```go
// internal/storage/memory/memory.go
// Get value by key
func (s *Storage) Get(key string) ([]byte, error) {
	if len(key) <= 0 {
		return nil, nil
	}
	s.mux.RLock()
	v, ok := s.db[key]
	s.mux.RUnlock()
	if !ok || v.expiry != 0 && v.expiry <= time.Now().Unix() {
		return nil, nil
	}

	return v.data, nil
}
```

ReneWerner87:
@Fenny can take a look

wirepair:
+1 to this, I also ran into this when trying to use https://github.com/dgraph-io/ristretto as a backend cache, only by returning nil for the object when a key doesn't exist will it return a valid session object.

PR #1570, opened 2021-10-09
🩹 #1406: Added ErrNotFound on return when Storage Cache didn't found th…

mr-aboutin:
As mentioned in #1406 Return for the `Storage.Get` was inconsistent with the documentation from the interface `fiber.Storage`

Current: 
If `store.Get("some_key_that_doesnt_exist")` get called and no Unexpired key is found. return is `nil, nil`

After Fix:
If `store.Get("some_key_that_doesnt_exist")` get called and no Unexpired key is found. return is `nil, fiber.ErrNotFound`

### DIFF
--- a/internal/storage/memory/memory.go
+++ b/internal/storage/memory/memory.go
@@ -46,7 +46,6 @@ func (s *Storage) Get(key string) ([]byte, error) {
 	s.mux.RUnlock()
 	if !ok || v.expiry != 0 && v.expiry <= uint32(time.Now().Unix()) {
 		return nil, fiber.ErrNotFound
-		//return nil, nil
 	}
 
 	return v.data, nil


ReneWerner87:
@mr-aboutin please check the test cases and the code of use

mr-aboutin:
Well technically, the Storage struct in `gofiber/storage` is consistent with Documentation:
```Go
// Storage interface for communicating with different database/key-value
// providers. Visit https://github.com/gofiber/storage for more info.
type Storage interface {
	// Get gets the value for the given key.
	// `nil, nil` is returned when the key does not exist
	Get(key string) ([]byte, error)
```

Wich was changed with [Commit ✏ return nil for notfound](https://github.com/gofiber/storage/commit/ab94351a60f368d316c83f6d937a6b2e16f4a8cf) wich makes me questioned if the correct aproach is to bring back `ErrNotFound` or it was causing more problem in the first place. Maybe @Fenny Could shed some light on the change.

ReneWerner87:
> and don´t forget https://github.com/gofiber/storage/blob/main/memory/memory.go#L48

- [ ] this must still be observed

ReneWerner87:
@Fenny https://github.com/gofiber/storage/pull/41/files

do you remember why you had made this pull request, there you had removed the behavior that is currently requested

PR #1590, opened 2021-10-21
Add ctx.BodyParser protobuf content-type support and json protobuf unmarshal surport

li1234yun:

Enhanch `ctx.BodyParser()` function to support protobuf content-type body and support json unmarshal to protobuf.

Protobuf is google serialization protocol like json. Many microservice use the protobuf as request body, so I added the protobuf support.

**Commit formatting** 

**Notice**

There are no established content-type specifications for protobuf, I just vist stackoverflow and find `application/x-protobuf` content-type for now.

[KNOWN MEDIA TYPES](https://www.iana.org/assignments/media-types/media-types.xhtml)


ReneWerner87:
Currently not sure if we accept this, as we are always quite critical with new dependencies to other packages

If this could be solved via a third party package, we would certainly accept it (e.g. in our contrib repository  https://github.com/gofiber/contrib). 

So I have to ask the other maintainers about the core customization first.

So it might take a bit until we all have made a decision

li1234yun:
> I agree with Rene, it would be better to make a PR to the https://github.com/gofiber/contrib repo instead of this. We don't want to introduce additional dependencies, especially for users who aren't interested in protobufs

Thank you! I will try it later.

Issue #1536, opened 2021-09-19
🚀 Add Bind method to fiber.Ctx for adding var map for binding to template engine


aztecrabbit:
**Is your feature request related to a problem?**
Yes, I want add var map for template engine from middleware.

**Describe the solution you'd like**
```go
app.Use(func(c *fiber.Ctx) error {
  c.Bind(fiber.Map{
    "DEBUG": true,
    "title": "App",
  })
  return c.Next()
})
app.Use(func(c *fiber.Ctx) error {
  c.Bind(fiber.Map{
    "PATH": c.Path(),
  })
  return c.Next()
})
app.Get("/", func(c *fiber.Ctx) error {
  return c.Render("index", fiber.Map{
    "title": "Home",
  })
})
```
```
DEBUG: true
 PATH: /
title: Home
```
Optional
```go
app.Use(func(c *fiber.Ctx) error {
  c.PutBind("SITE_NAME", "App Name")
  c.PutBind("IP", c.IP())
  return c.Next()
})
app.Get("/", func(c *fiber.Ctx) error {
  fmt.Println(c.GetBind("IP"))
  return c.Render("index", fiber.Map{
    "title": "Home",
  })
})
```
```
SITE_NAME: App Name
       IP: 127.0.0.1
    title: Home
```

gitMitos:
c.Context().SetUserValue("IP",  c.IP() ) in middleware
then 
app.Get("/", func(c *fiber.Ctx) error {
  val:=c.Context().Value("IP") // plus maybe some wrap functions if needed
  return c.Render("index", fiber.Map{
    "IP": val,
  })
})

these are not droids you're looking for? or you want better "sugar" with specifically "PutBind" and "GetBind"?

aztecrabbit:
No. `c.Context().SetUserValue` and `c.Context().Value` is same like `c.Locals` do. https://docs.gofiber.io/api/ctx#locals.

ReneWerner87:
Exactly, please use the locals to populate a map and then when rendering you use the values set via the locals

aztecrabbit:
Like this?
```go
app.Get("/", func(c *fiber.Ctx) error {
  return c.Render("index", fiber.Map{
    "DEBUG": c.Locals("DEBUG"),
    "SITE_NAME": c.Locals("SITE_NAME"),
    "PATH": c.Path(),
  })
})
```
Or maybe like this
```go
app.Use(func(c *fiber.Ctx) error {
[...]

PR #1604, opened 2021-10-29
🔥 Feature: Add ctx.Bind for set default var map for template engine

aztecrabbit:

#1536 

Hi, i'm changing binding var type from `interface{}` to `Map`, is that ok?

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -239,15 +239,13 @@ func (c *Ctx) BaseURL() string {
 	return c.baseURI
 }
 
-// Set or combine default view var map for binding to template engine
-func (c *Ctx) Bind(viewVarMap Map) {
-	if c.viewVarMap != nil {
-		for k, v := range viewVarMap {
-			c.viewVarMap[k] = v
-		}
-	} else {
-		c.viewVarMap = viewVarMap
+// Set or get default view var map for binding to template engine
+func (c *Ctx) Bind(key string, value ...interface{}) interface{} {
+	if len(value) == 0 {
+		return c.viewVarMap[key]
 	}
+	c.viewVarMap[key] = value[0]
+	return value[0]
 }
 
 // Body contains the raw body submitted in a POST request.


liaohongxing:
can render be changed to:   

```go
Render(*Ctx, io.Writer, string, Map, ...string) error
```

ReneWerner87:
> can render be changed to:
> 
> ```go
> Render(*Ctx, io.Writer, string, Map, ...string) error
> ```
 why ?

ReneWerner87:
@liaohongxing why ?

aztecrabbit:
@ReneWerner87 Because we can easily implement for this feature (set default var map for tempate engine), and also no need to check `binding` arg type in every `Render` func in `gofiber/template` repo.

https://github.com/gofiber/template/blob/359ca25289df6bd214320ff864103d18051e5b01/jet/jet.go#L206-L252

ReneWerner87:
the same topic I had just answered here
https://github.com/gofiber/fiber/issues/1622#issuecomment-1002128469
think it is unnecessary, because currently everything works and simply the bindings can be extended with the values that are needed

always having everything and introducing the context here is not really in line with clean code concepts and also makes writing unittests harder and using them outside the fiber context

aztecrabbit:
> think it is unnecessary, because currently everything works and simply the bindings can be extended with the values that are needed

I'm using fiber not for api, pure using template engine. Sending session data to template engine in every route is too much work for me.

I'll close this pull request. But the feature like in this pull request will be added to fiber or not (maybe in fiber v3)? Because i think feature like this is very important for project that only using template engine.

ReneWerner87:
do not misunderstand, the "why" referred to the extension/adjustment of the render function with the context (first parameter)

the bind method is already useful and good, the only thing that prevents me from merging it is the adjustment of the type in the declaration, don't know if that causes a breaking change and then users have to make adjustments, even though it wouldn't really be necessary

aztecrabbit:
I missunderstand. I'm not seeing the source code of this repo for long time. I'm very sorry for this. Maybe i drinking too much coffee today, hahaha. Again, i'm so sorry @ReneWerner87

ReneWerner87:
is okay, does not matter :D , am always glad about every pull request

ReneWerner87:
@aztecrabbit can you check my last comments to the pull request, maybe we can make this backward compatible

liaohongxing:
I hope to change Render(io.Writer, string, interface(), ...string) error to Render(*Ctx, io.Writer, string, interface(), ...string) error in the next major version

Add a Ctx parameter, interface() can be changed to Map, it doesn’t matter

This is very helpful for me to write a custom template engine project, mainly to help me inject session, ctx.Get into the template

@ReneWerner87

aztecrabbit:
@ReneWerner87 The `interface{}` to `Map` think? I think not possible, because `interface{}` is too abstract for this. I think add this feature in fiber v3 is the best option right now

PR #1397, opened 2021-06-21
🔥 Feature: Add TrustedProxies feature

oroshnivskyy:
Add TrustedProxies feature to prevent HTTP header spoofing. This PR is intended to close #1300 issue.

PR #1614, opened 2021-11-08
Fix using IP ranges in config.TrustedProxies (#1607)

### LINKED ISSUE #1607: 🐛 Fiber will never start with big IP networks in config.TrustedProxies
**Fiber version**
2.21.0

**Issue description**
Recently I asked for a feature: #1600. Next day it was implemented by #1602. Thanks, but I think this solution is very dangerous.

Let's see how current solution works:

```go
// Iterates IP address which is between range
for ip := ip.Mask(ipnet.Mask); ipnet.Contains(ip); utils.IncrementIPRange(ip) {
    app.config.trustedProxiesMap[ip.String()] = struct{}{}
}
```

Actually, this solution just adds all IP adresses that networks contains in trustedProxiesMap, one by one.

Look at Cloudflare – most popular proxy for websites and, I think, the most popular case. What happens if we add all IPv4 Cloudflare networks to `config.TrustedProxies`? `trustedProxiesMap` will contain **1 786 880 IPs**! Although it is extremely inefficient on memory, it can still be OK. But try to add at least one IPv6 Cloudflare IP network, and fiber will never finish adding IPs to `trustedProxiesMap`.
<details>

  <summary>Spoiler</summary>
    If we add all Cloudflare IPv6 networks, fiber  need to process <b>1 109 194 275 199 700 726 309 617 091 584 IPs</b>.

</details>

**Code snippet**
Just add all cloudflare (or any other big) IP networks to config to reproduce.

<details>
<summary>Cloudflare networks</summary>

```go
[]string{
	"103.21.244.0/22",
	"103.22.200.0/22",
	"103.31.4.0/22",
	"104.16.0.0/12",
	"108.162.192.0/18",
	"131.0.72.0/22",
	"141.101.64.0/18",
	"162.158.0.0/15",
	"172.64.0.0/13",
	"173.245.48.0/20",
	"188.114.96.0/20",
	"190.93.240.0/20",
	"197.234.240.0/22",
	"198.41.128.0/17",
	"2400:cb00::/32",
	"2606:4700::/32",
	"2803:f800::/32",
	"2405:b500::/32",
	"2405:8100::/32",
	"2a06:98c0::/29",
	"2c0f:f248::/32",
}
```

</details>

hi019:
Closes #1607


app.go
```diff
@@ -348,8 +348,9 @@ type Config struct {
 	// Read EnableTrustedProxyCheck doc.
 	//
 	// Default: []string
-	TrustedProxies    []string `json:"trusted_proxies"`
-	trustedProxiesMap map[string]struct{}
+	TrustedProxies        []string `json:"trusted_proxies"`
+	trustedProxiesMap     map[string]struct{}
```
hi019, line 352: @ReneWerner87 do you know the reasoning for using `map[string]struct{}` over `[]string` here?

ReneWerner87: not really, we should ask the creator 

https://github.com/gofiber/fiber/pull/1397/files
@oroshnivskyy

OlehRoshnivskyi: To decrease the time complexity of IsProxyTrusted function.

app.go
```diff
@@ -348,8 +348,9 @@ type Config struct {
...
+	trustedProxyRangesMap []*net.IPNet
```
sadfun, line 353: Actually it's not map, it's slice/array/list or something :)

ReneWerner87 (approved):

renanbastos93 (approved):

PR #1616, opened 2021-11-09
(readme): fixed broken basic routing examples

yusufpapurcu:

Basic routing example is outdated.


ReneWerner87 (changes requested):
https://github.com/gofiber/fiber/blob/master/app.go#L40
nope
its not outdated

PR #1623, opened 2021-11-14
Recover middleware: Added, fiber.Ctx as a first paremeter to StackTra…

turgayozgur:

Added, fiber.Ctx as a first paremeter of the recover middleware StackTraceHandler.

Adding a fiber.Ctx as a parameter of StackTraceHandler function allow us to set stack trace string to c.Locals(). So, we will be able to log the stack trace on the ErrorHandler instead of putting that log twice for the stack trace.

### DIFF
--- a/middleware/recover/README.md
+++ b/middleware/recover/README.md
@@ -50,7 +50,7 @@ type Config struct {
 	// StackTraceHandler defines a function to handle stack trace
 	//
 	// Optional. Default: defaultStackTraceHandler
-	StackTraceHandler func(e interface{})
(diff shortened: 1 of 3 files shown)


turgayozgur:
https://github.com/gofiber/docs/pull/216

hi019:
Wouldn't this be a breaking change? Maybe we could create a new method like `StackTraceHandlerWithContext`

ReneWerner87:
Yes, should be a breaking change, would mention that in the release notes and refrain from using another method as it will only affect a few

Issue #825, opened 2020-09-24
🐛 CSRF cookie is SameSite=Lax


falsandtru:
**Fiber version**
2.0.2
**Issue description**
Should be Strict by default.
**Code snippet**

```
Set-Cookie: _csrf=0e854b50-dc6e-46c7-a6f2-30b1dffa49f1; expires=Fri, 25 Sep 2020 08:20:04 GMT; path=/; SameSite=Lax
```

hi019:
I'll make a PR for this soon. For anyone wondering


> Lax
> Cookies are allowed to be sent with top-level navigations and will be sent along with GET request initiated by third party website. This is the default value in modern browsers.

> Strict
> Cookies will only be sent in a first-party context and not be sent along with requests initiated by third party websites.

Fenny:
If the `SameSite` attribute is set to `Strict`, then the browser will not include the cookie in any requests that originate from another site. This is the most defensive option, but it can impair the user experience, because if a logged-in user follows a third-party link to a site, then they will appear not to be logged in, and will need to log in again before interacting with the site in the normal way. See more here: https://portswigger.net/web-security/csrf/samesite-cookies

I made a PR <https://github.com/gofiber/fiber/compare/master...Fenny:master#diff-a1d2da43df7c00f48723bb4751a6c393>  that allows you to override the default `SameSite` attribute in the `csrf` middleware.

```go
app.Use(csrf.New(csrf.Config{
    Cookie: &fiber.Cookie{
        SameSite: "Strict",
    }
}))
```

falsandtru:
> if a logged-in user follows a third-party link to a site, then they will appear not to be logged in, and will need to log in again before interacting with the site in the normal way.

CSRF cookie's SameSite value shoudn't affect the other cookies. So the session cookie should independently have the different SameSite value.

Fenny:
@falsandtru you are right, I went over it too quickly and forgot about that. I will make a PR so the csrf mw will default to `Strict` 👍

falsandtru:
Why not fixed? `Lax` can't prevent clickjacking.

falsandtru:
Looks like the source code is fixed but the release note is wrong.

https://github.com/gofiber/fiber/releases/tag/v2.0.3
> csrf middleware now sets SameSite=Lax by default #825

PR #1640, opened 2021-12-01
fix(middleware/csrf): CookieSameSite default "Lax"

### LINKED ISSUE #1639: 🐛 error(middleware/csrf): CookieSameSite default should be "Lax"
Default is "Strict", it should be "Lax"

Browser defaults are Lax. Fiber using a default of "Strict" is unadvisable as it will cause unexpected behavior. Strict requires HTTP first-party context and disregards requests initiated by third parties. So after navigating to the site from a third party site the safe HTTP methods (GET, HEAD, or OPTIONS) will not set the cookie. For example, a login page would not work by default when navigating to it from google or any other non-first party link.

"With Strict, the cookie is only sent to the site where it originated. Lax is similar, except that cookies are sent when the user navigates to the cookie's origin site. For example, by following a link from an external site. None specifies that cookies are sent on both originating and cross-site requests, but only in secure contexts (i.e., if SameSite=None then the Secure attribute must also be set). If no SameSite attribute is set, the cookie is treated as Lax."

https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#samesite_attribute

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite

The behaviour the team likely sought to prevent is SameSite=None, which blocks cookie set on cross-origin requests.

sixcolors:
Default is "Strict", it should be "Lax"

Browser defaults are Lax. Fiber using a default of "Strict" is unadvisable as it will cause unexpected behavior. Strict requires HTTP first-party context and disregards requests initiated by third parties. So after navigating to the site from a third party site the safe HTTP methods (GET, HEAD, or OPTIONS) will not set the cookie. For example, a login page would not work by default when navigating to it from google or any other non-first party link.

"With Strict, the cookie is only sent to the site where it originated. Lax is similar, except that cookies are sent when the user navigates to the cookie's origin site. For example, by following a link from an external site. None specifies that cookies are sent on both originating and cross-site requests, but only in secure contexts (i.e., if SameSite=None then the Secure attribute must also be set). If no SameSite attribute is set, the cookie is treated as Lax."

https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#samesite_attribute

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite

The behaviour the team likely sought to prevent is SameSite=None, which blocks cookie set on cross-origin requests.

This PR closes #1639

### DIFF
--- a/middleware/csrf/README.md
+++ b/middleware/csrf/README.md
@@ -46,7 +46,7 @@ app.Use(csrf.New()) // Default config
 app.Use(csrf.New(csrf.Config{
 	KeyLookup:      "header:X-Csrf-Token",
 	CookieName:     "csrf_",
-	CookieSameSite: "Strict",
+	CookieSameSite: "Lax",
 	Expiration:     1 * time.Hour,
 	KeyGenerator:   utils.UUID,
 }))
@@ -106,7 +106,7 @@ type Config struct {
 	CookieHTTPOnly bool
 
 	// Indicates if CSRF cookie is requested by SameSite.
-	// Optional. Default value "Strict".
+	// Optional. Default value "Lax".
 	CookieSameSite string
 
 	// Expiration is the duration before csrf token will expire
@@ -138,7 +138,7 @@ type Config struct {
 var ConfigDefault = Config{
 	KeyLookup:      "header:X-Csrf-Token",
 	CookieName:     "csrf_",
-	CookieSameSite: "Strict",
+	CookieSameSite: "Lax",
 	Expiration:     1 * time.Hour,
 	KeyGenerator:   utils.UUID,
 }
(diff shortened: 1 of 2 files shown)


ReneWerner87:
https://github.com/gofiber/fiber/issues/825

it was lax before and was changed here to strict

sixcolors:
> #825
> 
> it was lax before and was changed here to strict

also per issue:

Ack prior; I still think arguments in comment are valid. Further default for other mw using CookieSameSite, session, is Lax.

ReneWerner87:
doc changed https://github.com/gofiber/docs/commit/b67c5313f322a56b08874e17f292aef5b125f32a

PR #1667, opened 2021-12-23
🔧 fix(middleware/csrf): unmatched token returns nil error

### LINKED ISSUE #1666: 🐛 bug(middleware/csrf): unmatched token returns nil error
**Fiber version**
v2.23.0
**Issue description**
middleware/csrf: unmatched token returns nil error

sixcolors:
This PR closes #1666 

when the following is true

``if manager.getRaw(token) == nil ``

err will be nil otherwise it would have returned on line 40.

So we must set the error before calling `cfg.ErrorHandler(c, err)`

### DIFF
--- a/middleware/csrf/csrf.go
+++ b/middleware/csrf/csrf.go
@@ -7,6 +7,10 @@ import (
 	"github.com/gofiber/fiber/v2"
 )
 
+var (
+	errTokenNotFound = errors.New("csrf token not found")
+)
+
 // New creates a new middleware handler
 func New(config ...Config) fiber.Handler {
 	// Set default config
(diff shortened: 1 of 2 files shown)


ReneWerner87:
thx, could you please. provide a unittest for this

sixcolors:
The current unit tests work unchanged as the error handler was still called, but just had a nil err value. The default err handler still returned 403 as it did not check the err value. 

Do you have a suggestion of what unit test change you’d like for retiring a non nil error?

ReneWerner87:
since you have made an adjustment, the expectation should also change, otherwise the unittests that passed before are incomplete, if despite the change the external parameters do not change

i'm for extending the test to check the error text, because that's what you expect, if the token doesn't match, you want to be informed about that too

liaohongxing:
I hope to change Render(io.Writer, string, interface(), ...string) error to Render(*Ctx, io.Writer, string, interface(), ...string) error in the next major version

Add a Ctx parameter, interface() can be changed to Map, it doesn’t matter

@ReneWerner87

sixcolors:
> since you have made an adjustment, the expectation should also change, otherwise the unittests that passed before are incomplete, if despite the change the external parameters do not change
> 
> i'm for extending the test to check the error text, because that's what you expect, if the token doesn't match, you want to be informed about that too

ack

ReneWerner87 (approved):

PR #1671, opened 2021-12-28
🔥Feature: add function to check if request came from localhost

### LINKED ISSUE #1629: 🚀 Function to check if request came from localhost
Could you build into the library a simple function that guarantees that the request came from a local host?
That would be handy than writing your own bike every time, which often isn't quite right enough yet.
Let's say this option doesn't work correctly.
```go
func GetIPs(c *fiber.Ctx) []string {
    ips := c.IPs()
    if len(ips) == 0 {
        ips = append(ips, c.IP())
    }
    return ips
}

func NextIfLocal(c *fiber.Ctx) bool {
    return GetIPs(c)[0] == "127.0.0.1"
}
```

This one's better:
```go
func IsLocalHost(address string) bool {
	/*
		fmt.Println(IsLocalHost("::1"))
		fmt.Println(IsLocalHost("::1:3001"))
		fmt.Println(IsLocalHost("127.0.0.1:80"))
		fmt.Println(IsLocalHost("0.0.0.0"))
		fmt.Println(IsLocalHost("178.10.10.10"))  // false
	*/

	localHosts := []string{"localhost", "127.0.0.1", "0.0.0.0", "::1"}
	for _, h := range localHosts {
		if strings.Contains(address, h) {
			return true
		}
	}
	return false
}

func NextIfLocal(c *fiber.Ctx) bool {
	return IsLocalHost(GetIPs(c)[0])
}
```

vecpeng:
Close #1629 .
Usages
```
func handleSomething(c *fiber.Ctx) error {
	// Print true when request came from localhost
	fmt.Println(c.IsFromLocal())
	return nil
}
```

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -1289,3 +1289,23 @@ func (c *Ctx) IsProxyTrusted() bool {
 	_, trustProxy := c.app.config.trustedProxiesMap[c.fasthttp.RemoteIP().String()]
 	return trustProxy
 }
+
+// IsLocalHost will return true if address is a localhost address.
+func (c *Ctx) isLocalHost(address string) bool {
+	localHosts := []string{"127.0.0.1", "0.0.0.0", "::1"}
+	for _, h := range localHosts {
+		if strings.Contains(address, h) {
+			return true
+		}
+	}
+	return false
+}
+
+// IsFromLocal will return true if request came from local.
+func (c *Ctx) IsFromLocal() bool {
+	ips := c.IPs()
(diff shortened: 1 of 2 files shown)


ReneWerner87:
thanks for the feature

please provide a part in our docs(@vecpeng )
https://github.com/gofiber/docs/blob/master/api/ctx.md

vecpeng:
OK, I open a pull request here. https://github.com/gofiber/docs/pull/218

PR #1672, opened 2021-12-28
Add method the get registered handler count.

efectn:
We need to get handler count in some cases (like custom startup message). So i added method to get handler count.


ReneWerner87 (approved):

ReneWerner87:
Thx, don't forget the docs repository

PR #1684, opened 2022-01-02
✨feature: add default address for app.Listen()

vecpeng:
I add a default address for app.Listen().
If there is no given address, it will use environment variable or serve at port ":8080" by default.
You can use this:
```go
app = fiber.New()
app.Listen() // Serve at default port ":8080"
```


ReneWerner87:
Thanks for your effort, but I think a default port is not necessary and only causes errors or confusion

PR #1693, opened 2022-01-07
✨feature: Pass all locals to ctx.Render

kingdevnl:
**✨feature: Pass all locals to ctx.Render**

This will by default pass all locals set user ctx.Local to the Template engine.

This could be usefull for setting variables using a middleware.

### DIFF
--- a/app.go
+++ b/app.go
@@ -183,6 +183,11 @@ type Config struct {
 	// Default: ""
 	ViewsLayout string `json:"views_layout"`
 
+	// PassLocalsToViews Enables passing of the locals set on a fiber.Ctx to the template engine
+	//
+	// Default: false
+	PassLocalsToViews bool `json:"pass_locals_to_views"`
+
 	// The amount of time allowed to read the full request including body.
 	// It is reset after the request handler has returned.
 	// The connection's read deadline is reset when the connection opens.
--- a/ctx.go
+++ b/ctx.go
@@ -1017,16 +1017,20 @@ func (c *Ctx) Render(name string, bind interface{}, layouts ...string) error {
 	buf := bytebufferpool.Get()
 	defer bytebufferpool.Put(buf)
 
-	// Safely cast the bind interface to a ma
-	bindMap, ok := bind.(Map)
-	// Check if the bind is a map
-	if ok {
-		// Loop through each local and set it in the map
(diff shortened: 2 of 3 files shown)


ReneWerner87 (commented):
Please add a testcase and something in the documentation (https://github.com/gofiber/docs)

ReneWerner87 (approved):

app.go
```diff
@@ -183,6 +183,11 @@ type Config struct {
 	// Default: ""
 	ViewsLayout string `json:"views_layout"`
 
+	// PassLocalsToViews Enables passing of the locals set on a fiber.Ctx to the template engine
```
hi019, line 186: We should probably mention that the Locals will be accessible to the template just like any other passed value, and that they will overwrite keys manually passed to `Render`.

hi019 (changes requested):

ctx.go
```diff
@@ -1017,6 +1017,22 @@ func (c *Ctx) Render(name string, bind interface{}, layouts ...string) error {
...
 
+	// Check if the PassLocalsToViews option is enabled (By default it is disabled)
+	if c.app.config.PassLocalsToViews {
+		// Safely cast the bind interface to a map
+		bindMap, ok := bind.(Map)
+		// Check if the bind is a map
+		if ok {
+			// Loop through each local and set it in the map
+			c.fasthttp.VisitUserValues(func(key []byte, val interface{}) {
+				bindMap[string(key)] = val
```
ReneWerner87, line 1028: the map that is being handed in should only be expanded, not completely replaced

furthermore the values which are given directly to the render function should have more value

i.e. a value which is in the locals with key "abc" and one which is directly passed to the render function, the one which is passed to the render function should win

ReneWerner87 (changes requested):

ctx.go
ReneWerner87: @kingdevnl

ReneWerner87:
@kingdevnl

kingdevnl:
Will have a look once I'm home, had quite a busy week so wasn't able to check it before

ReneWerner87:
thx

ReneWerner87 (approved):

PR #1677, opened 2021-12-29
✨feature: print all routes message when server starts

vecpeng:
Closes #1675 
It will print something like this:
image

PR #1702, opened 2022-01-10
change json tag for EnablePrintRoutes

vecpeng:
`json:"print_routes"` => `json:"enable_print_routes"`

### DIFF
--- a/app.go
+++ b/app.go
@@ -356,7 +356,7 @@ type Config struct {
 
 	//If set to true, will print all routes with their method, path and handler.
 	// Default: false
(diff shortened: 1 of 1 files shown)


ReneWerner87:
this is a breaking change, please always consider at the beginning
(initial implementation https://github.com/gofiber/fiber/pull/1677)

this time it will still work, because not many use this feature and also don't use json representation, but next time please fix it right away

vecpeng:
I'm so sorry for for carelessness. I won't make same mistake next time.

PR #1713, opened 2022-01-14
:fire: Feature: Add Route() method like Chi router.

efectn:
Add Route method like [Chi router](https://github.com/go-chi/chi#router-interface).

**Signature:**

`Route(prefix string, fn func(router Router), name ...string) Router`

**Using Example:** 

```
app.Route("/test", func(router fiber.Router) {
	router.Get("/john", func(c *fiber.Ctx) error {
		return c.SendString("doe")
	}).Name("john")
}, "test.")
```

### DIFF
--- a/app.go
+++ b/app.go
@@ -693,6 +693,21 @@ func (app *App) Group(prefix string, handlers ...Handler) Router {
 	return &Group{prefix: prefix, app: app}
 }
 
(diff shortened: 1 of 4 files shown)


ReneWerner87:
don´t forget the doc repository

@gofiber/maintainers what do you think?

PR #1728, opened 2022-01-21
Fix NewErrors() and Improve NewError()

balcieren:
If error message's type happens interface{}, It can return multiple error strings etc.

### NewErrors()
```go
func main() {
	app := New(fiber.Config{
		ErrorHandler: func(c *fiber.Ctx, err error) error {
			code := fiber.StatusInternalServerError
			var messages interface{}

			if e, ok := err.(*fiber.Error); ok {
				code = e.Code
				messages = e.Message
			}

			return c.Status(code).JSON(fiber.Map{
				"status_code": code,
				"status_text": utils.StatusMessage(code),
				"messages":    messages,
			})
		},
	})

		return NewErrors(fiber.StatusBadRequest, "validatien error 1", "validation error 2")
	})

	app.Listen(":3000")
}
```

### JSON is going to be 
```json
{
  "messages": [
    "validatien error 1",
    "validation error 2"
  ],
  "status_code": 400,
  "status_text": "Bad Request"
}
```
---
### Different Example
```go
func main() {
	app := New(fiber.Config{
		ErrorHandler: func(c *fiber.Ctx, err error) error {
			code := fiber.StatusInternalServerError
			var messages interface{}

			if e, ok := err.(*fiber.Error); ok {
				code = e.Code
				messages = e.Message
			}

			return c.Status(code).JSON(fiber.Map{
				"status_code": code,
				"status_text": utils.StatusMessage(code),
				"messages":    messages,
			})
		},
	})

		return NewErrors(StatusBadRequest,
			fiber.Map{
				"type":        "validation",
				"description": "password must be longer than 6 characters",
			},
			fiber.Map{
				"type":        "validation",
				"description": "invalid e-mail",
			})
	})

	app.Listen(":3000")
}
```

### JSON is going to be 
```json
{
  "messages": [
    {
      "description": "password must be longer than 6 characters",
      "type": "validation"
    },
    {
      "description": "invalid e-mail",
      "type": "validation"
    }
  ],
  "status_code": 400,
  "status_text": "Bad Request"
}
```

### DIFF
--- a/app.go
+++ b/app.go
@@ -705,7 +705,18 @@ func NewError(code int, message ...interface{}) *Error {
 		Message: utils.StatusMessage(code),
 	}
 	if len(message) > 0 {
-		e.Message = message
+		e.Message = message[0]
+	}
+	return e
+}
+
+func NewErrors(code int, messages ...interface{}) *Error {
+	e := &Error{
+		Code:    code,
+		Message: utils.StatusMessage(code),
+	}
+	if len(messages) > 0 {
+		e.Message = messages
 	}
 	return e
 }
--- a/app.go
+++ b/app.go
@@ -84,13 +84,8 @@ type ErrorHandler = func(*Ctx, error) error
 
 // Error represents an error that occurred while handling a request.
 type Error struct {
-	Code    int    `json:"code"`
-	Message string `json:"message"`
-}
-
-type Errors struct {
-	Code     int      `json:"code"`
-	Messages []string `json:"messages"`
+	Code    int         `json:"code"`
+	Message interface{} `json:"message"`
 }
 
 // App denotes the Fiber application.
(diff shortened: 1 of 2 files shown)


vecpeng:
It seems good but the only problem is that the type of message has been changed, which may cause backward compatibility issues.
Maybe we should wait maintainers' review.
@ReneWerner87 @efectn @hi019 what do you think?

balcieren:
> It seems good but the only problem is that the type of message has been changed, which may cause backward compatibility issues. Maybe we should wait maintainers' review. @ReneWerner87 @efectn @hi019 what do you think?

Echo also uses interface{} for message

app.go
```diff
@@ -84,8 +84,8 @@ type ErrorHandler = func(*Ctx, error) error
 
 // Error represents an error that occurred while handling a request.
 type Error struct {
-	Code    int    `json:"code"`
-	Message string `json:"message"`
+	Code    int         `json:"code"`
```
Mridul11, line 87: nit: we can get rid of extra spaces

balcieren: interface{} is longer than string so editor has aligned automatically it.

balcieren:
@ReneWerner87 I think it is not big update and all checks have passed. What do you think ?

ReneWerner87:
I look at it today, had the week other tasks to do

ReneWerner87 (approved):

balcieren:
@ReneWerner87 I have updated doc, [PR](https://github.com/gofiber/docs/pull/232)

ReneWerner87:
Thx

PR #1730, opened 2022-01-23
added HEAD method to caching

jessequinn:
When using the curl -I statement, the HEAD method, is invoked. However, only the GET method is being cached. To overcome this limitation an additional clause was added to the `cache.go` to all the HEAD method to also cache.


ReneWerner87:
Unfortunately, the adaptation has addressed a wider memory allocation, making the process slower

I am in exchange with the author of the pr and we are trying to reduce this allocation

ReneWerner87 (approved):

PR #1735, opened 2022-01-28
Swagger middleware added

ziyaozclk:
The added middleware handles the Swagger JSON file. I use [Fiber](https://github.com/gofiber/fiber) in many projects and I need it very much. I'm sure it will be useful to many people.


efectn:
Can you create this PR under the https://github.com/gofiber/contrib instead of here. Adding this to the core isn't good idea i think. Thanks

ziyaozclk:
@efectn You're right. I didn't know this library. Thanks for your suggestion

PR #1737, opened 2022-01-29
:sparkles: feature: add Next to Pprof and Expvar middlewares.

efectn:


ReneWerner87 (commented):
Don't forget the readme files inside of the middleware folder and the document repository.

Thought by the way it is not necessary since we have the new skip middleware.
But if you have made the effort, I can take over the sense of unification.

ReneWerner87 (approved):

Issue #1398, opened 2021-06-22
🤗  Is there a possibility for an internal redirect ?


fufuok:
**Question description**

Need to redirect some dynamic interfaces to the specified route internally, but it is not a fixed rewrite middleware.

I did not find an example, please help, thanks.

**Code snippet**

```go
package main

import (
	"log"

	"github.com/gofiber/fiber/v2"
)

func main() {
	app := fiber.New()

	v1 := app.Group("/v1/:name")
	{
		v1.Post("", V1APIHandler)
		v1.Get("", V1APIHandler)
	}

	app.Post("/test", func(c *fiber.Ctx) error {
		log.Println("/test")
		return c.SendString(string(c.Body()))
	})

	_ = app.Listen(":3000")
}

func V1APIHandler(c *fiber.Ctx) error {
	if c.Method() == "POST" && c.Params("name") == "test" {
		log.Println("post and test")
		// Need internal redirection, similar to gin, instead of 307 Temporary Redirect, ths.
		// c.Request.URL.Path = "/test"
		// app.HandleContext(c)
		return c.Redirect("/test", fiber.StatusTemporaryRedirect)
	}
	return c.SendString("Hello, " + c.Path() + " 👋!")
}
```

**curl test**

```shell
curl -v -XPOST 'http://127.0.0.1:3000/v1/test' -d '{"a":123}'
Note: Unnecessary use of -X or --request, POST is already inferred.
*   Trying 127.0.0.1:3000...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 3000 (#0)
> POST /v1/test HTTP/1.1
> Host: 127.0.0.1:3000
> User-Agent: curl/7.68.0
> Accept: */*
> Content-Length: 9
> Content-Type: application/x-www-form-urlencoded
> 
* upload completely sent off: 9 out of 9 bytes
* Mark bundle as not supporting multiuse
< HTTP/1.1 307 Temporary Redirect
< Date: Tue, 22 Jun 2021 03:31:30 GMT
< Content-Length: 0
< Location: /test
< 
* Connection #0 to host 127.0.0.1 left intact
```

***307 Temporary Redirect, requires the client to actively re-initiate the request.***

This is not what I expected, I hope it looks like this:

------

**gin demo**

```go
package main

import (
	"log"

	"github.com/gin-gonic/gin"
)

func main() {
	app := gin.Default()

	v1 := app.Group("/v1/:name", V1APIHandler(app))
	{
[...]

PR #1739, opened 2022-01-31
🔥 Feature: Add ability to restart route handling

### LINKED ISSUE #1731: 🐛 Internal redirect does not reset route and handler index
**Fiber version**

v2.25.0

**Issue description**

I found this issue while searching for a way to do internal redirects. #1398 seems to suggest that this is not supported, however, there is the out-of-tree [rewrite middleware](https://github.com/gofiber/rewrite) that is exactly doing that. It essentially uses the `c.Path()` method to set a new path. The code below shows a simplified example.

As also mentioned in #1398, an internal redirect would need to reset `c.indexRoute` (and `c.indexHandler`). This is currently not done which could lead to unexpected routing decisions. In the example below requesting `GET /old` would return `Not Found` while I would expect it to be `new`. This is because the `/new` route was already checked before the internal redirect and is thus skipped. If instead the `/new` route would be registered after the redirect handler it would be executed.

I initially thought of resetting `c.indexRoute` and `c.indexHandler` in `c.Path()`, but this could be unexpected as well, e.g. handlers might be executed twice. Making this an explicit choice by the user might be more appropriate.

What do you think about adding a `Restart` method to the context that resets `c.indexRoute` and then executes `c.app.next()`?

**Code snippet**

```go
package main

func main() {
	app := fiber.New()

	app.Get("/new", func(c *fiber.Ctx) error {
		return c.SendString("new")
	})

		c.Path("/new")
		return c.Next()
	})

		// e.g. catch-all for 404
		return fiber.ErrNotFound
	})

	app.Listen(":3000")
}
```

mtneug:

Adds the ability to restart the route handling e.g. after an internal redirect.

Fixes #1731

The example from #1731 is changed in the following way:

```go
package main

func main() {
	app := fiber.New()

	app.Get("/new", func(c *fiber.Ctx) error {
		return c.SendString("new")
	})

		c.Path("/new")
		// instead of c.Next() as this would execute the catch-all handler
		// note that this example works because GET /new is registered first. Infinit loops are possible with RestartRouting.
		return c.RestartRouting()
	})

		// e.g. catch-all for 404
		return fiber.ErrNotFound
	})

	app.Listen(":3000")
}
```

Edit 1: changed method name to `RestartRouting`

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -782,6 +782,14 @@ func (c *Ctx) Next() (err error) {
 	return err
 }
 
+// RestartRouting instead of going to the next handler. This may be usefull after
+// changing the request path. Note that handlers might be executed again.
+func (c *Ctx) RestartRouting() error {
+	c.indexRoute = -1
+	_, err := c.app.next(c)
+	return err
+}
+
 // OriginalURL contains the original request URL.
 // Returned value is only valid within the handler. Do not store any references.
 // Make copies or use the Immutable setting to use the value outside the Handler.
--- a/ctx_test.go
+++ b/ctx_test.go
@@ -2065,8 +2065,8 @@ func Test_Ctx_RenderWithLocalsAndBinding(t *testing.T) {
 	utils.AssertEqual(t, "<h1>Hello, World!</h1>", string(c.Response().Body()))
 }
 
-// go test -run Test_Ctx_Restart
-func Test_Ctx_Restart(t *testing.T) {
+// go test -run Test_Ctx_RestartRouting
+func Test_Ctx_RestartRouting(t *testing.T) {
 	app := New()
 	calls := 0
 	app.Get("/", func(c *Ctx) error {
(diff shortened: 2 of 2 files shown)


ctx.go
```diff
@@ -782,6 +782,14 @@ func (c *Ctx) Next() (err error) {
 	return err
 }
 
+// Restart request handling instead of going to next handler. This may be usefull
+// after changing the request path. Note that handlers might be executed again.
+func (c *Ctx) Restart() error {
```
ReneWerner87, line 787: can we use a more descriptive name

at first glance you can not see what should be restarted

maybe RestartRouting

ReneWerner87:
thanks for the pull request, had such an idea too
https://github.com/gofiber/fiber/issues/1398#issuecomment-865626045
in my imagination i would have combined the restart and the setting of the route in one method and called the whole thing internalRedirect

ctx.go
mtneug: Sure. I thought of this as an alternative to `return c.Next()`; therefore the shorter name.

ReneWerner87:
@mtneug

mtneug:
@ReneWerner87: I already renamed it to `RestartRouting`. Have I missed something?

ReneWerner87:
oh I had not seen

can we rename the test cases too ?

ReneWerner87 (approved):

ReneWerner87:
thx

PR #1742, opened 2022-02-02
Doc(limiter): clarify variable name 'expire'

NorbertHauriel:
- The variable 'expire' doesn't imply why it is different from 'expiration', so I made its name more specific.
- Comments mentioning it were changed accordingly.

### DIFF
--- a/middleware/limiter/limited_fixed.go
+++ b/middleware/limiter/limited_fixed.go
@@ -64,7 +64,7 @@ func (FixedWindow) New(cfg Config) fiber.Handler {
 		e.currHits++
 
 		// Calculate when it resets in seconds
-		expire := e.exp - ts
+		resetInSec := e.exp - ts
 
 		// Set how many hits we have left
 		remaining := cfg.Max - e.currHits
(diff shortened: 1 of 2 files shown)


middleware/limiter/limited_fixed.go
```diff
@@ -79,7 +76,7 @@ func (FixedWindow) New(cfg Config) fiber.Handler {
 		if remaining < 0 {
 			// Return response with Retry-After header
 			// https://tools.ietf.org/html/rfc6584
-			c.Set(fiber.HeaderRetryAfter, strconv.FormatUint(expire, 10))
+			c.Set(fiber.HeaderRetryAfter, strconv.FormatUint(expiration, 10))
```
ReneWerner87, line 79: think this is wrong, here should be the remaining time and not the complete time 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After

NorbertHauriel: It is wrong indeed. 122cbfeeaca88dd8abf78fe7702d7a978bdf09bf is the only commit now, it holds the proposed doc fix for both window limiters.

ReneWerner87 (approved):

Issue #1276, opened 2021-04-10
🚀 Mounted app views


yekurtal:
**Is your feature request related to a problem?**
Yes, when mounting an app with views the views are not rendered. See #996.

**Describe the solution you'd like**
Mounted app routes use the mounted app views.

**Describe alternatives you've considered**
Setting views in main app.

**Additional context**
None?

ReneWerner87:
Can you provide a reproducible example, that would speed up troubleshooting tremendously

ReneWerner87:
@yekurtal

yekurtal:
Github didn't notify me about your first request, sorry about that.

Here is a simple mounted app with embedded template:

`/web/template/hello_world.gohtml`
`<h1>Hello {{ .Name }}!</h1>`

`/main.go`
`
//go:embed web/template/*
var template embed.FS

func main() {
	sub := fiber.New(fiber.Config{
		Views: html.NewFileSystem(tool.SubAndWrapFS(template, "web/template"), ".gohtml"),
	})

	sub.Get("/:name", func(ctx *fiber.Ctx) error {
		return ctx.Render("hello_world", fiber.Map{
			"Name": ctx.Params("name"),
		})
	})

	app := fiber.New()

	app.Mount("/hello", sub)

	if err := app.Listen("localhost:5000"); err != nil {
		log.Fatal(err)
	}
}
`

Hope this helps.

ReneWerner87:
okay we will check this

gzuidhof:
I'd like to add that I ran into the same issue/shortcoming. It seems that after an app is mounted the main app's config.Views is used.

Peeking into `app.Mount`'s implementation it looks like only the routes get copied over, does this mean that the config of the mounted app gets ignored in general? Note I may be misunderstanding!

gzuidhof:
This may be useful for you as well @yekurtal, I pulled the code from `ctx.Render` into its own small renderer:
```go
package templaterenderer

import (
	"github.com/gofiber/fiber/v2"
	"github.com/valyala/bytebufferpool"
)

type TemplateRenderer struct {
	views fiber.Views
}

func NewRenderer(v fiber.Views) TemplateRenderer {
	return TemplateRenderer{
		views: v,
	}
}

func (r *TemplateRenderer) Render(ctx *fiber.Ctx, name string, bind interface{}, layouts ...string) error {
	buf := bytebufferpool.Get()
	defer bytebufferpool.Put(buf)

[...]

PR #1749, opened 2022-02-04
:bug: fix: mounted app views

efectn:
Fixes https://github.com/gofiber/fiber/issues/1276


ReneWerner87 (changes requested):
instead of adding a list for the properties we want to take from the mounted apps, we should solve this differently

had some communication with the author of the pull request and we are working on a better variant

something like this
image
image
image
image

ReneWerner87 (approved):

PR #1752, opened 2022-02-06
Feature: Session Only Cookies

### LINKED ISSUE #1610: 🤗 why it's not possible to set  cookie for CSRF middleware expire date to session?
Hello. Why in  https://docs.gofiber.io/api/middleware/csrf middleware impossible to set session cookie? In secure context from XXS/CSRF attacks session cookies are more secure.

Or is it possible to set cookies as session in this middleware?

If I set Expiration as 0, Expiration field goes to header, but if I want to make cookies as session I should not set headers Max-Age and Expires.

Session cookie - https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#define_the_lifetime_of_a_cookie

abhi12299:
Allows setting cookies that last for just the browser session by using `SessionOnly` property in `fiber.Cookie` which ignores `Max-Age` and `Expires` headers if set to true.
Now you can set a session only cookie like so:
```go
c.Cookie(&fiber.Cookie{
    //...
    SessionOnly: true,
    // ...
})
```
This should close #1610.


ReneWerner87:
- [ ] don´t forget the documentation repository

ReneWerner87:
hmm with the feature i actually only meant the csrf middleware

thought you could set the cookie to session without this switch by setting the expires to 0

but this does not work with the csrf middleware, because there the default is used

we need the check there but a check to take the default value in case of doubt

abhi12299:
> don´t forget the documentation repository

Wanted to update after your review.

> thought you could set the cookie to session without this switch by setting the expires to 0

Oh that is actually right.. So should I revert it from there?

> we need the check there but a check to take the default value in case of doubt

Didn't get you.. do you want me to keep the `SessionOnly` key in middleware `Config` struct? If we remove this key, then in order to put Expires to 0 in the csrf middleware, we need to compare the `cfg.Expiration` to some special value (maybe like -1 or 0) and set Expires to 0, but that is probably not the right. Or alternatively, we could keep `SessionOnly` and set Expires to 0 that way. What do you suggest?

ReneWerner87:
> Or alternatively, we could keep SessionOnly and set Expires to 0 that way. What do you suggest?

Yeah, keep the config settings and do it in this way

abhi12299:
@ReneWerner87 I read about how cookies can be made session only, and I found that Expires and Max-Age have to be omitted. Setting Expires to 0 will make browser delete the cookie immediately, which you can read about [here](https://en.wikipedia.org/wiki/HTTP_cookie#Expires_and_Max-Age) and [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expires?retiredLocale=id). So there is one observation:
> thought you could set the cookie to session without this switch by setting the expires to 0

For a cookie to be session only, we cannot just set expiration to 0 which means we have to keep a flag in `ctx.Cookie` function. 
Also in case of the csrf middleware, we have to omit Expiration when `SessionOnly` is true.

What are your thoughts on this?

ReneWerner87:
https://github.com/gofiber/fiber/blob/ad1a925d38188d33c44b80c6a720c7a9d1e921cf/middleware/csrf/config.go#L150

The line I meant, for that we would need the setting

We need to know that the user differs from the default

abhi12299:
> https://github.com/gofiber/fiber/blob/ad1a925d38188d33c44b80c6a720c7a9d1e921cf/middleware/csrf/config.go#L150
> 
> The line I meant, for that we would need the setting
> 
> We need to know that the user differs from the default

@ReneWerner87 understood. I will use the setting to bypass setting expiration altogether. 
By the way, did you read through the reference links that I gave above? I think we might need a similar setting for `ctx.Cookie` struct as well.

ReneWerner87:
ok perfect

https://github.com/gofiber/fiber/pull/1752#issuecomment-1030836136

only this task is missing

Issue #1371, opened 2021-06-08
Performance issue with simple App.Use of CSRF middleware 


gitMitos:
By simply adding 

`app.Use(csrf.New(csrf.Config{
	 	KeyLookup:  "form:form_csrf",
		CookieName: "csrf_",
	 	ContextKey: "csrf",
	 }))`

i have significant drop of request/second on helloworld like app

With it i have 65k-75k req/sec on endpoint

`app.Get("/hi", func(c *fiber.Ctx) error {
		return c.SendString("hi")
	})`

if i delete it - i will be getting about 179k-181k req/sec

Is it price of having csrf middleware in general or this specific implementation is not a best one? Does anyone else have same issue?

ReneWerner87:
> Is it price of having csrf middleware in general or this specific implementation is not a best one? Does anyone else have same issue?

yes right, security usually comes at the cost of performance, because checks have to be made and data has to be stored

probably we can optimize the middleware even further, for this you or anyone else like to deposit a pull request

ReneWerner87:
here you could use the sync pool
https://github.com/gofiber/fiber/blob/master/middleware/csrf/csrf.go#L68

think storing in storage is also a big performance eater

GibranHL0:
Hey 👋 , It seems like an interesting issue.
Can I give it a shot?

hi019:
Sure @GibranHL0 , I'll assign you

GibranHL0:
Great, I'll make the proposal as soon as possible.

GibranHL0:
Hey, I've been doing some stress tests using the same [tool](https://github.com/smallnest/go-web-framework-benchmark) as in the README file.

I created a hello world like app as in the issue statement.
```go
app.Get("/hello", func(c *fiber.Ctx) error { return c.SendString("hi") })
```

The scenarios I used were

1. Fiber without CSRF enabled
2. CSRF enabled
```go
app.Use(csrf.New(csrf.Config{ KeyLookup: "form:form_csrf", CookieName: "csrf_", ContextKey: "csrf", }))
```
3. CSFR with sync.pool (I had to assign the values to the Cookie, otherwise it will fill them with the zero values)
```go
var cPool = sync.Pool{
	New: func() interface{} { return new(fiber.Cookie) },
}

cookie := cPool.Get().(*fiber.Cookie)

defer cPool.Put(cookie)

cookie.Name = cfg.CookieName
cookie.Domain = cfg.CookieDomain
[...]

PR #1756, opened 2022-02-06
:zap: change default gc interval of internal/memory

efectn:
10ms interval time causes big performance degradation. So i made it 1 second. Potentially closes https://github.com/gofiber/fiber/issues/1371. 

**Before:**
```
Running 10s test @ http://127.0.0.1:5000
  12 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     2.99s     1.91s    6.69s    52.54%
    Req/Sec       -nan      -nan   0.00      0.00%
  396049 requests in 10.00s, 94.05MB read
Requests/sec:  39612.49
Transfer/sec:      9.41MB
```

**After:**
```
Running 10s test @ http://127.0.0.1:5000
  12 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     2.12s     1.24s    4.61s    55.31%
    Req/Sec       -nan      -nan   0.00      0.00%
  602860 requests in 10.00s, 143.16MB read
Requests/sec:  60310.52
Transfer/sec:     14.32MB
```

### DIFF
--- a/internal/memory/memory.go
+++ b/internal/memory/memory.go
@@ -23,7 +23,7 @@ func New() *Storage {
 		data: make(map[string]item),
 		ts:   uint32(time.Now().Unix()),
 	}
-	go store.gc(10 * time.Millisecond)
+	go store.gc(1 * time.Second)
 	go store.updater(1 * time.Second)
 	return store
 }


ReneWerner87:
Will do the deep check tomorrow.

ReneWerner87:
@efectn can you check the memory storage in the storages repository
https://github.com/gofiber/storage/blob/main/memory/memory.go#L99
looks that the adapter is outdated ... maybe we can refresh it with the last changes

efectn:
> @efectn can you check the memory storage in the storages repository https://github.com/gofiber/storage/blob/main/memory/memory.go#L99 looks that the adapter is outdated ... maybe we can refresh it with the last changes

I'll check it weekend

Issue #1715, opened 2022-01-15
🚀 Add mTLS and fix deprecation in TLS config


ltns35:
**Is your feature request related to a problem?**
Yes, when starting the app with ListenTLS it has the following default configuration where **PreferServerCipherSuites** is deprecated by the Go team.
```go
config := &tls.Config{
			MinVersion:   tls.VersionTLS12,
			PreferServerCipherSuites: true, // deprecated and ignored
			Certificates: []tls.Certificate{
				cert,
			},
		}
```

It could be interesting to allow mTLS in the configuration because many times we have to use certificate authorization for our microservices.

**Describe the solution you'd like**
We should update the **MinVersion** to **TLS13**, remove **PreferServerCipherSuites** because is deprecated and allow to **add** the **CA**.

```go
config := &tls.Config{
			MinVersion:  tls.VersionTLS13, // changed to TLS13
			Certificates: []tls.Certificate{
				cert,
			},
                        ClientCAs:    caCertPool,
			ClientAuth:   tls.RequireAndVerifyClientCert
		}
```

**Describe alternatives you've considered**
 Maybe adding the CA for mTSL could be done with another function called ListenWithMutualTLS.

**Additional context**
Go version: 1.17.5
Fiber version: 2.24.0

PR #1761, opened 2022-02-08
Update TLS Config & Add ListenMutualTLS

efectn:
- Remove deprecations on tls.Config.
- Add ListenMutualTLS method.

**Closes:** https://github.com/gofiber/fiber/issues/1715


ReneWerner87:
- [ ] Don't forget the documentation repository

ReneWerner87 (approved):

app.go
```diff
@@ -814,8 +817,7 @@ func (app *App) ListenTLS(addr, certFile, keyFile string) error {
 			return fmt.Errorf("tls: cannot load TLS key pair from certFile=%q and keyFile=%q: %s", certFile, keyFile, err)
 		}
 		config := &tls.Config{
-			MinVersion:               tls.VersionTLS12,
-			PreferServerCipherSuites: true,
```
hi019, line 818: would this be a breaking change?

app.go
```diff
@@ -841,6 +843,66 @@ func (app *App) ListenTLS(addr, certFile, keyFile string) error {
 	return app.server.ServeTLS(ln, certFile, keyFile)
 }
 
+// ListenMutualTLS serves HTTPs requests from the given addr.
+// certFile, keyFile and clientCertFile are the paths to TLS certificate and key file.
+
+//  app.ListenMutualTLS(":8080", "./cert.pem", "./cert.key", "./client.pem")
+//  app.ListenMutualTLS(":8080", "./cert.pem", "./cert.key", "./client.pem")
```
hi019, line 850:
```suggestion
//  app.ListenMutualTLS(":8080", "./cert.pem", "./cert.key", "./client.pem")
```

app.go
```diff
@@ -814,8 +817,7 @@ func (app *App) ListenTLS(addr, certFile, keyFile string) error {
 			return fmt.Errorf("tls: cannot load TLS key pair from certFile=%q and keyFile=%q: %s", certFile, keyFile, err)
 		}
 		config := &tls.Config{
-			MinVersion:               tls.VersionTLS12,
-			PreferServerCipherSuites: true,
+			MinVersion: tls.VersionTLS13,
```
hi019, line 820: would this be a breaking change?

app.go
hi019, line 850:
```suggestion
//  app.ListenMutualTLS(":8080", "./cert.pem", "./cert.key", "./client.pem")
```

hi019 (changes requested):

app.go
efectn: I'm not sure but it would be breaking change for old certficates may be.

balcieren: `MinVersion: tls.VersionTLS12`
`MaxVersion: tls.VersionTLS13`

can it prevent breaking change ?

efectn: I think we dont need it. Also i've seen https://tip.golang.org/doc/go1.18#tls10. I think we can use tls1.2 as minimum version instead. I can revert it. @ReneWerner87 @hi019

ReneWerner87:
@hi019 please check again

hi019:
Looks good. I'll test it out today or tomorrow

hi019 (approved):

PR #1766, opened 2022-02-11
✨ feature: custom validator 

balcieren:
Usage:

```go
package main

import (  
	"github.com/go-playground/validator/v10"
	"github.com/gofiber/fiber/v2"
)

type User struct {
    Name  string `json:"name" validate:"required"`
    Email string `json:"email" validate:"required,email"`
}

type CustomValidator struct {
    validator *validator.Validate
}


func (cv *CustomValidator) Validate(i interface{}) error {
  return cv.validator.Struct(i)
}

func main() {
     Validator: &CustomValidator{validator: validator.New()},
  })

  app.Post("/users", func(c *fiber.Ctx) error {
    u := new(User)

    if err := c.BodyParser(u); err != nil {
      return fiber.NewError(StatusBadRequest, err.Error())
    }

    if err := c.Validate(u); err != nil {
      return fiber.NewError(StatusBadRequest, err.Error())
    }

    return c.JSON(u)
  })

  app.Listen(":3000")
}
```

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -1485,3 +1485,10 @@ func (c *Ctx) IsFromLocal() bool {
 	}
 	return c.isLocalHost(ips[0])
 }
+
+func (c *Ctx) Validate(i interface{}) error {
+	if c.app.config.Validator == nil {
+		return ErrValidatorNotRegistered
+	}
+	return c.app.config.Validator.Validate(i)
+}
--- a/app.go
+++ b/app.go
@@ -88,6 +88,11 @@ type Error struct {
 	Message interface{} `json:"message"`
 }
 
+// Validator is the interface that wraps the Validate function.
+type Validator interface {
+	Validate(i interface{}) error
+}
+
 // App denotes the Fiber application.
 type App struct {
 	mutex sync.Mutex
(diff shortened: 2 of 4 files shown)


ctx.go
```diff
@@ -1485,3 +1485,10 @@ func (c *Ctx) IsFromLocal() bool {
 	}
 	return c.isLocalHost(ips[0])
 }
+
+func (c *Ctx) Validate(i interface{}) error {
+	if c.app.config.Validator == nil {
+		return ErrValidatorNotRegistered
+	}
+	return c.app.config.Validator.Validate(i)
```
efectn, line 1493: We need to check here is it Ptr or Struct like https://github.com/gin-gonic/gin/blob/master/binding/default_validator.go#L55

efectn: I'm resolving this comment because of adding default validator bloats fiber core.

balcieren:
@ReneWerner87 what do you think about this pr ?

ReneWerner87:
not sure, from the point of view this is good, but i am also only one of many maintainers 

so we will probably first need a bit and discuss this

can imagine that someone could say that it is unnecessary, because you can implement it yourself without much code

so PR is in a feedback round

ReneWerner87:
regardless, thank you for your work

no matter what the decision is
unfortunately we always have to evaluate and see if the features are understandable and useful for the masses

ReneWerner87:
Feature was rejected for the time being, we thank you for the effort anyway

Issue #1767, opened 2022-02-11
🐛Backwards Incompatible Change to fiber.Error


golangaddict:
**Fiber version**
2.26.0

**Issue description**
The change referenced here [CLICK HERE](https://github.com/gofiber/fiber/commit/7cf1886df5f6041449ca082701c05bb5eb13100e#r66439045) in #1728 which was applied to a MINOR update is backwards incompatible. This change will break underlying structures that reference fiber.Error.Message using the concrete type "string".

One cannot simply imply that although the string can now be accessed via `fiber.Message.Error()` that this is a backwards-compatible change. The fact that code modification is needed by consumers in order to recorrect this issue is the exact representation of incompatibility.

This is not an argument of the effort required to fix or resolve this issue. It is merely a statement that semantic versioning is not being followed by this project if decisions/actions like this are being supported by the maintaining contributors.

Symantic versioning implies that both MINOR & PATCH updates must be done in a backwards compatible manner. This change is not backwards compatible in a typical scenario.

Changing the concrete type of publicly exposed API struct fields is bad practice for anything other than MAJOR backwards incompatible updates.

**Code snippet**

```go
package main

import "github.com/gofiber/fiber/v2"

func ErrorHandler(c *fiber.Ctx, err error) error {
	var status int
	var message string

	switch e := err.(type) {
	case *fiber.Error:
		status = e.Code
		message = e.Message // This now throws an error because the concrete type of `Message` is no longer `string`
	default:
		status = http.StatusInternalServerError
		message = err.Error()
	}

	return c.Status(status).JSON(fiber.Map{"message": message})
}
```

balcieren:
> **Fiber version** 2.26.0
> 
> **Issue description** The change referenced here [CLICK HERE](https://github.com/gofiber/fiber/commit/7cf1886df5f6041449ca082701c05bb5eb13100e#r66439045) in #1728 which was applied to a MINOR update is backwards incompatible. This change will break underlying structures that reference fiber.Error.Message using the concrete type "string".
> 
> One cannot simply imply that although the string can now be accessed via `fiber.Message.Error()` that this is a backwards-compatible change. The fact that code modification is needed by consumers in order to recorrect this issue is the exact representation of incompatibility.
> 
> This is not an argument of the effort required to fix or resolve this issue. It is merely a statement that semantic versioning is not being followed by this project if decisions/actions like this are being supported by the maintaining contributors.
> 
[...]

PR #1768, opened 2022-02-11
🩹 fix: backwards incompatible change to fiber.Error

balcieren:
fix https://github.com/gofiber/fiber/issues/1767

backward compatible

### DIFF
--- a/app.go
+++ b/app.go
@@ -84,9 +84,8 @@ type ErrorHandler = func(*Ctx, error) error
 
 // Error represents an error that occurred while handling a request.
 type Error struct {
-	Code       int         `json:"code"`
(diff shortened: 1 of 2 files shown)


liaohongxing:
This broken change should wait for V3, AnyMessage is not elegant

hi019:
I agree with the previous reply, I think we should remove the AnyMessage changes and then merge this PR. Users can always return their own custom error in a handler and then convert it to JSON in the ErrorHandler.

After that, we can do a quick release to restore compatibility.

ReneWerner87:
Am of the same opinion, let's correct the error and work backward compatible

hi019 (approved):

PR #1777, opened 2022-02-14
:sparkles: feature: add initial support for hooks

efectn:
#### Simple Usage:
```go
func main() {
	app := fiber.New()

	app.OnResponse(func(c *fiber.Ctx, m fiber.Map) error {
		fmt.Print(c.Route())

		return nil
	})

	app.OnRoute(func(c *fiber.Ctx, m fiber.Map) error {
		fmt.Print(m)

		return nil
	})

	app.OnName(func(c *fiber.Ctx, m fiber.Map) error {
		fmt.Print(m)

		return nil
	})

		return c.SendString("-")
	}).Name("index")

	app.Listen(":5000")
}
```

#### To-Do List:
- [x] Add detailed tests.
- [x] Comment lines.
- [x] Create documentations.


app.go
```diff
@@ -1302,3 +1315,101 @@ func (app *App) printRoutesMessage() {
...
+	app.hookList["onRequest"] = append(app.hookList["onRequest"], handler...)
+	app.mutex.Unlock()
+}
+
+func (app *App) executeOnRouteHooks(route Route) {
+	for _, v := range app.hookList["onRoute"] {
+		ctx := app.AcquireCtx(&fasthttp.RequestCtx{})
+		defer app.ReleaseCtx(ctx)
+
+		v(ctx, Map{"route": route})
```
efectn, line 1374: I'm not sure about what to do here. Should i skip error handling or do anything else?

ReneWerner87:
gave him some tasks to move all the code into a separate struct called hooks.go(+tests) and thus reduce the size of the file for the app

@hi019 what do you think about the different hooks

hi019:
Will look tomorrow

hi019:
Looks good, I'll test it out tomorrow. But why do we need OnResponse/OnRequest when the same thing can be done with a middleware? Also what about changing the hook signature to not return anything, so that the hook handles any errors?

efectn:
> Looks good, I'll test it out tomorrow. But why do we need OnResponse/OnRequest when the same thing can be done with a middleware? Also what about changing the hook signature to not return anything, so that the hook handles any errors itself?

OnX methods don't return anything already. 
Right about onResponse/onRequest, i can remove them.

hooks.go
```diff
@@ -0,0 +1,118 @@
+package fiber
+
+import (
+	"github.com/valyala/fasthttp"
+)
+
+// Handler defines a function to create hooks for Fibe.
```
hi019, line 7:
```suggestion
// Handler defines a function to create hooks for Fiber.
```

efectn:
> I mean HookHandler:
> 
> ```go
>  type HookHandler = func(*Ctx, Map) error
> ```

i thought handling OnRoute and OnListen errors would be useful but if you're not OK, i can remove them.

ReneWerner87 (approved):

hi019:
Yeah I think they should be removed. After all we're not doing much with the error, just panicking.

hi019:
Actually, I think the current implementation is fine. I'll finish the review tomorrow

efectn:
> What about using a struct instead of a map for the hook function params, to provide better type safety? It would be nice if each hook had its own struct. Also this print statement seems to be executed twice:
> 
> ```go
> package main
> 
> import (
> 	"fmt"
> 	"log"
> 
> 	"github.com/gofiber/fiber/v2"
> )
> 
> func main() {
> 	app := fiber.New(fiber.Config{DisableStartupMessage: true})
> 
> 	app.Hooks().OnRoute(func(c *fiber.Ctx, m fiber.Map) error {
> 		fmt.Println(m["route"].(fiber.Route).Path)
> 
> 		return nil
> 	})
> 
> 	app.Get("/hello", h)
> 	log.Fatal(app.Listen(":3001"))
> }
> 
> func h(c *fiber.Ctx) error {
> 	return c.Status(fiber.StatusOK).SendString("Hello")
> }
> ```
> 
> It's because app.register is being called twice somehow. Still looking into that

I think it's not bug because of Get creates GET and HEAD methods.

efectn:
> What about using a struct instead of a map for the hook function params, to provide better type safety? It would be nice if each hook had its own struct.

done

hi019:
I meant changing the hook handlers to be something like this:

```go
// Old
type HookHandler = func(*Ctx, Map) error

// New
type OnRouteHandler = func(*Ctx OnRouteParams)
```

This will prevent typecasting:

```go
// Old
fmt.Println(m["route"].(fiber.Route).Path)

// New
fmt.Println(m.Route.Path)
```

efectn:
> I meant changing the hook handlers to be something like this:
> 
> ```go
> // Old
> type HookHandler = func(*Ctx, Map) error
> 
> // New
> // naming could be improved
> type OnRouteHandler = func(*Ctx OnRouteParams)
> ```
> 
> This will prevent typecasting:
> 
> ```go
> app.Hooks().OnRoute(func(c *fiber.Ctx, m fiber.OnRouteParams) error {
> 	// Old
> 	fmt.Println(m["route"].(fiber.Route).Path)
> 
> 	// New
> 	fmt.Println(m.Route.Path)
> 
> 	return nil
> })
> ```

Good idea. I'll change it. Also i plan to add OnGroup. Thanks for suggestions.

hi019 (approved):

hi019:
Also we'll need to update your docs pr with the struct changes

efectn:
> Also we'll need to update your docs pr with the struct changes

I'll update docs tomorrow.

efectn:
> > Also we'll need to update your docs pr with the struct changes
> 
> I'll update docs tomorrow.

done

app.go
```diff
@@ -461,11 +457,25 @@ func New(config ...Config) *App {
...
+		config:      Config{},
+		getBytes:    utils.UnsafeBytes,
+		getString:   utils.UnsafeString,
+		appList:     make(map[string]*App),
+		latestRoute: &Route{},
+		latestGroup: &Group{},
+	}
+
+	// Define hooks
+	app.hooks = Hooks{
```
ReneWerner87, line 469: please move this part to the hooks struct by creating a new method

ReneWerner87 (approved):

hooks.go
```diff
@@ -25,9 +25,21 @@ type Hooks struct {
 	onShutdown  []OnShutdownHandler
 }
 
+func newHooks(app *App) hooks {
```
ReneWerner87, line 28: use a pointer struct

Issue #1786, opened 2022-02-19
🐛 proxy.Do is changing the originalURL() and not restoring it to the originalURL of the handler


geet:
**Fiber version**
github.com/gofiber/fiber/v2 v2.27.0
**Issue description**
I am invoking proxy.Do(c,proxy_url) inside a fiber handler. Inside the handler, 
```
c.OriginalURL() 
```
returns a non-immutable string due to zero allocation in fiber. 
Now, when i pass the fiber context to the proxy.Do(ctx,addr), the originalURL is modified due to the [setRequestURI](https://github.com/gofiber/fiber/blob/master/middleware/proxy/proxy.go#L124) inside proxy.go
Here is the code snippet showing this.
**Code snippet**

```go
package main

import (
	"fmt"

	"github.com/gofiber/fiber/v2"
	"github.com/gofiber/fiber/v2/middleware/proxy"
)

func main() {

	app := fiber.New()
	app.Get("/test", func(c *fiber.Ctx) error {
		originalUrl := c.OriginalURL()
		fmt.Println(originalUrl) // '/test'
		if err := proxy.Do(c, "http://www.google.com/search"); err != nil {
			return err
		}
                 fmt.Println(c.OriginalURL()) // `/search'
		fmt.Println("And here: ", originalUrl) // '/sear'
		return c.SendString("tested")
	})
	app.Listen(":3002")
}

```
From the browser try hitting localhost:3002/test two times. In the first occurrence, the results are expected due to zero allocation implementation of c.OriginalURL() method in fiber. 

The above problem can be resolved by using the immutable method provided by the fiber utils to persist values.
```
	originalURL := utils.ImmutableString(c.OriginalURL())
```
But c.OriginalURL() will still point to the URL of the proxy and not to the original handler.

here is the snippet.
```
package main

import (
	"fmt"

	"github.com/gofiber/fiber/v2"
	"github.com/gofiber/fiber/v2/middleware/proxy"
	"github.com/gofiber/utils"
)

func main() {

	app := fiber.New()
	app.Get("/test", func(c *fiber.Ctx) error {
		//originalUrl := c.OriginalURL()
		originalUrl := utils.ImmutableString(c.OriginalURL())
		fmt.Println(originalUrl) // '/test'
		if err := proxy.Do(c, "http://www.google.com/search"); err != nil {
			return err
		}
		fmt.Println(c.OriginalURL()) // `/search`
		fmt.Println("And here: ", originalUrl) // '/sear'
[...]

Issue #1499, opened 2021-08-20
🐛 `c.OriginalURL()` changed after `SendFile`


Trim21:
**Fiber version**

github.com/gofiber/fiber/v2 v2.17.0

**Issue description**

## `c.OriginalURL()` changed after `c.SendFile`

**Code snippet**

```go
package main

import (
	"fmt"
	"log"

	"github.com/gofiber/fiber/v2"
)

func main() {
	app := fiber.New()

	// Steps to reproduce
	app.Get("/index", func(c *fiber.Ctx) error {
		fmt.Println("original", c.OriginalURL()) // "/index?page=1"
		err := c.SendFile(fmt.Sprintf("files/thread/%s.html", c.Query("page", "1")))
		fmt.Println(c.OriginalURL()) // "$CWD/files/thread/1.html"
		return err
	})

	log.Fatal(app.Listen(":3000"))
}
```

ReneWerner87:
https://github.com/gofiber/fiber/blob/master/ctx.go#L1053-L1075

right , we change the original url so we can use the fasthttp internal functionality

probably we should change this back

maybe you can support us with a pull request and a unittest for it

Trim21:
I found that there is an extra bug here, even default config `KeyGenerator: func(c *fiber.Ctx) string {	return c.Path() }` will make cache confused.

Trim21:
this should be separated into 2 issues maybe.

ReneWerner87:
feel free to create a pull request with the solution for both things

Trim21:
I'm not sure what kind of fix do you expected for this cache issues.

Is it ok to make a copy string of generated key as real cache key?

geet:
@ReneWerner87  I think we need to reopen this issue. If I try  the code snippet mentioned in the [comment](https://github.com/gofiber/fiber/issues/1499#issue-975160251) , I can still reproduce the bug. Try hitting the `/index` endpoint twice.
i gave a fix for this in the [PR](https://github.com/gofiber/fiber/pull/1788)

PR #1788, opened 2022-02-20
Restore Original URL of the context after changing it

geet:
This PR fixes two issues.

1.  Restore original URL after proxy.Do [issue](https://github.com/gofiber/fiber/issues/1786)
The proxy.Do method changes the c.OriginalURL() while setting the requestURI  `req.SetRequestURI(addr)`
Since  `proxy.Do` can be used inside the fiber handlers, it will change the origialURL() since the reference to the context is passed.  To fix this behaviour, I am resetting it once the proxy is done.

2. The similar [issue](https://github.com/gofiber/fiber/issues/1499) was raised for `sendFile` and  a fix was merged. However. I find that the fix is not working.

Because of the current behaviour, if we use `c.OriginalURL()`  after invoking the proxy.Do() or sendFile() it will cause side effects as `c.OriginalURL()` will continue to point to the url changed in the proxy.Do() or sendFile() method.

**Note**: using ImmutableString for persisting strings has a slight **performance cost**. Mentioned in the welcome page of fiber under Zero allocation
(This is my first PR in this project. Please let me know if I need to make some changes)

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -1301,7 +1301,7 @@ func (c *Ctx) SendFile(file string, compress ...bool) error {
 		}
 	}
 	// Restore the original requested URL
-	originalURL := utils.ImmutableString(c.OriginalURL())
+	originalURL := utils.CopyString(c.OriginalURL())
 	defer c.fasthttp.Request.SetRequestURI(originalURL)
 	// Set new URI for fileHandler
 	c.fasthttp.Request.SetRequestURI(file)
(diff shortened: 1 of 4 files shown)


efectn (changes requested):
You should use CopyString instead of ImmutableString due to it's deprecated.

geet:
> CopyString

@efectn Thanks for the suggestion. CopyString is better approach. Changing ImmutableString to CopyString

efectn (approved):
Looks good to me.

ReneWerner87 (approved):

PR #1790, opened 2022-02-20
Fixing a mistake of the comment in the limiter package

### LINKED ISSUE #1789: 📚 A mistake in the limiter middleware source code comment
There is the *config.go* file in the limiter package. The file contains definition of config for limiter middleware. It has a field `Max` which sets maximum amount of requests. According to the comment beside it implies number of connections during `Duration` seconds, but `Duration` field is deprecated and `Expiration` field to be used instead. The comment is also confusing because the [doc page](https://docs.gofiber.io/api/middleware/limiter) about the limiter middleware does not have any mention for the `Duration` field at all except for the comment itself.

qracer:
closes #1789

### DIFF
--- a/middleware/limiter/config.go
+++ b/middleware/limiter/config.go
@@ -14,7 +14,7 @@ type Config struct {
 	// Optional. Default: nil
 	Next func(c *fiber.Ctx) bool
 
-	// Max number of recent connections during `Duration` seconds before sending a 429 response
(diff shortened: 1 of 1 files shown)


efectn:
Thanks. Can you create same PR for https://github.com/gofiber/docs/blob/master/api/middleware/limiter.md?

qracer:
Created [PR for docs](https://github.com/gofiber/docs/pull/245) too

PR #1798, opened 2022-03-01
:link: Fix goreportcard & pkg.go.dev links in READMEs

jfcg:
also replace `golang.org` links with `go.dev`


ReneWerner87:
Thanks for the tip, can you correct the link in the other readme's for the other languages as well

hi019 (approved):

ReneWerner87 (approved):

PR #1809, opened 2022-03-05
:zap: Optimize App.buildTree()

jfcg:
- Reduces reads from / writes to slices
- Reduces reads from maps
- Increases sorting speed

### DIFF
--- a/router.go
+++ b/router.go
@@ -447,27 +447,28 @@ func (app *App) buildTree() *App {
 	}
 	// loop all the methods and stacks and create the prefix tree
 	for m := range intMethod {
-		app.treeStack[m] = make(map[string][]*Route)
+		tsMap := make(map[string][]*Route)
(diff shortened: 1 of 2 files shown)


ReneWerner87:
@jfcg hi, can you post a benchmark report here of the times before and after ?

jfcg:
Hi @ReneWerner87,
Added `Benchmark_Startup_Process()`. When run with
```
go test -run=^$ -bench=Benchmark_Startup_Process -benchmem -count=9
```
It gets 1427122 ns/op (median) and 1406289 ns/op (mean) on this branch.
It gets 1509394 ns/op (median) and 1521415 ns/op (mean) on master.

Diff is 82.3 µs/op (median) and 115.1 µs/op (mean) 😃

Cheers..

Note: My CPU is Core i5 4210M

ReneWerner87 (approved):

PR #1813, opened 2022-03-07
🩹 Fix limiter middleware db connection

### LINKED ISSUE #1812: 🐛 Limiter middleware works improperly with any database
Limiter middleware works fine with default in-memory storage. The same for db storage but only if `SkipSuccessfulRequests` and `SkipFailedRequests` set to `false`. 

If, for example, `SkipSuccessfulRequests` set to `true`, the middleware will still ban a user for specified amount of requests even if they have status codes less than 400.

The bug caused by the `set` method in [manager.go](https://github.com/gofiber/fiber/blob/master/middleware/limiter/manager.go):
```go
func (m *manager) set(key string, it *item, exp time.Duration) {
	if m.storage != nil {
		if raw, err := it.MarshalMsg(nil); err == nil {
			_ = m.storage.Set(key, raw, exp)
		}
		// we can release data because it's serialized to database
		m.release(it)
	} else {
		m.memory.Set(key, it, exp)
	}
}
```
If any database is used, `m.release(it)` to be executed what will set all fields of `it` structure to 0.

In its turn, [limiter_fixed.go](https://github.com/gofiber/fiber/blob/master/middleware/limiter/limiter_fixed.go) has code to check for successful requests:
```go
if (cfg.SkipSuccessfulRequests && c.Response().StatusCode() < fiber.StatusBadRequest) ||
			(cfg.SkipFailedRequests && c.Response().StatusCode() >= fiber.StatusBadRequest) {
			mux.Lock()
			e.currHits--
			remaining++
			mux.Unlock()
		}
```
Unfortunately, this actions do not have any sense, because `manager.get()` is called before:
```go
manager.set(key, e, cfg.Expiration)
``` 
Thereby, all fields values of `e` equals to 0 and checking for successful requests is meaningless.

The simplest and straightforward way to fix the bug is to remove `m.release(it)` from `manager.go`.

qracer:
Closes #1812

This PR just removes `release` method from [manager.go](https://github.com/gofiber/fiber/blob/master/middleware/limiter/manager.go) so `SkipSuccessfulRequests` and `SkipFailedRequests` from the limiter config will work properly.

Close my PR if you find my approach inappropriate to merge. I just used the simplest way to fix the bug but it can be fixed via refactoring code in [limited_config.go](https://github.com/gofiber/fiber/blob/master/middleware/limiter/limiter_fixed.go)


ReneWerner87:
i will check this tomorrow

ReneWerner87:
@qracer I think releasing the memory and putting it back into the pool is important for the performance.

You are right, if another storage is used, the following construct makes no sense

I think we should execute the get on the key  in the body of the block and set it again after the change, then the releasing can also remain in it.

Should work or?

qracer:
@ReneWerner87 I reckon we can do without calling `get` twice. A more convenient solution would be to rearrange the mutex schedule and postpone `set` execution with `defer`. I will commit a new bugfix this evening.

ReneWerner87:
Ok.

Thought one the position is binding, because one has only after the next the results.
I'm curious how your fix looks like.
Best would also be more advanced tests that test just that.

qracer:
@ReneWerner87 I've updated the pull request and changed the bugfix approach. Sorry for no test, I don't know  how to mock every available database efficiently to test

ReneWerner87:
Ok, no problem

middleware/limiter/limiter_fixed.go
```diff
@@ -69,12 +72,6 @@ func (FixedWindow) New(cfg Config) fiber.Handler {
 		// Set how many hits we have left
 		remaining := cfg.Max - e.currHits
 
-		// Update storage
-		manager.set(key, e, cfg.Expiration)
```
ReneWerner87, line 73: think moving the set function must also be done in the part where the limit was reached

middleware/limiter/limiter_fixed.go
```diff
@@ -44,6 +44,9 @@ func (FixedWindow) New(cfg Config) fiber.Handler {
 
 		// Lock entry
 		mux.Lock()
+		defer func() {
+			mux.Unlock()
```
ReneWerner87, line 48: this is not a good idea, the mutex in the defer would block the reading for too long

Because with c.Next() all possible handlers are executed

ReneWerner87 (changes requested):

ReneWerner87:
@qracer i added some comments, can you please check it again

ReneWerner87 (approved):

ReneWerner87 (approved):

PR #1832, opened 2022-03-22
:up: bump goccy/go-json to 0.9.6

jfcg:
The updates in json library [seems](https://github.com/goccy/go-json/blob/master/CHANGELOG.md#v096---20220322) to justify a patch release as well.


app.go
```diff
@@ -37,7 +37,7 @@ import (
 )
 
 // Version of current fiber package
-const Version = "2.30.0"
+const Version = "2.30.1"
```
efectn, line 40: You shouldn't edit this line.

efectn (changes requested):

app.go
jfcg: removed 👍🏼

efectn (approved):

PR #1834, opened 2022-03-22
✨ feature: add callShutdownHooksOnSigInt function

Streamer272:
* Adds `callShutdownHooksOnSigInt` function to App
* Edits `OnShutdownHandler` to take `*os.Signal` argument

When fiber app is killed by SIGINT or SIGTERM, fiber doesn't call shutdown hooks when should.
This pull request solves it by adding a channel to `Listen`, `ListenTLS` and `ListenMutualTLS` and calling `callShutdownHooksOnSigInt` as a goroutine that listens on SIGINT and SIGTERM signals thanks to go's build in `os/signal`.
After calling the shutdown hooks, function exists the program with `os.Exit(0)`.

When calling all the on shutdown handlers, they might now want to execute when a specific signel is used to kill the app, so now they take `*os.Signal` as argument and can implement their own logic.

### DIFF
--- a/app.go
+++ b/app.go
@@ -792,9 +792,9 @@ func (app *App) Listen(addr string) error {
 	// prepare the server for the start
 	app.startupProcess()
 
-	sigint := make(chan os.Signal)
-	signal.Notify(sigint, syscall.SIGINT, syscall.SIGTERM)
-	go app.callShutdownHooksOnSigInt(sigint)
+	sigChan := make(chan os.Signal)
+	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
+	go app.callShutdownHooksOnSigInt(sigChan)
 
 	// Print startup message
 	if !app.config.DisableStartupMessage {
@@ -845,9 +845,9 @@ func (app *App) ListenTLS(addr, certFile, keyFile string) error {
 	// prepare the server for the start
 	app.startupProcess()
 
-	sigint := make(chan os.Signal)
-	signal.Notify(sigint, syscall.SIGINT, syscall.SIGTERM)
-	go app.callShutdownHooksOnSigInt(sigint)
+	sigChan := make(chan os.Signal)
+	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
+	go app.callShutdownHooksOnSigInt(sigChan)
 
 	// Print startup message
 	if !app.config.DisableStartupMessage {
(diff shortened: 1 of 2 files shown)


hi019:
I think this is a good addition, it allows for graceful shutdown -- things like closing DB connections or removing temporary files. 

Uber's Fx does the same thing: https://pkg.go.dev/go.uber.org/fx#App.Start

Streamer272:
Should I add an option to OnShutdown hook?

Streamer272:
I added `*os.Signal` argument to `OnShutdownHandler`, so now `OnShutdownHandler` is `func (*os.Signal)`

hi019:
I'll look at this later today

hooks.go
```diff
@@ -1,12 +1,14 @@
...
+import "os"
+
 // Handlers define a function to create hooks for Fiber.
 type OnRouteHandler = func(Route) error
 type OnNameHandler = OnRouteHandler
 type OnGroupHandler = func(Group) error
 type OnGroupNameHandler = OnGroupHandler
 type OnListenHandler = func() error
-type OnShutdownHandler = OnListenHandler
+type OnShutdownHandler = func(*os.Signal)
```
hi019, line 11: This would break backwards compatibility :(. Honestly I think not showing `os.Signal` is fine for now. Are there any situations where a hook would need to change its functionality based on the signal type?

hi019 (changes requested):
Thanks for the PR. What about putting the signal code in app.startupProcess? Just to reduce duplication

hooks.go
efectn: I think we can add DisableDefaultGraceful 
config instead. Otherwise, i dont think hardcoded things are good. What about? @ReneWerner87 @hi019 @Streamer272

hi019: Sure, works for me. What's hardcoded, though?

efectn: > Sure, works for me. What's hardcoded, though?

Some users may want to add their custom signals.

app.go
```diff
@@ -780,21 +782,30 @@ func (app *App) Listen(addr string) error {
...
 	ln, err := net.Listen(app.config.Network, addr)
 	if err != nil {
 		return err
 	}
+
 	// prepare the server for the start
 	app.startupProcess()
+
+	sigChan := make(chan os.Signal)
+	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
```
renanbastos93, line 796: I believe that we need to give autonomy to the people to treat signals according to their needs it. 
Maybe we can create a method to explicit this signals default.
I created a package treat signals Golang maybe it can help you.
https://github.com/renanbastos93/ossignals

Fukutu:
I believe graceful shutdown should be implemented by the user, maybe this should be middleware...

efectn:
Hi @Streamer272. v3 now has more detailed and user-defined automatic graceful shutdown processing with https://github.com/gofiber/fiber/pull/1930. Thanks for the great idea :rocket: 
https://github.com/gofiber/fiber/blob/v3-beta/listen.go#L58

PR #1847, opened 2022-04-03
RFC: Return an instance of `*fiber.Error` when no handler found

codemicro:
## What does this change?

**This pull request alters the routing behaviour of Fiber to return an instance of `*fiber.Error` when a handler cannot be found to service a given request.** Prior to this commit, Fiber formed a plain-text response on-the-fly that cannot be modified by users of the framework.

This change means that errors resulting from no matching handlers being found would be routed via `fiber.App.config.ErrorHandler`, just like other errors are within Fiber.

## Why might we want this change?

**This change allows users to customise the look of their 404 page**, for example, to serve a proper HTML page instead of serving plain-text.

**Additionally, this would remove the need to add a custom 404 handler to the end of the call stack [as documented here](https://docs.gofiber.io/extra/faq#how-do-i-handle-custom-404-responses), which prevents `405 Method Not Allowed` responses [from here](https://github.com/gofiber/fiber/blob/16b8717a29cf9606f58df601f595f82decd1c23a/router.go#L143=) being correctly served.** This is due to the 404 handler matching every request that otherwise exhausts the list of currently registered handlers. By switching to returning an error instead of making an on-the-fly response, we remove the need for the catch-all 404 handler to exist, hence meaning `405 Method Not Allowed` responses will be dealt with correctly.

## Precedent

Returning `*fiber.Error` types to form an appropriate error response is already done in the codebase, for example when serving [`405 Method Not Allowed` responses](https://github.com/gofiber/fiber/blob/16b8717a29cf9606f58df601f595f82decd1c23a/router.go#L143=).

## Considerations

1. **Could this count as a breaking change?** Theorietically, any user using the framework should already have an adequate error handler in place. If they chose to continue using the default error handler, nothing would appear to change - responses generated by this change and by the previous code are identical when using the default error handler. 

   It is not impossible to think that this could be a breaking change, as the `fiber.App.config.ErrorHandler` function is now being called with errors that it was not once called with. However, I believe that this change should not cause an issue as it is assumed that any user of the framework has a functional error handler already impemented, be it the default one or a custom one.

2. **This error cannot be checked for using `errors.Is`**. Instead, you'd have to do something like this:
   ```go
   
   if e, ok := err.(*fiber.Error); ok && e.Code == fiber.StatusNotFound {
       // ...
   }
   ```
---

I'd be interested to hear everyone's opinions and feedback on this proposal.

:)

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -1089,9 +1089,9 @@ func Test_App_Next_Method(t *testing.T) {
 
 	app.Use(func(c *Ctx) error {
 		utils.AssertEqual(t, MethodGet, c.Method())
-		c.Next()
+		err := c.Next()
 		utils.AssertEqual(t, MethodGet, c.Method())
-		return nil
+		return err
 	})
 
 	resp, err := app.Test(httptest.NewRequest(MethodGet, "/", nil))
--- a/middleware/cache/cache_test.go
+++ b/middleware/cache/cache_test.go
@@ -391,32 +391,32 @@ func Test_Cache_WithHead(t *testing.T) {
 func Test_Cache_WithHeadThenGet(t *testing.T) {
 	app := fiber.New()
 	app.Use(New())
-	app.Get("/get", func(c *fiber.Ctx) error {
+	app.Get("/", func(c *fiber.Ctx) error {
 		return c.SendString(c.Query("cache"))
 	})
 
-	headResp, err := app.Test(httptest.NewRequest("HEAD", "/head?cache=123", nil))
+	headResp, err := app.Test(httptest.NewRequest("HEAD", "/?cache=123", nil))
 	utils.AssertEqual(t, nil, err)
 	headBody, err := ioutil.ReadAll(headResp.Body)
 	utils.AssertEqual(t, nil, err)
 	utils.AssertEqual(t, "", string(headBody))
 	utils.AssertEqual(t, cacheMiss, headResp.Header.Get("X-Cache"))
 
-	headResp, err = app.Test(httptest.NewRequest("HEAD", "/head?cache=123", nil))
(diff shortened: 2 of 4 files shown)


codemicro:
It appears that the `Test_App_Next_Method` test is failing as a result of a poorly written test.

```
--- FAIL: Test_App_Next_Method (0.00s)
    app_test.go:1099: 
        Test:            Test_App_Next_Method
        Trace:           app_test.go:1099
        Description:     Status code
        Expect:          404     (int)
        Result:          200     (int)
```

```go
// what's in the test
app.Use(func(c *Ctx) error {
    utils.AssertEqual(t, MethodGet, c.Method())
    c.Next()
    utils.AssertEqual(t, MethodGet, c.Method())
    return
})
// changing the test to this makes the test pass
app.Use(func(c *Ctx) error {
    utils.AssertEqual(t, MethodGet, c.Method())
    err := c.Next()
    utils.AssertEqual(t, MethodGet, c.Method())
    return err
})

codemicro:
I am unsure as to what to do with the failing `Test_Cache_WithHeadThenGet` test. Since `c.Next()` returns an error on a route that's not found, none of the cache headers are being applied to the response, as the test expects.

codemicro:
As far as I can tell, `Test_Cache_WithHeadThenGet` is supposed to check that once a HEAD request to a specific endpoint is cached, a GET request to the same endpoint doesn't return the HEAD request's cached content. I've altered it to fit that bill, as before it was making requests to different endpoints, one of which just didn't exist.

ReneWerner87 (approved):

ReneWerner87:
For me it would be okay, just want to ask the other maintainers for their opinion

In the next release it will be then in any case.

renanbastos93 (approved):

Issue #1785, opened 2022-02-18
🚀 A custom setting for the CacheDuration in SendFile()?


hakoptak:
**Is your feature request related to a problem?**
I have a use case where I have to peform some actions before sending out a static html file:

```go
app.Get("/home", func(c *fiber.Ctx) error {

	logInfo("GET /home")

	// Some actions

	return c.SendFile("home.html", true)

})
```
Currently, the `SendFile()` method accepts a compress flag and internally the `CacheDuration` is set to a fixed 10 seconds. But since the `home.html` is static by nature, I would like to define a `CacheDuration` beyond 10 seconds (like in `Static()`).

**Describe the solution you'd like**
Could we
1. Extend the SendFile method to `SendFile(file string, compress ...bool, duration ...time.Duration)`?
2. Or, more generally, define a new method `SendStaticFile(file string, config ...Static)`?
3. Or, which breaks the API, redefine `SendFile(file string, config ...Static)`?

**Describe alternatives you've considered**
None

**Additional context**
None

ReneWerner87:
@hi019 @efectn what do you mean ?
I prefer the 2nd method

1st method is not possible, because 2 times rest parameters can not come after each other

the 3rd method breaks the api

efectn:
> @hi019 @efectn what do you mean ? I prefer the 2nd method
> 
> 1st method is not possible, because 2 times rest parameters can not come after each other
> 
> the 3rd method breaks the api

Me too. I think we can do 3. one on v3.

hi019:
Yeah we can do 2, Or `SendFileWithDuration`

hi019:
2 sounds good

ReneWerner87:
@hakoptak

hakoptak:
Hi @ReneWerner87, I am looking into it. I guess that solution 2 has its troubles: the `static` struct contains fields that do not apply to the send function (as far as I understand it). It concerns `ByteRange, Browse, Download, Index, Next()`

As always: the devil is in the details...

To be honest: I misunderstood the meaning of `CacheDuration` (thought it was the duration of the cached file, but it turns out that it is the duration of the request caching). So this makes my use case less strong.

PR #1852, opened 2022-04-08
:sparkles: feature: add SendFileWithConfig method to ctx

efectn:
closes https://github.com/gofiber/fiber/issues/1785

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -1298,18 +1298,26 @@ func (c *Ctx) Send(body []byte) error {
 }
 
 var (
-	sendFileOnce              sync.Once
-	sendFileFS                *fasthttp.FS
-	sendFileHandler           fasthttp.RequestHandler
-	SendFileWithConfigOnce    sync.Once
-	SendFileWithConfigFS      *fasthttp.FS
-	SendFileWithConfigHandler fasthttp.RequestHandler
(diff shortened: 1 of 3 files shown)


ctx.go
```diff
@@ -1367,6 +1370,81 @@ func (c *Ctx) SendFile(file string, compress ...bool) error {
 	return nil
 }
 
+// SendFileWithConfig transfers the file from the given path.
+// Some config fields are overwritten. You can change them by config parameter.
+// Sets the Content-Type response HTTP header field based on the filenames extension.
+func (c *Ctx) SendFileWithConfig(file string, config ...SendFile) error {
```
ReneWerner87, line 1376: the code looks very similar to the sendFile method, can we swap the core of both methods into a private and control both with it

or at least that the SendFile method uses SendFileWithConfig 

```go
func (c *Ctx) SendFile(file string, compress ...bool) error {
     return c.SendFileWithConfig(file, SendFile{Compress: compress})
}
```


would be good if we could remove as many duplicate variables and keep the code small

efectn: Right. I think sendfile should use sendfilewithconfig.

efectn: done

ReneWerner87 (approved):

Issue #1816, opened 2022-03-08
How can I get all param variables?


cw-20020752:
For example, if my url is /test/:param1/test2/:param2
Isn't there a function that returns all path variables including param1 and param2 in middleware? 
(like url.Query())

What I expect is this.
--> 
url : test/1/test2/2
function returns: param1:1, param2:2

ReneWerner87:
we do not have such a method, since we assumed that if the route itself is defined, you know the parameters

you can still get the route from the context and there is a parameters node which contains all parameters in a slice

ReneWerner87:
@cw-20020752

long2ice:
Current I do like this:
```go
params := make(map[string][]string)
for _, param := range c.Route().Params {
	params[param] = append(params[param], c.Params(param))
}
```

ReneWerner87:
@long2ice perhaps a new method in context worth ?

cw-20020752:
Thanks for your kind replies!

PR #1853, opened 2022-04-09
:sparkles: feature: add AllParams method

efectn:
**Signature:**
`func (c *Ctx) AllParams() map[string]string`

**Benchmark Result:**
`Benchmark_Ctx_AllParams-4   	 2544381	       530.9 ns/op	     336 B/op	       2 allocs/op`

**Closes:**
https://github.com/gofiber/fiber/issues/1816

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -833,6 +833,17 @@ func (c *Ctx) Params(key string, defaultValue ...string) string {
 	return defaultString("", defaultValue)
 }
 
+// Params is used to get all route parameters.
+// Using Params method to get params.
+func (c *Ctx) AllParams() map[string]string {
+	params := make(map[string]string, len(c.route.Params))
+	for _, param := range c.route.Params {
+		params[param] = c.Params(param)
+	}
+
+	return params
+}
+
 // ParamsInt is used to get an integer from the route parameters
 // it defaults to zero if the parameter is not found or if the
 // parameter cannot be converted to an integer
--- a/ctx_test.go
+++ b/ctx_test.go
@@ -1318,6 +1318,44 @@ func Test_Ctx_Params(t *testing.T) {
 	utils.AssertEqual(t, StatusOK, resp.StatusCode, "Status code")
 }
 
+// go test -race -run Test_Ctx_AllParams
+func Test_Ctx_AllParams(t *testing.T) {
+	t.Parallel()
+	app := New()
+	app.Get("/test/:user", func(c *Ctx) error {
+		utils.AssertEqual(t, map[string]string{"user": "john"}, c.AllParams())
+		return nil
+	})
+	app.Get("/test2/*", func(c *Ctx) error {
+		utils.AssertEqual(t, map[string]string{"*1": "im/a/cookie"}, c.AllParams())
+		return nil
+	})
+	app.Get("/test3/*/blafasel/*", func(c *Ctx) error {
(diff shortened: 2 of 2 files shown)


darkweak:
I don't know if duplicate keys are supported (e.g. `http://my.app/path?mykey[]=first_value&mykey[]=second_value`).  Shouldn't it be a `map[string][]string` instead of a simple `map[string]string`?  
And you define the map length from the c.route.Params length. I guess it doesn't handle the duplicate keys right?

ReneWerner87:
> I don't know if duplicate keys are supported (e.g. `http://my.app/path?mykey[]=first_value&mykey[]=second_value`).  Shouldn't it be a `map[string][]string` instead of a simple `map[string]string`?  
> And you define the map length from the `c.route.Params` length. I guess it doesn't handle the duplicate keys right?

In your handling you describe query parameters and no resource(route) parameters

With the resource parameters or route parameters, we know beforehand how many and which ones they are, because they are predefined.

darkweak:
Sorry for the confusion 😔

renanbastos93:
Hi dude, sorry but I don't like this approach because we have a method `c.Route().Params` for this.

efectn:
> Hi dude, sorry but I don't like this approach because we have a method `c.Route().Params` for this.

I created PR because of "Accepting PR" label 🤔. What do you think @hi019 @ReneWerner87

ReneWerner87:
This access by route is not really intuitive, I am more in favor of access by method

renanbastos93:
> > Hi dude, sorry but I don't like this approach because we have a method `c.Route().Params` for this.
> 
> I created PR because of "Accepting PR" label 🤔. What do you think @hi019 @ReneWerner87

I am so sorry, but I commented on my opinion.

renanbastos93 (approved):
LGTM

ctx.go
```diff
@@ -833,6 +833,17 @@ func (c *Ctx) Params(key string, defaultValue ...string) string {
 	return defaultString("", defaultValue)
 }
 
+// Params is used to get all route parameters.
+// Using Params method to get params.
+func (c *Ctx) AllParams() map[string]string {
+	params := make(map[string]string, len(c.route.Params))
```
renanbastos93, line 839:
```suggestion
func (c *Ctx) AllParams() map[string][]string {
	params := make(map[string][]string, len(c.route.Params))
```

ctx.go
```diff
@@ -833,6 +833,17 @@ func (c *Ctx) Params(key string, defaultValue ...string) string {
...
+	for _, param := range c.route.Params {
+		params[param] = c.Params(param)
```
renanbastos93, line 841: please adjust it based on suggestion when having many params

ReneWerner87:
> > > Hi dude, sorry but I don't like this approach because we have a method `c.Route().Params` for this.
> > 
> > I created PR because of "Accepting PR" label 🤔. What do you think @hi019 @ReneWerner87
> 
> I am so sorry, but I commented on my opinion.
> 

All good, everyone can freely express his opinion here

Your opinions are all welcome, helps us think about it again and reassess the whole thing

ctx.go
efectn: I'll look at tomorrow.

ctx.go
ReneWerner87: there is no list, always only one parameter value for one parameter key

ReneWerner87 (approved):

PR #1860, opened 2022-04-14
Add missing header removal for SendFile

efectn:
i forgot add this on https://github.com/gofiber/fiber/pull/1852 :D


renanbastos93 (approved):

ReneWerner87 (changes requested):
need adjustments
revert: https://github.com/gofiber/fiber/pull/1861

since the global sendfile object is customized, but the requesthandler and this object are created in multiple handlers with different configurations, this can lead to errors in parallel usage

PR #1862, opened 2022-04-15
chore: remove incorrect links from supporter table

ankush:
I don't recall using this project or donating to it 2 years ago, it looks like someone with the same username donated it :smile:


ankush:
I don't know what emoji would be good for this 👀 whoever merges it please add it if required by squash merging.

ReneWerner87 (approved):

Issue #1874, opened 2022-04-22
🚀 Human readable BodyLimit


webdevium:
**Is your feature request related to a problem?**
No

**Describe the solution you'd like**

What about make ability to set body size limit as human readable string?
Say we have environment variable `BODY_SIZE=42MB`
Then we read and set it to the fiber configuration.

Lets imagine **BodyLimitString** property in the fiber.Config
```
app := fiber.New(fiber.Config{
	BodyLimitString: os.Getenv("BODY_SIZE"),
})
```
And this property automatic parsed into bytes size

ReneWerner87:
would treat the outside

integer numbers are perfect for config, everything else must be converted from outside

i.e. best you build a function that parses this or we could provide something in the utils, but in the config i wouldn't anchor it additionally

webdevium:
@ReneWerner87  Fine. Maybe we can provide some function into Config, like ParseBodyLimit() as utility?

Sorry for my persistence. I'm just starting my journey into golang development.
I want to get it right :)

webdevium:
Or you mean utils package?

ReneWerner87:
converting does not really have anything to do with the bodylimit

we could provide a function in the utils called "ConvertToBits".
```go 
func ConvertToBits(humanReadbleBytes string) int
```

maybe here https://github.com/gofiber/fiber/blob/master/utils/common.go

webdevium:
Sounds good!
Can I try to make this function?

ReneWerner87:
```go
utils.ConvertToByte("42MB") // 4194304
utils.ConvertToByte("42MegaByte") // 4194304
utils.ConvertToByte("42MByte") // 4194304
```

ReneWerner87:
> Sounds good! Can I try to make this function?

sure, but don´t forget 
unittests, documentation, benchmark method

ReneWerner87:
Looks good, will merge it soon

webdevium:
Tnx! 

Maybe I will be a good golang developer :)

gaby:
@VanoDevium How do you use this function? There's no documentation about it. I tried: 

```
ConvertToBytes("20MB")
utils.ConvertToBytes("20MB")
fiber.ConvertToBytes("20MB")
common.ConvertToBytes("20MB")
```

They all return function is "undefined"

ReneWerner87:
It's not released yet

But it's coming in the next days.

gaby:
Thank you! 👍🏻

PR #1875, opened 2022-04-22
utils: ConvertToBytes

webdevium:
According #1874

Utility method for converting human-readable strings into integer bytes.
Supports size syntax like docker or nginx configs.
Examples:
42m
42M
42MB
42 MB

Returns **0** if error occurred.


ReneWerner87 (approved):

utils/common.go
```diff
@@ -32,6 +40,8 @@ var (
 	uuidSeed    [24]byte
 	uuidCounter uint64
 	uuidSetup   sync.Once
+	unitsMap    = map[string]int64{"k": kb, "m": mb, "g": gb, "t": tb, "p": pb}
+	sizeRegex   = regexp.MustCompile(`(?i)^(\d+(\.\d+)*) ?([kmgtp])?b?$`)
```
renanbastos93, line 44: Do we really need to use regex here?
Can you explain more about this regex? I do like to make many tests using strings instead regex.

renanbastos93 (approved):

utils/common.go
webdevium: (based on my strong experience in regexp patterns)

This regexp is simple pattern for extracting float value and unit name

So, something like:

(strict begin)(int or float value)(maybe space symbol)(maybe unit letter)(maybe b symbol)(string end)

So, any next examples are correct:
```
42
42m
42 m
42 Mb
42.5 Mb
```

Pattern is correct and many developers use it in different packages, frameworks and languages.

And, I really don't know how parse human readable value using strings.

ReneWerner87: i tried the string recognition
https://github.com/gofiber/fiber/pull/1875/commits/4491602d7ef78869e269282705084e22680f3be9

```go
old:
Benchmark_ConvertToBytes/fiber-8                 3541161                317.9 ns/op          128 B/op          2 allocs/op
Benchmark_ConvertToBytes/fiber-8                 3465039             336.8 ns/op          128 B/op          2 allocs/op

new:
Benchmark_ConvertToBytes/fiber-12               32883782            33.76 ns/op              0 B/op          0 allocs/op
Benchmark_ConvertToBytes/fiber-12               36084900           33.47 ns/op              0 B/op          0 allocs/op
```

ReneWerner87: @webdevium are you okay with the new code ? is now a little more, but also 10 times faster

webdevium: @ReneWerner87 Awesome!
I don't have enough golang experience to do things like this without regular expressions :) 

Thank you for your help.

webdevium: @ReneWerner87 But method has panic if human string is empty :(

PR #1881, opened 2022-04-26
:bug:  Fix expiration time in cache middleware

### LINKED ISSUE #1869: 🤗 How to correctly configure custom ExpirationGenerator in cache middleware
**Question description**
How to correctly configure custom ExpirationGenerator in cache middleware? My config is show below. (I am using Redis for storage with package `github.com/gofiber/storage/redis`

**Code snippet**

```
app.Use(cache.New(cache.Config{
    Next: func(c *fiber.Ctx) bool {
        return c.Query("refresh") == "true"
    },
    ExpirationGenerator: func(c *fiber.Ctx, cfg *cache.Config) time.Duration {
        newCacheTime, _ := strconv.Atoi(c.GetRespHeader("Cache-Time", "0"))
        return time.Second * time.Duration(newCacheTime)
    },
    KeyGenerator: func(c *fiber.Ctx) string {
        return c.OriginalURL()
    },
    Storage: database.Redis,
}))
```
I have not done anything fancy here, its the exact same `ExpirationGenerator` func mentioned in the docs. But, for some reason, key's expiration is always set to the default i.e `Expiration: 1 * time.Minute`. 

I have tried printing the value of `newCacheTime` and it's correct i.e whatever I send using `c.Response().Header.Add("Cache-Time", "30")`
```
 ExpirationGenerator: func(c *fiber.Ctx, cfg *cache.Config) time.Duration {
        newCacheTime, _ := strconv.Atoi(c.GetRespHeader("Cache-Time", "0"))
        fmt.Println(newCacheTime)
        return time.Second * time.Duration(newCacheTime)
    },
```
This is something so trivial but can't figure it out. It's driving me crazy.

breakbuidl:
Even if `ExpirationGenerator` was provided in cache config, `Expiration` was used anyway.

fixes #1869

### DIFF
--- a/middleware/cache/cache_test.go
+++ b/middleware/cache/cache_test.go
@@ -19,6 +19,8 @@ import (
 )
 
 func Test_Cache_CacheControl(t *testing.T) {
+	t.Parallel()
+
 	app := fiber.New()
 
 	app.Use(New(Config{
@@ -77,6 +79,8 @@ func Test_Cache_Expired(t *testing.T) {
 }
 
 func Test_Cache(t *testing.T) {
+	t.Parallel()
+
 	app := fiber.New()
 	app.Use(New())
 
@@ -102,6 +106,8 @@ func Test_Cache(t *testing.T) {
 }
 
 func Test_Cache_WithSeveralRequests(t *testing.T) {
+	t.Parallel()
+
 	app := fiber.New()
 
 	app.Use(New(Config{
@@ -135,6 +141,8 @@ func Test_Cache_WithSeveralRequests(t *testing.T) {
 }
 
 func Test_Cache_Invalid_Expiration(t *testing.T) {
+	t.Parallel()
+
 	app := fiber.New()
 	cache := New(Config{Expiration: 0 * time.Second})
 	app.Use(cache)
@@ -161,6 +169,8 @@ func Test_Cache_Invalid_Expiration(t *testing.T) {
 }
 
 func Test_Cache_Invalid_Method(t *testing.T) {
+	t.Parallel()
+
 	app := fiber.New()
 
 	app.Use(New())
@@ -199,6 +209,8 @@ func Test_Cache_Invalid_Method(t *testing.T) {
 }
 
 func Test_Cache_NothingToCache(t *testing.T) {
+	t.Parallel()
+
 	app := fiber.New()
 
 	app.Use(New(Config{Expiration: -(time.Second * 1)}))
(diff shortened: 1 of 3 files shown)


breakbuidl:
Could any of the maintainers please approve running workflows?

middleware/cache/cache_test.go
```diff
@@ -291,15 +291,40 @@ func Test_CustomExpiration(t *testing.T) {
...
-	req := httptest.NewRequest("GET", "/", nil)
-	_, err := app.Test(req)
+	resp, err := app.Test(httptest.NewRequest("GET", "/", nil))
 	utils.AssertEqual(t, nil, err)
 	utils.AssertEqual(t, true, called)
-	utils.AssertEqual(t, 6000, newCacheTime)
+	utils.AssertEqual(t, 2, newCacheTime)
+
+	// Sleep until the cache is expired
+	time.Sleep(3 * time.Second)
```
ReneWerner87, line 305: please reduce the time for this, otherwise the test procedure will be greatly extended

you can also wait nanoseconds, micro or milliseconds, whichever is stable

if necessary i would decide for a few milliseconds, because the process also takes a bit of time

ReneWerner87 (changes requested):

middleware/cache/cache_test.go
breakbuidl: I tried setting `newCacheTime` to `999` and it failed.

breakbuidl: Also, I don't know if it's related or not. But, I played around with `Test_CacheExpired` and when `Expiration` is set to less than `1 Second`, somehow the expiration time is `1 Minute`

The below test fails.
```
t.Parallel()

	app := fiber.New()
	app.Use(New(Config{Expiration: 500 * time.Millisecond}))

		return c.SendString(fmt.Sprintf("%d", time.Now().UnixNano()))
	})

	resp, err := app.Test(httptest.NewRequest("GET", "/", nil))
	utils.AssertEqual(t, nil, err)
	body, err := ioutil.ReadAll(resp.Body)
	utils.AssertEqual(t, nil, err)

	// Sleep until the cache is expired
	time.Sleep(1000 * time.Millisecond)

	respCached, err := app.Test(httptest.NewRequest("GET", "/", nil))
	utils.AssertEqual(t, nil, err)
	bodyCached, err := ioutil.ReadAll(respCached.Body)
	utils.AssertEqual(t, nil, err)

	if bytes.Equal(body, bodyCached) {
		t.Errorf("Cache should have expired: %s, %s", body, bodyCached)
	}
```

ReneWerner87 (approved):

Issue #1136, opened 2021-01-25
🚀 cache middleware: set maximum size of cache


sweepyoface:
**Is your feature request related to a problem?**

In a way. My problem is that I want to set `Cache-Control: public, max-age=31536000` on my responses, but when using the cache middleware that also means I have to cache the response on the server side for the same duration. This brings up memory usage concerns. 

**Describe the solution you'd like**

I'd like to set the maximum size that the cache is allowed to be, and have the middleware start expiring the oldest entries when it reaches that limit.

**Describe alternatives you've considered**

I've considered just setting process memory limits, but that doesn't seem ideal. This issue could also be solved by decoupling the server side and client side caching durations and having an option to configure both, but this proposal solves two problems at the same time.

**Additional context**

The duration of my server side cache is not all that important, as it will be behind a CDN caching layer. It's simply there as a fallback, so I'm happy with the duration of 30 minutes. This is too short for `Cache-Control` however.

xenowits:
@ReneWerner87 can u plz assign me to this issue?

ReneWerner87:
Sure

xenowits:
@ReneWerner87 any ideas how this can be implemented? I have been thinking but couldn't get any definitive idea.

sweepyoface:
To take the same approach as many popular caching libraries, one could assign a weight to each entry (based on its size) and then configure the maximum weight of the cache. Lower-weighted items would be evicted first using some sort of logic.

ReneWerner87:
@xenowits i saw something in fasthttp library, will share it with you in some hours

ReneWerner87:
hmm had just looked, but find nothing

think you would have to look in the garbage collector loop again and again, how large the dataset is (unsafe.SiteOf) and then if necessary the oldest out, but for this they must first be marked or you create a slice with the keys in which order was added

would do it concurrently in the garbage collector, of course you can do it every time with the set, but then the time of the handler is extended

if you do it in the garbage collector, you can't be 100% sure that the limit won't be exceeded for a short time, but it would be good for the whole process, maybe you go down 5% from the defined max.

conorbros:
Is this still in progress?

ReneWerner87:
suspect @xenowits had no idea, so he did not continue working

@gofiber/contributors if someone has time and an idea how to solve this, we are always happy about pull requests

si3nloong:
@ReneWerner87 i can handle this

ReneWerner87:
Yes sure

[...]

PR #1892, opened 2022-05-03
🔥 Feature: Add max size to cache

dranikpg:
This PR is a proposed solution to #1136 and allows **limiting the cache size**.

As stated in the issue, the cache should expire old entries to make room for new entries. To find old entries, it has to keep track of their expirations in a separate data structure. 

### Finding a solution

In general, the data structure has to support three main operations:
- adding entries 
- removing arbitrary entries (in case of self-expiration)
- removing the oldest entry (with lowest expiration point)

At first, it looks like a binary search tree would be the perfect fit. However GO's standart library doesn't offer much for implementing them and I didn't want to add new dependencies/piles of code. An easier and equally performant solution exists: lets use a **binary heap and track node movements for arbitrary removals**. This is what `indexedHeap` is for.

### Benchmarks

Max size does not affect read performance in any way

`Benchmark_Cache_MaxSize` is a benchmark for three cases:
- No max size specified (= no performance penalty)
- A very large max size (= only insertions)
- A small max size (= constant removal & insertion)

My results are as follows:
```
Benchmark_Cache_MaxSize/Disabled-16              1626 ns/op             349 B/op          7 allocs/op
Benchmark_Cache_MaxSize/Disabled-16              1588 ns/op             392 B/op          7 allocs/op
Benchmark_Cache_MaxSize/Unlim-16                 1812 ns/op             665 B/op          8 allocs/op
Benchmark_Cache_MaxSize/Unlim-16                 1678 ns/op             629 B/op          8 allocs/op
Benchmark_Cache_MaxSize/LowBounded-16            1178 ns/op             256 B/op          8 allocs/op
Benchmark_Cache_MaxSize/LowBounded-16            1190 ns/op             256 B/op          8 allocs/op
```

We're **up one allocation from first to second case and about 10% slower** per insertion. The third case is the fastest because our cache always stays small.

This is a benchmark where the handler takes zero time to execute. 10% might sound like much, but the **difference is less than one microsecond in my case** (thats actually suspiciously small 🤔). I doubt any handler call that has to be cached would care about such a time period.

### Measuring real memory consumption

Instead of counting entries, we could count their body size to measure real memory consumption. 

### Other possible soltuions

* We could ignore expiration all together and remove just the oldest entry by time of insertion. This would be the easiest and most performant solution. I'm not sure whether this would be correct behaviour (what if we pre-cache a bunch of big pages on startup - should we drop them all?)
* Implement some garbage collection goroutine. This sounds more complex and is full of bad worst cases


ReneWerner87:
Thanks for this feature

will review it this weekend and if everything is ok will merge it

middleware/cache/config.go
```diff
@@ -59,6 +59,13 @@ type Config struct {
 	//
 	// Default: false
 	StoreResponseHeaders bool
+
+	// Max number of entries simultaneously stored in cache. When limit is reached,
+	// entries with the nearest expiration are deleted to make room for new.
+	// 0 means no limit
+	//
+	// Default: 0
+	MaxSize int
```
ReneWerner87, line 68: can you use uint -> negative sizes are not possible

middleware/cache/cache.go
```diff
@@ -59,6 +59,8 @@ func New(config ...Config) fiber.Handler {
 	)
 	// Create manager to simplify storage operations ( see manager.go )
 	manager := newManager(cfg.Storage)
+	// Create indexed heap for tracking expirations ( see heap.go )
+	heap := &indexedHeap{}
```
ReneWerner87, line 63: the allocations come from the fact that elements must be created here at runtime.

could you not create the array with fixed size already in the heapmap with the information about the max size ?

dranikpg:
I've just tracked down that one allocation 🥳 
It comes from promoting a `heapEntry` to an `interface{}` when pushing onto the heap.
Pushing manually (append + heap.Fix) avoids it.

ReneWerner87:
I just remembered that we should use mutex, sync map or atomic due to the concurrency processes to prevent race conditions in the access

middleware/cache/cache.go
ReneWerner87: Yes you are right

ReneWerner87:
Can you update the documentation for the middleware in our docs repository

dranikpg:
> I just remembered that we should use mutex, sync map or atomic due to the concurrency processes to prevent race conditions in the access

Sync is already in place for storage writes, heap modifications happen around it. There shouldn't be any issues

dranikpg:
> * [ ]  Pls update the documentation for the middleware in our docs repository
> * [ ]  update the config documentation in our README.md for the middleware

Before touching docs: are we sure we'll count entries and not bytes?

ReneWerner87:
> > I just remembered that we should use mutex, sync map or atomic due to the concurrency processes to prevent race conditions in the access
> 
> Sync is already in place for storage writes, heap modifications happen around it. There shouldn't be any issues
> 

Perfect

ReneWerner87:
> > * [ ]  Pls update the documentation for the middleware in our docs repository
> > * [ ]  update the config documentation in our README.md for the middleware
> 
> Before touching docs: are we sure we'll count entries and not bytes? 

Will look tomorrow again more closely and answer

ReneWerner87:
@dranikpg 
Ok then we should name the configuration flag differently

e.g. maxEntries

@efectn @hi019 what do you think about the count of bytes vs count of entries

efectn:
> @dranikpg Ok then we should name the configuration flag differently
> 
> e.g. maxEntries
> 
> @efectn @hi019 what do you think about the count of bytes vs count of entries

I think counting bytes is better option

middleware/cache/heap.go
```diff
@@ -0,0 +1,92 @@
+package cache
+
+import (
+	"container/heap"
+)
+
+type heapEntry struct {
+	key   string
+	exp   uint64
+	bytes uint
```
dranikpg, line 10: Entry size is stored in the heap to update the total size without reading entries on delete

efectn (approved):

ReneWerner87:
@dranikpg thx
https://github.com/gofiber/fiber/pull/1892#issuecomment-1117638432
don´t forget the doc repository

ReneWerner87 (approved):

PR #1897, opened 2022-05-06
Fix nothing to cache

breakbuidl:
There was no check for nothing to cache if `custom expiration < 0`. This PR adds that check.

Also, the current implementation checks if expiration < 0 for nothing to cache. However nothing must be cached also in the case `expiration = 0`. Hence, improved the current implementation to `expiration <= 0` for nothing to cache.

### DIFF
--- a/middleware/cache/cache_test.go
+++ b/middleware/cache/cache_test.go
@@ -345,6 +345,38 @@ func Test_CustomExpiration(t *testing.T) {
 	}
 }
 
+func Test_NothingToCache_CustomExpiration(t *testing.T) {
+	t.Parallel()
+
+	app := fiber.New()
+
+	app.Use(New(Config{ExpirationGenerator: func(c *fiber.Ctx, cfg *Config) time.Duration {
+		newCacheTime, _ := strconv.Atoi(c.GetRespHeader("Cache-Time", "600"))
+		return time.Second * time.Duration(newCacheTime)
+	}}))
+
+	app.Get("/", func(c *fiber.Ctx) error {
+		c.Response().Header.Add("Cache-Time", "-1")
+		return c.SendString(time.Now().String())
+	})
+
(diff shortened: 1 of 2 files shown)


ReneWerner87 (changes requested):
pls check my review comments

ReneWerner87:
@breakbuidl can you check my review comments ?

breakbuidl:
Hey @ReneWerner87, sorry for the late replay, just got back to this. Will finish this over the weekend.

ReneWerner87:
@breakbuidl any progress on the weekend ?

middleware/cache/cache_test.go
```diff
@@ -345,6 +345,38 @@ func Test_CustomExpiration(t *testing.T) {
...
+	}}))
+
+	app.Get("/", func(c *fiber.Ctx) error {
+		c.Response().Header.Add("Cache-Time", "-1")
+		return c.SendString(time.Now().String())
+	})
+
+	resp, err := app.Test(httptest.NewRequest("GET", "/", nil))
+	utils.AssertEqual(t, nil, err)
+	body, err := ioutil.ReadAll(resp.Body)
```
ishan16696, line 365:
```suggestion
	body, err := io.ReadAll(resp.Body)
```
see this https://go.dev/doc/go1.16#ioutil

ReneWerner87:
@breakbuidl?

gaby:
@breakbuidl Friendly bump

middleware/cache/cache_test.go
efectn: it's breaking change for 1.14 and 1.15

ReneWerner87:
due to the non-activity, combined with the fact that we don't really see a case for it, because the expiration can't be less than 0 and even if it is, the entry isn't really cached then

so we close the pull request until the code is updated and/or the need is described in more detail

PR #1909, opened 2022-05-16
Spliting dump request to add Query 🐛 🚩 

ancogamer:
Spliting []dump into dumps, and editing the dump[1] to = req.URL.String(), so we can tests itens that depends on querys from the request.

### DIFF
--- a/app.go
+++ b/app.go
@@ -9,6 +9,7 @@ package fiber
 
 import (
 	"bufio"
+	"bytes"
 	"crypto/tls"
 	"crypto/x509"
 	"errors"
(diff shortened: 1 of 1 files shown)


ReneWerner87:
hi thanks for the work

is this not already possible, just trying to understand what your customization is supposed to do

can you explain this in more detail and if necessary send an example, which shows what was not possible before and is now possible

ReneWerner87:
think i now understand the point of this pull requests

nevertheless please describe briefly in the pull request with an example, what is possible now, what did not work before

so that we have this stored here for later

can you please also extend the test with such a case
in: https://github.com/gofiber/fiber/blob/master/app_test.go

app.go
```diff
@@ -974,6 +975,11 @@ func (app *App) Test(req *http.Request, msTimeout ...int) (resp *http.Response,
 		return nil, err
 	}
 
+	// adding back the query from URL, since dump cleans it
+	dumps := bytes.Split(dump, []byte(" "))
```
ReneWerner87, line 979: please improve this construct with regard to performance and allocations

ancogamer: I will take a better a look on this. The problem is: imagine you have a query /foo?bar=boo, when the dump is create the query part is removed, staying /foo.

PR #1911, opened 2022-05-18
:construction_worker: ci: actions for MacOS

efectn:
Add macos-latest to platform of testing action.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,6 +12,22 @@ jobs:
       uses: actions/setup-go@v3
       with:
         go-version: ${{ matrix.go-version }}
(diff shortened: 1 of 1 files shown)


ReneWerner87:
don't know how much github actions we have free

maybe we should first optimize the caching of some actions, so that they don't take so long

efectn:
> don't know how much github actions we have free
> 
> maybe we should first optimize the caching of some actions, so that they don't take so long

Which steps should we cache?

ReneWerner87:
We should keep thr cache
https://github.com/gofiber/template/blob/master/.github/workflows/test.yml#L15

PR #1924, opened 2022-05-30
Improve code consistency

ad3n:
Remove named return in router


ReneWerner87:
we like the current writing, no improvement needed

Issue #1825, opened 2022-03-20
🚀 v3 Request: fiber.Error's message type of interface{}


balcieren:
If fiber.Error's message's type will be interface{}, we can return multiple error messages etc.

Example 

```go
fiber.NewError(fiber.StatusBadRequest, fiber.Map{
  "type": "validation",
  "message": "email is invalid",
})
```

efectn:
Same as https://github.com/gofiber/fiber/pull/1728

pwtitle:
i want this: fiber.NewError(fiber.StatusBadRequest, "validation", "email is invalid", "other message")

PR #1925, opened 2022-05-31
:sparkles: v3 (feature): use any as default Message type of Error struct

efectn:
closes https://github.com/gofiber/fiber/issues/1825


harry93848bb7:
On a side note for the v3 fiber `Error` type shouldn't we include a method added to the `Ctx` such as `ctx.SendError(err Error)` to encourage usage of the type (as an exported package defined variable) and avoid writing helpers such as:

```go
func SendError(c *fiber.Ctx, code int, message string) error {
	c.Status(code)
	return c.JSON(fiber.NewError(code, message))
}
```
Instead a method could simply be defined as:
```
func (c *Ctx) SendString(err Error) error {
	c.Status(err.Code)
	return c.JSON(err)
}
```

ReneWerner87 (approved):

PR #1943, opened 2022-06-27
Logging Middleware - TagLatency doesn't have standard format between modes

marcelogamba:

This change affects TagLatency in middleware logging. 
Currently if you don't change the default format the latency output is using millisecond as rounding, however if you configure a custom logging message format it outputs using time.Duration (string) serialization.

I can enumerate two problems:
- Standard between the two options (Default and Custom Format)
- time.Duration will output characters in UTF8 and UNICODE (UTF-16) according to its serialization unit (microseconds)

This PR suggests replacing time.Duration serialization to millisecond, the same used in DefaultFormat.

### DIFF
--- a/middleware/logger/logger.go
+++ b/middleware/logger/logger.go
@@ -229,7 +229,7 @@ func New(config ...Config) fiber.Handler {
 			case TagUA:
 				return buf.WriteString(c.Get(fiber.HeaderUserAgent))
 			case TagLatency:
-				return buf.WriteString(stop.Sub(start).String())
+				return buf.WriteString(fmt.Sprintf("%7v", stop.Sub(start).Round(time.Millisecond)))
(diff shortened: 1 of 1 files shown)


marcelogamba:
Hi there, apparently there is a problem with other modules compiling/testing on MacOs. 
Unfortunately I don't have a Mac to try a fix. As far as I noticed my change doesn't affected those tests failing (memory test).

Could anyone help?

efectn:
> Hi there, apparently there is a problem with other modules compiling/testing on MacOs. Unfortunately I don't have a Mac to try a fix. As far as I noticed my change doesn't affected those tests failing (memory test).
> 
> Could anyone help?

I don't think the problem is related this PR. Tests related timeout is falling randomly. I'll check them.

PR #1796, opened 2022-02-28
🚀 HTTP/2 Feature

balcieren:
Usage :
```go
	app := fiber.New(fiber.Config{
		EnableHTTP2: true, // Default: false
	})

	app.Get("/http11", func(c *fiber.Ctx) error {
		return c.JSON(c.IsHTTP11()) //return true without http2 client support
	})

	app.Get("/http2", func(c *fiber.Ctx) error {
		return c.JSON(c.IsHTTP2()) //return true via http2 client support
	})

	app.ListenTLS(":8080", ".github/testdata/public.crt", ".github/testdata/private.key")
```

Issue #1683

### To Do List

- [x] Add property to app's config.
- [x] Compatible with prefork.
- [x] ListenTLS with HTTP/2.
- [x] ListenMutualTLS with HTTP/2.
- [x] Add IsHTTP2 and IsHTTP11 functions to context.
- [x] HTTP2 client support.
- [x] Add tests.

ZAAAA

PR #1957, opened 2022-06-29
🔥 Added HTTP2 support to server.

HrachMD:
New config parameter added for the App - HttpProtocolVersion.
The possible values are ProtoV1 for Http/1.1 and ProtoV2 for Http/2.
Setting HttpProtocolVersion = ProtoV2 will take effect only when TLS is enabled (eg. ListenTLS).


HrachMD:
> #1796

it's in a draft for a 4mns (

efectn:
> 

Because dgrr/http2 is not ready to use in production

PR #1960, opened 2022-07-01
fix middleware/session update cookie.

ly020044:
Session middleware should always update client cookie so that cookie expires at the same time as session.

### DIFF
--- a/middleware/session/session.go
+++ b/middleware/session/session.go
@@ -144,10 +144,8 @@ func (s *Session) Save() error {
 		s.exp = s.config.Expiration
 	}
 
-	// Create session with the session ID if fresh
-	if s.fresh {
-		s.setSession()
-	}
+	// Update client cookie
+	s.setSession()
 
 	// Convert data to bytes
 	mux.Lock()


ReneWerner87:
should we refresh the expire time every time ? if you remove the fresh switch in the condition why is s.fresh still needed(think we are destroying the reason for introducing the property) ? would have expected some unittests outlining the problem or showing that your fix works -> otherwise it is not guaranteed that it is fixed forever

iredmail:
> should we refresh the expire time every time ?

Yes.

Let's say session created at 2:00PM and expire time is set to 2 hours, each time you request a web page, the session expiration time is updated to stay 2 hours. If you request a page at 2:30PM, session expire time will be (correctly) updated to 4:30PM, but cookie still has expiration time at 4PM. In this case when request at any time after 4PM, the web session is invalid because session ID stored in client cookie expired and server regenerates one.

I checked other session implementations like Gin's session middleware, they all update cookie expiration time. Sample code (Please click the `Set Count` link and check cookie in web browser):

```
package main

import (
	"fmt"
	"html/template"
	"net/http"

	"github.com/gin-contrib/sessions"
	"github.com/gin-contrib/sessions/memstore"
	"github.com/gin-gonic/gin"
)

func main() {
	router := gin.Default()
	store := memstore.NewStore([]byte("niuniuniu"))
	router.Use(sessions.Sessions("gin_session", store))

	tmpl := "<h1>Hello World!</h1> <h3>count: {{.Count}}</h3> <h3><a href=\"/s\">Set Count</a></h3>"
	t, err := template.New("test").Parse(tmpl)
	if err != nil {
		panic(err)
	}
	router.SetHTMLTemplate(t)

	router.GET("/", func(ctx *gin.Context) {
		session := sessions.Default(ctx)
		ctx.HTML(http.StatusOK, "test", gin.H{
			"Count": session.Get("count"),
		})
	})

	router.GET("/s", func(ctx *gin.Context) {
		session := sessions.Default(ctx)
		var count int
		v := session.Get("count")
		if v == nil {
			count = 0
		} else {
			count = v.(int)
			count++
		}
		session.Set("count", count)
		if err := session.Save(); err != nil {
			fmt.Println(err)
		}

		ctx.Redirect(http.StatusFound, "/")
	})

	router.Run(":8080")
}
```

ReneWerner87:
Ok, thanks

ReneWerner87 (approved):

iredmail:
Is it ok to merge this PR now? @ReneWerner87

PR #1964, opened 2022-07-05
add: params parse

olongfen:
**Hello, This PR is binds the param string to a struct**

**Parse path parameters and body parameters using a struct**
```go 
	app.Put("/user/:id", func(ctx *Ctx) error {
		type User struct {
			// uri
			ID uint `uri:"id" validate:"required"`
			// body
			Username string `json:"username" validate:"required"`
			Age      int    `json:"age"`
		}
		var (
			d = new(User)
		)
		// params parser
		_ = ctx.ParamsParser(d)
		// body parser
		_ = ctx.BodyParser(d)
		// then validate
		validate.Struct(d)
		return nil
	})
```

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -39,6 +39,7 @@ const (
 	queryTag     = "query"
 	reqHeaderTag = "reqHeader"
 	bodyTag      = "form"
+	uriTag       = "uri"
 )
 
 // userContextKey define the key name for storing context.Context in *fasthttp.RequestCtx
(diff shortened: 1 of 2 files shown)


ReneWerner87:
- [ ] please add a part in the documentation repository

liaohongxing:
uri tag should be renamed param, similar to Echo Gin in the go ecosystem

considering the existing params(), paramsint(), is it param or params that needs to be discussed

olongfen:
> uri tag should be renamed params , similar to Echo Gin in the go ecosystem
> 
> considering the existing params(), paramsint(), is it param or params that needs to be discussed

> gin  use struct tag   is uri tag ,  path param include many value,so  i think use paramsParser is better
> 
image

liaohongxing:
it is recommended to change to params, consistent with the existing API

```
func (c *Ctx) Params(key string, defaultValue ... string) string
func (c *Ctx) ParamsInt(key string) (int, error)
func (c *Ctx) ParamsParser(out interface{}) error // Tag is params

func (c *Ctx) Query(key string, defaultValue ... string) string
func (c *Ctx) QueryParser(out interface{}) error // Tag is query
```

liaohongxing:
@efectn  what do you think

ReneWerner87:
@olongfen what do you think ? its a valid point

olongfen:
ok

PR #1968, opened 2022-07-06
binds the param string to a struct use params tag

olongfen:
**binds the param string to a struct use params tag**
```go 
	app.Get("/test1/userId/role/:roleId", func(ctx *Ctx) error {
		type Demo struct {
			UserID uint `params:"userId"`
			RoleID uint `params:"roleId"`
		}
		var (
			d = new(Demo)
		)
		ctx.ParamsParser(d)
		return nil
```

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -39,7 +39,7 @@ const (
 	queryTag     = "query"
 	reqHeaderTag = "reqHeader"
 	bodyTag      = "form"
-	uriTag       = "uri"
+	paramsTag    = "params"
 )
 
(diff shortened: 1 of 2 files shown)


olongfen:
@ReneWerner87 This check sometime 1.17.x failed in macos, sometime 1.16.x failed in macos

ReneWerner87:
> @ReneWerner87 This check sometime 1.17.x failed in macos, sometime 1.16.x failed in macos

yes we know, we have to make this more stable somehow
the timeout tests are not really stable

Issue #1840, opened 2022-03-30
🚀  v3 Request: Retry Mechanism


gozeloglu:
**Is your feature request related to a problem?**

**Describe the solution you'd like**
I am planning to add the exponential backoff algorithm. It is a widely used retry policy. Basically, there will be 2 different functions. One is `Retry()` (or `Do()` ) method and `NextBackoff()`(or `NextTime()`) methods. Of course, there will be some helper methods like `NewBackoff()`, `Reset()`. This feature will be flexible. The users can be used `Retry()` method and everything will be handled by the package. They won't consider details. They will just call the method like this:

```go
backoff.Retry(func() error{
    // Do whatever you want    
    return err
})
```
Or, maybe some users want to handle the retry part themselves. In this case, we provide all functionalities other than `Retry()` method like `Reset()`, `NextTime()`, `Sleep()`. They use the package and call the methods like this:

```go
for i := 0; i < n; i++ {
    ...
    next := backoff.NextTime()
    backoff.Sleep(next)
}
...
backoff.Reset()
```
Also, we can add two options to check the retry count/time with `Context` and `retryCount`. In `Context` usage, we set timeout to `context` and retry goes until timeout expires. For `retryCount` usage, we set number of retry step like 10, and the method retries 10 times. 

**Describe alternatives you've considered**
[**Regular Interval**](https://docs.microsoft.com/en-us/azure/architecture/best-practices/transient-faults#:~:text=Regular%20intervals.%20The%20application%20waits%20for%20the%20same%20period%20of%20time%20between%20each%20attempt.%20For%20example%2C%20it%20may%20retry%20the%20operation%20every%203%20seconds.): Waits in regular period of time.
[**Incremental Intervals**](https://docs.microsoft.com/en-us/azure/architecture/best-practices/transient-faults#:~:text=Incremental%20intervals.%20The%20application%20waits%20a%20short%20time%20before%20the%20first%20retry%2C%20and%20then%20incrementally%20increasing%20time%20between%20each%20subsequent%20retry.%20For%20example%2C%20it%20may%20retry%20the%20operation%20after%203%20seconds%2C%207%20seconds%2C%2013%20seconds%2C%20and%20so%20on.): Waiting time is increasing incrementally. 
**Additional context**
* [cenkalti/backoff](https://github.com/cenkalti/backoff)
* https://aws.amazon.com/builders-library/timeouts-retries-and-backoff-with-jitter/ 
[...]

PR #1972, opened 2022-07-09
✨ v3 (feature): add retry mechanism

gozeloglu:
* General logic is implemented.
* Unit tests are added.

Signed-off-by: Gökhan Özeloğlu <gokhan.ozeloglu@deliveryhero.com>

It is related to #1840. 

I added a new package called `retry`. Exponential backoff algorithm is used. I can refactor the code like constant/default values, tests, etc. I am going to add more tests and maybe I'll do some small changes to the code.


middleware/retry/exponential_backoff_test.go
```diff
@@ -0,0 +1,66 @@
+package retry
+
+import (
+	"fmt"
+	"github.com/stretchr/testify/assert"
```
efectn, line 5: I think you should use fiber's assert instead of adding new deps.

gozeloglu: Fixed in cd5d0cd.

efectn: Dont forget running go mod tidy

efectn (commented):
- I think we should be able to change defaults by a config.
- Jitter backoff would be great.
- Hooks would be also great.
- A simple README for the middleware.
- Should we integrate this into the client directly?

gozeloglu:
Hi @efectn. I have two questions. What do you mean by **Should we integrate this into the client directly?** Can you explain it? Also, what should I do for hooks? As a note, jitter is already provided with adding randomness.

efectn:
> Hi @efectn. I have two questions. What do you mean by **Should we integrate this into the client directly?** Can you explain it? Also, what should I do for hooks? As a note, jitter is already provided with adding randomness.

Perhaps we would integrate this into the client like https://github.com/go-resty/resty#retries to use with client easier.
- Btw don't forget running go mod tidy to remove testify from go.mod and go.sum
- Also you can remove app.Use lines from README due to this is not handler middleware.
- Jitter seems ok to me

gozeloglu:
> Perhaps we would integrate this into the client like https://github.com/go-resty/resty#retries to use with client easier.

So, do you mean that we should use it as follow?
```go
app := fiber.New()
retry := app.NewExponentialBackoff())
err := retry.Retry(func() error)
// handle error
```
(Naming can be changed like `retry.Retry` might not be good.)
So, also you mentioned that 
> Also you can remove app.Use lines from README due to this is not handler middleware.

As I understood that I should remove it from `/middleware` directory and carry it to the root directory. If yes, it might be a problem for the config part. If I am wrong, please correct me.

efectn:
> As I understood that I should remove it from `/middleware` directory and carry it to the root directory. If yes, it might be a problem for the config part. If I am wrong, please correct me.

We may locate it in '/utils' as a directory. however i'm not sure.

> (Naming can be changed like retry.Retry might not be good.)

i wasn't talking about it. i was talking about adding some built-in retry methods for client that use this retry function. ref: https://github.com/go-resty/resty/blob/master/request.go#L763

middleware/retry/config.go
```diff
@@ -0,0 +1,66 @@
...
+	MaxRetryCount:   10,
+	currentInterval: 1 * time.Second,
+}
+
+// configDefault sets the config values if they are not set.
+func configDefault(config ...Config) Config {
+	if len(config) == 0 {
+		return DefaultConfig
+	}
+	cfg := config[0]
```
Ja7ad, line 49: Why do we need to get the config slice if we have set element 0 of the config slice?

gozeloglu:
> We may locate it in '/utils' as a directory. however i'm not sure.

I think, `utils` is not a good place to locate them. Because, in general, `utils` is a place for utility functions used across the codebase. Also, it has a flat structure. I am not going to change it until you give another suggestion.

> i wasn't talking about it. i was talking about adding some built-in retry methods for client that use this retry function. ref: https://github.com/go-resty/resty/blob/master/request.go#L763

I think I understood what you meant. Maybe, I can create a new directory called `retry` in the root directory. But, it might not good for the framework. 

> Hooks would be also great.

Also, what can I do with hooks? I don't have any idea what should I do.

middleware/retry/config.go
gozeloglu: I checked configs in other parts and I saw it as a general convention or something like that.

efectn: It's because of the make config parameter optional.

efectn:
> Also, what can I do with hooks? I don't have any idea what should I do.

For example, we can execute hooks in each retry if user defined them.

@ReneWerner87 what do you think about the location?

ReneWerner87:
Yeah its a good idea. I like hooks, its a good solution for customization and give the possibility to extend our functionality.

efectn:
> Yeah its a good idea. I like hooks, its a good solution for customization and give the possibility to extend our functionality.

Btw we're not sure about `middleware/` is true location for retry. What do you think?

middleware/retry/config.go
Ja7ad: > 

Usually get a slice of the configuration with a type options pattern.
For normal configuration, getting slice is not necessary.

Ja7ad: > It's because of the make config parameter optional.

Usually get a slice of the configuration with a type options pattern.
For normal configuration, getting slice is not necessary.

Ja7ad:
> > Yeah its a good idea. I like hooks, its a good solution for customization and give the possibility to extend our functionality.
> 
> Btw we're not sure about `middleware/` is true location for retry. What do you think?

Usually retry is a core option for framework, i think we can set retry as option for framework.

gozeloglu:
> Yeah its a good idea. I like hooks, its a good solution for customization and give the possibility to extend our functionality.

Can you provide an example of what should I do with hooks? It can be a code snippet, example code, or something like that. It can make more clear to understand your point.

efectn:
> > Yeah its a good idea. I like hooks, its a good solution for customization and give the possibility to extend our functionality.
> 
> Can you provide an example of what should I do with hooks? It can be a code snippet, example code, or something like that. It can make more clear to understand your point.

We may add it when to refactor client. Looks like we dont need at the moment.

efectn (approved):

gozeloglu:
Hi @ReneWerner87. This is a polite ping. Can you review the PR?

ReneWerner87:
Oh sorry, sure will have a look this week(currently much todo)

wangjq4214:
Now I need to use retry in client refactoring. I also think `middleware/` is not a good location, it might be a good idea to put it in `utils/` or `client/retry.go`?

ReneWerner87:
> Now I need to use retry in client refactoring. I also think `middleware/` is not a good location, it might be a good idea to put it in `utils/` or `client/retry.go`?

talked to efectn about it and we want to introduce a new folder

this should be called "_plugin_", "_addon_" or "_extension_"

in utils and helper should be only simple functionalities

please create an order "addon" and under it the order "retry" -> there you place your new functionality 
also helps go modules, so that only what is needed is checked out

please revise your readme -> the example confused me at first

efectn:
> > Now I need to use retry in client refactoring. I also think `middleware/` is not a good location, it might be a good idea to put it in `utils/` or `client/retry.go`?
> 
> talked to efectn about it and we want to introduce a new folder
> 
> this should be called "_plugin_", "_addon_" or "_extension_"
> 
> in utils and helper should be only simple functionalities
> 
> please create an order "addon" and under it the order "retry" -> there you place your new functionality also helps go modules, so that only what is needed is checked out
> 
> please revise your readme -> the example confused me at first

Hi @gozeloglu. Can you take a look when you're free? We'll merge this after these changes.

gozeloglu:
Hi @efectn. Thanks for informing me. I'll take a look and make changes as soon as possible.

gozeloglu:
Hi again,

Thanks for renaming @efectn. It seems the PR is ready to merge. I have a question to @ReneWerner87 about the latest comment. 

> please revise your readme -> the example confused me at first

Which part confused you? What is the exact problem? Other than this part, rest of the code seems fine to merge.

ReneWerner87:
At the time when I wrote this, there was still a completely wrong example stored there
Look at it tomorrow and then I merge

efectn:
> Hi again,
> 
> Thanks for renaming @efectn. It seems the PR is ready to merge. I have a question to @ReneWerner87 about the latest comment.
> 
> > please revise your readme -> the example confused me at first
> 
> Which part confused you? What is the exact problem? Other than this part, rest of the code seems fine to merge.

It was related `app.Use` part but i fixed that.

gozeloglu:
> It was related app.Use part but i fixed that

Thanks, @efectn. Looking forward to seeing the merge of the PR. 🚀  
Thanks, @ReneWerner87 as well.

ReneWerner87 (approved):

Issue #1821, opened 2022-03-18
🚀 v3 Request: Detailed Binding for Query, Form, Headers, Params...


efectn:
Fiber already has QueryParser, BodyParser and ReqHeadersParser but they're really limited and not extendable. Also i think we don't need gorilla/schema. 

I think creating binding sub-directory will solve this problem. https://github.com/gin-gonic/gin/tree/master/binding. It'll provide more more detailed binding specs. (without default validator, maybe custom validator function can be added, https://github.com/gofiber/fiber/pull/1766).

Also adding all methods to ctx would be make ctx really mess and big. So we should add these binding methods by sub-method.
```
Bind().URI()
Bind().Form()
Bind().Params()
Bind().Headers()
```

Look These:
- https://github.com/gin-gonic/gin#model-binding-and-validation
- https://echo.labstack.com/guide/binding/
- https://github.com/gin-gonic/gin/tree/master/binding
- https://github.com/labstack/echo/blob/master/bind.go
- https://github.com/labstack/echo/blob/master/binder.go

efectn:
> can we change validator function by passing another implemen?

Yes

hotrungnhan:
hope it release soon

efectn:
> hope it release soon

Hi. I'll start implementing this after closing some PRs for v3. https://github.com/orgs/gofiber/projects/1

PR #1981, opened 2022-07-26
:sparkles: v3 (feature): initial support for binding

efectn:
**Partially related:** https://github.com/gofiber/fiber/issues/1821

This PR aims to create detailed and moduler binding system for Fiber v3. Fiber ctx will be more clean thanks to this feature. Also Fiber will be able to bind into the map.

## To-Do List:
- [x] Embed bind into the DefaultCtx.
- [x] Params binder. 
- [x] Resp header binder.
- [x] Cookie binder.
- [x] Add support for custom binders in automatically body binding.
- [x] Support for map binding.
- [x] Make Should/Must functional.
- [x] Custom struct validator support.
- [x] Add comment lines and simple README for binding.

**Note:** I won't change **gorilla/schema** in this PR. But next PRs, i'll create more minimal and useful parser for Fiber.


bind.go
```diff
@@ -0,0 +1,83 @@
...
+		if binder.Name() == name {
+			return binder.Parse(b.ctx, dest)
+		}
+	}
+
+	return errors.New("fiber: custom binder not found, please be sure to enter right name!")
+}
+
+func (b *Bind) Header(out any) error {
+	return binder.HeaderBinder.Bind(b.ctx.Request(), out)
```
ReneWerner87, line 41: shouldn't the Bind.Query also check the customBinder ? otherwise the functionalities are not changeable with custom binders

ctx.go
```diff
@@ -1576,3 +1359,9 @@ func (c *DefaultCtx) IsFromLocal() bool {
 	}
 	return c.isLocalHost(ips[0])
 }
+
+func (c *DefaultCtx) Binding() *Bind {
+	return &Bind{
```
ReneWerner87, line 1364: is it ok to reinitialize the binder every time we call the bind method ? or should we put the instance in the ctx class ?

bind.go
efectn: I think we don't need it at the moment. We can take a look it next pr

ctx.go
efectn: I'll add it into the ctx

binder/cookie.go
```diff
@@ -0,0 +1,45 @@
...
+
+	reqCtx.Request.Header.VisitAllCookie(func(key, val []byte) {
+		if err != nil {
+			return
+		}
+
+		k := utils.UnsafeString(key)
+		v := utils.UnsafeString(val)
+
+		if strings.Contains(v, ",") && equalFieldType(out, reflect.Slice, k) {
```
efectn, line 29: do we need here for cookies?

ctx.go
```diff
@@ -1576,3 +1360,12 @@ func (c *DefaultCtx) IsFromLocal() bool {
 	}
 	return c.isLocalHost(ips[0])
 }
+
+func (c *DefaultCtx) Binding() *Bind {
+	if c.bind == nil {
```
efectn, line 1365: Bind() would be better i suppose.

Trim21:
I have another idea about this, we may not need that many internal temporary `map`

https://gist.github.com/Trim21/a3994fcd182f4d4f936738f5acb86559

with go 1.18 generic, we can have a zero(or maybe one) alloc and almost same with hand writing performance and good API.

efectn:
> I have another idea and POC about this feature, we do not need internal temporary `map` and schema (or schema-lite), and `gorilla/schema` is really slow.
> 
> https://github.com/Trim21/inj/blob/master/example/main.go#L27
> 
> with go 1.18 generic, we can have a 2 allocs and good performance and (IMO) easy to use API like this:
> 
> ```go
> app.Get("/:id", Wrap(func(c *fiber.Ctx, req Req) error {
> 	return c.JSON(req)
> }))
> ```
> 
> and a 5 fields benchmark, (`BenchmarkGorillaSchema_map_to_struct_only` simple use schema to convert map to struct, not even contains code to build map from ctx)
> 
> ```
> cpu: AMD Ryzen 7 5800X 8-Core Processor
> Benchmark_unmarshal_by_hand-16                          18636231                63.80 ns/op            0 B/op          0 allocs/op
> Benchmark_lib_unmarshal-16                               6516075               183.0 ns/op            96 B/op          2 allocs/op
> BenchmarkGorillaSchema_map_to_struct_only-16              435432              2818 ns/op             904 B/op         49 allocs/op
> PASS
> ```

I checked your code and it seems good idea to use context methods instead of creating temporarily map. I'll check it in the next PR

ReneWerner87 (approved):

PR #1983, opened 2022-07-26
Closes #1931 "🤗 How to get path param before a custom verb?"

### LINKED ISSUE #1931: 🤗 How to get path param before a custom verb?
**Question description**

The http request is `GET http://localhost:3000/resources/123:verb`

How to get `123` from request path? I tried `\\:customVerb` as the document said but `c.Params("id")` return nothing.

**Version**

v2.34.0

**Code snippet**
```go
package main

import (
	"github.com/gofiber/fiber/v2"
)

func main() {
	app := fiber.New()
	app.Get("/resources/:id\\:verb", func(c *fiber.Ctx) error {
		return c.SendString(c.Params("id"))
	})
	_ = app.Listen(":3000")
}
```

ReneWerner87:
Add possibility for parameters before custom verb

### DIFF
--- a/middleware/proxy/proxy_test.go
+++ b/middleware/proxy/proxy_test.go
@@ -124,7 +124,7 @@ func Test_Proxy_Balancer_WithTlsConfig(t *testing.T) {
 
 // go test -run Test_Proxy_Forward_WithTlsConfig_To_Http
 func Test_Proxy_Forward_WithTlsConfig_To_Http(t *testing.T) {
-	t.Parallel()
+	//t.Parallel()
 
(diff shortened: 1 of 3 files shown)


efectn (commented):
I think we should add a test case for Ctx_Params.

ReneWerner87:
> I think we should add a test case for Ctx_Params.

not really needed
https://github.com/gofiber/fiber/pull/1983/files#diff-e65edbcd5a006378e11ec33d7b121f21a20ec2631433f0b75bfa54e11768e9caR185 tested with this line

efectn (approved):

PR #1977, opened 2022-07-13
:sparkles: feature: customizable colors

efectn:
close https://github.com/gofiber/fiber/issues/1890

PR #1987, opened 2022-07-27
:art: Fix padding around app name in startup message when containing non-ascii characters

Maxi-Mega:
When the app name contains non-ascii chars (that are stored on more than 1 byte) , the padding around it in the startup message becomes incorrect. The veritable length of the app name can be obtained by getting the length of the underlying rune slice, like `len([]rune(s))`.

I added a test (`Test_App_Master_Process_Show_Startup_MessageWithAppNameNonAscii`) that creates an app with a name containing accentuated letters.

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -1528,6 +1528,16 @@ func Test_App_Master_Process_Show_Startup_MessageWithAppName(t *testing.T) {
 	utils.AssertEqual(t, true, strings.Contains(startupMessage, app.Config().AppName))
 }
 
+func Test_App_Master_Process_Show_Startup_MessageWithAppNameNonAscii(t *testing.T) {
+	appName := "Serveur de vérification des données"
+	app := New(Config{Prefork: true, AppName: appName})
+	startupMessage := captureOutput(func() {
+		app.startupMessage(":3000", false, "")
+	})
+	fmt.Println(startupMessage)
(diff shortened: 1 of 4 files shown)


ReneWerner87:
@Maxi-Mega can you refresh ? after the merge of https://github.com/gofiber/fiber/pull/1977 there are conflics

Maxi-Mega:
> @Maxi-Mega can you refresh ? after the merge of #1977 there are conflicts

Done, tell me if everything's ok

app_test.go
```diff
@@ -1401,6 +1401,16 @@ func Test_App_Master_Process_Show_Startup_MessageWithAppName(t *testing.T) {
 	utils.AssertEqual(t, true, strings.Contains(startupMessage, app.Config().AppName))
 }
 
+func Test_App_Master_Process_Show_Startup_MessageWithAppNameNonAscii(t *testing.T) {
```
efectn, line 1404: Can you move all startup message, route list tests to listen_test.go

efectn (changes requested):

app_test.go
Maxi-Mega: > Can you move all startup message, route list tests to listen_test.go

Sure

efectn (approved):

Issue #1996, opened 2022-08-06
🤗 [Question]: `app.Test` may hang infinitely with handler calling `runtime.Goexit`


Trim21:
### Question Description

I'm using `github.com/stretchr/testify/mock` to create a mock, and this library may call `t.FailNow()` if caller try to do a unexpected method call.

But `(*testing.T).FailNow()` will only call `runtime.Goexit` but not `panic`.

Therefore, the handler can't panic if the test fail. and I can only get a timeout error instead of panic error.

and, if `app.Test` is called without timeout, this test will hang **infinitely** .

I want to know is there a way to handle this problem?

### Code Snippet (optional)

```go
package main_test

import (
	"net/http"
	"net/http/httptest"
	"testing"

	"github.com/gofiber/fiber/v2"
)

func TestFailNow(t *testing.T) {
	app := fiber.New()

	app.Get("/", func(c *fiber.Ctx) error {
		t.FailNow() // or runtime.Goexit()
		return nil
	})

	req := httptest.NewRequest(http.MethodGet, "/", http.NoBody)

	res, err := app.Test(req, -1)
	if err != nil {
		t.Error(err)
		t.FailNow()
	}

	if res.StatusCode != 502 {
		t.FailNow()
	}
}
```

### Checklist:

- [X] I agree to follow Fiber's [Code of Conduct](https://github.com/gofiber/fiber/blob/master/.github/CODE_OF_CONDUCT.md).
- [X] I have checked for existing issues that describe my questions prior to opening this one.
- [X] I understand that improperly formatted questions may be closed without explanation.

Trim21:
There is a workaround, add a test-only middleware to handle this:

```golang
			app.Use(func(c *fiber.Ctx) error {
				var returned bool
				defer func() {
					if !returned {
						panic("t.FailNow() called in handler")
					}
				}()

				err := c.Next()
				returned = true
				return err
			})
```

`runtime.Goexit()` will call all `defer`, so in this way fiber can get the control flow.

wangjq4214:
`runtime.Goexit()` will call all `unexecuted defer` functions and **exit the current goroutine**, causing the error channel in Test function to not receive any messages.

And I think it's important to check the parameters at the controller level and make sure that the mock covers these test cases, rather than making the mock function call t.FailNow.

Changing app.Test will introduce unnecessary complexity.

Trim21:
this is more a developer experience question.

[...]

PR #1997, opened 2022-08-06
fix infinitely `app.Test`

Trim21:

details are explained in #1996

add a `defer` in `app.Test` so `runtime.Goexit` can be handled.


Trim21:
There is a data race error caused by `Client`, nothing to do with this PR.

ReneWerner87 (approved):

Issue #1903, opened 2022-05-08
 🚀 Request: Route Constraints 


CerealKiller97:
Route constraints execute when a match has occurred to the incoming URL and the URL path is tokenized into route values. I found this very useful when defining routes. ASP NET Core was inspiration for this feature.

# Example

```go
  app.Get("/users/{id:int}", func (ctx *fiber.Ctx) {
      // will come here if id is int
     // otherwise return 404
  })
```

Proposed constraints:

| Constraint  | Example           | Example matches              |
| ----------- | --------------  | ------------------------- |
| int                          |  {id:int}             | 123456789, -123456789 |
| bool                      |  {active:bool}   | true,false                            |
| datetime              | {dob:datetime} | 2016-12-31, 2016-12-31 7:32pm |
| guid                      | {id:guid}                        | CD2C1638-1638-72D5-1638-DEADBEEF1638|
| float                      | {weight:float}               | 1.234, -1,001.01e8 |
| minlength(value) | {username:minlength(4)} | Test (must be at least 4 characters) |
| maxlength(value) | {filename:maxlength(8)} | MyFile (must be no more than 8 characters |
| length(length) | {filename:length(12)} | somefile.txt (exactly 12 characters) |
| min(value) | {age:min(18)} | 19 (Integer value must be at least 18) |
| max(value) | {age:max(120)} | 91 (Integer value must be no more than 120) |
| range(min,max) | {age:range(18,120)} | 91 (Integer value must be at least 18 but no more than 120) |
| alpha | {name:alpha} | Rick (String must consist of one or more alphabetical characters, a-z and case-insensitive) |
| regex(expression) | {ssn:regex(^\\d{{3}}-\\d{{2}}-\\d{{4}}$)} | 123-45-6789 (Must match regular expression) |

More details: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/routing?view=aspnetcore-6.0#route-constraints

MohabMohamed:
hello @CerealKiller97, are you working on this feature?
if not, can you or @efectn assign it to me, I would love to work on it

efectn:
> hello @CerealKiller97, are you working on this feature? if not, can you or @efectn assign it to me, I would love to work on it

Assigned. If @CerealKiller97 isn't working on this, you can start working 🚀🚀

balcieren:
It is looking good. This feature can open from fiber.Config for v2.
Example:
```go
fiber.New(fiber.Config{
[...]

PR #1998, opened 2022-08-06
:sparkles: feature: route constraints 

efectn:
**Close:** https://github.com/gofiber/fiber/issues/1903

**Example:**
```go
app.Get("/:test<min(5)>", func(c *fiber.Ctx) error {
	return c.SendString(c.Params("test"))
})

app.Get("/:test<int;min(5)>", func(c *fiber.Ctx) error {
	return c.SendString(c.Params("test"))
})
```

**To-Do:**
- [x] Tests and benchmarks.
- [x] Multiple constraints.
- [x] Add customization for regex matching.
- [x] Performance checks.


path.go
```diff
@@ -368,3 +472,135 @@ func RemoveEscapeChar(word string) string {
...
+		fmt.Print(time.Parse(c.Data[0], param))
+		_, err = time.Parse(c.Data[0], param)
+	case regexConstraint:
+		match, _ := regexp.MatchString(c.Data[0], param)
+		if !match {
+			err = errors.New("")
+		}
+	}
+
+	return err == nil
```
efectn, line 605: should we skip or return false when consantraint not found

path_test.go
```diff
@@ -416,6 +416,124 @@ func Test_Path_matchParams(t *testing.T) {
...
+		{url: "/api/v1/5", params: []string{"5"}, match: true},
+		{url: "/api/v1/15", params: []string{"15"}, match: false},
+	})
+	testCase("/api/v1/:param<range(5,10)>", []testparams{
+		{url: "/api/v1/ent", params: []string{"ent"}, match: false},
+		{url: "/api/v1/9", params: []string{"9"}, match: true},
+		{url: "/api/v1/5", params: []string{"5"}, match: true},
+		{url: "/api/v1/15", params: []string{"15"}, match: false},
+	})
+	testCase("/api/v1/:param<datetime(2006\\-01\\-02)>", []testparams{
```
efectn, line 503: don't know why it fails on github actiıons

path.go
```diff
@@ -370,3 +479,134 @@ func RemoveEscapeChar(word string) string {
 	}
 	return word
 }
+
+func getParamConstraintType(constraintPart string) TypeConstraint {
+	switch constraintPart {
+	case "int":
```
ReneWerner87, line 485: add constants for these wordings

utils/regex.go
```diff
@@ -0,0 +1,4 @@
+package utils
+
+// RegexMatch returns the matching s by regex pattern.
+type RegexMatch func(pattern string, s string) (matched bool, err error)
```
ReneWerner87, line 4: good idea

router.go
```diff
@@ -61,7 +62,13 @@ type Route struct {
 	Handlers []Handler `json:"-"`      // Ctx handlers
 }
 
-func (r *Route) match(detectionPath, path string, params *[maxParams]string) (match bool) {
+func (r *Route) match(detectionPath, path string, params *[maxParams]string, matcher ...utils.RegexMatch) (match bool) {
```
ReneWerner87, line 65: don't like having a functionality with too many parameters 

cleancode says that this should be rethought and reimplemented

app.go
```diff
@@ -323,6 +324,13 @@ type Config struct {
 	// Default: json.Unmarshal
 	JSONDecoder utils.JSONUnmarshal `json:"-"`
 
+	// When set by an external client of Fiber it will use the provided implementation of a
+	// RegexMatcher
+	//
+	// Allowing for flexibility in using another regex library for matching
+	// Default: regexp.MatchString
+	RegexMatcher utils.RegexMatch `json:"-"`
```
ReneWerner87, line 332: please set the "regexp.MatchString" as default matcher, then we don't need optional parameters in the match functionality anymore

path.go
```diff
@@ -250,7 +345,13 @@ func findNextNonEscapedCharsetPosition(search string, charset []byte) int {
 }
 
 // getMatch parses the passed url and tries to match it against the route segments and determine the parameter positions
-func (routeParser *routeParser) getMatch(detectionPath, path string, params *[maxParams]string, partialCheck bool) bool {
+func (routeParser *routeParser) getMatch(detectionPath, path string, params *[maxParams]string, partialCheck bool, matcher ...utils.RegexMatch) bool {
+	// Override matcher
+	var regexMatch = regexp.MatchString
```
ReneWerner87, line 350: this is likely to lead to allocations

router.go
```diff
@@ -61,7 +62,13 @@ type Route struct {
...
+	// Override matcher
+	var regexMatch = regexp.MatchString
```
ReneWerner87, line 67: this is likely to lead to allocations

path.go
```diff
@@ -370,3 +479,134 @@ func RemoveEscapeChar(word string) string {
...
+		data2, _ := strconv.Atoi(c.Data[1])
+		num, err = strconv.Atoi(param)
+
+		if !(num >= data && num <= data2) {
+			err = errors.New("")
+		}
+	case datetimeConstraint:
+		_, err = time.Parse(c.Data[0], param)
+	case regexConstraint:
+		match, _ := matcher(c.Data[0], param)
```
ReneWerner87, line 605: precompiled matchers are faster

when writing the contrains into the route, we should take the regex based on the passed regex matcher in the app configuration and precompile the regular expression, so that the additional parameter which is injected via the "match" can be omitted and by precompiling (which happens when building the routes -> so is not performance critical) the performance increases

app.go
efectn: What about adding private var? We can assign app config to that variable. It'd be easier to get it outsuide of app methods.

path.go
```diff
@@ -370,3 +479,134 @@ func RemoveEscapeChar(word string) string {
...
+	case intConstraint:
+		_, err = strconv.Atoi(param)
+	case boolConstraint:
+		_, err = strconv.ParseBool(param)
+	case floatConstraint:
+		_, err = strconv.ParseFloat(param, 32)
+	case alphaConstraint:
+		for _, r := range param {
+			if !unicode.IsLetter(r) {
+				err = errors.New("")
```
ReneWerner87, line 549: please come up with something else for the error messages

would say that we create predefined errors with placeholders, which can be overlaid from the outside and then rendered in this function

the errors must then be passed up and output 
of course only when the route is normally matched without contrains check and then the contrains will be checked and any errors will be printed out 

otherwise you leave the errors out of the contrains check completely

ReneWerner87 (changes requested):
Please add more benchmark tests, so we can keep an eye on the most important thing -> the performance

app.go
ReneWerner87: just do it like this 
https://github.com/gofiber/fiber/blob/e5eb8d3c98b33d8047cdc27bc165330f4f8b5312/app.go#L506-L510

router.go
efectn: may be we can add `defaultMatcher` variable (local) and set it inside of the `New()` instead of adding new param.

path.go
efectn: yes. it's good idea to use Compile for performance. But it'll be difficult to add customization for regex section it after using Compile. It seems there's a deal :smiley:

path.go
```diff
@@ -370,3 +505,142 @@ func RemoveEscapeChar(word string) string {
...
+		data2, _ := strconv.Atoi(c.Data[1])
+		num, err = strconv.Atoi(param)
+
+		if num < data || num > data2 {
+			return false
+		}
+	case datetimeConstraint:
+		_, err = time.Parse(c.Data[0], param)
+	case regexConstraint:
+		if c.RegexCompiler == nil {
```
ReneWerner87, line 631: why not precompile the regex when the route is specified in the registers

if then the requests are parsed this goes on the performance

ReneWerner87 (approved):

PR #2003, opened 2022-08-08
:sparkles: feature: add XML to context.

efectn:

**Commit formatting** 

### DIFF
--- a/app.go
+++ b/app.go
@@ -330,7 +330,7 @@ type Config struct {
 	//
 	// Allowing for flexibility in using another XML library for encoding
 	// Default: xml.Marshal
-	XMLEncoder utils.JSONMarshal `json:"-"`
+	XMLEncoder utils.XMLMarshal `json:"-"`
 
 	// Known networks are "tcp", "tcp4" (IPv4-only), "tcp6" (IPv6-only)
 	// WARNING: When prefork is set to true, only "tcp4" and "tcp6" can be chose.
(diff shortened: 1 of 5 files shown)


ctx.go
```diff
@@ -736,6 +731,18 @@ func (c *Ctx) JSONP(data interface{}, callback ...string) error {
 	return c.SendString(result)
 }
 
+// XML converts any interface or string to XML.
+// This method also sets the content header to application/xml.
+func (c *Ctx) XML(data interface{}) error {
```
Trim21, line 736: use `any` considering we are already using go1.18?

efectn: This is v2

Trim21: oh sorry 😥

efectn: No problem. I'll convert it to any when merging master into the v3-beta.

app.go
```diff
@@ -323,6 +325,13 @@ type Config struct {
 	// Default: json.Unmarshal
 	JSONDecoder utils.JSONUnmarshal `json:"-"`
 
+	// XMLEncoder set by an external client of Fiber it will use the provided implementation of a
+	// XMLMarshal
+	//
+	// Allowing for flexibility in using another XML library for encoding
+	// Default: xml.Marshal
+	XMLEncoder utils.JSONMarshal `json:"-"`
```
ReneWerner87, line 333: you mean utils.XMLMarshal

ReneWerner87 (changes requested):

app.go
efectn: Done

ReneWerner87 (approved):

PR #2004, opened 2022-08-08
🐛 bug: fix proxy overwrote the wrong scheme

### LINKED ISSUE #2001: 🐛 [Bug]: Middleware proxy.Do overwrote the wrong scheme
### Bug Description

I am making a simple proxy that you can append the proxied url after /, for example, you can request http://localhost:3333/http://ip-api.com/json/1.1.1.1 , which will proxy request to http://ip-api.com/json/1.1.1.1.

Here is my code:

```golang
package main

import (
	"strings"

	"github.com/gofiber/fiber/v2"
	"github.com/gofiber/fiber/v2/middleware/proxy"
)

func main() {
	app := fiber.New()
	app.Get("/*", func(c *fiber.Ctx) error {
		path := c.OriginalURL()
		url := strings.TrimPrefix(path, "/")
		if err := proxy.Do(c, url); err != nil {
			return err
		}
		c.Response().Header.Del(fiber.HeaderServer)
		return nil
	})
	app.Listen(":3333")
}
```

When I request the url like: http://localhost:3333/http://ip-api.com/json/1.1.1.1 , it will show this error:
> unsupported protocol "ttp". http and https are supported

By debugging and following the code, the scheme has been overwritten to "ttp" at this line: https://github.com/gofiber/fiber/blob/master/middleware/proxy/proxy.go#L131 , since at some point the `addr` variable has changed to "ttp://xxxxx" (without the leading "h"), the parsed scheme changed to "ttp". It seems related to the memory allocation, the problem will be gone, if in my code I either:

* change the fiber config to use `Immutable: true`, or 
* change `path := c.OriginalURL()` to `path := utils.CopyString(c.OriginalURL())` .

But since the code above doesn't violate fiber's rule:
> As a rule of thumb, you must only use context values within the handler, and you must not keep any references.

Even if I don't use `Immutable: true` or `utils.CopyString`, the problem shouldn't be there. 

Also, this bug doesn't exist when I was using version v2.10.0, which doesn't have these 3 lines: https://github.com/gofiber/fiber/blob/master/middleware/proxy/proxy.go#L130

### How to Reproduce

Copy the above code and run it with fiber v2.36.0, then visit: http://localhost:3333/http://ip-api.com/json/1.1.1.1 .
You should be able to see the error: 
> unsupported protocol "ttp". http and https are supported

I can reproduce this in my 3 laptops: Macbook m1, Mackbook m2 and Macbook Intel.

### Expected Behavior

The error shouldn't be there (the scheme shouldn't be incorrectly overwritten) even without "Immutable: true" config or CopyString

### Fiber Version

v2.36.0

### Code Snippet (optional)

See above

### Checklist:

wangjq4214:

Fix proxy overwrote the wrong scheme

close #2001

**Commit formatting** 

### DIFF
--- a/middleware/proxy/proxy.go
+++ b/middleware/proxy/proxy.go
@@ -123,11 +123,13 @@ func Do(c *fiber.Ctx, addr string) error {
 	res := c.Response()
 	originalURL := utils.CopyString(c.OriginalURL())
 	defer req.SetRequestURI(originalURL)
-	req.SetRequestURI(addr)
+
+	copiedURL := utils.CopyString(addr)
+	req.SetRequestURI(copiedURL)
 	// NOTE: if req.isTLS is true, SetRequestURI keeps the scheme as https.
 	// issue reference:
 	// https://github.com/gofiber/fiber/issues/1762
-	if scheme := getScheme(utils.UnsafeBytes(addr)); len(scheme) > 0 {
+	if scheme := getScheme(utils.UnsafeBytes(copiedURL)); len(scheme) > 0 {
 		req.URI().SetSchemeBytes(scheme)
 	}
 
--- a/middleware/proxy/proxy_test.go
+++ b/middleware/proxy/proxy_test.go
@@ -2,7 +2,6 @@ package proxy
 
 import (
 	"crypto/tls"
-	"io"
 	"io/ioutil"
 	"net"
 	"net/http"
@@ -387,7 +386,7 @@ func Test_Proxy_Do_HTTP_Prefix_URL(t *testing.T) {
 
 	resp, err := app.Test(httptest.NewRequest(http.MethodGet, "/http://"+addr, nil))
 	utils.AssertEqual(t, nil, err)
-	s, err := io.ReadAll(resp.Body)
+	s, err := ioutil.ReadAll(resp.Body)
 	utils.AssertEqual(t, nil, err)
 	utils.AssertEqual(t, "hello world", string(s))
 }
--- a/middleware/proxy/proxy_test.go
+++ b/middleware/proxy/proxy_test.go
@@ -2,8 +2,10 @@ package proxy
 
 import (
 	"crypto/tls"
+	"io"
 	"io/ioutil"
 	"net"
+	"net/http"
 	"net/http/httptest"
 	"strings"
 	"testing"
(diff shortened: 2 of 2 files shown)


middleware/proxy/proxy.go
```diff
@@ -123,11 +123,13 @@ func Do(c *fiber.Ctx, addr string) error {
 	res := c.Response()
 	originalURL := utils.CopyString(c.OriginalURL())
 	defer req.SetRequestURI(originalURL)
-	req.SetRequestURI(addr)
+
+	copiedURL := utils.CopyString(addr)
+	req.SetRequestURI(copiedURL)
```
ReneWerner87, line 128:
```suggestion
	req.SetRequestURI(utils.CopyString(addr))
```
this should be enough, because within the string is changed

middleware/proxy/proxy.go
```diff
@@ -123,11 +123,13 @@ func Do(c *fiber.Ctx, addr string) error {
...
 	// NOTE: if req.isTLS is true, SetRequestURI keeps the scheme as https.
 	// issue reference:
 	// https://github.com/gofiber/fiber/issues/1762
-	if scheme := getScheme(utils.UnsafeBytes(addr)); len(scheme) > 0 {
+	if scheme := getScheme(utils.UnsafeBytes(copiedURL)); len(scheme) > 0 {
```
ReneWerner87, line 132:
```suggestion
	if scheme := getScheme(utils.UnsafeBytes(addr)); len(scheme) > 0 {
```
and this back to before

ReneWerner87 (changes requested):

ReneWerner87:
thanks for the work, can you look over my comments

middleware/proxy/proxy.go
wangjq4214: This modification still has the original problem, and I haven't explored the cause. The `addr` will change after running `req.SetRequestURI` regardless of the data passed in.

I'll probably explore the reasons over the weekend.

wangjq4214:
> thanks for the work, can you look over my comments

In this scenario, addr is a slice reference to the `[]byte c.fasthttp.Request.Header.requestURI`, so when **SetRequestURI** is executed the requestURI changes causing addr to change as well. 

More specifically, for the above example, addr is equivalent to requestURI[1:], so after the requestURI changes from /http://ip-api.com/json to http://ip-api.com/json addr will change from http://ip-api.com/json to ttp://ip-api.com/json, causing SetSchema to fail.

ReneWerner87:
since this removes the link to the variable with the addr, the following change is then no longer necessary

wangjq4214:
> since this removes the link to the variable with the addr, the following change is then no longer necessary

image

image

When I use the modifications you mentioned above, I can't pass the test.

In SetRequestURI, `h.requestURI = append(h.requestURI[:0], requestURI...)` will be executed. Because the resulting []byte is shorter than the original one, he will execute in the original []byte. And no memory will be allocated.

middleware/proxy/proxy.go
ReneWerner87: i was wrong -> your code is okay

middleware/proxy/proxy.go
ReneWerner87: i was wrong -> your code is okay

ReneWerner87 (approved):

PR #2010, opened 2022-08-11
🐛 bug: Case sensitivity for parameters in GetRouteURL

### LINKED ISSUE #1907: 🤗 Case sensitivity for parameters in GetRouteURL
**Question description**

When using capitalized letters in `fiber.Map{}` with `GetRouteURL`, the case sensitivity of the keys is ignored.

**Code snippet** 

```go
package main

func main() {
        app := fiber.New()

	// app := fiber.New(fiber.Config{
	// 	CaseSensitive: true,
	// })

		idUrl, err := c.GetRouteURL("id", fiber.Map{
			"testid": "ABC",
			"testId": "DEF",
		})

		if err != nil {
			panic(err)
		}

		return c.SendString(idUrl)
	})

	app.Get("/:testid/:testId/other_paths", func(c *fiber.Ctx) error {
		return c.SendString("Hello, World!\ntestid: " + c.Params("testid") + "\ntestId: " + c.Params("testId"))
	}).Name("id")

	app.Listen(":3001")
}

```
For example, GET `/`, you will receive `/ABC/ABC/other_paths`. The case sensitivity for GetRouteURL is ignored

On the other hand, c.Params can tell the difference between `testid` and `testId`.

I think that `Params` and `GetRouteURL` should behave the same behavior on case sensitivity.

---

Meanwhile, enabling `CaseSensitive` in the config will receive `/ABC/DEF/other_paths`

wangjq4214:

Try to fix case sensitivity for parameters in GetRouteURL.

close #1907 

**Commit formatting** 

### DIFF
--- a/ctx_test.go
+++ b/ctx_test.go
@@ -1350,6 +1350,11 @@ func Test_Ctx_Params(t *testing.T) {
 		utils.AssertEqual(t, "", c.Params("optional"))
 		return nil
 	})
+	app.Get("/test5/:id/:Id", func(c *Ctx) error {
+		utils.AssertEqual(t, "first", c.Params("id"))
+		utils.AssertEqual(t, "first", c.Params("Id"))
+		return nil
+	})
 	resp, err := app.Test(httptest.NewRequest(MethodGet, "/test/john", nil))
 	utils.AssertEqual(t, nil, err, "app.Test(req)")
 	utils.AssertEqual(t, StatusOK, resp.StatusCode, "Status code")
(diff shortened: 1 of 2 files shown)


ctx_test.go
```diff
@@ -1365,6 +1370,32 @@ func Test_Ctx_Params(t *testing.T) {
 	resp, err = app.Test(httptest.NewRequest(MethodGet, "/test4", nil))
 	utils.AssertEqual(t, nil, err, "app.Test(req)")
 	utils.AssertEqual(t, StatusOK, resp.StatusCode, "Status code")
+
+	resp, err = app.Test(httptest.NewRequest(MethodGet, "/test5/first/second", nil))
+	utils.AssertEqual(t, nil, err)
+	utils.AssertEqual(t, StatusOK, resp.StatusCode, "Status code")
+}
+
+func Test_Ctx_Params_Case_Sencitive(t *testing.T) {
```
efectn, line 1379: Can you fix typo

efectn (approved):

ReneWerner87:
- [ ] before the merge I will check the performance

wangjq4214:
image

The `EqualFold` took a lot of time, and according to my Beckmark data, its performance dropped by 50% with a lot of overhead.

A better way to solve this problem may be needed.

ReneWerner87:
yes I have also noticed, only I can not think of anything better right now

so already, only unfortunately with much overhead
you could keep a list where the parameter names are already uppercase and make the key for the check uppercase before the pass through
but that is pretty overhead

wangjq4214:
> yes I have also noticed, only I can not think of anything better right now
> 
> so already, only unfortunately with much overhead you could keep a list where the parameter names are already uppercase and make the key for the check uppercase before the pass through but that is pretty overhead

Do you mean that when calling Param I should check if the key contains uppercase letters? If it contains uppercase letters then should I use the already uppercase list to match param?

I think if it is as I understand it, if the list only contains the path param name defined in the user uri then it is still case sensitive and if the list contains all possible combinations then is the memory overhead higher.

If I don't understand it correctly, can you explain it in more detail?

ReneWerner87:
Yes you have misunderstood

One way to speed it up would be at the beginning when the route is registered and the parameter names are extracted, create a list with them where they are already all uppercase and then use this list when comparing them

Before of course convert the parameter string which is used in the method(ctx.Params) to uppercase, after that you can test without EqualFold with normal string comparison.

ReneWerner87:
But the effort is not compared to the benefit

ReneWerner87:
https://github.com/gofiber/fiber/blob/master/router.go#L263
Normally we only need the params from the prettyied route

wangjq4214:
> But the effort is not compared to the benefit

I agree with you, I'll go back and try it after I basically finish the testing work and documentation for the client refactor this weekend.

wangjq4214:
> Yes you have misunderstood
> 
> One way to speed it up would be at the beginning when the route is registered and the parameter names are extracted, create a list with them where they are already all uppercase and then use this list when comparing them
> 
> Before of course convert the parameter string which is used in the method(ctx.Params) to uppercase, after that you can test without EqualFold with normal string comparison.

There is still a 30% drop in performance and additional memory and time overhead is introduced during route registration.

ReneWerner87:
Ok

ReneWerner87:
I think the last stand is okay. Will merge it in a few hours.

wangjq4214:
> I think the last stand is okay. Will merge it in a few hours.

Do we need commit 4c97ad843b3fdd13d09625a64eba6a821faf1237 ? Otherwise I need to roll him back.

PR #2011, opened 2022-08-12
✨ update: tls.ClientHelloInfo in Ctx

thomasdseao:
This PR is related to a request : https://github.com/orgs/gofiber/projects/1/views/4

Add the function ClientHelloInfo to the fiber Ctx so everybody can access it from handler.

### DIFF
--- a/app.go
+++ b/app.go
@@ -112,6 +112,8 @@ type App struct {
 	// Latest route & group
 	latestRoute *Route
 	latestGroup *Group
+	// TLS handler
+	tlsHandler *tlsHandler
 }
 
 // Config is a struct holding the server settings.
@@ -366,9 +368,6 @@ type Config struct {
 	//
 	// Optional. Default: DefaultColors
 	ColorScheme Colors `json:"color_scheme"`
-
-	// TLS handler
-	tlsHandler *tlsHandler
 }
 
 // Static defines configuration options when defining static assets.
--- a/app.go
+++ b/app.go
@@ -366,6 +366,9 @@ type Config struct {
 	//
 	// Optional. Default: DefaultColors
 	ColorScheme Colors `json:"color_scheme"`
+
+	// TLS handler
+	tlsHandler *tlsHandler
 }
 
 // Static defines configuration options when defining static assets.
--- a/app.go
+++ b/app.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	"encoding/json"
+	"encoding/xml"
 
 	"github.com/gofiber/fiber/v2/utils"
 	"github.com/valyala/fasthttp"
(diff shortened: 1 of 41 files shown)


app.go
```diff
@@ -366,6 +366,9 @@ type Config struct {
 	//
 	// Optional. Default: DefaultColors
 	ColorScheme Colors `json:"color_scheme"`
+
+	// TLS handler
+	tlsHandler *tlsHandler
```
efectn, line 371: This can be a part of App struct

ctx_test.go
```diff
@@ -1247,6 +1247,28 @@ func Test_Ctx_Method(t *testing.T) {
...
 }
 
+// go test -run Test_Ctx_ClientHelloInfo
+func Test_Ctx_ClientHelloInfo(t *testing.T) {
+	t.Parallel()
+	app := New()
+	app.Post("/test", func(c *Ctx) error {
+		result := c.ClientHelloInfo()
+		if result != nil {
+			panic("Invalid ClientHelloInfo, it should be nil")
```
efectn, line 1257: Return an error with NewError instead of `panic()`. And we may need more tests i suppose

app.go
thomasdseao: Changed 👍

ctx_test.go
thomasdseao: I add more tests on this, please check it and tell me if it looks good for you 👍

efectn:
I think this method should be in App. Can you move it into the app?

wireguard-dev:
> I think this method should be in App. Can you move it into the app?

@efectn Why? I think it must be in context directly, not in App()

efectn:
> > I think this method should be in App. Can you move it into the app?
> 
> @efectn Why? I think it must be in context directly, not in App()

It doesn't have context-specific thing

wireguard-dev:
> > > I think this method should be in App. Can you move it into the app?
> > 
> > 
> > @efectn Why? I think it must be in context directly, not in App()
> 
> It doesn't have context-specific thing

tls.ClientHelloInfo, is sent by the browser when the user connects to the website, so I think it should be in context.

thomasdseao:
Yes the CHI is sent by the client, that's why I put it in the context.
But if you think this is more compliant with Fiber logical to have it in App, I can move it.

ReneWerner87:
@thomasdseao 
can you please add a pull request to our documentation repository so that everyone knows what the point of it is and how to use it?
https://github.com/gofiber/docs/pulls

thomasdseao:
@ReneWerner87 That's done : https://github.com/gofiber/docs/pull/271

ReneWerner87 (approved):

ctx.go
```diff
@@ -130,6 +143,8 @@ func (app *App) AcquireCtx(fctx *fasthttp.RequestCtx) *Ctx {
 	c.fasthttp = fctx
 	// reset base uri
 	c.baseURI = ""
+	// Attach tlsHandler object to context
+	c.tlsHandler = app.tlsHandler
```
efectn, line 147: i think we can get it as `c.app.tlsHandler` instead of adding new field.

PR #2020, opened 2022-08-17
fixes #2016 - make IP() and IPs() more reliable

### LINKED ISSUE #2016: 🐛 [Bug]: `IP()` function does not return client IP reliably
### Bug Description

According to documentation for `func (c *Ctx) IP() string`: 

> IP returns the remote IP address of the request.

However, the logic in this function is too basic and cannot be trusted. Here are some issues for example:

- when setting `ProxyHeader: "X-Forwarded-For"` (or any other value), if this header is missing or empty then fiber will return an empty value for `c.IP()`. I think it would be better if the value defaulted to `fasthttp.RemoteIP().String()` when the value of this header is missing/empty
- no validation or modifications are done on the return value of `c.Get(c.app.config.ProxyHeader)`. This means that it's possible to return any string that does not even represent an IP

### How to Reproduce

Example App
```
package main

import (
	"github.com/gofiber/fiber/v2"
	"log"
)

func main() {
	app := fiber.New(fiber.Config{ProxyHeader: "X-Forwarded-For"})
	app.Get("/ip", func(c *fiber.Ctx) error {
		return c.JSON(fiber.Map{"ip": c.IP()})
	})
	log.Fatal(app.Listen(":19876"))
}
```

Testing
```
# Empty when proxy header is missing
❯ curl http://127.0.0.1:19876/ip
{"ip":""}

# Populated when proxy header has a value
❯ curl -H "X-Forwarded-For: 1.1.1.1" http://127.0.0.1:19876/ip
{"ip":"1.1.1.1"}

# The value is not checked, just simply passed through
❯ curl -H "X-Forwarded-For: whatever" http://127.0.0.1:19876/ip
{"ip":"whatever"}

❯ curl -H "X-Forwarded-For: 1.1.1.1, 2.2.2.2" http://127.0.0.1:19876/ip
{"ip":"1.1.1.1, 2.2.2.2"}
```

### Expected Behavior

```
# default to TCP IP address that you received the request from
❯ curl http://127.0.0.1:19876/ip
{"ip":"127.0.0.1"}

# same default if the proxy header does not make sense
❯ curl -H "X-Forwarded-For: whatever" http://127.0.0.1:19876/ip
{"ip":"127.0.0.1"}

# attempt to interpret what the IP of the actual client is
❯ curl -H "X-Forwarded-For: 1.1.1.1, 2.2.2.2" http://127.0.0.1:19876/ip
{"ip":"1.1.1.1"}
```

### Fiber Version

v2.36.0

### Code Snippet (optional)

```go
package main

import (
	"github.com/gofiber/fiber/v2"
	"log"
)

func main() {
	app := fiber.New(fiber.Config{ProxyHeader: "X-Forwarded-For"})
	app.Get("/ip", func(c *fiber.Ctx) error {
		return c.JSON(fiber.Map{"ip": c.IP()})
	})
	log.Fatal(app.Listen(":19876"))
}
```

### Checklist:

gbolo:
see #2016

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -649,33 +649,49 @@ func (c *Ctx) Port() string {
 }
 
 // IP returns the remote IP address of the request.
+// If ProxyHeader is configured, it will parse that header and return the first valid IP address
 // Please use Config.EnableTrustedProxyCheck to prevent header spoofing, in case when your app is behind the proxy.
 func (c *Ctx) IP() string {
 	if c.IsProxyTrusted() && len(c.app.config.ProxyHeader) > 0 {
-		return c.Get(c.app.config.ProxyHeader)
+		return c.extractIPFromHeader(c.app.config.ProxyHeader)
 	}
 
 	return c.fasthttp.RemoteIP().String()
 }
 
-// IPs returns an string slice of IP addresses specified in the X-Forwarded-For request header.
-func (c *Ctx) IPs() (ips []string) {
-	header := c.fasthttp.Request.Header.Peek(HeaderXForwardedFor)
-	if len(header) == 0 {
-		return
-	}
-	ips = make([]string, bytes.Count(header, []byte(","))+1)
-	var commaPos, i int
-	for {
-		commaPos = bytes.IndexByte(header, ',')
-		if commaPos != -1 {
-			ips[i] = utils.Trim(c.app.getString(header[:commaPos]), ' ')
-			header, i = header[commaPos+1:], i+1
-		} else {
-			ips[i] = utils.Trim(c.app.getString(header), ' ')
-			return
+// extractValidIPs will return a slice of strings that represent valid IP addresses
+// in the input string. The order is maintained. The separator is a comma
+func extractValidIPs(input string) (validIPs []string) {
+	unvalidatedIps := strings.Split(input, ",")
+	for _, ip := range unvalidatedIps {
+		if parsedIp := net.ParseIP(strings.TrimSpace(ip)); parsedIp != nil {
+			validIPs = append(validIPs, parsedIp.String())
 		}
 	}
+	return
+}
+
+// extractIPFromHeader will attempt to pull the real client IP from the given header
+// currently it will return the first valid IP address in header
+func (c *Ctx) extractIPFromHeader(header string) string {
+	// extract only valid IPs from the header's value
+	validIps := extractValidIPs(c.Get(header))
+
+	// since X-Forwarded-For has no RFC, it's really up to the proxy to decide whether to append
+	// or prepend IPs to this list. For example, the AWS ALB will prepend but the F5 BIG-IP will append ;(
+	// for now lets just go with the first value in the list...
+	if len(validIps) > 0 {
+		return validIps[0]
+	}
+
+	// return the IP from the stack if we could not find any valid Ips
+	return c.fasthttp.RemoteIP().String()
+}
+
+// IPs returns a string slice of IP addresses specified in the X-Forwarded-For request header.
+// Only valid IP addresses are returned
+func (c *Ctx) IPs() (ips []string) {
+	return extractValidIPs(c.Get(HeaderXForwardedFor))
 }
 
 // Is returns the matching content type,
--- a/ctx.go
+++ b/ctx.go
@@ -662,10 +662,38 @@ func (c *Ctx) IP() string {
 // extractValidIPs will return a slice of strings that represent valid IP addresses
 // in the input string. The order is maintained. The separator is a comma
 func extractValidIPs(input string) (validIPs []string) {
-	unvalidatedIps := strings.Split(input, ",")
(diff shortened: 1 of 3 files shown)


ctx.go
```diff
@@ -649,33 +649,49 @@ func (c *Ctx) Port() string {
...
-		if commaPos != -1 {
-			ips[i] = utils.Trim(c.app.getString(header[:commaPos]), ' ')
-			header, i = header[commaPos+1:], i+1
-		} else {
-			ips[i] = utils.Trim(c.app.getString(header), ' ')
-			return
+// extractValidIPs will return a slice of strings that represent valid IP addresses
+// in the input string. The order is maintained. The separator is a comma
+func extractValidIPs(input string) (validIPs []string) {
+	unvalidatedIps := strings.Split(input, ",")
```
ReneWerner87, line 665: this leads to further allocations
what does the benchmark say about the adjustment ?

please try to minimize the allocations so that we keep our perfomance promise for the consumers

ReneWerner87 (changes requested):

gbolo:
> @gbolo can you try to make it more allocation less ? to improve the performance

@ReneWerner87 I can try. Do you have any suggestions on an approach? Previously for the length of the ips slice, it was easy to figure out because you simply just counted the numbers of commas + 1:
```
ips = make([]string, bytes.Count(header, []byte(","))+1)
```

However, it's a little trickier now because we need to validate the input first. I will try another approach and run benchmarks to measure the impact.

ReneWerner87:
image
the old approach is not possible ? 
or something like this
https://github.com/gofiber/fiber/blob/1fec8757c4115a5c1405733fb5cd61e66a364223/ctx.go#L173-L180
without storing the ip in a separate slice

gbolo:
I have improved the performance and reduced allocations of this functionality. I will continue to improve it more this week. I was also thinking that it may be a good idea to allow IP validation to be a configuration option. What do you think @ReneWerner87 ?

Current performance:
```
❯ go test -v -run=ctx_test.go -bench=Benchmark_Ctx_IP -benchmem -count=4
goos: linux
goarch: amd64
cpu: 11th Gen Intel(R) Core(TM) i7-11700 @ 2.50GHz
Benchmark_Ctx_IPs
Benchmark_Ctx_IPs-16                    	 4877410	       278.6 ns/op	      96 B/op	       4 allocs/op
Benchmark_Ctx_IPs-16                    	 4073689	       271.5 ns/op	      96 B/op	       4 allocs/op
Benchmark_Ctx_IPs-16                    	 3561628	       316.0 ns/op	      96 B/op	       4 allocs/op
Benchmark_Ctx_IPs-16                    	 3180038	       323.1 ns/op	      96 B/op	       4 allocs/op
```

Previous performance with high allocations:
```
❯ go test -v -run=ctx_test.go -bench=Benchmark_Ctx_IPs -benchmem -count=4
goos: linux
goarch: amd64
cpu: 11th Gen Intel(R) Core(TM) i7-11700 @ 2.50GHz
Benchmark_Ctx_IPs
Benchmark_Ctx_IPs-16    	 2446430	       769.3 ns/op	     256 B/op	      10 allocs/op
Benchmark_Ctx_IPs-16    	 1783593	       675.9 ns/op	     256 B/op	      10 allocs/op
Benchmark_Ctx_IPs-16    	 1708234	       587.8 ns/op	     256 B/op	      10 allocs/op
Benchmark_Ctx_IPs-16    	 1299441	       797.3 ns/op	     256 B/op	      10 allocs/op
```

Existing performance with NO ip validation
```
❯ go test -v -run=ctx_test.go -bench=Benchmark_Ctx_IP -benchmem -count=4
goos: linux
goarch: amd64
cpu: 11th Gen Intel(R) Core(TM) i7-11700 @ 2.50GHz
Benchmark_Ctx_IPs
Benchmark_Ctx_IPs-16    	15274425	       119.9 ns/op	      48 B/op	       1 allocs/op
Benchmark_Ctx_IPs-16    	11612115	       118.6 ns/op	      48 B/op	       1 allocs/op
Benchmark_Ctx_IPs-16    	 9998650	       132.9 ns/op	      48 B/op	       1 allocs/op
Benchmark_Ctx_IPs-16    	11643238	       137.8 ns/op	      48 B/op	       1 allocs/op
```

ReneWerner87:
thank you, would say ip validation is not needed for now

ReneWerner87 (approved):

ReneWerner87:
are you finished or do you want to make adjustments ? otherwise i would merge

gbolo:
> are you finished or do you want to make adjustments ? otherwise i would merge

please don't merge yet, I want to make some more adjustments

ReneWerner87:
For new features don't forget to create a doc entry in our documentation repository.

gbolo:
> For new features don't forget to create a doc entry in our documentation repository.

done: https://github.com/gofiber/docs/pull/272

PR #2024, opened 2022-08-19
🐛 monitor middleware - fix ignore custom settings

### LINKED ISSUE #2019: 🐛 [Bug]: Monitor middleware ignores custom settings if default title is used.
### Bug Description

When using the monitor middleware I noticed that if I don't change the title or refresh rate the other custom settings don't get applied. 

The bug is on this line: https://github.com/gofiber/fiber/blob/master/middleware/monitor/config.go#L99

Where if title or refresh rate are the `Default Values` everything else gets ignored in the `else`.

### How to Reproduce

Run code

### Expected Behavior

Custom settings should be applied regardless of title/refresh rate.

### Fiber Version

v2.36.0

### Code Snippet (optional) (Bug, Font/Chart URLs are ignored)

```go
package main

import (
	"log"

	"github.com/gofiber/fiber/v2"
	"github.com/gofiber/fiber/v2/middleware/monitor"
)

func main() {
	app := fiber.New()

	app.Get("/metrics", monitor.New(monitor.Config{
	    FontURL: "blah/blah.css",
	    ChartJsURL: "blah/blah.js",
    }))

}
```

### Code Snippet (Working, Font/Chart URLs get applied)

```go
package main

import (
	"log"

	"github.com/gofiber/fiber/v2"
	"github.com/gofiber/fiber/v2/middleware/monitor"
)

func main() {
	app := fiber.New()

	app.Get("/metrics", monitor.New(monitor.Config{
	    Title: "MyService Metrics Page",
	    FontURL: "blah/blah.css",
	    ChartJsURL: "blah/blah.js",
    }))

}
```

This bug doesn't show up in monitor_test.go, because the test is changing the `Title` to "New + DefaultTitle".

### Checklist:

wangjq4214:

Fix monitor middleware ignores custom settings if default title is used.

close #2019

### DIFF
--- a/middleware/monitor/config.go
+++ b/middleware/monitor/config.go
@@ -96,15 +96,8 @@ func configDefault(config ...Config) Config {
 	if cfg.ChartJsURL == "" {
 		cfg.ChartJsURL = defaultChartJsURL
 	}
-	if cfg.Title == ConfigDefault.Title && cfg.Refresh == ConfigDefault.Refresh {
-		cfg.index = ConfigDefault.index
-	} else {
-		if cfg.Refresh < minRefresh {
-			cfg.Refresh = minRefresh
-		}
-		// update cfg.index with custom title/refresh
-		cfg.index = newIndex(viewBag{cfg.Title,
-			cfg.Refresh, cfg.FontURL, cfg.ChartJsURL, cfg.CustomHead})
+	if cfg.Refresh < minRefresh {
+		cfg.Refresh = minRefresh
 	}
 
 	if cfg.Next == nil {
@@ -113,6 +106,14 @@ func configDefault(config ...Config) Config {
 
 	if !cfg.APIOnly {
 		cfg.APIOnly = ConfigDefault.APIOnly
+	}
+
+	if cfg.Title == ConfigDefault.Title && cfg.Refresh == ConfigDefault.Refresh {
+		cfg.index = ConfigDefault.index
+	} else {
+		// update cfg.index with custom title/refresh
+		cfg.index = newIndex(viewBag{cfg.Title,
+			cfg.Refresh, cfg.FontURL, cfg.ChartJsURL, cfg.CustomHead})
 	}
 
 	return cfg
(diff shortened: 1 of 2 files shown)


efectn (commented):
Can you add some test cases

ReneWerner87:
> Can you add some test cases

would be nice 
then we probably would have noticed it in the beginning and no one can break it without us noticing it

middleware/monitor/config.go
```diff
@@ -115,5 +108,13 @@ func configDefault(config ...Config) Config {
 		cfg.APIOnly = ConfigDefault.APIOnly
 	}
 
+	if cfg.Title == ConfigDefault.Title && cfg.Refresh == ConfigDefault.Refresh {
+		cfg.index = ConfigDefault.index
+	} else {
+		// update cfg.index with custom title/refresh
+		cfg.index = newIndex(viewBag{cfg.Title,
+			cfg.Refresh, cfg.FontURL, cfg.ChartJsURL, cfg.CustomHead})
```
gaby, line 116: These changes will not get applied if the `IF` above is True.

gaby (changes requested):

gaby (changes requested):
These changes are not fixing the problem. The else statement never runs if the default `Title` and `Refresh` are used which is the original issue

wangjq4214:
> These changes are not fixing the problem. The else statement never runs if the default `Title` and `Refresh` are used which is the original issue

To be honest I don't really understand what the point of this judgment is, it seems to me to be unnecessary.

gaby:
> > These changes are not fixing the problem. The else statement never runs if the default `Title` and `Refresh` are used which is the original issue
> 
> To be honest I don't really understand what the point of this judgment is, it seems to me to be unnecessary.

No judging, just wanted to point the issue with the URL's. The original commit fixed the issue for the "refresh" field. Looks better now! 👍🏻

wangjq4214:
@ReneWerner87 @efectn @gaby Please check out the new changes when you have time.

gaby (approved):
Looks good to me! Thanks! 🚀

ReneWerner87 (approved):

PR #2028, opened 2022-08-19
🧹 remove deprecated `io/ioutil` package

jfcg:


efectn:
Fiber supports 1.14-1.19. If we change them, Fiber wont work on 1.14. However, they were changed in v3.

efectn:
> 1.14 seems quite old, is there any rationale for supporting it on v2 ?

I think not but it would be backward-compatibility problem for 1.14 users. @ReneWerner87 what do you think

ReneWerner87:
yes we should keep our promise on the start page and stay downward compatible with version 2

image

jfcg:
There have been 11 security-related [releases](https://groups.google.com/g/golang-announce/) to Go 1.16+ since the release of Go 1.15.15. It is probably horribly dangerous to use anything below latest Go 1.16 as the Golang project provides updates only to the lastest two versions, which are at the moment 1.18 and 1.19.

PR #2033, opened 2022-08-19
:bug: bug: fix route constraints problems

efectn:
**Bugs & Improvements:**
- `strings.Split` doesn't take care of escape char. It's potaintial bug for regex, datetime constraints.
- Constraints weren't parsed properly when `CaseSensitive` is false.
- I converted `exactLen` to `len`. It's shorter and better.

I couldn't see another problem. If you see another problems, you can share them to fix before releasing stable `v2.37.0` version.


ReneWerner87 (approved):

ReneWerner87:
- [ ] there is one more task which efectn want to fix

Issue #2025, opened 2022-08-19
🤗 [Problem]: tls.Config in ListenTLS


wireguard-dev:
### Question Description

Can do you, please, set NextProto: []string{"h2"} in tls.Config of ListenTLS?
Without that I can't use dgrr/http2 :( 

Fiber version is v2.37.0-rc.1

### Code Snippet (optional)

```go
package main

import "github.com/gofiber/fiber/v2"
import "log"

func main() {
  app := fiber.New()

  http2.ConfigureServer(app.Server(), http2.ServerConfig{})

  log.Fatal(app.ListenTLS(":3000", ...))
}
```

### Checklist:

- [X] I agree to follow Fiber's [Code of Conduct](https://github.com/gofiber/fiber/blob/master/.github/CODE_OF_CONDUCT.md).
- [X] I have checked for existing issues that describe my questions prior to opening this one.
- [X] I understand that improperly formatted questions may be closed without explanation.

efectn:
I think we don't need it because of Fiber doesn't have default HTTP/2 support. You can use Listener to pass custom TLS config.

wireguard-dev:
> I think we don't need it because of Fiber doesn't have default HTTP/2 support. You can use Listener to pass custom TLS config.

In my opinion, this might kill the ClientHelloInfo feature, because I don't have access to the non-exported fields in fiber.

efectn:
> Author

Right. We may need to get TLS from https://github.com/gofiber/fiber/blob/master/listen.go#L31 and pass it to CHI to resolve this.

PR #2034, opened 2022-08-19
:bug: bug: make tlsHandler public to use it with Listener

efectn:
should fix https://github.com/gofiber/fiber/issues/2025 but not tested yet.

### DIFF
--- a/listen.go
+++ b/listen.go
@@ -26,15 +26,9 @@ import (
 
 // Listener can be used to pass a custom listener.
 func (app *App) Listener(ln net.Listener) error {
-	// ClientHelloInfo support
-	tlsHandler := &tlsHandler{}
-	addr, tlsConfig := lnMetadata(app.config.Network, ln)
-	if tlsConfig != nil {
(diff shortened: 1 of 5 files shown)


efectn:
@wireguard-dev can you test it

listen.go
```diff
@@ -183,7 +199,9 @@ func (app *App) ListenMutualTLS(addr, certFile, keyFile, clientCertFile string)
 	}
 
 	// Attach the tlsHandler to the config
+	app.mutex.Lock()
```
ReneWerner87, line 202: just use the method app.SetTLSHandler ? instead of repeating ourselves

listen.go
```diff
@@ -183,7 +199,9 @@ func (app *App) ListenMutualTLS(addr, certFile, keyFile, clientCertFile string)
...
 	app.tlsHandler = tlsHandler
+	app.mutex.Unlock()
```
ReneWerner87, line 204:
```suggestion
	app.SetTLSHandler(tlsHandler)
```

ReneWerner87 (approved):

PR #2036, opened 2022-08-20
v3(tests): use testify for assertion

### LINKED ISSUE #2035: 🚀 v3 Request: use https://github.com/stretchr/testify for testing
### Feature Description

remove current `utils.AssertEqual` and use https://github.com/stretchr/testify for testing assertion

_No response_

### Code Snippet (optional)

_No response_

### Checklist:

Trim21:
close #2035

replace all `utils.AssertEqual` with `testify/require`


Trim21:
I'm not very familiar with our PR merging workflow, can I just merge it or I should wait another maintainer to do it 🤔

ReneWerner87:
Just wait for the maintainers. You can merge little things like dependencies updates, documentation or test update, but should wait for big decisions.

ReneWerner87:
And best never merge your own code without feedback from others, otherwise no one can give feedback.

efectn:
Testify is OK to me since it has more detailed output and better methods.

ReneWerner87:
Just don't know if we should do that
The idea of fenny then, was to have as little dependence as possible

Hope in version 3 we can follow this idea as good as possible, so if it is not absolutely necessary, we should not make ourselves dependent

efectn:
I think v3 has less deps than v2 at the moment. We can also write custom extractor instead of using `msgp` i suppose.

https://github.com/gofiber/fiber/blob/v3-beta/go.mod

Trim21:
I want say that vendoring go package in `internal/` is not a good idea .....

ReneWerner87:
Yes I agree, that's why I have already allowed some dependencies

My only question is if we got along with our native test functions so far, do we really need an extra package?
Should only be questioned so we have as few dependencies as possible

Trim21:
that dependent on why we need to "have as little dependence as possible" I think 🤔

To make user download less code or something else?

ReneWerner87:
One dependency quickly becomes several, when the next developer wants to write his test with other libs we suddenly have a lot of dependencies.

But if we don't allow these from the beginning and extend our internal test function a bit, we would be free from these decisions.

I'll look at the code later which assertion is not possible with our internal function

Trim21:
> when the next developer wants to write his test with other libs we suddenly have a lot of dependencies.

I think we can have some code style guide about this...

PR #2047, opened 2022-08-26
🔥Feature: add Match() function

KiddoV:
## Description
Match() function will help users to choose between adding one or more HTTP methods instead of just, currently, one or all method(s).

## Type of change

### DIFF
--- a/app.go
+++ b/app.go
@@ -723,6 +723,18 @@ func (app *App) Patch(path string, handlers ...Handler) Router {
 	return app.Add(MethodPatch, path, handlers...)
 }
 
+// Match registers one or more HTTP methods to register a route
+func (app *App) Match(methods []string, path string, handlers ...Handler) Router {
+	for _, method := range methods {
+		for _, m := range intMethod {
+			if method == m {
+				_ = app.Add(method, path, handlers...)
+			}
+		}
+	}
+	return app
+}
+
 // Add allows you to specify a HTTP method to register a route
 func (app *App) Add(method, path string, handlers ...Handler) Router {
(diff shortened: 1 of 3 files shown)


ReneWerner87:
thanks for the work
but i would not include this helper method in the core, because this is not necessary due to the method chaining and there are hardly any use cases for it

alternatively you can use the app.All method, if the handler should be registered everywhere

KiddoV:
Correct me if I'm wrong, but I think this is necessary since in some cases, users only need to deal with, for example, 2 methods, say "GET" and "POST". So instead of bringing all with ``app.All()`` (could waste some memory for unnecessary methods), they have more option to choose.
Also, we can filter out the wrong method in the handler if the client try to request, for example, with "PUT" or something else.

Thanks,

ReneWerner87:
yes right, but that is easily possible with 
```go
app.Get(path, handler).Post(path, handler)
```

ReneWerner87:
we just have to make sure that the core is not too bloated with methods that are not for the majority, so i would decide against this helper method

in v3 you have the possibility to extend the context with your own methods so we can easily extend helper methods and have them in separate repositories

efectn:
```go
app.route('/events')
  .all(function (req, res, next) {
    // runs for all HTTP verbs first
    // think of it as route specific middleware!
  })
  .get(function (req, res, next) {
    res.json({})
  })
  .post(function (req, res, next) {
    // maybe add a new event...
  })
```

Express has a method like this. We were planning to add something like this for v3

KiddoV:
Sounds good. Thanks for the clarification!

Issue #2053, opened 2022-08-26
🚀 [Feature]: adding Environment Variable expose middleware


mstrYoda:
### Feature Description

It is a common requirement while debugging/viewing environmental informations of our apps, so it would be nice to have an env var expose middleware like expvar.

### Code Snippet (optional)

```go
package main

import "github.com/gofiber/fiber/v2"
import "github.com/gofiber/fiber/v2/middleware/envvar"
import "log"

func main() {
  app := fiber.New()
  app.Use(envvar.New())

  // An example to describe the feature

  log.Fatal(app.Listen(":3000"))
}
```

### Checklist:

- [X] I agree to follow Fiber's [Code of Conduct](https://github.com/gofiber/fiber/blob/master/.github/CODE_OF_CONDUCT.md).
- [X] I have checked for existing issues that describe my suggestion prior to opening this one.
- [X] I understand that improperly formatted feature requests may be closed without explanation.

PR #2054, opened 2022-08-26
🚀 [Feature]: add envvar expose middleware

mstrYoda:
## Description

This pr adds environment variable expose middleware.

Fixes https://github.com/gofiber/fiber/issues/2053

## Type of change

## Checklist:


efectn:
Please update awesome-fiber and https://github.com/gofiber/fiber#-internal-middleware

middleware/envvar/envvar.go
```diff
@@ -0,0 +1,85 @@
...
+
+	return func(c *fiber.Ctx) error {
+		path := c.Path()
+		if len(path) < len(cfg.Path) || !strings.HasPrefix(path, cfg.Path) {
+			return c.Next()
+		}
+
+		if path == cfg.Path {
+			envVar := newEnvVar(cfg)
+			varsByte, err := json.Marshal(envVar)
```
efectn, line 49: Get this from app config instead of using hardcoded stdlib

efectn (changes requested):

middleware/envvar/envvar.go
mstrYoda: Thx, done 👍

mstrYoda:
> Please update awesome-fiber and https://github.com/gofiber/fiber#-internal-middleware

All done 👍

efectn (changes requested):
Dont forget adding middleware section to other READMEs

efectn:
Check actions please. Lgtm

mstrYoda:
> Check actions please. Lgtm

I changed `t.Setenv` to `os.Setenv` to provide backward compatibility with older Go versions.

ReneWerner87:
please check again the methods of envvar (some of them are not use and lowercase -> so not exposed )

can you please also provide a better example and link where such functionality is used in other languages or frameworks so i can better understand the use of middleware

mstrYoda:
> please check again the methods of envvar (some of them are not use and lowercase -> so not exposed )
> 
> can you please also provide a better example and link where such functionality is used in other languages or frameworks so i can better understand the use of middleware

Well, I quickly searched but could not find an exact example of this. People are trying to reimplement their middleware/components for this work. We always wonder about our env vars that contain feature flags too in our APIs.
This is why I thought that it would be good to have built-in one.

middleware/envvar/envvar.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/gofiber/fiber/v2"
+	"github.com/gofiber/fiber/v2/utils"
+	"os"
+	"strings"
+)
+
+// Config defines the config for middleware.
+type Config struct {
+	// Path specifies the path of handler that exposes environment variables
+	Path string
```
ReneWerner87, line 13: specifying the path is not necessary in my opinion
you can say outside where the middleware should run

```go
app.Use("/expose/envvars", envvar.New(envvar.Config{
    ...
}))
```

mstrYoda: So we should not support app.Use(envvars.New()) approach? User must provide path?

efectn: Like monitor middleware

efectn (approved):

ReneWerner87 (approved):

Issue #1829, opened 2022-03-21
🚀 v3 Request: New ideas for routing and mounting


balcieren:
New Router instead of `Group` , `Route` and `Mount` (should be deprecate)

[Express Router](https://expressjs.com/en/5x/api.html#express.router)

### Routing
```go
func UserRouter() *fiber.Router {
	router := fiber.NewRouter()

	router.Get("/", func(c *fiber.Ctx) error {
		return c.JSON("fiber v3")
	})

	return router
}

func main() {
	app := fiber.New()

	app.Use("/user", UserRouter)

	app.Listen(":3000")
}
```
### RouterConfig can contain 
![router-config](https://user-images.githubusercontent.com/46889813/159263046-53081a3d-30cc-4298-8123-0c16ad8369db.jpg)

---

### Mounting (https://github.com/gofiber/fiber/pull/2022)
[Express Mount](http://expressjs.com/en/5x/api.html#app.onmount)
```go
func main() {
	app := fiber.New()

	user := fiber.New()

	user.Get("/", func(c *fiber.Ctx) error {
		return c.JSON("fiber-v3")
	})

	user.On("mount", func(parent *fiber.App) {
		fmt.Println("user mounted")
	})

	app.Use("/user", user)

	app.Listen(":3000")
}
```
---
### fiber.Route() (#2065)
[Express Route](http://expressjs.com/en/5x/api.html#app.route)
```go
func main() {
	app := fiber.New()

	app.Route("/events").
		All(func(c *fiber.Ctx) error {
			...
		}).
		Get(func(c *fiber.Ctx) error {
			...
		}).
		Post(func(c *fiber.Ctx) error {
			...
		})

	app.Listen(":3000")
}
```

vishjain:
@ReneWerner87 Can I work on this feature?

efectn:
Hi @vishjain. Are there any progress?

vishjain:
Working on it right now. Thanks!

m13t:
I'm quite keen to see if this would support my use case as I was surprised it doesn't work in v2. I'm essentially dynamically constructing an additional router in a piece of middleware and then handing off to that router. This is something I've done similar in express for complex dynamic multi-tenanted apps but short of creating it all in a `fiber.Handler` and branching that in the middleware I'm not sure how else to do it. Would be nice to see this in v3.

[...]

PR #2065, opened 2022-08-30
✨ v3 (feature): new Route method

wangjq4214:
related to #1829


wangjq4214:
> @wangjq4214 Why you changed `Route` with `RouteInfo` ?

I think the method returned by this function is probably more commonly used, so a simpler name is needed. Changing the previous name is not quite right, I will change it back, do you have any idea about the name of this type?

register.go
```diff
@@ -0,0 +1,106 @@
...
+// 🤖 Github Repository: https://github.com/gofiber/fiber
+// 📌 API Documentation: https://docs.gofiber.io
+
+package fiber
+
+// Register struct
+type Register struct {
+	app *App
+
+	Path string
```
efectn, line 11: This can be private i suppose.

register.go
```diff
@@ -0,0 +1,106 @@
...
+//	     return c.Next()
+//	})
+//	app.Use("/api", func(c fiber.Ctx) error {
+//	     return c.Next()
+//	})
+//	app.Use("/api", handler, func(c fiber.Ctx) error {
+//	     return c.Next()
+//	})
+//
+// This method will match all HTTP verbs: GET, POST, PUT, HEAD etc...
```
efectn, line 27: Needs to update comments

efectn (changes requested):

efectn:
Router interface is not compatible for this stuff. It has path param in methods

balcieren:
> Router interface is not compatible for this stuff. It has path param in methods

That's true. You are right.

balcieren:
I think we should change Route struct name

efectn (approved):

balcieren:
Actually `RouteFunc` might be a good name to seperate with `Route`

ReneWerner87:
@wangjq4214 thx for the PR

can you create some test with route which have a deeper level

so for example 3 level 

so we can be sure that there can be scaled infinitely

wangjq4214:
@efectn thx

register.go
andrijana44: const chunks = [];

readable.on('readable', () => {
  let chunk;
  while (null !== (chunk = readable.read())) {
    chunks.push(chunk);
  }
});

readable.on('end', () => {
  const content = chunks.join('');
});

register.go
andrijana44: const chunks = [];

readable.on('readable', () => {
  let chunk;
  while (null !== (chunk = readable.read())) {
    chunks.push(chunk);
  }
});

readable.on('end', () => {
  const content = chunks.join('');
});

PR #2071, opened 2022-09-02
fix unhandled errors in app_test.go

Kamandlou:


efectn:
I don't think we need to handle errors for `Send()`. It always returns `nil`. https://github.com/gofiber/fiber/blob/master/ctx.go#L1437

Kamandlou:
> I don't think we need to handle errors for `Send()`. It always returns `nil`. https://github.com/gofiber/fiber/blob/master/ctx.go#L1437

Yes, it is not needed now, but there may be changes in this method that are needed in the future

PR #2073, opened 2022-09-04
fix: remove deprecated package

dozheiny:
In Go 1.19, ioutil is a deprecated package, so I changed it into io.ReadAll.


efectn:
Fiber supports go 1.14-1.20. so we cant remove ioutil package. But https://github.com/gofiber/fiber/tree/v3-beta uses io and os instead of ioutil

PR #2081, opened 2022-09-09
:broom: update: add methods configuration for cache middleware

efectn:
## Description
We may need to cache some methods according to our needs. I think we should be able to set methods to cache.

## Type of change

## Checklist:


ReneWerner87 (approved):

ReneWerner87:
thanks @efectn 
LGTM

Issue #2087, opened 2022-09-14
🚀 [Feature]: Shutdown() add optional timeout


pjebs:
### Feature Description

Currently when I call `Shutdown()`, it seems to wait a long time (many minutes).

It would be good for it to pass in an optional **ctx** or a **max wait time**. If it hasn't gracefully shutdown by the max wait time, it hard kills the server connections etc.

### Checklist:

- [X] I agree to follow Fiber's [Code of Conduct](https://github.com/gofiber/fiber/blob/master/.github/CODE_OF_CONDUCT.md).
- [X] I have checked for existing issues that describe my suggestion prior to opening this one.
- [X] I understand that improperly formatted feature requests may be closed without explanation.

efectn:
As different as Shutdown(), we can add ShutdownWithContext(ctx) that checks ctx before shutdowning

efectn:
> As different as Shutdown(), we can add ShutdownWithContext(ctx) that checks ctx before graceful shutdowning

pjebs:
@li-jin-gou https://github.com/valyala/fasthttp/issues/1389#issuecomment-1267351046

efectn:
You can access fasthttp.Server with https://github.com/gofiber/fiber/blob/master/app.go#L852. But it's also OK to me add Close() method to Fiber.

PR #2092, opened 2022-09-15
:sparkles: Feature: add ShutdownWithContext

li-jin-gou:
## Description

add ShutdownWithContext

## Type of change

## Checklist:

## Commit formatting:

### DIFF
--- a/app.go
+++ b/app.go
@@ -10,6 +10,7 @@ package fiber
 import (
 	"bufio"
 	"bytes"
+	"context"
 	"errors"
 	"fmt"
 	"net"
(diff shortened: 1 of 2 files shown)


ReneWerner87:
thanks for the feature
please remember the checkpoints and remove everything that does not apply and check off everything already done

li-jin-gou:
> thanks for the feature please remember the checkpoints and remove everything that does not apply and check off everything already done

OK，thanks

pjebs:
This doesn't address #2087 at all. Perhaps my feature request was misunderstood.

li-jin-gou:
> This doesn't address #2087 at all. Perhaps my feature request was misunderstood.
> 
> It needs to force the shutdown.

sorry,

> This doesn't address #2087 at all. Perhaps my feature request was misunderstood.
> 
> It needs to force the shutdown.

Sorry, maybe I misunderstood it

efectn:
I think **context with timeout** would be good idea for force shutdown. Fasthttp waits closing of connections. So this should be the best solution.

pjebs:
There needs to be a way force the shutdown (ungracefully), if it can't be shutdown gracefully within a certain time limit.

efectn:
> There needs to be a way force the shutdown (ungracefully), if it can't be shutdown gracefully within a certain time limit.

Why dont you use `os.Exit()`

li-jin-gou:
Is this pr still needed? :roll_eyes:

pjebs:
No

PR #2107, opened 2022-09-20
Feat: Register custom methods

### LINKED ISSUE #2096: 🐛 [Bug]: panic: add: invalid http method LOAD
### Bug Description

I want to create a handler with a custom http method called "LOAD".

```
app.Add("LOAD", "/hello", func(c *fiber.Ctx) error {
	return c.SendString("Hello, World 👋!")
})
```

It crashes the application:

`panic: add: invalid http method LOAD`

### How to Reproduce

app.Add("LOAD", "/hello", func(c *fiber.Ctx) error {
	return c.SendString("Hello, World 👋!")
})

### Expected Behavior

It should allow for custom http methods

### Fiber Version

 Fiber v2.37.1

### Code Snippet (optional)

_No response_

### Checklist:

rafimuhammad01:
## Description
This PR is to support the custom method feature. I updated register function so that function won't be giving validation if the method is unsupported yet, so user can add any method that they want to. I also make changes to methodInt function to return the index instead of using hardcoded with switch-case statement.

Fixes #2096
## Type of change

## Checklist:

- [X] For new code I have written benchmarks so that they can be analyzed and improved

## Commit formatting:


efectn:
There should be 2 ways to allow custom methods:
- Allow adding custom methods and validate them when to add new routes.
- Allow unknown methodds without any custom method registration.

Maybe the first one would be overdesign. Current implementation seems OK but i'll review it then. What do you think @ReneWerner87

ReneWerner87:
sorry very busy week, will answer tomorrow

ReneWerner87 (approved):

ReneWerner87:
current implementation seems to be okay

- [ ] please also update the documentation repository

feature will be in the next release

helpers.go
```diff
@@ -332,28 +332,13 @@ var getBytesImmutable = func(s string) (b []byte) {
...
-		return 5
-	case MethodOptions:
-		return 6
-	case MethodTrace:
-		return 7
-	case MethodPatch:
-		return 8
-	default:
-		return -1
+	for i, v := range intMethod {
```
ReneWerner87, line 335: did you check the performance, we had added the switch at that time because the access is faster there than with the map
can you make a comparison here

we should keep the functions which are used for the main path of the application as fast as possible

router.go
```diff
@@ -216,7 +216,8 @@ func (app *App) register(method, pathRaw string, handlers ...Handler) Router {
 	method = utils.ToUpper(method)
 	// Check if the HTTP method is valid unless it's USE
 	if method != methodUse && methodInt(method) == -1 {
-		panic(fmt.Sprintf("add: invalid http method %s\n", method))
+		intMethod = append(intMethod, method)
+		app.updateStack()
```
ReneWerner87, line 220: shouldn't we rather call it "increase routing stack", so that you know what is done there?

ReneWerner87:
@rafimuhammad01 pls check the last comments

router.go
rafimuhammad01: I agree with this, will change the function name so it will be more representative

helpers.go
rafimuhammad01: not quite sure about the performance, but I think I need this implementation for the custom methods. If we using switch case statement it will only work for static methods but for this feature it will be dynamic.

ReneWerner87: Right, maybe we should switch back to a map as well

can you test if the slice-for or the map access is faster by creating a benchmark and posting metrics

rafimuhammad01: this is the benchmark for forloop implementation and switch-case statement:

- For Loop

```
goos: darwin
goarch: amd64
cpu: Intel(R) Core(TM) i5-5350U CPU @ 1.80GHz
Benchmark_MethodInt
Benchmark_MethodInt-4   	55310788	        20.03 ns/op
PASS
```

- Switch-Case Statement

```
goos: darwin
goarch: amd64
cpu: Intel(R) Core(TM) i5-5350U CPU @ 1.80GHz
Benchmark_MethodInt
Benchmark_MethodInt-4   	1000000000	         0.3897 ns/op
PASS
```

Switch case is far way better, I think we can keep the switch case and handle the dynamic route on default statement, it will be like this:
```go
func methodInt(s string) int {
	switch s {
	case MethodGet:
		return 0
	case MethodHead:
		return 1
	case MethodPost:
		return 2
	case MethodPut:
		return 3
	case MethodDelete:
		return 4
	case MethodConnect:
		return 5
	case MethodOptions:
		return 6
	case MethodTrace:
		return 7
	case MethodPatch:
		return 8
	default:
		for i, v := range intMethod[8:] {
			if s == v {
				return i
			}
		}
		return -1
	}
}
```

I did the benchmark for this and this is the result:
- for method that hardcoded:
```
goos: darwin
goarch: amd64
cpu: Intel(R) Core(TM) i5-5350U CPU @ 1.80GHz
Benchmark_MethodInt
Benchmark_MethodInt-4   	1000000000	         0.3713 ns/op
PASS
```
- for dynamic method
```
goos: darwin
goarch: amd64
cpu: Intel(R) Core(TM) i5-5350U CPU @ 1.80GHz
Benchmark_MethodInt
Benchmark_MethodInt-4   	244620964	         4.902 ns/op
PASS
```

with this approach, the performance of the hardcoded method will be the same and the performance of dynamic method will be better from the first approach although it wont be better than the hardcoded method. What do you think?

rafimuhammad01:
@ReneWerner87 please check the last comment

helpers.go
ReneWerner87: this is okay for me

ReneWerner87:
will do this today

app.go
```diff
@@ -555,6 +555,11 @@ func New(config ...Config) *App {
 	return app
 }
 
+func (app *App) updateStack() {
```
ReneWerner87, line 558: we can rename the method and make it more useful by adding the extension of the intMethod list here as well

perhaps rename it to 
addCustomRequestMethod
and the parameter is the name of the new request method, after that the other 2 lists will be expanded

ReneWerner87 (changes requested):
after the last adjustments i will merge it

app_test.go
```diff
@@ -1647,3 +1659,11 @@ func Test_App_SetTLSHandler(t *testing.T) {
 
 	utils.AssertEqual(t, "example.golang", c.ClientHelloInfo().ServerName)
 }
+
+func Test_App_UpdateStack(t *testing.T) {
```
ReneWerner87, line 1663: can you rename this method and add a check for the intMethod values

ReneWerner87:
> current implementation seems to be okay
> 
> * [ ]  please also update the documentation repository
> 
> feature will be in the next release

don´t forget this @rafimuhammad01 
we need somewhere a hint in the routing section that we support custom request methods

app_test.go
ReneWerner87: fixed

ReneWerner87 (approved):

ReneWerner87 (changes requested):
through the test run i found out that there is one more problem

because if you add a custom method and run the USE or ALL before, the new method will not be considered there, because it is not inital in the list of those to be considered

should we maybe not approach the feature so dynamically and add the methods on-the-fly and rather maintain a slice inside the fiber.config with the allowed methods ?

so the intMethod slice as config property with the 9 default methods? 
then you could also reduce or extend the default if you don't want TRACE or CONNECT to work or LOAD to be added

but then you would have to consider the mounting, because these methods are set per fiber app and would have to be merged then

@efectn @rafimuhammad01 what do you think ?

efectn:
> through the test run i found out that there is one more problem https://github.com/gofiber/fiber/actions/runs/3198073219/jobs/5222134276 image
> 
> because if you add a custom method and run the USE or ALL before, the new method will not be considered there, because it is not inital in the list of those to be considered
> 
> should we maybe not approach the feature so dynamically and add the methods on-the-fly and rather maintain a slice inside the fiber.config with the allowed methods ?
> 
> so the intMethod slice as config property with the 9 default methods? then you could also reduce or extend the default if you don't want TRACE or CONNECT to work or LOAD to be added
> 
> but then you would have to consider the mounting, because these methods are set per fiber app and would have to be merged then
> 
> @efectn @rafimuhammad01 what do you think ?

Sounds good idea

ReneWerner87:
@rafimuhammad01 okay for you ? should i do this or you want todo this ?

rafimuhammad01:
Hi @ReneWerner87 sorry for the late reply, I've been busy this week. I still didn't understand about the other approach we will implement, could you please elaborate more?

ReneWerner87:
the problem
```go
func main() {
	app := fiber.New(fiber.Config{})

	app.Use(func(ctx *fiber.Ctx) error { // is always called only not at LOAD
		// doing something
		return ctx.Next()
	})
	app.Add("LOAD", "/test", func(ctx *fiber.Ctx) error {
		return ctx.SendString("hello")
	})

}

```


the solution
```go
func main() {
		RequestMethods: []string{ // default is the normal list
                      fiber.MethodGet,
                      fiber.MethodHead,
                      fiber.MethodPost,
                      fiber.MethodPut,
                      fiber.MethodDelete,
                      fiber.MethodConnect,
                      fiber.MethodOptions,
                      fiber.MethodTrace,
                      fiber.MethodPatch,
                      "LOAD",
              },
        // with this approach we can reduce the list and only allow what we want
	})

	app.Use(func(ctx *fiber.Ctx) error { // is always called
		// doing something
		return ctx.Next()
	})
	app.Add("LOAD", "/test", func(ctx *fiber.Ctx) error {
		return ctx.SendString("hello")
	})

}
```

rafimuhammad01:
I'm sorry for the late response, have been super busy right now. I see the problem and will try to fix this issue. Thank you!

ReneWerner87 (approved):

ReneWerner87:
- [ ] don´t forget to update the documentation repository

Issue #2116, opened 2022-09-23
🚀 [Feature]: Add ability to configure client for proxy.Do


Vovcharaa:
### Feature Description

Add ability to provide custom client configuration to client that used inside proxy.Do, or make client from proxy package public.

### Additional Context (optional)

Right now i faced issue where i was out of free connections (ErrNoFreeConns) and couldn`t provide custom number of connection per host or wait for connection timeout and was forced to copy-paste proxy implementation to my code to provide custom client settings.

### Checklist:

- [X] I agree to follow Fiber's [Code of Conduct](https://github.com/gofiber/fiber/blob/master/.github/CODE_OF_CONDUCT.md).
- [X] I have checked for existing issues that describe my suggestion prior to opening this one.
- [X] I understand that improperly formatted feature requests may be closed without explanation.

li-jin-gou:
Referring to the proxy usage scenario of the `Hertz` , I found it useful to set the client.

PR #2117, opened 2022-09-24
:sparkles: feat:  support to set client in proxy mw

li-jin-gou:
## Description

Fixes https://github.com/gofiber/fiber/issues/2116
- In some scenarios the user need set up some complex client config via custom client
- The case of a global client only gives the user no way to set up special configurations for only one middleware

so this pr allow to use custom client and  distinguish global client and custom client in proxy mw and compatible with previous logic.

## Type of change

## Checklist:

- [x]  update of the docs repository

## Commit formatting:

### DIFF
--- a/middleware/proxy/proxy.go
+++ b/middleware/proxy/proxy.go
@@ -24,34 +24,39 @@ func Balancer(config Config) fiber.Handler {
 	cfg := configDefault(config)
 
 	// Load balanced client
-	var lbc fasthttp.LBClient
-	// Set timeout
-	lbc.Timeout = cfg.Timeout
+	var lbc = &fasthttp.LBClient{}
+	// Note that Servers, Timeout, WriteBufferSize, ReadBufferSize and TlsConfig
+	// will not be used if the client are set.
+	if config.Client == nil {
+		// Set timeout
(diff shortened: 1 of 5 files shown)


ReneWerner87:
please don´t forget 

- [ ] the update of the docs repository

li-jin-gou:
> please don´t forget
> 
> * [ ]  the update of the docs repository

Thanks for reminding. :sob:

middleware/proxy/proxy.go
```diff
@@ -24,34 +24,37 @@ func Balancer(config Config) fiber.Handler {
...
-	for _, server := range cfg.Servers {
-		if !strings.HasPrefix(server, "http") {
-			server = "http://" + server
-		}
+	var lbc = &fasthttp.LBClient{}
+	if config.Client == nil {
+		// Set timeout
+		lbc.Timeout = cfg.Timeout
+		// Scheme must be provided, falls back to http
+		for _, server := range cfg.Servers {
```
ReneWerner87, line 32: we should make clear in the readme or readme configuration that if the clients are set, then the server addresses and read and write buffer settings are not processed

li-jin-gou: done

Issue #2066, opened 2022-08-30
🐛 [Bug]: Mounting does not work if you use it before declaring the route


balcieren:
### Bug Description

I found critical bugs about mount. 

1.  If app's router codes (Get, Post, ...) write after `app.Use()`, app wouldn't register route stack.
```go
func main() {
	app := fiber.New()
	sub := fiber.New()

	app.Mount("/sub", sub)

	sub.Get("/", func(c *fiber.Ctx) error {
		return c.JSON("this is sub app")
	})

	app.Listen(":3000")
}
```
2. Sub in sub mounting does not work
```go
func main() {
	app := fiber.New()
	sub := fiber.New()
	sub2 := fiber.New()
	sub2.Get("/", func(c *fiber.Ctx) error {
		return c.JSON("this is sub2 app")
	})

	app.Mount("/sub", sub)

	sub.Mount("/sub2", sub2)

	app.Listen(":3000")
}
```

### Expected Behavior

Application should register sub app's stack in `app.Listen()`  (`startupProcess()`)

need good refactor and fix bugs

### Fiber Version

v2.37.0

### Checklist:

- [X] I agree to follow Fiber's [Code of Conduct](https://github.com/gofiber/fiber/blob/master/.github/CODE_OF_CONDUCT.md).
- [X] I have checked for existing issues that describe my problem prior to opening this one.
- [X] I understand that improperly formatted bug reports may be closed without explanation.

ReneWerner87:
```go
func main() {
	app := fiber.New()
	sub := fiber.New()

	sub.Get("/", func(c *fiber.Ctx) error {
		return c.JSON("this is sub app")
	})
	app.Mount("/sub", sub) // <- must be at the end

	log.Fatal(app.Listen(":3000"))
}
```
the current implementation takes the route at the time of the mount command and transfers it to the parent app

just as the order of routes is important in the declaration of routes, it is also important here

probably we could also change this and only pull the routes from the sub app in the startup process
(but would remove the "critical" part of the bug, as it is a known intentional behavior)

```go
func main() {
	app := fiber.New()
	sub := fiber.New()
	sub2 := fiber.New()
	sub2.Get("/", func(c *fiber.Ctx) error {
		return c.JSON("this is sub2 app")
	})

	sub.Mount("/sub2", sub2) // <- first mount sub2 in sub
	app.Mount("/sub", sub) // <- then mount sub in app

	log.Fatal(app.Listen(":3000"))
}
```
[...]

PR #2120, opened 2022-09-25
:bug: bug: improve mounting behavior 

efectn:
## Description
Fixes https://github.com/gofiber/fiber/issues/2066 and adds some features to mounting feature :rocket: 

- Add OnMount hook that is fired when sub-app is mounted on a parent app. 
- Add MountPath method to app. It returns mount prefix of sub-app. If app not mounteed, it returns empty string.
- Don't execute onRoute/OnGroup/OnName/OnGroupName hooks when to add routes of sub-apps to parent app. Return pathes with prefix for sub-app routes instead.
- Don't add routes in Mount. Add them in startupProcess. We're now able to Mount app before adding routes (supoorting nested mounting).
- Fix views when both of parent app and sub-app have different view engines

## Type of change

## Checklist:

## Commit formatting:


router.go
```diff
@@ -436,19 +442,23 @@ func (app *App) addRoute(method string, route *Route) {
...
 
-	app.mutex.Lock()
-	app.latestRoute = route
-	if err := app.hooks.executeOnRouteHooks(*route); err != nil {
-		panic(err)
+	// Execute onRoute hooks & change latestRoute if not adding mounted route
+	if !mounted {
+		app.mutex.Lock()
+		app.latestRoute = route
+		if err := app.hooks.executeOnRouteHooks(*route); err != nil {
```
efectn, line 449: Not much sure about here. Maybe we should add mounted prefix to route path. otherwise route hooks won't work properly.

app.go
```diff
@@ -1024,7 +1033,66 @@ func (app *App) startupProcess() *App {
...
+
+		if len(parent) > 0 {
+			prefix = parent[0] + prefix
+		}
+
+		if _, ok := app.appList[prefix]; !ok {
+			app.appList[prefix] = subApp
+		}
+
+		if len(subApp.appList) > 1 {
```
ReneWerner87, line 1068: please add a comment here "why it should be greater than 1"

i know its because of the real app -> but the next dev should also know this

app.go
```diff
@@ -1024,7 +1033,66 @@ func (app *App) startupProcess() *App {
 	}
 
 	app.mutex.Lock()
+	defer app.mutex.Unlock()
+
+	// add routes of sub-apps
+	if !app.subAppsRoutesAdded {
+		app.appendSubAppLists(app.appList)
+		app.addSubAppsRoutes(app.appList)
```
ReneWerner87, line 1041: who adds the stack when no name is added to a route? i miss this in a central location

app.go
```diff
@@ -584,30 +590,33 @@ func (app *App) SetTLSHandler(tlsHandler *TLSHandler) {
...
-		for r := range stack[m] {
-			route := app.copyRoute(stack[m][r])
-			app.addRoute(route.Method, app.addPrefixToRoute(prefix, route))
-		}
-	}
-
 	// Support for configs of mounted-apps and sub-mounted-apps
 	for mountedPrefixes, subApp := range fiber.appList {
 		app.appList[prefix+mountedPrefixes] = subApp
-		subApp.init()
```
ReneWerner87, line 603: who is initalizing the app now ?
https://github.com/gofiber/fiber/blob/ce2d087394dc74dbe45e6c9f9bb263ae6f105a26/app.go#L952

this is necessary so that the render features also work together with the subapp
https://github.com/gofiber/fiber/blob/ce2d087394dc74dbe45e6c9f9bb263ae6f105a26/ctx.go#L1401

but maybe we should do this later in the init function of the parent app

if this has not been noticed by then, it seems as if we still need a test for this case. -> pls add a testcase for this mount app case from #1749

app.go
```diff
@@ -116,6 +116,12 @@ type App struct {
 	latestGroup *Group
 	// TLS handler
 	tlsHandler *TLSHandler
+	// check added routes of sub-apps
+	subAppsRoutesAdded bool
+	// Returns parent app if app was mounted
+	parentApp *App
```
ReneWerner87, line 122: property is only set in the whole code, never read out 

do we really need this then?

app.go
```diff
@@ -116,6 +116,12 @@ type App struct {
 	latestGroup *Group
 	// TLS handler
 	tlsHandler *TLSHandler
+	// check added routes of sub-apps
+	subAppsRoutesAdded bool
```
ReneWerner87, line 120: property is used in one place and set directly from false to true without actually being used anywhere else 

do we really need this property ?

ReneWerner87 (changes requested):
some little adjustments needed

app.go
efectn: app.Test executes buildTree in each usage. We shouldn't mount app again in tests

ReneWerner87: Okay make sense
Why not using sync.Once instead of the boolean property?

app.go
ReneWerner87: parentApp property is still not really in use

ctx_test.go
```diff
@@ -3107,6 +3176,23 @@ func (t *testTemplateEngine) Load() error {
 	return nil
 }
 
+type testTemplateEngine2 struct {
```
ReneWerner87, line 3179: Instead of copying them, please make the path in testTemplateEngine configurable

ctx_test.go
```diff
@@ -2761,6 +2761,75 @@ func Test_Ctx_Render_Mount(t *testing.T) {
 	utils.AssertEqual(t, "<h1>Hello a!</h1>", string(body))
 }
 
+// go test -run Test_Ctx_Render_Mount_ParentOrSubHasViews
+func Test_Ctx_Render_Mount_ParentOrSubHasViews(t *testing.T) {
```
ReneWerner87, line 2765: please. move the mount parts in a mount_test.go file

app_test.go
```diff
@@ -316,13 +316,67 @@ func Test_App_Mount(t *testing.T) {
...
 	app.Mount("/john", micro)
-
 	resp, err := app.Test(httptest.NewRequest(MethodGet, "/john/doe", nil))
 	utils.AssertEqual(t, nil, err, "app.Test(req)")
 	utils.AssertEqual(t, 200, resp.StatusCode, "Status code")
 	utils.AssertEqual(t, uint32(2), app.handlersCount)
 }
 
+// go test -run Test_App_Mount_Nested
+func Test_App_Mount_Nested(t *testing.T) {
```
ReneWerner87, line 326: please move the mount parts in a mount_test.go file

app.go
```diff
@@ -107,15 +108,23 @@ type App struct {
...
-	// Mounted and main apps
-	appList map[string]*App
 	// Hooks
 	hooks *Hooks
 	// Latest route & group
 	latestRoute *Route
 	latestGroup *Group
 	// TLS handler
 	tlsHandler *TLSHandler
+	// Mounted and main apps
```
ReneWerner87, line 118: can you please encapsulate all mount relevant properties that are needed in a struct which is located in mount.go

app.go
```diff
@@ -584,30 +594,33 @@ func (app *App) SetTLSHandler(tlsHandler *TLSHandler) {
 // any of the fiber's sub apps are added to the application's error handlers
 // to be invoked on errors that happen within the prefix route.
 func (app *App) Mount(prefix string, fiber *App) Router {
```
ReneWerner87, line 596: move it together with the other mount relevant functions into a mount.go file

ctx_test.go
```diff
@@ -2761,6 +2761,75 @@ func Test_Ctx_Render_Mount(t *testing.T) {
...
+	t.Parallel()
+
+	engine := &testTemplateEngine{}
+	err := engine.Load()
+	utils.AssertEqual(t, nil, err)
+
+	engine2 := &testTemplateEngine2{}
```
ReneWerner87, line 2772: please remove the loading function in the Render function 
https://github.com/gofiber/fiber/blob/5b1885a44c059b567e2fab9104df04ec9e0065e3/internal/template/html/html.go#L186-L192

ReneWerner87 (changes requested):
code looks good, but i have made a few comments which should clean up the code and make it easier to read and understand

ReneWerner87 (approved):

PR #2132, opened 2022-10-02
utils: use HTTP status messages from stdlib instead of keeping own list

leonklingele:
There's no need to keep an outdated copy of the list in this package as the "net/http" import is used by other files as well.

### DIFF
--- a/utils/http.go
+++ b/utils/http.go
@@ -4,7 +4,10 @@
 
 package utils
 
-import "strings"
+import (
+	"net/http"
+	"strings"
+)
 
 const MIMEOctetStream = "application/octet-stream"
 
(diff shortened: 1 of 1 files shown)


efectn:
```sh
Benchmark_StatusMessage/fiber-16                                        1000000000               0.6744 ns/op          0 B/op          0 allocs/op
Benchmark_StatusMessage/fiber-16                                        1000000000               0.6788 ns/op          0 B/op          0 allocs/op
Benchmark_StatusMessage/default-16                                      446818872                2.664 ns/op           0 B/op          0 allocs/op
Benchmark_StatusMessage/default-16                                      447009616                2.661 ns/op           0 B/op          0 allocs/op
```
Fiber's StatusMessage is faster than StatusText of `net/http`. I don't think we need to replace it

gaby:
Is there a way to use the const variable from net pkg? Seems the overhead comes fron the switch statement

jozsefsallai:
> Is there a way to use the const variable from net pkg? Seems the overhead comes fron the switch statement

We could refer to the `http.Status[whatever]` consts instead of using raw ints in that map, but I don't think it's really worth it, as these status codes are unlikely to ever change, and if a new one gets added, we can easily just add it to Fiber too. Personally I don't think there's really any benefit for this change. If anyone can bring up a valid real-world benefits, than we can reconsider it.

gaby:
> > Is there a way to use the const variable from net pkg? Seems the overhead comes fron the switch statement
> 
> We could refer to the `http.Status[whatever]` consts instead of using raw ints in that map, but I don't think it's really worth it, as these status codes are unlikely to ever change, and if a new one gets added, we can easily just add it to Fiber too. Personally I don't think there's really any benefit for this change. If anyone can bring up a valid real-world benefits, than we can reconsider it.

Totally agree with your statement. I'm curious about the performance difference between that and what Fiber is currently doing, and if it's substantial.

gaby:
@efectn @jozsefsallai @leonklingele So fasthttp already defines status codes the same way Fiber does. It even provides the same function: https://github.com/valyala/fasthttp/blob/master/status.go

jozsefsallai:
True, but it looks like fasthttp will return the string "Unknown Status Code" if the status code is outside the valid range, or a status message is not defined for the provided status code. Fiber, on the other hand, returns an empty string in these scenarios, so it's not 1:1 on how it works and it may be a breaking change for certain applications.

PR #2133, opened 2022-10-02
utils: update mime extensions

leonklingele:


efectn:
Can you also create a PR for https://github.com/gofiber/utils

efectn:
> > Can you also create a PR for https://github.com/gofiber/utils
> 
> Is this just a literal copy of the `utils` package used in here?! Why keep it duplicate?

We'll use it in v3 https://github.com/gofiber/fiber/blob/v3-beta/go.mod#L6

Issue #2145, opened 2022-10-08
🚀 [Feature]: missing session SessionOnly (cookie)


pjebs:
### Feature Description

See: https://pkg.go.dev/github.com/gofiber/fiber/v2@v2.38.1/middleware/session#Config

There seems to be no option to set SessionOnly.

See: https://pkg.go.dev/github.com/gofiber/fiber/v2#Cookie

SessionOnly deletes the cookies when the user closes the browser window.

If you set Expiration to 0, it deletes it immediately.

There is also ambiguity: Does `Expiration` refer to how long the data is stored in the session data or how long the cookie lasts for?

### Checklist:

- [X] I agree to follow Fiber's [Code of Conduct](https://github.com/gofiber/fiber/blob/master/.github/CODE_OF_CONDUCT.md).
- [X] I have checked for existing issues that describe my suggestion prior to opening this one.
- [X] I understand that improperly formatted feature requests may be closed without explanation.

li-jin-gou:
Refer to https://stackoverflow.com/questions/14196671/session-only-cookies-with-javascript

pjebs:
No, it can be done via backend.
You have to omit expiry value. Prob is you can't omit because a value of 0 ( the "zero value") means something else i.e. Delete session immediately.

li-jin-gou:
![image](https://user-images.githubusercontent.com/97824201/194757621-2902bfbc-9225-454e-8f9d-e2b0361d8943.png)
It also seems to be fine without setting.

pjebs:
You can't unset it. A value of zero is not the same

efectn:
> Hello can I jump in and help?

Sure. You can create a PR

PR #2152, opened 2022-10-13
🚀 [Feature]: SessionOnly when cookie.Expires is 0

yvestumushimire:
## Description

When Cookie Expires is not set default to zero (time.Time{}) is the same as `SessionOnly: true`

Fixes https://github.com/gofiber/fiber/issues/2145

## Type of change

## Checklist:

## Commit formatting:


pjebs:
https://stackoverflow.com/questions/70715904/golang-cookie-max-age-vs-expire

yvestumushimire:
> https://stackoverflow.com/questions/70715904/golang-cookie-max-age-vs-expire

@pjebs Thank you for reference I also find this definition:
>  Cookies are session cookies if they do not specify the Expires or Max-Age attribute.
I updated my changes.

ctx.go
```diff
@@ -456,8 +456,15 @@ func (c *Ctx) Cookie(cookie *Cookie) {
 	// i.e. cookie supposed to last beyond browser session
 	// refer: https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#define_the_lifetime_of_a_cookie
 	if !cookie.SessionOnly {
-		fcookie.SetMaxAge(cookie.MaxAge)
-		fcookie.SetExpire(cookie.Expires)
+		// Cookies are session cookies if they do not specify the Expires or Max-Age attribute.
+		// refer: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie
+		// phrase from MDN: "If both Expires and Max-Age are set, Max-Age has precedence."
+		// Skipping setting 0 time and seconds to MaxAge and Expires
+		if cookie.MaxAge != 0 {
```
ReneWerner87, line 463: now the question is how to make sure that you send an expire with 0 when defining the cookie 

shouldn't it rather be adjusted within the session middleware, that as soon as you send 0, the sessionOnly propertie is used when setting the cookie, instead of adjusting the general feature to set the cookie ?

ReneWerner87 (changes requested):
shouldn't it rather be adjusted within the session middleware, that as soon as you send 0, the sessionOnly property is used when setting the cookie, instead of adjusting the general feature to set the cookie ?

ReneWerner87:
@yvestumushimire can you check my last comment

yvestumushimire:
> shouldn't it rather be adjusted within the session middleware, that as soon as you send 0, the sessionOnly property is used when setting the cookie, instead of adjusting the general feature to set the cookie ?

Hello @ReneWerner87 can you please guide where " as soon as you send 0" happens I was thinking `setSession` method can solve it.

ReneWerner87:
currently we are changing the way cookies are handled throughout fiber so that it is no longer possible to set the expire of a cookie to 0

by doing this we are limiting the usage for all consumers and this may lead to unexpected behavior 

would rather change the logic in setSession 
https://github.com/gofiber/fiber/blob/c187c6a2f5d5e71bfc9ac799f7f8f1ab187aa8a4/middleware/session/session.go#L188
so that when the expire is set to 0 there instead of maxAge and expire the cookie is marked as session only
https://github.com/gofiber/fiber/blob/c187c6a2f5d5e71bfc9ac799f7f8f1ab187aa8a4/middleware/session/session.go#L198-L199

and maybe this place
https://github.com/gofiber/fiber/blob/c187c6a2f5d5e71bfc9ac799f7f8f1ab187aa8a4/middleware/session/session.go#L143-L145

yvestumushimire:
So if I understand you correctly instead of doing this:
```go
if int(s.exp.Seconds()) != 0 {
			fcookie.SetMaxAge(int(s.exp.Seconds()))
		} else if s.exp != 0 {
			fcookie.SetExpire(time.Now().Add(s.exp))
		}
```
I should set `SessionOnly: true`

ReneWerner87:
> So if I understand you correctly instead of doing this:
> 
> ```go
> if int(s.exp.Seconds()) != 0 {
> 			fcookie.SetMaxAge(int(s.exp.Seconds()))
> 		} else if s.exp != 0 {
> 			fcookie.SetExpire(time.Now().Add(s.exp))
> 		}
> ```
> 
> I should set `SessionOnly: true`

yes but in the session middleware files

pjebs:
Technically, setting Expire to 0 is a deliberate setting to immediately delete the session.

ReneWerner87:
Yes i know, but maybe someone is doing this instead of the delete and then we produce a breaking change

For me the way to solve the bug is in the files of the middleware

yvestumushimire:
I also saw `SessionOnly` is fields for Cookie struct and it is not implemented in `fasthttp`

ReneWerner87:
i think the feature is not completely thought out and also not 100% perfectly possible

how long should we keep the data in the storage if the cookie is set to the session only setting ?

pjebs:
> how long should we keep the data in the storage if the cookie is set to the session only setting ?

The expiration for the session data is a different concept. There should be a separate setting for that since the cookie expiration and session data expiration are different.
The only requirement should be cookie expiration duration should be <= session data expiration.

ReneWerner87:
then we would need another setting and should not mix it with the expire flag

and the kosumer then decides how long it keeps the data in the storage with sessionOnly?

ReneWerner87:
@yvestumushimire any progress on the subject ? last statement from me was that we need another setting not to destroy the other because they are actually different settings 

and we can't control 2 different things with expire

expire is used to tell the storage how long it should be kept in them 
Since sessionOnly or 0 is not a valid value for the storage expire time, this will not be feasible with one config setting.

yvestumushimire:
Hello @ReneWerner87 I actually get confused and I really need to understand these concepts more. cause the `SessionOnly` config is already available, and by definition if `Expires` is not specified **"the cookie becomes a session cookie. A session finishes when the client shuts down, after which the session cookie is removed."** - https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie 
the idea (my changes) was to make expires as alias for SessionOnly config. I would like your suggestions & guidance on how to do it.

ReneWerner87:
"Session only" is set if the expire is set to 0, but then the data is not stored in the server.

In my opinion you can't use the expire config setting for the session only feature, because both are different things.

Currently expire time controls how long the data is kept in the backend memory/storage and at the same time how long the session is valid or exists in the browser

"Session only" means, as you said, that the sessionid referencing the backend data is only present until the browser is closed.
But we can't set the expire to 0 for this feature, because at every request of the browser the data behind the referenced session id in the backend storage would be empty.

So we should make expire controllable for storage and session with more than one configuration property

Hope now it is more understandable, otherwise just report again

yvestumushimire:
@ReneWerner87 Does these latest changes make sense?

middleware/session/config.go
```diff
@@ -57,6 +57,9 @@ type Config struct {
 
 	// The session name
 	sessionName string
+
+	// Session only cookies
```
leonklingele, line 61: This and the next line should move up to the exported fields.

leonklingele:
What's the status of this @efectn @ReneWerner87. PR lgtm.

ReneWerner87 (approved):

PR #2153, opened 2022-10-13
:bug:  fix: solve data race in middleware/proxy's test

li-jin-gou:
## Description

Fixes # (issue)

## Type of change

## Checklist:

## Commit formatting:

### DIFF
--- a/middleware/proxy/proxy.go
+++ b/middleware/proxy/proxy.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/url"
 	"strings"
+	"sync"
 
 	"github.com/gofiber/fiber/v2"
 	"github.com/gofiber/fiber/v2/utils"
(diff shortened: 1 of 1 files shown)


li-jin-gou:
I found that the unit test I wrote before had `datarace` .

li-jin-gou:
> > I found that the unit test I wrote before had `datarace` between use `WithClient()` and `cli = client`.
> > 
> > * https://github.com/gofiber/fiber/blob/master/middleware/proxy/proxy.go#L114  set client
> > * https://github.com/gofiber/fiber/blob/master/middleware/proxy/proxy_test.go#L401 get client
> > 
> > can I use `mutex` or `atomic` here or remove `t.Parallel()`?
> 
> I think mutex

thanks

efectn (approved):

ReneWerner87 (approved):

Issue #2151, opened 2022-10-12
🐛 [Bug]: X-Forwarded-Host, X-Forwarded-Proto, X-Forwarded-Protocol can have a comma-separated list


supakornbabe:
### Bug Description

The implement of [c.Protocol()](https://github.com/gofiber/fiber/blob/9a6002056c4c1e76a44f0c68a0d7c27c39e09507/ctx.go#L1017) and [c.Hostname()](https://github.com/gofiber/fiber/blob/9a6002056c4c1e76a44f0c68a0d7c27c39e09507/ctx.go#L663) is wrong and the problem is happen when request is go through more than 1 proxies.

Reference: https://httpd.apache.org/docs/2.2/mod/mod_proxy.html#page-header

> Be careful when using these headers on the origin server, since they will contain more than one (comma-separated) value if the original request already contained one of these headers. For example, you can use %{X-Forwarded-For}i in the log format string of the origin server to log the original clients IP address, but you may get more than one address if the request passes through several proxies.

https://github.com/xp-framework/xp-framework/issues/162

### How to Reproduce

1. Set up go fiber behind 2 proxies
2. Use c.BaseURL(), c.Protocol(), c.Hostname() then these return comma separated list
```
BaseURL: https, https://hello.world.com, hello.world.com
Protocol: https, https
Hostname: hello.world.com, hello.world.com
```

### Expected Behavior

c.BaseURL(), c.Protocol(), c.Hostname() return only first item in list

Reference: https://github.com/xp-framework/xp-framework/issues/162

### Fiber Version

v2.38.1

### Code Snippet (optional)

```go
package main

import "github.com/gofiber/fiber/v2"
import "log"

func main() {
  app := fiber.New()

  // Steps to reproduce

  log.Fatal(app.Listen(":3000"))
}

func LoginHandler(c *fiber.Ctx) error {
	fmt.Println("BaseURL: "+c.BaseURL())
	fmt.Println("Protocol: "+c.Protocol())
        fmt.Println("Hostname: "+c.Hostname())
}

```
```
Result
BaseURL: https, https://hello.world.com, hello.world.com
Protocol: https, https
Hostname: hello.world.com, hello.world.com
```

### Checklist:

- [X] I agree to follow Fiber's [Code of Conduct](https://github.com/gofiber/fiber/blob/master/.github/CODE_OF_CONDUCT.md).
- [X] I have checked for existing issues that describe my problem prior to opening this one.
[...]

PR #2154, opened 2022-10-14
🩹 Fix: handle multiple X-Forwarded header

supakornbabe:
## Description

Fix the problem when the server is behind more than 2 proxies and c.BaseURL(), c.Protocol(), c.Hostname() will return the list instead of the first variable.
```go
func LoginHandler(c *fiber.Ctx) error {
	fmt.Println("BaseURL: "+c.BaseURL())
	fmt.Println("Protocol: "+c.Protocol())
        fmt.Println("Hostname: "+c.Hostname())
}
```

### Result
```
BaseURL: https, https://hello.world.com, hello.world.com
Protocol: https, https
Hostname: hello.world.com, hello.world.com
```
So this PR will fix these functions to return only first variable from list
### Result
```
BaseURL: https://hello.world.com
Protocol: https
Hostname: hello.world.com
```
Fixes https://github.com/gofiber/fiber/issues/2151

## Type of change

## Checklist:

## Commit formatting:

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -661,7 +661,7 @@ func (c *Ctx) GetRespHeaders() map[string]string {
 func (c *Ctx) Hostname() string {
 	if c.IsProxyTrusted() {
 		if host := c.Get(HeaderXForwardedHost); len(host) > 0 {
-			return host
+			return strings.Split(host, ",")[0]
 		}
 	}
 	return c.app.getString(c.fasthttp.Request.URI().Host())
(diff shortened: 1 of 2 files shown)


ctx.go
```diff
@@ -661,7 +661,7 @@ func (c *Ctx) GetRespHeaders() map[string]string {
 func (c *Ctx) Hostname() string {
 	if c.IsProxyTrusted() {
 		if host := c.Get(HeaderXForwardedHost); len(host) > 0 {
-			return host
+			return strings.Split(host, ",")[0]
```
efectn, line 664: I think you can use strings.Index instead. Split causes extra alloc

supakornbabe: Thanks for suggestion, applied already 😄

efectn (approved):

ReneWerner87 (approved):

PR #2159, opened 2022-10-18
🚀 [Feature]: Cache-Control: no-cache

marcmartin13:
Added noCache function for checking Cache-Control: no-cache in the request header

## Description
If the request header contains Cache-Control: no-cache it will skip getting the cached entry and proceeds to cache the response.

## Type of change

## Checklist:

### DIFF
--- a/middleware/cache/README.md
+++ b/middleware/cache/README.md
@@ -132,12 +132,13 @@ type Config struct {
 	// Default: []string{fiber.MethodGet, fiber.MethodHead}
 	Methods []string
 
-	// Check if the request header contains Cache-Control: no-cache
-	//
-	// Default: func(c *fiber.Ctx) bool {
-	// 	return strings.Contains(c.Get(fiber.HeaderCacheControl), "no-cache")
+	// Allows you to use no-cache and no-store request directives
+	// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control
+	// 
+	// Default: func(c *fiber.Ctx, directive string) bool {
+	// 	return strings.Contains(c.Get(fiber.HeaderCacheControl), directive)
 	// },
-	NoCache func(c *fiber.Ctx) bool
+	RequestDirective func(c *fiber.Ctx, directive string) bool
 }
 ```
 
@@ -158,8 +159,8 @@ var ConfigDefault = Config{
 	Storage:              nil,
 	MaxBytes:             0,
 	Methods:              []string{fiber.MethodGet, fiber.MethodHead},
-	NoCache: func(c *fiber.Ctx) bool {
-		return strings.Contains(c.Get(fiber.HeaderCacheControl), "no-cache")
+	RequestDirective: func(c *fiber.Ctx, directive string) bool {
+		return strings.Contains(c.Get(fiber.HeaderCacheControl), directive)
 	},
 }
 ```
--- a/middleware/cache/README.md
+++ b/middleware/cache/README.md
@@ -4,12 +4,7 @@ Cache middleware for [Fiber](https://github.com/gofiber/fiber) designed to inter
 
 Request Directives
 The no-cache request directive will return the up-to-date response but still caches it. You will always get a "miss" cache status.
-Usage:
-Cache-Control: no-cache
-
 The no-store request directive will refrain from caching. You will always get the up-to-date response.
-Usage:
-Cache-Control: no-store
 
 ## Table of Contents
 
--- a/middleware/cache/README.md
+++ b/middleware/cache/README.md
@@ -131,6 +131,13 @@ type Config struct {
 	//
 	// Default: []string{fiber.MethodGet, fiber.MethodHead}
 	Methods []string
+
+	// Check if the request header contains Cache-Control: no-cache
+	//
+	// Default: func(c *fiber.Ctx) bool {
+	// 	return strings.Contains(c.Get(fiber.HeaderCacheControl), "no-cache")
+	// },
+	NoCache func(c *fiber.Ctx) bool
 }
 ```
 
@@ -151,5 +158,8 @@ var ConfigDefault = Config{
 	Storage:              nil,
 	MaxBytes:             0,
 	Methods:              []string{fiber.MethodGet, fiber.MethodHead},
+	NoCache: func(c *fiber.Ctx) bool {
+		return strings.Contains(c.Get(fiber.HeaderCacheControl), "no-cache")
+	},
 }
 ```
--- a/middleware/cache/README.md
+++ b/middleware/cache/README.md
@@ -138,7 +138,7 @@ type Config struct {
 	// Default: func(c *fiber.Ctx, directive string) bool {
 	// 	return strings.Contains(c.Get(fiber.HeaderCacheControl), directive)
 	// },
-	RequestDirective func(c *fiber.Ctx, directive string) bool
+	requestDirective func(c *fiber.Ctx, directive string) bool
 }
 ```
 
@@ -159,7 +159,7 @@ var ConfigDefault = Config{
 	Storage:              nil,
 	MaxBytes:             0,
 	Methods:              []string{fiber.MethodGet, fiber.MethodHead},
-	RequestDirective: func(c *fiber.Ctx, directive string) bool {
+	requestDirective: func(c *fiber.Ctx, directive string) bool {
 		return strings.Contains(c.Get(fiber.HeaderCacheControl), directive)
 	},
 }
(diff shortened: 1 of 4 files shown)


ReneWerner87:
@marcmartin13 thanks for the work

can you tell me what is the difference of this function with the Next function? this can also be done with it or ?

thought rather we integrate the fixed without possible change of functionality

can you add more tests so we can make sure this functionality works and we don't have to manually test each time for minor adjustments

middleware/cache/cache.go
```diff
@@ -109,43 +109,47 @@ func New(config ...Config) fiber.Handler {
...
-			c.Response().SetStatusCode(e.status)
-			c.Response().Header.SetContentTypeBytes(e.ctype)
-			if len(e.cencoding) > 0 {
-				c.Response().Header.SetBytesV(fiber.HeaderContentEncoding, e.cencoding)
-			}
-			if e.headers != nil {
-				for k, v := range e.headers {
-					c.Response().Header.SetBytesV(k, v)
+		// Check if no-cache
+		if !cfg.noCache(c) {
```
ReneWerner87, line 113: please flatten

simply turn over and make return, of course via the start of the mutex lock

marcmartin13:
> @marcmartin13 thanks for the work
> 
> can you tell me what is the difference of this function with the Next function? this can also be done with it or ?
> 
> thought rather we integrate the fixed without possible change of functionality
> 
> can you add more tests so we can make sure this functionality works and we don't have to manually test each time for minor adjustments

Hi @ReneWerner87

Sorry for the late reply

first: Next function does not execute if the entry is already cached (I already tested it and also used skip middleware but that's not how it works).  It will always get hit response because it will not reach the Next function. That's why adding Cache-Control: no-cache checker will be viable if we want to control the cache. Sample what if I don't want to get hit reponse but I want to store the cached (always miss).

test: ETag + Cache
I still researching on how can show my test here but I already tested it on my local project. I found out that If the entry is already cached It will always get the hit response even using ETag "if-none-match". So I used skip middleware as a workaround since Next function will do nothing if the entry is already cached.

       app.Use(cache.New(settings.CacheConfig), func(c *fiber.Ctx) bool {
		return strings.Contains(c.Get(fiber.HeaderCacheControl), "no-cache")
	}))
	
but thats not it should be. Using "no-cache" in request directive means I will always get an updated response but I still can cached it.

Please see this.
image
https://www.cloudflare.com/learning/cdn/glossary/what-is-cache-control/

I hope you understand. Sorry for my bad english.

Thanks!

middleware/cache/cache.go
marcmartin13: Do you mean to include these lines inside the check if no-cache condition?

// Lock entry
mux.Lock()

// Get timestamp
ts := atomic.LoadUint64(&timestamp)

marcmartin13: I've updated the file. Instead of enclosing the check/get entry, I put no-cache check condition on line 199 since it blocks the check if the entry is expired condition.

image

middleware/cache/cache_test.go
```diff
@@ -107,6 +107,36 @@ func Test_Cache(t *testing.T) {
 	utils.AssertEqual(t, cachedBody, body)
 }
 
+func Test_CacheWithCacheControlNoCacheRequestHeader(t *testing.T) {
```
marcmartin13, line 110: Hi @ReneWerner87

I've added a test method here Test_CacheWithCacheControlNoCacheRequestHeader

middleware/cache/cache_test.go
```diff
@@ -107,6 +108,110 @@ func Test_Cache(t *testing.T) {
 	utils.AssertEqual(t, cachedBody, body)
 }
 
+func Test_Cache_WithCacheControlNoCacheRequestHeader(t *testing.T) {
```
marcmartin13, line 111: added Test_Cache_WithCacheControlNoCacheRequestHeader and Test_Cache_WithETagAndCacheControlNoCacheRequestHeader

ReneWerner87:
okay sounds good
can you still add some test cases for the new code though

middleware/cache/config.go
```diff
@@ -72,6 +73,9 @@ type Config struct {
 	//
 	// Default: []string{fiber.MethodGet, fiber.MethodHead}
 	Methods []string
+	
+	// Check if the request header contains Cache-Control: no-cache
+	noCache func(c *fiber.Ctx) bool
```
efectn, line 78: Why not to make it public?

marcmartin13: Oh, now that you mention it.

efectn:
Please update readme and create a PR for docs

marcmartin13:
> Please update readme and create a PR for docs

Hi @efectn, I'm sorry but where I can find it, and how to create PR for docs? Thank you

efectn:
> > Please update readme and create a PR for docs
> 
> Hi @efectn, I'm sorry but where I can find it, and how to create PR for docs? Thank you

https://github.com/gofiber/docs/tree/master/api/middleware you can take a look here

middleware/cache/cache.go
ReneWerner87: i see no change in the structure ?

ReneWerner87 (approved):

marcmartin13:
Hi @ReneWerner87 and @efectn

may ask if, is it ok to add one more request directive?

image
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control

it tells the server not to cache

middleware/cache/README.md
```diff
@@ -131,6 +131,14 @@ type Config struct {
...
 	// Default: []string{fiber.MethodGet, fiber.MethodHead}
 	Methods []string
+
+	// Allows you to use no-cache and no-store request directives
+	// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control
+	// 
+	// Default: func(c *fiber.Ctx, directive string) bool {
+	// 	return strings.Contains(c.Get(fiber.HeaderCacheControl), directive)
+	// },
+	RequestDirective func(c *fiber.Ctx, directive string) bool
```
ReneWerner87, line 141: @marcmartin13 
not sure yet if we really should expose the RequestDirective method

what can i configure as a consumer of the middleware other than what is configured there, if this method is used in two different places

ReneWerner87 (changes requested):
not sure yet if we really should expose the RequestDirective method

ReneWerner87:
> Please update readme and create a PR for docs

- [ ]  update readme and create a PR for docs

middleware/cache/README.md
marcmartin13: oh right. i'll make it private

marcmartin13: @ReneWerner87 
oh right, I'll make it private for now

marcmartin13:
> not sure yet if we really should expose the RequestDirective method

I've changed it to a normal private function.

Question: is updating readme and docs necessary since it is just a private function?

thanks!

marcmartin13 (commented):
changed to normal private function

efectn:
> I've changed it to a normal private function.
> 
> Question: is updating readme and docs necessary since it is just a private function?
> 
> thanks!

No

ReneWerner87:
> I've changed it to a normal private function.
> 
> Question: is updating readme and docs necessary since it is just a private function?
> 
> thanks!

we could already describe the new behavior so that everyone knows when they use the middleware and don't have to look in the code, but maybe you already implicitly assume this behavior

marcmartin13:
> not sure yet if we really should expose the RequestDirective method



> not sure yet if we really should expose the RequestDirective method



> > I've changed it to a normal private function.
> > Question: is updating readme and docs necessary since it is just a private function?
> > thanks!
> 
> we could already describe the new behavior so that everyone knows when they use the middleware and don't have to look in the code, but maybe you already implicitly assume this behavior

yep, you are right. but I don't know where to put the description of this function since readme and docs have only the config description.

ReneWerner87:
then just start with the first descriptions

look at the other middlewares and copy the style or look where the description i

marcmartin13:
> then just start with the first descriptions
> 
> look at the other middlewares and copy the style or look where the description i

on it

ReneWerner87:
https://github.com/gofiber/fiber/blob/master/middleware/cache/README.md?plain=1#L3
https://github.com/gofiber/docs/blob/master/api/middleware/cache.md?plain=1#L3

ReneWerner87 (approved):

marcmartin13:
@ReneWerner87 thanks for helping me

PR #2164, opened 2022-10-20
[Bug] Fix query string parameter pass to fiber context

joseroberto:
## Description

Remove this change (https://github.com/gofiber/fiber/pull/1909).

Fix test when using query string parameter.   

When you use `/foo?bar=boo`, your test case should able to QueryParser parameter from fiber.Ctx:

```
func ListResource(c *fiber.Ctx) error {
        mrf := new(models ResouceFilter)
	if err = c.QueryParser(mrf); err != nil {
		return sendError(c, http.StatusBadRequest, err)
	}
```

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/app.go
+++ b/app.go
@@ -883,11 +883,6 @@ func (app *App) Test(req *http.Request, msTimeout ...int) (resp *http.Response, 
 		return nil, err
 	}
 
-	// adding back the query from URL, since dump cleans it
-	dumps := bytes.Split(dump, []byte(" "))
-	dumps[1] = []byte(req.URL.String())
-	dump = bytes.Join(dumps, []byte(" "))
-
 	// Create test connection
 	conn := new(testConn)
 
--- a/app.go
+++ b/app.go
@@ -9,7 +9,6 @@ package fiber
 
 import (
 	"bufio"
-	"bytes"
 	"errors"
 	"fmt"
 	"net"


app.go
```diff
@@ -883,11 +882,6 @@ func (app *App) Test(req *http.Request, msTimeout ...int) (resp *http.Response,
 		return nil, err
 	}
 
-	// adding back the query from URL, since dump cleans it
```
li-jin-gou, line 886: 我认为 PR 的描述和标题应该按照 pull request template 来填写

ReneWerner87:
@ancogamer the change in the pull request before is from you, can you check if this is correct here or give background information

app.go
```diff
@@ -883,11 +882,6 @@ func (app *App) Test(req *http.Request, msTimeout ...int) (resp *http.Response,
...
-	dumps := bytes.Split(dump, []byte(" "))
```
li-jin-gou, line 887: I think the description and title of the PR should be filled in according to [pull request template ](https://github.com/gofiber/fiber/blob/master/.github/pull_request_template.md).

li-jin-gou (changes requested):

ReneWerner87:
but is this change okay or not ?

ancogamer:
Dont see any problem, but the response of Test wont have anymore the query on it.  @joseroberto  can you show a example of how to get those values on test case ?

joseroberto:
@ancogamer Dear, your change suppresses query string parameter's to be pass throw the context.

dumps[1] = []byte(req.URL.String()). <---- this remove all the URI query string

Before:   `/foo?bar=boo`
After:  `/foo`

```
func xxx(c *fiber.Ctx) error {
	err = c.Query("bar");  <-- It show not be empty and it is with your modification
}
```

app.go
joseroberto: Sorry about that!!!

ancogamer:
Well, @ReneWerner87  to me this change looks fine. Sorry for the problens, i dont know why, but when i add it, the DumpRequest, wasn't returning the  ?query, looks like it was fixed.

ReneWerner87:
Ok thx for the clarification

ReneWerner87 (approved):

Issue #2160, opened 2022-10-18
🚀 [Feature]: v3 Router interface change


pjebs:
### Feature Description

**v2 definition:**
```go
type Router interface {
	Use(args ...interface{}) Router

	Get(path string, handlers ...Handler) Router
	Head(path string, handlers ...Handler) Router
	Post(path string, handlers ...Handler) Router
	Put(path string, handlers ...Handler) Router
	Delete(path string, handlers ...Handler) Router
	Connect(path string, handlers ...Handler) Router
	Options(path string, handlers ...Handler) Router
	Trace(path string, handlers ...Handler) Router
	Patch(path string, handlers ...Handler) Router

	Add(method, path string, handlers ...Handler) Router
	Static(prefix, root string, config ...Static) Router
	All(path string, handlers ...Handler) Router

	Group(prefix string, handlers ...Handler) Router

	Route(prefix string, fn func(router Router), name ...string) Router

	Mount(prefix string, fiber *App) Router

	Name(name string) Router
}
```

**v3 definition:**

It would be good if the signature could be changed for these:
```
	Get(path string, handler Handler, middleware ...Handler) Router
	Head(path string, handler Handler, middleware ...Handler) Router
	Post(path string, handler Handler, middleware ...Handler) Router
	Put(path string, handler Handler, middleware ...Handler) Router
	Delete(path string, handler Handler, middleware ...Handler) Router
	Connect(path string, handler Handler, middleware ...Handler) Router
	Options(path string, handler Handler, middleware ...Handler) Router
	Trace(path string, handler Handler, middleware ...Handler) Router
	Patch(path string, handler Handler, middleware ...Handler) Router

	Add(method, path string, handler Handler, middleware ...Handler) Router
	All(path string, handler Handler, middleware ...Handler) Router
```

**Secondly,**

```go
Add(methods []string, path string, handler Handler, middleware ...Handler) Router
```
OR alternatively, 

`method string` interprets methods with pipes: eg. `"GET|POST"`

### Additional Context (optional)

It will be operationally equivalent to (but cleaner):
```go
func WithMW(handler fiber.Handler, middleware ...fiber.Handler) []fiber.Handler {
	stack := make([]fiber.Handler, 0, len(middleware)+1)
	stack = append(stack, middleware...)
	return append(stack, handler)
}

app.Get("/hello", WithMW(func(c *fiber.Ctx) error {
	return c.SendString("Hello, World 👋!")
})...)
```

### Checklist:

[...]

PR #2175, opened 2022-10-27
Add(methods []string, path string, handlers ...Handler) Router

pjebs:
See: https://github.com/gofiber/fiber/issues/2160

This is PR Part 1


efectn:
I think we can do this in https://github.com/gofiber/fiber/pull/2176

PR #2176, opened 2022-10-27
:sparkles: v3 (feature): router interface changes

pjebs:
See: https://github.com/gofiber/fiber/issues/2160

This is PR Part 2 (requires Part 1: https://github.com/gofiber/fiber/pull/2175)

### DIFF
--- a/register.go
+++ b/register.go
@@ -17,7 +17,7 @@ type Register interface {
 	Trace(handlers ...Handler) Register
 	Patch(handlers ...Handler) Register
 
-	Add(method string, handlers ...Handler) Register
+	Add(methods []string, handlers ...Handler) Register
 
 	Static(root string, config ...Static) Register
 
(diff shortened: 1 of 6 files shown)


register.go
```diff
@@ -6,18 +6,18 @@ package fiber
...
+	Head(handler Handler, middleware ...Handler) Register
+	Post(handler Handler, middleware ...Handler) Register
+	Put(handler Handler, middleware ...Handler) Register
+	Delete(handler Handler, middleware ...Handler) Register
+	Connect(handler Handler, middleware ...Handler) Register
+	Options(handler Handler, middleware ...Handler) Register
+	Trace(handler Handler, middleware ...Handler) Register
+	Patch(handler Handler, middleware ...Handler) Register
+
+	Add(methods []string, handler Handler, middleware ...Handler) Register
```
ReneWerner87, line 20: we can allow simple strings and lists of strings here with generics ?

pjebs: I don't think generics is allowed in interface.

pjebs: One possibility is:

```
Add(methods interface{}, handler Handler, middleware ...Handler) Register calls
add(methods []string, handler Handler, middleware ...Handler) Register
```

ReneWerner87 (approved):

PR #2178, opened 2022-10-28
:construction_worker: ci: replace snyk with govulncheck

efectn:
## Description

https://github.com/jackc/pgx/issues/1352#issuecomment-1293908662


li-jin-gou (approved):

.github/workflows/vulncheck.yml
```diff
@@ -0,0 +1,20 @@
+on: [push, pull_request_target]
+name: Vulnerability Check
+jobs:
+  Security:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Install Go
+      uses: actions/setup-go@v3
+      with:
+        go-version: 1.19.x
```
gaby, line 10: Shouldn't this run with 1.16.x and 1.18.x too?

gaby (changes requested):

.github/workflows/vulncheck.yml
efectn: Benchmark also runs on 1 go version. I think we don't need another versions in this workflow

gaby (approved):

ReneWerner87 (approved):

Issue #2180, opened 2022-10-31
🐛 [Bug]: unsafeString2Bytes may not be future proof


pjebs:
### Bug Description

https://github.com/gofiber/fiber/blob/master/utils/convert.go#L24
https://github.com/gofiber/fiber/blob/master/internal/fasttemplate/unsafe.go#L15

See: https://stackoverflow.com/questions/59209493/how-to-use-unsafe-get-a-byte-slice-from-a-string-without-memory-copy

The current implementations may not be as future proof as it could be according to the discussion in the link above.

They propose:
```go
func unsafeGetBytes(s string) []byte {
    if s == "" {
        return nil // or []byte{}
    }
    return (*[0x7fff0000]byte)(unsafe.Pointer(
        (*reflect.StringHeader)(unsafe.Pointer(&s)).Data),
    )[:len(s):len(s)]
}
```

### How to Reproduce

N/A

### Expected Behavior

N/A

### Fiber Version

N/A 

### Checklist:

- [X] I agree to follow Fiber's [Code of Conduct](https://github.com/gofiber/fiber/blob/master/.github/CODE_OF_CONDUCT.md).
- [X] I have checked for existing issues that describe my problem prior to opening this one.
- [X] I understand that improperly formatted bug reports may be closed without explanation.

efectn:
Agree with you. I think we should use recommend solution.

ReneWerner87:
The initial functions were copies of the fasthttp functions, we should open a topic there too

Furthermore we should comment the line with the magic limit and determine this number outside on a constant one

PR #2204, opened 2022-11-11
:broom: update: replace UnsafeBytes util with suggested way

efectn:
## Description
Fixes https://github.com/gofiber/fiber/issues/2180
This is also 2 ns/op faster than old way.

## Type of change
- [x] Just an update.

## Checklist:

## Commit formatting:

### DIFF
--- a/utils/convert.go
+++ b/utils/convert.go
@@ -13,6 +13,8 @@ import (
 	"unsafe"
 )
 
+const MaxStringLen = 0x7fff0000 // Maximum string length for UnsafeBytes. (decimal: 2147418112)
+
 // #nosec G103
 // UnsafeString returns a string pointer without allocation
 func UnsafeString(b []byte) string {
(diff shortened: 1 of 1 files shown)


ReneWerner87 (approved):

utils/convert.go
```diff
@@ -13,21 +13,25 @@ import (
...
-	sh := (*reflect.StringHeader)(unsafe.Pointer(&s))
-	bh := (*reflect.SliceHeader)(unsafe.Pointer(&bs))
-	bh.Data = sh.Data
-	bh.Len = sh.Len
-	bh.Cap = sh.Len
-	return
+// UnsafeBytes returns a byte pointer without allocation.
+// String length shouldn't be more than 2147418112.
+func UnsafeBytes(s string) []byte {
+	if s == "" {
```
leonklingele, line 28: What happens if the MaxStringLen limit is exceeded? Why not check for this in here as well?

efectn: Will panic like `panic: runtime error: slice bounds out of range `

ReneWerner87: okay for me, since this limit is high enough

ReneWerner87 (approved):

PR #2214, opened 2022-11-14
router: return status 501 instead of 400 on unknown method

leonklingele:


efectn:
This is breaking-change i think. v3 would be better for this improvement.

leonklingele:
Replaced by https://github.com/gofiber/fiber/pull/2220.

PR #2221, opened 2022-11-14
Track Configured Values

calebcase:
## Description

Between version v2.23.0 and v2.24.0 the behavior of Mount changed with regard to how the child's ErrorHandler is used. Before v2.24.0 if the child's error handler was not set, then it used the parent's error handler. From v2.24.0 to v2.39.0 (most recent as of this PR) the behavior is to use the child's error handler (which, even when it isn't explicitly set, is set to the DefaultErrorHandler by `New()`).

This change fixes that issue and provides a way for others to fix similar issues as they arise by checking if the value was explicitly configured.

## Type of change

## Checklist:

## Commit formatting:

### DIFF
--- a/app.go
+++ b/app.go
@@ -114,6 +114,8 @@ type App struct {
 	customMethod bool
 	// Mount fields
 	mountFields *mountFields
+	// True if the ErrorHandler is explicitly configured
+	errorHandlerConfigured bool
 }
 
 // Config is a struct holding the server settings.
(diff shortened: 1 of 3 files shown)


app.go
```diff
@@ -114,6 +114,8 @@ type App struct {
 	customMethod bool
```
efectn, line 114: I think we can do the same for customMethod

calebcase: Done

ReneWerner87 (approved):

PR #2228, opened 2022-11-21
feat: add ShutdownWithTimeout function

pjebs:
Waiting for new tag upstream


pjebs:
See: https://github.com/valyala/fasthttp/pull/1383

app.go
```diff
@@ -839,12 +840,30 @@ func (app *App) HandlersCount() uint32 {
...
+}
+
+// ShutdownWithTimeout gracefully shuts down the server without interrupting any active connections. However, if the timeout is exceeded,
+// ShutdownWithTimeout will forcefully close any active connections.
+// ShutdownWithTimeout works by first closing all open listeners and then waiting for all connections to return to idle before shutting down.
+//
+// Make sure the program doesn't exit and waits instead for ShutdownWithTimeout to return.
+//
+// ShutdownWithTimeout does not close keepalive connections so its recommended to set ReadTimeout to something else than 0.
+func (app *App) ShutdownWithTimeout(timeout time.Duration) error {
```
efectn, line 859: I think we can make ShutdownWithContext public instead. Or add variadic context.Context param to Shutdown method. What do you think @ReneWerner87 @li-jin-gou

li-jin-gou: I think so.

pjebs: I think allowing context is overkill. Context allows for cancellation, deadline or timeout. This function only makes sense with a timeout.

efectn: I think we may need to pass context for some cases

li-jin-gou (changes requested):

app.go
```diff
@@ -839,12 +840,30 @@ func (app *App) HandlersCount() uint32 {
...
+	ctx, cancelFunc := context.WithTimeout(context.Background(), timeout)
+	defer cancelFunc()
+	return app.shutdownWithContext(ctx)
+}
+
+// shutdownWithContext shutsdown the server including by force if the context's deadline is exceeded.
+func (app *App) shutdownWithContext(ctx context.Context) error {
```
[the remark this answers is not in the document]
pjebs: Are you really sure you want to export it?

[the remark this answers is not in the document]
li-jin-gou: I am not referring to this line code specifically, but to this pr.😂

app.go
ReneWerner87: > I think we may need to pass context for some cases

Can you produce some realworld example scenarios so I can understand the background better?

functionality seems to me actually okay, don't know if you really still need the context or maybe you have already prepared the context outside and want to use it then, but don't know if this is a real world scenario or just a very theoretical one

efectn: I think it would be really useful to use https://pkg.go.dev/os/signal#NotifyContext easily

go.mod
```diff
@@ -6,8 +6,8 @@ require (
 	github.com/mattn/go-colorable v0.1.13
 	github.com/mattn/go-isatty v0.0.16
 	github.com/mattn/go-runewidth v0.0.14
+	github.com/valyala/fasthttp v1.42.0
 	github.com/valyala/bytebufferpool v1.0.0
-	github.com/valyala/fasthttp v1.41.0
```
leonklingele, line 10: Why did this line move upwards? It should stay sorted alphabetically.

app.go
pjebs: I think there is a misunderstanding of what this function does. `https://pkg.go.dev/os/signal#NotifyContext` can/should be used to _trigger_ the `Shutdown` function. It has nothing to do with the timeout for when the force shutdown should be activated.

pjebs:
This is now ready for release.

app.go
```diff
@@ -839,12 +840,30 @@ func (app *App) HandlersCount() uint32 {
 }
 
 // Shutdown gracefully shuts down the server without interrupting any active connections.
-// Shutdown works by first closing all open listeners and then waiting indefinitely for all connections to return to idle and then shut down.
+// Shutdown works by first closing all open listeners and then waiting indefinitely for all connections to return to idle before shutting down.
```
leonklingele, line 843: Why did this change? It should stay in sync with upstream.

app.go
```diff
@@ -839,12 +840,30 @@ func (app *App) HandlersCount() uint32 {
...
 //
 // Make sure the program doesn't exit and waits instead for Shutdown to return.
 //
 // Shutdown does not close keepalive connections so its recommended to set ReadTimeout to something else than 0.
 func (app *App) Shutdown() error {
+	return app.shutdownWithContext(context.Background())
+}
+
+// ShutdownWithTimeout gracefully shuts down the server without interrupting any active connections. However, if the timeout is exceeded,
```
leonklingele, line 852: Why is this a different comment than the one you provided upstream? imo, we should stay in sync.

pjebs: Upstream doesn't explain it properly.

app.go
pjebs: Prior comments is not grammatically correct.

leonklingele: This should be fixed upstream then as well.

app.go
leonklingele: This should be fixed upstream then as well.

pjebs: Not much I can do about that.

efectn:
> Can this be merged and tagged?

Can you create a PR for docs repo

pjebs:
Here is PR for docs https://github.com/gofiber/docs/pull/309

efectn (approved):

app.go
efectn, line 859: I think you can add a testcase

pjebs: I don't know how to add a test case and test force shutdown feature.

```go
func Test_App_Shutdown(t *testing.T) {
	t.Run("success", func(t *testing.T) {
		app := New(Config{
			DisableStartupMessage: true,
		})
		utils.AssertEqual(t, true, app.Shutdown() == nil)
	})

	t.Run("no server", func(t *testing.T) {
		app := &App{}
		if err := app.Shutdown(); err != nil {
			if err.Error() != "shutdown: server is not running" {
				t.Fatal()
			}
		}
	})
}
```

li-jin-gou: I think we can refer to https://github.com/valyala/fasthttp/blob/master/server_test.go#L3598 😁

pjebs: That test requires too much access to the fasthttp internals which is probably not wise to use from Fiber since the internals could change anytime.

li-jin-gou: Yes, but we can refer to the idea.

pjebs: Can you have a crack at it?

li-jin-gou: ok

li-jin-gou: Added a unit  test cc@pjebs  @efectn

pjebs: awesome work. One of the imports should not be grouped with standard lib imports.

pjebs:
awesome work @li-jin-gou

li-jin-gou (approved):

li-jin-gou:
LGTM

PR #2246, opened 2022-11-29
:bug: bug: handle url encoded space/values in middleware/filesystem

Simerax:
## Description

Make the filesystem middleware handle url encoded values in the url path. 
Main purpose of the change is to handle url encoded space.

Fixes # (issue)

## Type of change

## Checklist:

- [x] New and existing unit tests pass locally with my changes (`go test -run Test_FileSystem ./middleware/filesystem`, normal `go test ./...` failed before and after my change)

## Commit formatting:

### DIFF
--- a/.github/testdata/fs/css/style with space.css
+++ b/.github/testdata/fs/css/style with space.css
@@ -1,0 +1,4 @@
+h1 {
+    color: red;
+    text-align: center;
+}
--- a/middleware/filesystem/filesystem.go
+++ b/middleware/filesystem/filesystem.go
@@ -2,6 +2,7 @@ package filesystem
 
 import (
 	"net/http"
+	"net/url"
 	"os"
 	"strconv"
 	"strings"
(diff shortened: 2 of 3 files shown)


ReneWerner87:
Thanks for the fix

I think it will affect the performance though

Can you please make a before and after performance measurement and share the result with us

Here is an interesting link with a fasthttp functionality we might use

https://github.com/gofiber/fiber/blob/master/ctx.go#L1717

ReneWerner87:
btw have you tested the config setting `UnescapePath: true` ?

Simerax:
> btw have you tested the config setting `UnescapePath: true` ?

Looks like this setting fixes the problem. The path gets unescaped. I guess this pr is obsolete then. I think its a little weird that this setting is false by default though. Maybe it would make sense to document this behaviour in the middleware? something like "hey if you serve files with spaces make sure fiber config has `UnescapePath` set to true"

ReneWerner87:
> > btw have you tested the config setting `UnescapePath: true` ?
> 
> Looks like this setting fixes the problem. The path gets unescaped. I guess this pr is obsolete then. I think its a little weird that this setting is false by default though. Maybe it would make sense to document this behaviour in the middleware? something like "hey if you serve files with spaces make sure fiber config has `UnescapePath` set to true"

yes more documentation helps here in the case
can you do that?

Simerax:
see #2247

Issue #2163, opened 2022-10-19
🚀 [Feature]: Idempotency Middleware


GalvinGao:
### Feature Description

Idempotency is a guarantee that enables the client to have safe retries if no response (or a 500 response) is received from the server, by using an "idempotency key" that is supplied by client to denote the initial request and its retry requests as a singular set of intention (instead of multiple intentions).

Production examples of idempotency:
- [Stripe](https://stripe.com/docs/api/idempotent_requests): Idempotency on all mutable requests
- [Square](https://developer.squareup.com/docs/build-basics/common-api-patterns/idempotency) 
- [Shopify](https://shopify.dev/api/usage/idempotent-requests)

### Additional Context (optional)

When the middleware received a request...
- ...without an idempotency key
	- It will just call the `Next()` handler and that's all.
- ...with an idempotency key, but such key does not exist/already expired
	- Lock the idempotency mutex lock with idempotency key as the key. The lock is typically a distributed redis lock. We should just provide an interface for outer libraries/users to implement.
		- After acquired lock, check the existence of idempotency key again in storage.
	- The middleware will call `Next()` handler, and
		- If returned non-nil error, return the error and skip idempotency saving.
	- Save the response status, body, and all headers specified in `PersistResponseHeaders`
	- Unlock the idempotency mutex lock with idempotency key as the key.
- ...with an idempotency key, and such key exists/not expired
	- Return the saved response (status, body, and all headers saved) for such key.

### Code Snippet (optional)

```go
package main

import (
	"log"
	"time"

	"github.com/gofiber/fiber/v2"
	"github.com/gofiber/fiber/v2/middleware/idempotency"
)

func main() {
	app := fiber.New()

	app.Use(idempotency.New(idempotency.Config{
		// LifeTime is the backtrack time window of the presence of an idempotency key. Default is 24h.
		LifeTime: time.Hour * 24,

		// KeyHeader is the header for looking up the idempotency key from. Default is "X-Idempotency-Key".
		KeyHeader: "X-Idempotency-Key",

		// Storage is the storage of idempotency responses. Default is a in-memory fiber.Storage.
		Storage: fiberstore.NewRedis(), // this is to just illustrate a storage that implements fiber.Storage

		// PersistResponseHeaders is a set of headers to persist. Default is nil.
		PersistResponseHeaders: []string{
			fiber.HeaderContentType,
			fiber.HeaderContentLength,
			fiber.HeaderSetCookie,
		},
	}))

[...]

PR #2253, opened 2022-12-03
🚀 Feature: Add idempotency middleware

leonklingele:
## Description

Add idempotency middleware.
Implements https://github.com/gofiber/fiber/issues/2163.

## Type of change

## Checklist:


middleware/idempotency/storage.go
```diff
@@ -0,0 +1,87 @@
+package idempotency
+
+import (
+	"sync"
+	"time"
+)
+
+type Storage interface {
```
efectn, line 8: Why did you create new storage interface. I think you can make this like csrf, cache middlewares

leonklingele: We don't need the following `fiber.Storage` methods:

- `Delete(key string) error`
- `Reset() error`
- `Close() error`

All types satisfying the `idempotency.Storage` interface also implement the `fiber.Storage` interface.

efectn: Storage is generic interface for various fiber middlewares. I think we shouldnt create new interface. I can't see any desadvantage of using fiber.Storage. We should make code as simple as possible and we should avoid much copying i think.

GalvinGao: Would suggest the same here, and maybe also document it as something like:

> Those methods are unused so simply put a `panic("unexpected call")` would be sufficient.

GalvinGao (changes requested):
I would argue the same as in the conversation https://github.com/gofiber/fiber/pull/2253/files#r1038919642 to have consistency with `fiber.Storage`. Otherwise wonderful work!

efectn:
Why did you target `v3-beta` branch? I think this can be a part of v2.

Also apply PR template into your PRs please.

leonklingele:
@efectn I'll backport this to v2 once merged.

middleware/idempotency/storage.go
leonklingele: Using a `fiber.Storage` now.

leonklingele:
Soo, finally this is now ready for review. I removed the `Storage` interface, switched to `msgp` for marshalling, and improved some other things.

leonklingele:
Tests seem to be flaky: https://github.com/gofiber/fiber/actions/runs/3615636468/jobs/6092857426
Is this a known issue?

leonklingele:
Another, maybe different failure: https://github.com/gofiber/fiber/actions/runs/3615669563/jobs/6092912570

middleware/idempotency/idempotency.go
```diff
@@ -0,0 +1,153 @@
...
+		if err := c.Next(); err != nil {
+			// If the request handler returned an error, return it and skip idempotency
+			return err
+		}
+
+		// Construct response
+		res := &Response{
+			StatusCode: c.Response().StatusCode(),
+
+			Body: c.Response().Body(),
```
leonklingele, line 116: I wonder if we need to use `utils.CopyBytes` here.

efectn:
Can you change branch? Then, i'll review it

leonklingele:
Let's make this go in v3 first and I'll open up another PR targeting v2 later on.

leonklingele:
Anyone gotten the chance to review this already? I'm working on a project which might benefit from having this.

ReneWerner87:
Will do it tomorrow

ReneWerner87 (approved):
implementation looks very good

a few small comments/improvements

- benchmark for the middleware usage would be good, so we can improve it later on
- instead of the clousure i would maybe try to use a function that doesn't need to be created on-the-fly every time
- with the locals thing i'm not quite sure, what happens e.g. if you define the idempotency middleware multiple times, normally there shouldn't be any collisions as they should be used on different routes then ... currently quite not sure with that
- in some places we can optimize and save allocations, like in the ToLower( there exists a fn which does not cause allocations) and other places where only comparisons are done
-> for comparisons no copies need to be made, only if the data should be kept

ReneWerner87:
@leonklingele will wait for the others for the feedback and then i merge it

@efectn @GalvinGao do you have any other points, otherwise i would merge the pull request

leonklingele:
Thanks for the review! I’ll push the updates later today.

Which calls can be optimized? I’ve only found the single call to strings.ToLower used when doing the map lookup which we could replace by utils.ToLower

Also, did you see my question regarding using utils.CopyBytes when storing the request body? I think we need to use it.

ReneWerner87:
Yes use the copy there, i was not sure about this

GalvinGao (approved):
lgtm!

leonklingele:
I've rebased & addressed @ReneWerner87's points. ptal

efectn (approved):

ReneWerner87 (approved):

PR #2273, opened 2022-12-17
:sparkles: Feature: Add Protobuf method to ctx

tomasweigenast:
## Description

I was working on a Fiber fork for my server, and I've added a method to `Ctx` to allow sending Protobuf messages. I thought it was nice to include in the next release for any who want to support Protobuf encoding.  

## Type of change

## Checklist:


efectn:
Hi @tomasweigenast . Thanks for the PR but we don't want to add extra dependencies into the core, we should avoid bloats as possible as. 

Maybe you should add something useful into the our https://github.com/gofiber/contrib repo.

PR #2296, opened 2023-01-14
Fix: logger color output

cloudwindy:
## Description

Fixes logger color output

## Type of change

## Checklist:

- [X] For new code I have written benchmarks so that they can be analyzed and improved

## Commit formatting:

### DIFF
--- a/middleware/logger/config.go
+++ b/middleware/logger/config.go
@@ -136,7 +136,7 @@ func configDefault(config ...Config) Config {
 	}
 
 	// Enable colors if no custom format or output is given
-	if cfg.Output == nil && checkColorEnable(cfg.Format) {
+	if cfg.Output == ConfigDefault.Output && checkColorEnable(cfg.Format) {
 		cfg.enableColors = true
 	}
 


ReneWerner87 (changes requested):
Do not think that this change fixes a problem
Since I can also unfortunately not get any details from the description

cloudwindy:
This fixes the problem that because of the following code, the logger middleware does not show colored output in any cases.
```
if cfg.Output == nil {
	cfg.Output = ConfigDefault.Output
}

// Enable colors if no custom format or output is given
if cfg.Output == nil && checkColorEnable(cfg.Format) {
	cfg.enableColors = true
}
```
Since cfg.Output will be ConfigDefault.Output in case of it is nil, it does not enable colors.

ReneWerner87:
Okay I see, we will see if the output writer is the default one

Can you add a comment for this logic
// check for std streams

ReneWerner87 (approved):

li-jin-gou:
lgtm

Issue #2254, opened 2022-12-03
🚀 Feature: Make all tests run using "t.Parallel()"


leonklingele:
### Feature Description

In order to speed up test execution, all test funcs should make a call to `t.Parallel()`. This should be safe to implement as I have seen no tests where this might result in a data race so far.

### Checklist:

- [X] I agree to follow Fiber's [Code of Conduct](https://github.com/gofiber/fiber/blob/master/.github/CODE_OF_CONDUCT.md).
- [X] I have checked for existing issues that describe my suggestion prior to opening this one.
- [X] I understand that improperly formatted feature requests may be closed without explanation.

gaby:
Excellent idea, I can work on this one.

PR #2299, opened 2023-01-15
:broom: chore: make most tests parallel

efectn:
## Description
Fixes https://github.com/gofiber/fiber/issues/2254
Note: Some of tests not still parallel since they have some problems. Maybe we refactor them in the future.

## Type of change
- [x] Just test refactoring

## Checklist:

## Commit formatting:

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -51,6 +51,7 @@ func testErrorResponse(t *testing.T, err error, resp *http.Response, expectedBod
 }
 
 func Test_App_MethodNotAllowed(t *testing.T) {
+	t.Parallel()
 	app := New()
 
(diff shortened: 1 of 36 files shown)


gaby:
I started working on #2254 but tests were failing randomly when using `t.Parallel()` in all of them.

efectn:
> I started working on #2254 but tests were failing randomly when using `t.Parallel()` in all of them.

Yeah. Some tests -especially prefork tests- not working with parallel. I think we don't need to run them parallel. Currently tests are already faster than master.

gaby (approved):
Looks good!

li-jin-gou (approved):

PR #2315, opened 2023-01-29
🚀 Feature: Update golang.org/x/sys to tagged release

leonklingele:
Dependabot doesn't report updates to untagged commits which are referred to by their commit hash.


leonklingele:
The Go 1.16 support is preventing this, unfortunately: `note: module requires Go 1.17`

PR #2316, opened 2023-01-29
openssl rand -base64 32

meehow:
It's simpler to generate encryption key on CLI

### DIFF
--- a/middleware/encryptcookie/README.md
+++ b/middleware/encryptcookie/README.md
@@ -70,7 +70,7 @@ type Config struct {
 	// Base64 encoded unique key to encode & decode cookies.
 	//
 	// Required. Key length should be 32 characters.
-	// You may use `encryptcookie.GenerateKey()` to generate a new key.
+	// You may run `openssl rand -base64 32` or use `encryptcookie.GenerateKey()` to generate a new key.
 	Key string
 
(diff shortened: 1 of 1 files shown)


efectn (approved):

efectn:
Don't forget to change https://github.com/gofiber/fiber/blob/master/middleware/encryptcookie/config.go#L22

And please create a PR for https://github.com/gofiber/docs repo

middleware/encryptcookie/README.md
```diff
@@ -70,7 +70,7 @@ type Config struct {
 	// Base64 encoded unique key to encode & decode cookies.
 	//
 	// Required. Key length should be 32 characters.
```
leonklingele, line 72: Can you update this comment to mention that the key length should be 32 bytes _after_ base64-decoding:

`Required. The key should be 32 bytes of random data in base64-encoded form.`

leonklingele: Should be done in other parts of the README and the code itself as well.

ReneWerner87:
```suggestion
	// Required. The key should be 32 bytes of random data in base64-encoded form.
```

ReneWerner87 (approved):

PR #2317, opened 2023-01-29
revert: pr 2286

li-jin-gou:
## Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change. 
Explain the *details* for making this change. What existing problem does the pull request solve?

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
## Checklist:

- [ ] For new functionalities I follow the inspiration of the express js framework and built them similar in usage
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation - https://github.com/gofiber/docs for https://docs.gofiber.io/
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] If new dependencies exist, I have checked that they are really necessary and agreed with the maintainers/community (we want to have as few dependencies as possible)
- [ ] I tried to make my code as fast as possible with as few allocations as possible
- [ ] For new code I have written benchmarks so that they can be analyzed and improved

## Commit formatting:

Use emojis on commit messages so it provides an easy way of identifying the purpose or intention of a commit. Check out the emoji cheatsheet here: https://gitmoji.carloscuesta.me/

PR #2319, opened 2023-02-01
🐛 Bug: Fix issues introduced in linting PR

leonklingele:
## Description

Fix issues introduced in https://github.com/gofiber/fiber/pull/2286.
See the individual commit messages for details.

Refs https://github.com/gofiber/fiber/pull/2286#issuecomment-1407566424 (et al.) and https://github.com/gofiber/fiber/pull/2317.

## Type of change

## Checklist:


internal/dictpool/pool.go
```diff
@@ -10,7 +10,7 @@ var defaultPool = sync.Pool{
 
 // AcquireDict acquire new dict.
 func AcquireDict() *Dict {
-	return defaultPool.Get().(*Dict)
+	return defaultPool.Get().(*Dict) // nolint:forcetypeassert
```
renanbastos93, line 13: Oops, why do you ignore lint here? Another point, case it was nil here will it raise a panic?

internal/gopsutil/mem/mem.go
```diff
@@ -6,7 +6,8 @@ import (
 	"github.com/gofiber/fiber/v2/internal/gopsutil/common"
 )
 
-var invoke common.Invoker = common.Invoke{} //nolint:unused // We use this only for some OS'es
+//lint:ignore U1000 we need this elsewhere
```
renanbastos93, line 9: 🤔

internal/gopsutil/mem/mem_freebsd.go
```diff
@@ -86,6 +86,7 @@ func SwapMemory() (*SwapMemoryStat, error) {
 }
 
 // Constants from vm/vm_param.h
+// nolint: golint
```
renanbastos93, line 89: 🤔

internal/storage/memory/memory_test.go
```diff
@@ -149,13 +149,13 @@ func Benchmark_Storage_Memory(b *testing.B) {
 		b.ResetTimer()
 		for n := 0; n < b.N; n++ {
 			for _, key := range keys {
-				_ = d.Set(key, value, ttl)
+				d.Set(key, value, ttl)
```
renanbastos93, line 152: Doesn't it return an error here? I prefer it to be explicit that you ignore returns using underline because it stays readable and creates curiosity when to analysis tests.

internal/template/html/html.go
```diff
@@ -156,6 +156,7 @@ func (e *Engine) Load() error {
 		name = strings.TrimSuffix(name, e.extension)
 		// name = strings.Replace(name, e.extension, "", -1)
 		// Read the file
+		// #gosec G304
```
renanbastos93, line 159: 🆗

middleware/monitor/monitor.go
```diff
@@ -105,34 +105,34 @@ func New(config ...Config) fiber.Handler {
 
 func updateStatistics(p *process.Process) {
 	pidCPU, err := p.CPUPercent()
-	if err != nil {
+	if err == nil {
```
renanbastos93, line 108: As you changed it here. I suggest you make the comment below in line 109.

renanbastos93 (commented):
Fantastic, you did a good job dude. Even thou, I believe that's need to do little things, also we need to create unit tests to pkg `internal/dictpool`.

leonklingele:
This change is reverting all the good things we did for internal. We should not have touched it as it’s not our code. 💔

gaby:
@leonklingele I'm about to test the `monitor`. My main concern is with the fields used in `index.go` which are getting bypassed by all linters since it's all a string.

gaby:
The monitor middleware is working again:

image

Tested using the following:

```go
package main

import (
	"log"
	"time"

	"github.com/gofiber/fiber/v2"
	"github.com/gofiber/fiber/v2/middleware/monitor"
)

func main() {
	app := fiber.New()

	var myCustomConfig = monitor.Config{
		Title:      "Fiber Monitor Test",
		Refresh:    3 * time.Second,
	}

	app.Get("/metrics", monitor.New(myCustomConfig))
}
```

internal/dictpool/pool.go
gaby: @renanbastos93 The internal directory contains third-party libraries that were edited by a previous PR. This PR is reverting those changes.

internal/gopsutil/common/common.go
```diff
@@ -366,7 +366,7 @@ func HostDev(combineWith ...string) string {
 // getSysctrlEnv sets LC_ALL=C in a list of env vars for use when running
 // sysctl commands (see DoSysctrl).
 func getSysctrlEnv(env []string) []string {
-	var foundLC bool
```
gaby, line 369: @leonklingele Is here a way to just ignore all of `internal/gopsutil` directory via the new lint tool? Instead of having to add all these  `lint:ignore` directives?

gaby (approved):
So far the monitor middleware looks good!

ReneWerner87:
tasks:

- [ x ] fix for the monitor mw
- [ ] maybe some nolint stuff directly in the yaml and remove the comments
- [ ] change the ChartJsUrl back rename
- [ ] changes to the internal code copies back
- [ ] ignore internal copies for the linting

ReneWerner87 (changes requested):
pls check the task list

efectn:
I think we should prefer go.mod for dictpool, uuid, msgp deps. Gopsutil's current version has so much deps, i think it should stay as vendored dep.

ReneWerner87:
> I think we should prefer go.mod for dictpool, uuid, msgp deps. Gopsutil's current version has so much deps, i think it should stay as vendored dep.

ok, will try to implement it in this PR

ReneWerner87 (approved):

efectn:
We can remove https://github.com/leonklingele/fiber/tree/fixup-linting/internal/fwd since it's added as indirect dep.

efectn (approved):

PR #2325, opened 2023-02-03
use goccy/go-json

kokizzu:
## Description

replace encoding/json with goccy/go-json, because it's the only one that fastest but also truly compatible with encoding/json

## Type of change

- [x] Performance optimization

## Checklist:

- [x] For new code I have written benchmarks so that they can be analyzed and improved:  https://github.com/kokizzu/kokizzu-benchmark/tree/master/ser-deser

## Commit formatting:

### DIFF
--- a/app.go
+++ b/app.go
@@ -10,7 +10,6 @@ package fiber
 import (
 	"bufio"
 	"context"
-	"encoding/json"
 	"encoding/xml"
 	"errors"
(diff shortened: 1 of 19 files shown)


efectn:
Here's why we don't use any external json library as default -> https://docs.gofiber.io/guide/faster-fiber

PR #2306, opened 2023-01-19
🔥 write integer Query Parser.

dozheiny:
This pull request adds a new feature to the codebase - an Integer Query URL Parser. This parser extracts all integers found in the query parameters of a given URL. The parser takes in a single argument, the URL, and returns a list of integers found in the query parameters. This feature can be useful for extracting and analyzing data from web analytics and other online tracking systems. The code has been tested and is ready for review and integration into the main branch.

PR #2328, opened 2023-02-09
🔥 Feature: add queryFloat parser

dozheiny:
Hi!
Before this PR, I wrote this feature for integer variable #2306.
Now I wrote this feature for float64 variables,
With this parser, you can easily extract float values from URL queries. Pass the desired key and the URL string to the parser, and it will return the float value associated with that key. This eliminates the need for manual string parsing and makes retrieving the values you need easier.

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -1126,6 +1126,28 @@ func (c *Ctx) QueryInt(key string, defaultValue ...int) int {
 	return value
 }
 
+// QueryFloat returns float64 value of key string parameter in the url.
+// Default to empty or invalid key is 0.
+//
+//	GET /?name=alex&amount=32.23&id=
+//	QueryFloat("amount") = 32.23
+//	QueryFloat("amount", 3) = 32.23
+//	QueryFloat("name", 1) = 1
+//	QueryFloat("name") = 0
+//	QueryFloat("id", 3) = 3
+func (c *Ctx) QueryFloat(key string, defaultValue ...float64) float64 {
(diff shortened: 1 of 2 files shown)


gaby (approved):
LGTM

ctx.go
```diff
@@ -1126,6 +1126,28 @@ func (c *Ctx) QueryInt(key string, defaultValue ...int) int {
...
+//
+//	GET /?name=alex&amount=32.23&id=
+//	QueryFloat("amount") = 32.23
+//	QueryFloat("amount", 3) = 32.23
+//	QueryFloat("name", 1) = 1
+//	QueryFloat("name") = 0
+//	QueryFloat("id", 3) = 3
+func (c *Ctx) QueryFloat(key string, defaultValue ...float64) float64 {
+	// use strconv.ParseFloat to convert the param to a float or return zero and an error.
+	value, err := strconv.ParseFloat(c.app.getString(c.fasthttp.QueryArgs().Peek(key)), 64)
```
li-jin-gou, line 1140: There seems to be no need to use c.app.getString here because strconv.ParseFloat is already being used

li-jin-gou (approved):

ctx.go
dozheiny: Need to remove c.app.getString?

li-jin-gou: > Need to remove c.app.getString?

No need

PR #2329, opened 2023-02-09
🔥 Feature: add queryBool parser

dozheiny:
Hi!
Before this PR, I wrote this feature for integer variables https://github.com/gofiber/fiber/pull/2306 And float variables #2328.
Now I wrote this feature for boolean variables,
With this parser, you can easily extract bool values from URL queries. You can pass the desired key and the URL string to the parser, which will return the boolean value associated with that key. This eliminates manual string parsing and makes retrieving the values you need easier.

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -1142,10 +1142,8 @@ func (c *Ctx) QueryBool(key string, defaultValue ...bool) bool {
 		if len(defaultValue) > 0 {
 			return defaultValue[0]
 		}
-
 		return true
 	}
-
 	return value
 }
 
@@ -1165,10 +1163,8 @@ func (c *Ctx) QueryFloat(key string, defaultValue ...float64) float64 {
 		if len(defaultValue) > 0 {
 			return defaultValue[0]
 		}
-
 		return 0
 	}
-
 	return value
 }
 
--- a/ctx.go
+++ b/ctx.go
@@ -1126,6 +1126,29 @@ func (c *Ctx) QueryInt(key string, defaultValue ...int) int {
 	return value
 }
 
+// QueryBool returns bool value of key string parameter in the url.
+// Default to empty or invalid key is true.
+//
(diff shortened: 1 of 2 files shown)


li-jin-gou:
Next time PR can be merged, as the functions are similar

dozheiny:
> Next time PR can be merged, as the functions are similar

Yes, you are right. I will resolve conflict files.

dozheiny:
I don't know why PR closed; I think because pulled resolved conflicts.

gaby (approved):
LGTM

li-jin-gou (approved):

crdx:
This may not be the right place for this but I didn't want to clog up the issues list with a simple question.

What's the reason for `QueryBool()` returning `true` for a non-existent key? Since the default value of a bool in Go is `false`, I would have expected it to be `false` here too.

dozheiny:
Most frameworks (in other languages, not just in Go) will calculate the boolean value `true` actual if it is a string.

crdx:
Yes, in the string case `true` is returned and that's expected. I'm talking about in the case of the parameter not existing at all. [The documentation for QueryBool](https://docs.gofiber.io/api/ctx/#querybool) has a warning saying this, as well:

> Please note if that parameter is not in the request, true will be returned.

Compare to [QueryFloat](https://docs.gofiber.io/api/ctx/#queryfloat):

> Please note if that parameter is not in the request, zero will be returned.

`0` is the default value for a float, so that's what is returned. For a bool `false` is the default value, so shouldn't it be what's returned? (To me it makes as much sense as `QueryFloat` returning an arbitrary value like 123 if the key doesn't exist.)

I can't think of a situation where you want to query a non-existent key and get `true` back if it doesn't exist... what am I missing here?

dozheiny:
No. I think you are right.

dozheiny:
Sure, I'll do it.

PR #2351, opened 2023-03-01
allow preloaded certs with prefork

lublak:
## Description

Currently I have a function that generates or loads these keys according to different conditions.
Now I would have to save them temporarily as a file to be able to use the full functions of fiber.
The added functions allow me to specify the already loaded keys.
## Type of change

## Checklist:

## Commit formatting:

Oh! Its too late :c should i recreate my pull request?

### DIFF
--- a/listen.go
+++ b/listen.go
@@ -98,6 +98,14 @@ func (app *App) ListenTLS(addr, certFile, keyFile string) error {
 		return fmt.Errorf("tls: cannot load TLS key pair from certFile=%q and keyFile=%q: %w", certFile, keyFile, err)
 	}
 
+	return app.ListenTLSWithCertificate(addr, cert)
+}
+
+// ListenTLS serves HTTPS requests from the given addr.
+// cert is a tls.Certificate
+//
+//	app.ListenTLSWithCertificate(":8080", cert)
+func (app *App) ListenTLSWithCertificate(addr string, cert tls.Certificate) error {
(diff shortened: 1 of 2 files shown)


gaby (changes requested):
Can you add unit-tests to the new exported functions?

listen.go
```diff
@@ -161,6 +169,10 @@ func (app *App) ListenMutualTLS(addr, certFile, keyFile, clientCertFile string)
 	clientCertPool := x509.NewCertPool()
 	clientCertPool.AppendCertsFromPEM(clientCACert)
 
+	return app.ListenMutualTLSWithCertificate(addr, cert, clientCertPool)
+}
+
+func (app *App) ListenMutualTLSWithCertificate(addr string, cert tls.Certificate, clientCertPool *x509.CertPool) error {
```
gaby, line 175: Missing function comment here

lublak:
> Can you add unit-tests to the new exported functions?

does it make sense? I mean there are already unit tests for ListenMutual. And ListenMutual calls ListenMutualTLSWithCertificate with already loaded file. If i create a unit test for ListenMutualTLSWithCertificate  i just copy the content from ListenMutual to a unit test.

gaby:
> 

I think having an exported function without tests will lower the coverage % 🤔

lublak:
@gaby i created tests for it.
and true. Unit tests unfortunately do not know that it was actually already tested with other function.

PR #2374, opened 2023-03-15
:broom: chore: drop go 1.16 support & update dependencies

efectn:
## Description
1.16 is no longer supported by Fasthttp, let's drop it.

## Type of change

## Checklist:

## Commit formatting:


gaby:
Go `1.17` is EOL since August 2022, would it make more sense to make the jump to `1.18` as minimum like the contrib/storage middlewares?

https://endoflife.date/go

gaby (approved):

prefork_test.go
```diff
@@ -71,7 +71,6 @@ func Test_App_Prefork_Master_Process(t *testing.T) {
 }
 
 func Test_App_Prefork_Child_Process_Never_Show_Startup_Message(t *testing.T) {
-	t.Parallel()
```
leonklingele, line 74: This is most likely an unrelated change and should be reverted.

gaby:
> Go `1.17` is EOL since August 2022, would it make more sense to make the jump to `1.18` as minimum like the contrib/storage middlewares?
> 
> https://endoflife.date/go

I also noticed that action/setup-go tags stable as 1.19 and oldstable as 1.18. We can probably add 1.19 to the test suite. 

Src: https://github.com/actions/setup-go#using-stableoldstable-aliases

ReneWerner87:
> Go `1.17` is EOL since August 2022, would it make more sense to make the jump to `1.18` as minimum like the contrib/storage middlewares?
> 
> https://endoflife.date/go

we always support the latest version which supports fasthttp, in the new version of fiber we already make this jumpwe always support the latest version which supports fasthttp, in the new version of fiber we already make this jump

prefork_test.go
efectn: Just try to fix data race on tests

ReneWerner87 (approved):

PR #2385, opened 2023-03-26
docs: added code link to fiber config fields

shahriarsohan:
## Description

When reading the docs , It was hard to see what Colors and DefaultColors really mean in [Fiber.New()](https://docs.gofiber.io/api/fiber#config) config fields. I added direct code link, now we can click that and see the code.

## Checklist:


leonklingele:
Hey @shahriarsohan, thanks for providing this!
Please update the links so they don’t refer to the master branch. When we decide to update the file in the future, the line reference will no longer be correct.
Probably best to link to the latest tag instead (which pretty likely won’t change).

shahriarsohan:
Hey @leonklingele . It seems after every change in `color.go`, those links will become invalid. But now i added the permalink . It will always be the same because it referring to the current commit.

In Future, If we specifically update `Colors` and `DefaultColors`  then we just have to update those links.

Thanks

leonklingele:
hmm, why not link to the file without any line references? Would allow us to point to master and an always up-to-date version.

shahriarsohan:
That also make sense because `color.go` is a small file. I am updating..

shahriarsohan:
@leonklingele any update on this?

leonklingele:
@shahriarsohan it seems the API docs are autogenerated nowadays, so you need to make this change at another place (the Markdown file which you modified will get overwritten and lose your changes after the next auto docs update).

Instead, do your change here: https://github.com/gofiber/fiber/blob/e9541945e553268e3d17b3b7276218ae0b9a0eef/app.go#L384-L392

And use this kind of reference style:

https://github.com/fluhus/godoc-tricks/blob/4591e94c7b265f433db7aa658fa914beb532b9af/doc.go#L262
and
https://github.com/fluhus/godoc-tricks/blob/4591e94c7b265f433db7aa658fa914beb532b9af/doc.go#L276

ReneWerner87 (approved):

leonklingele (approved):
Sorry for the confusing, was under the impression we auto-gen the docs nowadays.

I've opened https://github.com/gofiber/fiber/issues/2401 as a feature request for this.

PR #2386, opened 2023-03-27
Fix #2383, accepts mimeType

### LINKED ISSUE #2383: 🐛 [Bug]: ctx.Accepts("text", "json") with headerAccept, "text/*, application/json" fails on macOS
### Bug Description

ctx.Accepts("text", "json") with headerAccept, "text/*, application/json" is not the same on macOS and linux

On macOS it returns "text"

On linux it returns "json"

### How to Reproduce

Run test on macOS and linux

### Expected Behavior

consistent return and tests pass

### Fiber Version

v2.43.0

### Code Snippet (optional)

```go
func Test_Ctx_Accepts(t *testing.T) {
	t.Parallel()
	app := New()
	defer app.ReleaseCtx(c)

	c.Request().Header.Set(HeaderAccept, "text/*, application/json")
	utils.AssertEqual(t, "json", c.Accepts("json", "text"))
}

// called with above params DEBUG comments added
func (c *Ctx) Accepts(offers ...string) string {
	if len(offers) == 0 {
		return ""
	}
	header := c.Get(HeaderAccept)
	if header == "" {
		return offers[0]
	}

	spec, commaPos := "", 0
	for len(header) > 0 && commaPos != -1 {
		commaPos = strings.IndexByte(header, ',')
		if commaPos != -1 {
			spec = utils.Trim(header[:commaPos], ' ')
		} else {
			spec = utils.TrimLeft(header, ' ')
		}
		if factorSign := strings.IndexByte(spec, ';'); factorSign != -1 {
			spec = spec[:factorSign]
		}

		var mimetype string
		for _, offer := range offers { // DEBUG first loop "text"
			if len(offer) == 0 {
				continue
				// Accept: */*
			} else if spec == "*/*" {
				return offer
			}

			if strings.IndexByte(offer, '/') != -1 {
				mimetype = offer // MIME type
			} else {
				mimetype = utils.GetMIME(offer) // DEBUG on linux: "application/octet-stream", on macOS: "text/plain; charset=utf-8"
			}

			if spec == mimetype {
				// Accept: <MIME_type>/<MIME_subtype>
				return offer
			}

			s := strings.IndexByte(mimetype, '/')
			// Accept: <MIME_type>/*
			if strings.HasPrefix(spec, mimetype[:s]) && (spec[s:] == "/*" || mimetype[s:] == "/*") { // DEBUG first loop mimeType = "text/*"
				return offer // DEBUG on macOS text spec matches the string before / (text), on linux text does not match string before / (application)
			}
		}
		if commaPos != -1 {
			header = header[commaPos+1:]
		}
	}

	return ""
}
```

### Further context

Caused by: utils.GetMIME("text"), which calls mime.TypeByExtension("." + extension) on macOS returns "text/plain; charset=utf-8" while on linux it returns ""

### Checklist:

ReneWerner87:
## Description

Correct the invalid behavior of the `ctx.Accepts` method

Fixes #2383

## Type of change

## Checklist:

### DIFF
--- a/helpers.go
+++ b/helpers.go
@@ -215,36 +215,82 @@ func getGroupPath(prefix, path string) string {
 	return utils.TrimRight(prefix, '/') + path
 }
 
-// return valid offer for header negotiation
-func getOffer(header string, offers ...string) string {
+// acceptsOffer This function determines if an offer matches a given specification.
+// It checks if the specification ends with a '*' or if the offer has the prefix of the specification.
+// Returns true if the offer matches the specification, false otherwise.
+func acceptsOffer(spec string, offer string) bool {
+	if len(spec) >= 1 && spec[len(spec)-1] == '*' {
+		return true
+	} else if strings.HasPrefix(spec, offer) {
+		return true
+	}
+	return false
+}
+
+// acceptsOfferType This function determines if an offer type matches a given specification.
+// It checks if the specification is equal to */* (i.e., all types are accepted).
+// It gets the MIME type of the offer (either from the offer itself or by its file extension).
+// It checks if the offer MIME type matches the specification MIME type or if the specification is of the form <MIME_type>/* and the offer MIME type has the same MIME type.
+// Returns true if the offer type matches the specification, false otherwise.
+func acceptsOfferType(spec string, offerType string) bool {
+	// Accept: */*
+	if spec == "*/*" {
+		return true
+	}
+
+	var mimetype string
+	if strings.IndexByte(offerType, '/') != -1 {
+		mimetype = offerType // MIME type
+	} else {
+		mimetype = utils.GetMIME(offerType) // extension
+	}
+
+	if spec == mimetype {
+		// Accept: <MIME_type>/<MIME_subtype>
+		return true
+	}
+
+	s := strings.IndexByte(mimetype, '/')
+	// Accept: <MIME_type>/*
+	if strings.HasPrefix(spec, mimetype[:s]) && (spec[s:] == "/*" || mimetype[s:] == "/*") {
+		return true
+	}
+
+	return false
+}
+
+// getOffer return valid offer for header negotiation
+func getOffer(header string, isAccepted func(spec string, offer string) bool, offers ...string) string {
 	if len(offers) == 0 {
 		return ""
 	} else if header == "" {
 		return offers[0]
 	}
 
-	spec, commaPos := "", 0
-	for len(header) > 0 && commaPos != -1 {
-		commaPos = strings.IndexByte(header, ',')
-		if commaPos != -1 {
-			spec = utils.Trim(header[:commaPos], ' ')
-		} else {
-			spec = header
+	for _, offer := range offers {
+		if len(offer) == 0 {
+			continue
 		}
-		if factorSign := strings.IndexByte(spec, ';'); factorSign != -1 {
-			spec = spec[:factorSign]
-		}
+		spec, commaPos := "", 0
+		for len(header) > 0 && commaPos != -1 {
+			commaPos = strings.IndexByte(header, ',')
+			if commaPos != -1 {
+				spec = utils.Trim(header[:commaPos], ' ')
+			} else {
+				spec = utils.TrimLeft(header, ' ')
+			}
+			if factorSign := strings.IndexByte(spec, ';'); factorSign != -1 {
+				spec = spec[:factorSign]
+			}
 
-		for _, offer := range offers {
-			// has star prefix
-			if len(spec) >= 1 && spec[len(spec)-1] == '*' {
-				return offer
-			} else if strings.HasPrefix(spec, offer) {
+			// isAccepted if the current offer is accepted
+			if isAccepted(spec, offer) {
 				return offer
 			}
-		}
-		if commaPos != -1 {
-			header = header[commaPos+1:]
+
+			if commaPos != -1 {
+				header = header[commaPos+1:]
+			}
 		}
 	}
 
--- a/helpers.go
+++ b/helpers.go
@@ -218,7 +218,7 @@ func getGroupPath(prefix, path string) string {
 // acceptsOffer This function determines if an offer matches a given specification.
 // It checks if the specification ends with a '*' or if the offer has the prefix of the specification.
 // Returns true if the offer matches the specification, false otherwise.
-func acceptsOffer(spec string, offer string) bool {
+func acceptsOffer(spec, offer string) bool {
 	if len(spec) >= 1 && spec[len(spec)-1] == '*' {
 		return true
 	} else if strings.HasPrefix(spec, offer) {
@@ -232,7 +232,7 @@ func acceptsOffer(spec string, offer string) bool {
 // It gets the MIME type of the offer (either from the offer itself or by its file extension).
 // It checks if the offer MIME type matches the specification MIME type or if the specification is of the form <MIME_type>/* and the offer MIME type has the same MIME type.
 // Returns true if the offer type matches the specification, false otherwise.
-func acceptsOfferType(spec string, offerType string) bool {
+func acceptsOfferType(spec, offerType string) bool {
 	// Accept: */*
 	if spec == "*/*" {
 		return true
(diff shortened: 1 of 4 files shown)


ReneWerner87:
current master:

```txt
Benchmark_Ctx_Accepts/run-[]string{".xml"}-2            	 	 	         7432663	        162.4 ns/op	       0 B/op	       0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{"json",_"xml"}-2     	                         5298415	        226.2 ns/op	       0 B/op	       0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{"application/json",_"application/xml"}-2         	 8494714	        141.1 ns/op	       0 B/op	       0 allocs/op
Benchmark_Ctx_AcceptsCharsets-2                                                     	19419148	        61.88 ns/op	       0 B/op	       0 allocs/op
Benchmark_Ctx_AcceptsEncodings-2                                                    	15028786	        79.98 ns/op	       0 B/op	       0 allocs/op
Benchmark_Ctx_AcceptsLanguages-2                                                    	18556627	        62.99 ns/op	       0 B/op	       0 allocs/op
```

ReneWerner87:
i have reversed the loop for all accepts methods, so that the arguments are passed first and then the header values are checked

@efectn @sixcolors @derkan 
do you agree that this was the right thing to do ? the logic of express leading to the outputs implies this to me

ReneWerner87:
after the fix:
```txt
Benchmark_Ctx_Accepts/run-[]string{".xml"}-2            	                         7367316	        162.8 ns/op	       0 B/op	       0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{"json",_"xml"}-2     	                         6219842	        192.5 ns/op	       0 B/op	       0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{"application/json",_"application/xml"}-2         	 8405518	        142.6 ns/op	       0 B/op	       0 allocs/op
Benchmark_Ctx_AcceptsCharsets-2                                                     	19289839	        62.33 ns/op	       0 B/op	       0 allocs/op
Benchmark_Ctx_AcceptsEncodings-2                                                    	14561808	        82.07 ns/op	       0 B/op	       0 allocs/op
Benchmark_Ctx_AcceptsLanguages-2                                                    	18876432	        63.73 ns/op	       0 B/op	       0 allocs/op
```

helpers.go
```diff
@@ -215,36 +215,82 @@ func getGroupPath(prefix, path string) string {
...
 
-// return valid offer for header negotiation
-func getOffer(header string, offers ...string) string {
+// acceptsOffer This function determines if an offer matches a given specification.
+// It checks if the specification ends with a '*' or if the offer has the prefix of the specification.
+// Returns true if the offer matches the specification, false otherwise.
+func acceptsOffer(spec, offer string) bool {
+	if len(spec) >= 1 && spec[len(spec)-1] == '*' {
+		return true
+	} else if strings.HasPrefix(spec, offer) {
```
renanbastos93, line 224: I believe that can we change that about one line. What do you think?

```suggestion
	if (len(spec) >= 1 && spec[len(spec)-1] == '*') || strings.HasPrefix(spec, offer) {
		return true
```

renanbastos93 (approved):
It's amazing, you made a good job, congrats dude.

sixcolors:
> i have reversed the loop for all accepts methods, so that the arguments are passed first and then the header values are checked
> 
> @efectn @sixcolors @derkan do you agree that this was the right thing to do ? the logic of express leading to the outputs implies this to me

@ReneWerner87 good job, this make the existing test cases behave as expected.

Reading the ctx.Accepts docs my understanding is that:

```
The Accepts function in the GoFiber context is used to check if the specified extensions or content types are acceptable based on the request's Accept HTTP header. It takes one or more strings as input arguments representing the content types or extensions and returns the most preferred one based on the Accept header.
```

Therefore, if I understand what the Accept function does, the following should hold true:

```
c.Request().Header.Set(HeaderAccept, "text/plain, application/json;q=0.5, text/html, */*;q=0.1")
	utils.AssertEqual(t, "txt", c.Accepts("json", "txt"))

	c.Request().Header.Set(HeaderAccept, "text/plain;charset=UTF-8;q=0.8,text/plain;charset=US-ASCII;q=0.4,text/plain;q=0.2,text/html")
	utils.AssertEqual(t, "text/plain;charset=US-ASCII", c.Accepts("text/plain", "text/plain;charset=US-ASCII"))
	utils.AssertEqual(t, "text/plain;charset=UTF-8", c.Accepts("text/plain;charset=US-ASCII", "text/plain;charset=UTF-8", "text/plain"))
	utils.AssertEqual(t, "text/html", c.Accepts("text/plain", "text/html"))
```

For example: It is possible that a browser could use the Accept header with the value text/plain;charset=UTF-8;q=0.8,text/plain;charset=US-ASCII;q=0.4,text/html. While text/html is typically assumed to have a default quality value of 1 if not specified, the browser could still explicitly specify it with a quality value of 1, as it is allowed by the HTTP specifications.

Furthermore, the browser could be indicating to the server that it prefers text/plain with a charset of UTF-8 over text/plain with a charset of US-ASCII, but it is still willing to accept both with different quality values. This could be useful in scenarios where the server has content in both character sets and wants to serve the content that is preferred by the client.

However, it is worth noting that the Accept header sent by the browser may vary depending on the user agent and its configuration. Therefore, it is important for the server to properly parse and interpret the Accept header to determine the preferred content type for the client.

## Background/Refs

The relevant RFCs for the HTTP/1.1 specification are RFC 7231 and RFC 7232.

Regarding the q parameter and its default value of 1, Section 5.3.1 of RFC 7231 states:

```
If no "*" is present in an Accept header field, then all
media types are acceptable, assuming that any media range
syntax is understood by the recipient.  If an Accept header
field is present, and if the server cannot send a response
which is acceptable according to the combined Accept field
value, then the server SHOULD send a 406 (not acceptable)
response.

[...]

Each media-range might be followed by one or more accept-
params, beginning with the "q" parameter for indicating a
relative quality factor.  The first "q" parameter (if any)
separates the media-range parameter(s) from the accept-
params.  Quality factors allow the user or user agent to
indicate the relative degree of preference for that media-
type, on a scale of 0 to 1.
```

Regarding the order of the MIME types in the Accept header, Section 5.3.2 of RFC 7231 states:

```
The order in which the content-codings are listed in the
header field is significant.  Typically, the first encoding
that is applied is the one listed as the first content-coding;
subsequent encodings are applied in the order listed.  The
order of charset values in the header field is significant,
whereas the order of language-tag values is not.  Specifically,
it does not mean that the first charset listed is preferred,
but rather that it represents the default charset if none is
otherwise specified.

An "Accept" header field can be present in a request to
indicate the formats that are acceptable for the response.
Field content negotiation is described in Section 5.3.  An
origin server that is capable of generating a response
message which is subject to content negotiation MAY vary the
response that is sent, based on the value of the Accept header
field supplied in the request.  For example, the server might
include a different Content-Type header field, or might
omit the content entirely.

The order in which the media type values are listed is
significant.  The first media type that is listed as acceptable
determines the format of the entity-body that is returned.  A
more precise or generally usable media type listed earlier in
the Accept header field SHOULD be given preference over those
that are listed later.
```

MDN on Quality Values https://developer.mozilla.org/en-US/docs/Glossary/Quality_values

```
The importance of a value is marked by the suffix ';q=' immediately followed by a value between 0 and 1 included, with up to three decimal digits, the highest value denoting the highest priority. When not present, the default value is 1.

[Examples](https://developer.mozilla.org/en-US/docs/Glossary/Quality_values#examples)
The following syntax

text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8

indicates the order of priority:

[...]

If there is no priority defined for the first two values, the order in the list is irrelevant. Nevertheless, with the same quality, more specific values have priority over less specific ones:

text/html;q=0.8,text/*;q=0.8,*/*;q=0.8

Value	Priority
text/html	0.8 (but totally specified)
text/*	0.8 (partially specified)
*/*	0.8 (not specified)

Some syntax, like the one of Accept, allow additional specifiers like text/html;level=1. These increase the specificity of the value. Their use is extremely rare.
```

helpers.go
ReneWerner87: yes good point, had only split it up for readability and would keep it that way as it is more understandable

sixcolors:
A full solution would likely involve a specificity function and sorting of a struct that looks something like:

```
type acceptedType struct {
	mime string
	q float64
	charset string
}
```

A performance cost seems likely.

Finally; If my understanding of ctx.Accpets() func is correct: May I suggest updating the docs with a similar text to my previous comment, so that there is no ambiguity?

sixcolors:
> are you ok with the current fix and you then provide a subsequent solution that takes into account the quantity parameter ? we should also test how expressjs behaves with these
> 
> new unittest etc.

Yes, the current PR is an improvement, so I'll approve and then submit a PR to deal with quality value and specificity.

PR #2391, opened 2023-03-30
💊 Change default value of Querybool from true to false.

dozheiny:
As discussed, we have [here](https://github.com/gofiber/fiber/pull/2329#issuecomment-1487527990); changed the default value from true to false for the QueryBool function.


ReneWerner87 (approved):

leonklingele:
This is a breaking (and severe) change btw.

ReneWerner87:
its okay, since it is only the default value for an edge case

ReneWerner87:
i will inform in the release notes, but will not make a big section for it

PR #2394, opened 2023-03-30
✨ feat(cors): Added new 'AllowOriginsFunc' function.

### LINKED ISSUE #2390: 🚀 [Feature]: Cors AllowOrigin func
### Feature Description

It's often helpful to set `Access-Control-Allow-Origin` response header value based on some conditions, for example, if the application is running in a test environment (specified by some arbitrary environment variable) then allow any port on localhost, you would be unable to use `*` as the value if you're also using `Access-Control-Allow-Credentials: true`.

In this scenario rather than having to specify every localhost with port origin you could potentially check whether the origin contains `localhost` which would return true and the implementation can then set the response header to the origin based on if this function returns true or false.

I was utilizing the fiber implementation of what `*` origin meant which was to not apply `*` but set whatever `Origin` header was set. This has been fixed as part of issue https://github.com/gofiber/fiber/issues/2338.

So I am proposing a way that could be used to set the `Access-Control-Allow-Origin` header to whatever the origin is via a function. I am aware Gin has the `AllowOriginFunc`.

The implementation of using both `AllowOrigins` and this function could either be the function overwrites whatever is stored within `AllowOrigins` or it performs the origin checks on values specified in `AllowOrigins` first and the origin does not match any of these values then run the function.

If this feature gets accepted I am happy to work on it.

### Code Snippet (optional)

```go
package main

import "github.com/gofiber/fiber/v2/middleware/cors"
import "log"

func main() {
  app := fiber.New()
  
  // Example 1 - check whether the origin is any scheme or port on localhost, this would result in any localhost app to access this app from the browser.
  app.Use(cors.New(cors.Config{
	AllowCredentials: true,
	AllowOriginsFunc: func(origin string) bool {
            return strings.Contains(origin, ":://localhost")
        }
  }))

  // Example 2 - check whether the environment variable is set to "development", this would result in any origins being able to access this app from the browser.
  app.Use(cors.New(cors.Config{
	AllowCredentials: true,
	AllowOriginsFunc: func(origin string) bool {
            return os.Getenv("ENVIRONMENT") == "development"
        }
  }))

}
```

In both examples the `Access-Control-Allow-Origin` response header would be set to whatever the `Origin` request header is.

### Checklist:

Jamess-Lucass:
## Description

Adds `AllowOriginsFunc` function to cors config which adds the ability to dynamically chose whether to allow the origin via CORS or not.

Fixes #2390

## Type of change

## Checklist:

## Commit formatting:


docs/api/middleware/cors.md
```diff
@@ -35,52 +35,72 @@ app.Use(cors.New(cors.Config{
 }))
```
leonklingele, line 35: btw: docs are auto-generated and shouldn't be modified by hand :)

leonklingele (changes requested):

docs/api/middleware/cors.md
ReneWerner87: will not be auto-generated, have already cleared it with him

middleware/cors/cors.go
```diff
@@ -126,6 +133,15 @@ func New(config ...Config) fiber.Handler {
...
 		}
 
+		// Run AllowOriginsFunc if the logic for
+		// handling the value in 'AllowOrigins' does
+		// not result in allowOrigin being set.
+		if allowOrigin == "" && cfg.AllowOriginsFunc != nil {
+			if cfg.AllowOriginsFunc(origin) {
+				allowOrigin = origin
+			}
+		}
```
ReneWerner87, line 143:
```suggestion
		if allowOrigin == "" && cfg.AllowOriginsFunc != nil && cfg.AllowOriginsFunc(origin) {
				allowOrigin = origin
		}
```

docs/api/middleware/cors.md
leonklingele: Sorry for the confusing, was under the impression we auto-gen the docs nowadays.

I've opened https://github.com/gofiber/fiber/issues/2401 as a feature request for this.

ReneWerner87 (approved):

middleware/cors/cors.go
leonklingele: Actually, I prefer the current code.

ReneWerner87: ok

leonklingele (changes requested):
We might want to issue a warning to stderr when both `config.AllowOrigins` and the new `config. AllowOriginsFunc` are set.

ReneWerner87:
@Jamess-Lucass can you do the suggested change ? (use the log package https://pkg.go.dev/log with "[Warning] ")
https://github.com/search?q=repo%3Agofiber%2Ffiber+%5BWarning%5D&type=code

ReneWerner87:
https://github.com/search?q=repo%3Agofiber%2Ffiber%20log.Printf(%22%5B&type=code

maybe with the middelware name in uppercase 
"[CORS] - [Warning] ...."

Jamess-Lucass:
Yeah I can add a warning log when both are set. I will add that I think it's a valid scenario for both to be set, eg.
```go
app.Use(cors.New(cors.Config{
    AllowCredentials: true,
    AllowOrigins: "https://my-domain.com", // for production
    AllowOriginsFunc: func(origin string) bool {
        return os.Getenv("ENVIRONMENT") == "development" // in development allow CORS from any origin
    }
}))
```

or even

```go
cfg := cors.Config{
    AllowCredentials: true,
    AllowOrigins: "https://my-domain.com", // for production
}

if os.Getenv("ENVIRONMENT") == "development" {
    cfg.AllowOriginsFunc = func(origin string) bool {
        // some logic
    },
}
  
app.Use(cors.New(cfg))
```

As long as people are aware this log does not indicate there is a miss configuration on their part and they do not need to resolve anything then I am happy.

ReneWerner87:
@Jamess-Lucass the idea was to allow both but to give the user the information to choose one or the other

with express only one setting is possible at the same time, because it is on the same property
https://expressjs.com/en/resources/middleware/cors.html#configuration-options

do you think it makes sense to use both at the same time ?
know that it is also possible with gin, but the question is whether it is useful

Jamess-Lucass:
> @Jamess-Lucass the idea was to allow both but to give the user the information to choose one or the other
> 
> with express only one setting is possible at the same time, because it is on the same property https://expressjs.com/en/resources/middleware/cors.html#configuration-options
> 
> do you think it makes sense to use both at the same time ? know that it is also possible with gin, but the question is whether it is useful

Ahh okay I see, I personally would be setting both values but if we encourage users to just set one of them the above code block example would then look like.

```go
cfg := cors.Config{
    AllowCredentials: true    
}

if os.Getenv("ENVIRONMENT") == "development" {
    cfg.AllowOriginsFunc = func(origin string) bool {
        // some logic
    },
}

if os.Getenv("ENVIRONMENT") == "production" {
    cfg.AllowOrigins = "https://my-domain.com", // for production
}
  
app.Use(cors.New(cfg))
```

I think either approach is fine. At the moment if it finds a match for `AllowOrigins` it won't then run `AllowOriginsFunc`. 

Whatever you think is best for developer experience

ReneWerner87:
since we are oriented to express and one would achieve both with the functionality of the function

I think that the current warning message is okay

what do you think @leonklingele @efectn

efectn:
I'm OK to the PR. But i'd recommend to add a description to avoid using this feature in production. It can be added to config and docs

Jamess-Lucass:
> I'm OK to the PR. But i'd recommend to add a description to avoid using this feature in production. It can be added to config and docs

See most recent commit, I updated docs with a notice about discouraging the use of this in production. Just let me know if any of the wording wants to be changed at all 👍

efectn (approved):

PR #2407, opened 2023-04-07
:rocket: Feature: Public ShutdownWithContext

jcyamacho:
## Description

The Fiber app offers two shutdown options: `Shutdown` and `ShutdownWithTimeout`. Both of these options internally use a private `shutdownWithContext` method, which is more versatile for those using Fiber. By making `shutdownWithContext` public, we allow developers to create the best strategy for shutting down the app.

## Type of change

## Checklist:

## Commit formatting:

### DIFF
--- a/app.go
+++ b/app.go
@@ -847,7 +847,7 @@ func (app *App) HandlersCount() uint32 {
 //
 // Shutdown does not close keepalive connections so its recommended to set ReadTimeout to something else than 0.
 func (app *App) Shutdown() error {
-	return app.shutdownWithContext(context.Background())
+	return app.ShutdownWithContext(context.Background())
 }
 
 // ShutdownWithTimeout gracefully shuts down the server without interrupting any active connections. However, if the timeout is exceeded,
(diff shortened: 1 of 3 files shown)


app.go
```diff
@@ -860,11 +860,15 @@ func (app *App) Shutdown() error {
...
 }
 
-// shutdownWithContext shuts down the server including by force if the context's deadline is exceeded.
-func (app *App) shutdownWithContext(ctx context.Context) error {
+// ShutdownWithContext shuts down the server including by force if the context's deadline is exceeded.
+//
+// Make sure the program doesn't exit and waits instead for ShutdownWithTimeout to return.
+//
+// ShutdownWithContext does not close keepalive connections so its recommended to set ReadTimeout to something else than 0.
+func (app *App) ShutdownWithContext(ctx context.Context) error {
```
gaby, line 871: If the function is public, we may need to add a test.

gaby (changes requested):
Now sure why the spacing was changed in the .md files

docs/api/app.md
```diff
@@ -145,19 +145,19 @@ func (app *App) MountPath() string
 
 ```go title="Examples"
 func main() {
-	app := fiber.New()
```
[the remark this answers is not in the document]
jcyamacho: thanks!! done

app.go
jcyamacho: done! the test is almost equal to `Test_App_ShutdownWithTimeout` 😅

jcyamacho:
> Now sure why the spacing was changed in the .md files

it was an auto-formatter in my local.. 😅

gaby (approved):
LGTM

efectn (approved):

PR #2418, opened 2023-04-15
🚀 [Feature]: Split fiber.Storage to multiple interfaces

### LINKED ISSUE #2399: 🚀 [Feature]: Split fiber.Storage to multiple interfaces
### Feature Description

Go interfaces are meant to be small. The `fiber.Storage` should be split to multiple interfaces:

Instead of (as currently done):

```go
type Storage interface {
	Get(key string) ([]byte, error)
	Set(key string, val []byte, exp time.Duration) error
	Delete(key string) error
	Reset() error
	Close() error
}
```

we should do:

```go
type Storage interface {
	StorageGetter
	StorageSetter
	StorageDeleter
	StorageResetter
	StorageCloser
}

type StorageGetter interface {
	Get(key string) ([]byte, error)
}
type StorageSetter interface {
	Set(key string, val []byte, exp time.Duration) error
}
type StorageDeleter interface {
	Delete(key string) error
}
type StorageResetter interface {
	Reset() error
}
type StorageCloser interface {
	Close() error
}
```

This even is a backwards-compatible way for `v2`.

See also https://github.com/gofiber/fiber/pull/2253#pullrequestreview-1203444213

_No response_

### Code Snippet (optional)

_No response_

### Checklist:

carmeloriolo:
## Description

Fixes #2399 
As suggested in the related issue this PR splits the already existing fiber.Storage interface becomes the embedding of smaller interfaces.

## Checklist:


ReneWerner87:
don´t forget the storage repository
https://github.com/gofiber/storage/blob/main/storage.go
https://github.com/gofiber/storage/blob/main/README.md

ReneWerner87:
and maybe you should adjust the usages in the places where the storage is used, so that only the functionality of the object/instance is needed, which is also used in the flow.

carmeloriolo:
> and maybe you should adjust the usages in the places where the storage is used, so that only the functionality of the object/instance is needed, which is also used in the flow.

this was something I was wondering if were related to the scope of the PR or not. I'll take this as a confirm to update all the usages.

leonklingele:
@ReneWerner87 any idea why CI benchmark is failing? `Performance alert! Previous value was 647.1 and current value is 1483. It is 2.2917632514294546x worse` doesn't look reasonable to me.

leonklingele (approved):

gaby:
> @ReneWerner87 any idea why CI benchmark is failing? `Performance alert! Previous value was 647.1 and current value is 1483. It is 2.2917632514294546x worse` doesn't look reasonable to me.

I was looking into this, it may be related to the logic we are using. I don't have a solution yet, but its not the runner.

carmeloriolo:
I have opened the same PR in https://github.com/gofiber/storage/pull/783

I have also find only one place in the project where an easy replacement from Storage to a granular one (StorageSetter) is possible. ([here](https://github.com/gofiber/fiber/pull/2418/commits/3f4de1a6f8ed2d8035f1410f842968fe35e5c942)).

gaby (approved):
LGTM

leonklingele:
There are a lot more uses of the big `fiber.Storage` interface, which only some methods are used of. For example:

- https://github.com/gofiber/fiber/blob/3e9575b0fe5dac95686b32ba7e996796d0b2eccf/middleware/idempotency/config.go#L48
- https://github.com/gofiber/fiber/blob/3e9575b0fe5dac95686b32ba7e996796d0b2eccf/middleware/csrf/config.go#L70
- https://github.com/gofiber/fiber/blob/3e9575b0fe5dac95686b32ba7e996796d0b2eccf/middleware/limiter/config.go#L54

carmeloriolo:
Yes I agree but the suggested  splitting is too much granular to cover that cases. Some interface usage relies only on Get, Set and Delete for example but this case fall always on the broader Storage interface.

leonklingele:
> Yes I agree but the suggested splitting is too much granular to cover that cases. Some interface usage relies only on Get, Set and Delete for example but this case fall always on the broader Storage interface.

I was thinking of doing something like this inside each of those middlewares:

```
type storage {
    fiber.StorageGetter
    fiber.StorageSetter
}
```

It's duplicated code, but makes everything so much more flexible.

carmeloriolo:
> > Yes I agree but the suggested splitting is too much granular to cover that cases. Some interface usage relies only on Get, Set and Delete for example but this case fall always on the broader Storage interface.
> 
> 
> 
> I was thinking of doing something like this inside each of those middlewares:
> 
> 
> 
> ```go
> 
> type storage interface {
> 
>     fiber.StorageGetter
> 
>     fiber.StorageSetter
> 
> }
> 
> ```
> 
> 
> 
> It's duplicated code, but makes everything so much more flexible.

Good idea! There will be a few duplication in each middleware but in the end by doing this way there is no more need to create useless methods just to adhere to the interface.

gaby (approved):
LGTM, awesome work!

leonklingele:
I‘m wondering whether we should export the Storage interfaces 🤔
Is there any downside to make them unexported?

carmeloriolo:
@leonklingele No downside at all, update in the last commit

ReneWerner87:
We should vote whether we really want to split out the storage interface.

I got the question from several sides if we really want to do that, because the storage interface is already small enough and not complex and there are probably not many advantages.

I like small units, but I'm still divided because I can't understand the questions about it.

So I would like to let the mass(community) decide. The majority can decide.

Please vote with thumbs up "👍" for the continuation and later merge of the pull request or with thumbs down "👎" for the stop of splitting, under the post.

efectn:
I understand the motivation of the PR and i agree, this PR may resolves some minor problems but causes code duplications. And i mostly want to reduce code duplications. So i vote for 👎

ReneWerner87:
Thanks for the effort

Issue #2419, opened 2023-04-15
Cloudflare and reverse proxy support 


justevery:
### Bug Description

A cloudflare and error issue.

### How to Reproduce

1. Go to '....'
3. Do '....'
4. See '....'

### Expected Behavior

1) The reverse proxy thing if any attack comes expose the reverse proxy IP address and it shows everyone "dial connection brb brb ip". So your ip address gone, i look the reverse proxy, error handler documents but i don't see how i can made all errors to custom html page.

2) Ratelimit are need to have a cloudflare mode for get CF-Connecting-IP it so annoying because of i tried 5-6 way for get CF-Connecting-IP and the end i do.

3) The logging thing has same problem what i said on 2.

### Fiber Version

latest

### Checklist:

- [X] I agree to follow Fiber's [Code of Conduct](https://github.com/gofiber/fiber/blob/master/.github/CODE_OF_CONDUCT.md).
- [X] I have checked for existing issues that describe my problem prior to opening this one.
- [X] I understand that improperly formatted bug reports may be closed without explanation.

ReneWerner87:
no 2 and 3 are fixable with 

```go
app := fiber.New(fiber.Config{
	ProxyHeader: "CF-Connecting-IP",
})
```` 
or
````go
app := fiber.New(fiber.Config{
	ProxyHeader: fiber.HeaderXForwardedFor,
})
````
according to https://developers.cloudflare.com/fundamentals/get-started/reference/http-request-headers/ and
https://developers.cloudflare.com/fundamentals/get-started/reference/http-request-headers/#cf-connecting-ip

after that the ctx.IP method is using this header for the ip determination
https://docs.gofiber.io/api/ctx#ip

and the logger will log the right ip and the limiter middleware is also using the right one.
![image](https://user-images.githubusercontent.com/7063188/232280980-0485cdeb-522f-45fd-821e-56dc3b0ff657.png)
https://docs.gofiber.io/api/middleware/limiter#config

no 1 is solveable with the ModifiedResponse method
here a small example:
```go
package main

import (
	"fmt"
	"github.com/gofiber/fiber/v2"
	"github.com/gofiber/fiber/v2/middleware/proxy"
)

func main() {
	app := fiber.New()


	app.Use(
		proxy.Balancer(proxy.Config{
		        Servers:  []string{
		              "http://localhost:3001",
		              "http://localhost:3002",
	                },
			ModifyResponse: func(c *fiber.Ctx) error {
[...]

PR #2421, opened 2023-04-17
🚀 Improve error handling for net error(s)

ReneWerner87:
Fixes: reverse proxy support #2419

With this change we want to improve the default error handler, so that it does not give out the internal network information of the dial in case of net.error.
So that fiber becomes more secure.

### DIFF
--- a/app.go
+++ b/app.go
@@ -470,12 +470,6 @@ var DefaultMethods = []string{
 
 // DefaultErrorHandler that process return errors from handlers
 func DefaultErrorHandler(c *Ctx, err error) error {
-	var netErr net.Error
-	if errors.As(err, &netErr) {
-		// 502 - Bad Gateway
-		err = ErrBadGateway
-	}
-
 	code := StatusInternalServerError
 	var e *Error
 	if errors.As(err, &e) {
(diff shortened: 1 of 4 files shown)


leonklingele (commented):
Is it really always an `502 Bad Gateway` error?

gaby:
> Is it really always an `502 Bad Gateway` error?

I'm wondering the same thing 🤔

ReneWerner87:
just give me good suggestions and i will change it

before it was always a 500 that was not specific and exposed the error message with the network information.

so i would either take the 500 again or 502 or 503

gaby:
> just give me good suggestions and i will change it
> 
> before it was always a 500 that was not specific and exposed the error message with the network information.
> 
> so i would either take the 500 again or 502 or 503

Is it possible to get the return code from the NetError object?

ReneWerner87:
not really, because it is not a http error
its a network connection error when trying the client request internally

gaby:
@ReneWerner87 This is what GPT-4 says:

```
This function appears to be a default error handler for some web application, and it is handling errors and returning appropriate HTTP status codes. Given the provided code, it makes sense to return a 500 status code (Internal Server Error) by default.

The function first checks if the error is a net.Error, and if it is, it sets the error to ErrBadGateway. This would typically correspond to a 502 status code (Bad Gateway). Then, the function checks if the error is of type *Error, and if it is, it retrieves the status code from that error.

However, it is worth noting that the function does not explicitly return 503 (Service Unavailable) at any point. To include that, you might want to add another condition to check for specific errors corresponding to a 503 status code.

In summary, based on the provided code, it makes sense for the function to return 500 (Internal Server Error) by default, and 502 (Bad Gateway) if the error is a net.Error. To return a 503 (Service Unavailable), you would need to add additional logic to handle that case.
```

I guess bad gateway makes sense in terms of a reverse proxy.

ReneWerner87:
- [ ] add an unittest

PR #2424, opened 2023-04-20
🚀 [Feature]: middleware/recover: ignore error from panics

tnako:
## Description

Once panic occur there is possibility to recover and continue serving request. Current `recover` middleware also transform panic message into error and pass into response body with status code 500. This can confuse users or disclose secure information.

With skipping error message response become good (code 200) and body is empty. Extended with custom `StackTraceHandler` allow to write any response and status code.

## Type of change

## Checklist:

### DIFF
--- a/middleware/recover/recover.go
+++ b/middleware/recover/recover.go
@@ -31,6 +31,10 @@ func New(config ...Config) fiber.Handler {
 					cfg.StackTraceHandler(c, r)
 				}
 
+				if cfg.SkipResponseError {
+					return
+				}
+
 				var ok bool
 				if err, ok = r.(error); !ok {
 					// Set error that will call the global error handler
--- a/docs/api/middleware/recover.md
+++ b/docs/api/middleware/recover.md
@@ -49,6 +49,11 @@ type Config struct {
     // Optional. Default: false
     EnableStackTrace bool
 
+	// SkipResponseError disable showing panics in response body
+	//
+	// Optional. Default: false
+	SkipResponseError bool
+
     // StackTraceHandler defines a function to handle stack trace
     //
     // Optional. Default: defaultStackTraceHandler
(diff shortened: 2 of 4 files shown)


ReneWerner87:
I understand the security concerns
but to set the response to "OK" in case of an error, i.e. successful, i don't think is right either

then rather get the 500 and not keep the body general

tnako:
> hen rather get the 500 and keep the body general "Internal server error"

500 is not an option as proxy servers will close connection, increasing communication costs

The same for default text. Raw English text confuse way more especially when browser/service expect json for example.

With **backward compatibility** such custom code and text can be done with
```
...
	app.Use(recover.New(recover.Config{EnableStackTrace: true, SkipResponseError: true, StackTraceHandler: processPanics}))
...
func processPanics(c *fiber.Ctx, _ interface{}) {
	_ = c.Status(http.StatusInternalServerError).SendString("Internal server error")
}
```

middleware/recover/recover.go
```diff
@@ -31,6 +31,10 @@ func New(config ...Config) fiber.Handler {
 					cfg.StackTraceHandler(c, r)
 				}
 
+				if cfg.SkipResponseError {
+					return
+				}
+
```
li-jin-gou, line 37: It seems better to add a panic handler to customize the user behavior, the default behavior is to return err.
- refer to https://github.com/cloudwego/hertz/commit/f687dac7225c070e85c2abbf87cd4605a01402f3

li-jin-gou: sry, there is already a StackTraceHandler and a handler can be used to do this, but there will be break changes.😞

ReneWerner87: would be ok for me, but with fiber error handler we already have this possibility, because the error which is caught in the middleware is passed on to the error handler and this is responsible for the output or the conversion of the response.

tnako:
Closing as custom response on panic can be done as part `ErrorHandler` in `fiber.Config{}`

ReneWerner87:
anyway thanks that you wanted to improve something 🚀

PR #2432, opened 2023-04-24
:rocket: Consistent way of logging and fix middleware log format

### LINKED ISSUE #2402: 🚀 [Feature]: Consistent way of issuing fiber log messages / warnings / errors
### Feature Description

Currently, some fiber warnings & errors are done using `fmt.Print*`, some using `log.Print*`:

Examples:
- https://github.com/gofiber/fiber/blob/5a8bc44a1f55149d2f6b78ad128cb1d9c4c1c761/internal/template/html/html.go#L115
- https://github.com/gofiber/fiber/blob/5a8bc44a1f55149d2f6b78ad128cb1d9c4c1c761/app.go#L524

There should only be a single way of doing this.

_No response_

### Code Snippet (optional)

_No response_

### Checklist:

kousikmitra:
## Description

- [x] Replace all fmt.Print* with log.Print*
- [x] Use `[MIDDLEWARENAME] - [LOG-LEVEL] text` in all middleware for logging

Fixes #2402 

## Type of change

## Checklist:

## Commit formatting:

### DIFF
--- a/internal/template/html/html.go
+++ b/internal/template/html/html.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"html/template"
 	"io"
+	"log"
 	"net/http"
 	"os"
 	"path/filepath"
@@ -112,7 +113,7 @@ func (e *Engine) Debug(enabled bool) *Engine {
 
 // Parse is deprecated, please use Load() instead
 func (e *Engine) Parse() error {
-	fmt.Println("Parse() is deprecated, please use Load() instead.")
+	log.Println("[Warning] Parse() is deprecated, please use Load() instead.")
 	return e.Load()
 }
 
@@ -169,7 +170,7 @@ func (e *Engine) Load() error {
 		}
 		// Debugging
 		if e.debug {
-			fmt.Printf("views: parsed template: %s\n", name)
+			log.Printf("views: parsed template: %s\n", name)
 		}
 		return err
 	}
--- a/middleware/cache/config.go
+++ b/middleware/cache/config.go
@@ -102,11 +102,11 @@ func configDefault(config ...Config) Config {
 
 	// Set default values
 	if cfg.Store != nil {
-		log.Printf("[CACHE] Store is deprecated, please use Storage\n")
+		log.Printf("[CACHE] - [Warning] Store is deprecated, please use Storage\n")
 		cfg.Storage = cfg.Store
 	}
 	if cfg.Key != nil {
-		log.Printf("[CACHE] Key is deprecated, please use KeyGenerator\n")
+		log.Printf("[CACHE] - [Warning] Key is deprecated, please use KeyGenerator\n")
 		cfg.KeyGenerator = cfg.Key
 	}
 	if cfg.Next == nil {
--- a/middleware/csrf/config.go
+++ b/middleware/csrf/config.go
@@ -132,15 +132,15 @@ func configDefault(config ...Config) Config {
 
 	// Set default values
 	if cfg.TokenLookup != "" {
-		log.Printf("[CSRF] TokenLookup is deprecated, please use KeyLookup\n")
+		log.Printf("[CSRF] - [Warning] TokenLookup is deprecated, please use KeyLookup\n")
 		cfg.KeyLookup = cfg.TokenLookup
 	}
 	if int(cfg.CookieExpires.Seconds()) > 0 {
-		log.Printf("[CSRF] CookieExpires is deprecated, please use Expiration\n")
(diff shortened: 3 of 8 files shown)


efectn:
We should add interface like https://github.com/labstack/echo/blob/master/log.go#L11 to make it customizable (maybe smaller). For example i may want to use zerolog, zap with fiber

efectn:
> > We should add interface like https://github.com/labstack/echo/blob/master/log.go#L11 to make it customizable (maybe smaller). For example i may want to use zerolog, zap with fiber
> 
> That would be a good addition. I can work on it. But I think a separate PR would be better instead of adding here.

Ok to me

leonklingele:
I'd prefer `[LEVEL] [SCOPE, e.g. MIDDLEWARE_NAME]` over the other way around.

Why don't we use a proper logging library like zerolog or zap though?

efectn:
> I'd prefer `[LEVEL] [SCOPE, e.g. MIDDLEWARE_NAME] [MESSAGE] [DETAILS]` over having `[SCOPE] [LEVEL] ..`.
> 
> Why don't we use a proper logging library like zerolog or zap though?

We shouldn't add extra dependencies for optional things

kousikmitra:
Totally agree with @efectn , I have opened an issue to track this. Please feel free to add additional context.

gaby:
I agree with @leonklingele it should be `[LEVEL] [MIDDLEWARE_NAME] [MESSAGE]`. It looks weird to have the level after the middleware name

leonklingele:
ftr: this‘ll soon be part of the Go stdlib 🫨🥳🚀

https://pkg.go.dev/golang.org/x/exp/slog

efectn:
> ftr: this‘ll soon be part of the Go stdlib 🫨🥳🚀
> 
> https://pkg.go.dev/golang.org/x/exp/slog

Seems slower than zerolog and zap. At least for now https://github.com/uber-go/zap#performance

kousikmitra:
> I agree with @leonklingele it should be `[LEVEL] [MIDDLEWARE_NAME] [MESSAGE]`. It looks weird to have the level after the middleware name

Ya agree, if others agree I will make the change here.

leonklingele:
Why did this get merged already? Didn’t we want to change the order of log level and scope info?

ReneWerner87:
Missed it, will revert it

next time it is better to store a task in the pull request or to reject it first, then I usually check it first 
but without this it is quite hard for me to keep track of the 10 repositories with all the communication in the pull requests and issues

leonklingele:
I‘d leave it as-is and just do a fixup.

ReneWerner87:
https://github.com/gofiber/fiber/pull/2444

ReneWerner87:
> We should add interface like https://github.com/labstack/echo/blob/master/log.go#L11 to make it customizable (maybe smaller). For example i may want to use zerolog, zap with fiber

what do you think about this kind of solution

```go
package main

import (
	"fmt"
	"log"
)

type Logger struct {
	topic string
}

func NewLogger(topic string) *Logger {
	return &Logger{topic: topic}
}

func (l *Logger) formatMessage(level, format string, args ...interface{}) string {
	message := fmt.Sprintf(format, args...)
	if l.topic != "" {
		return fmt.Sprintf("[%s] - [%s] %s", level, l.topic, message)
	}
	return fmt.Sprintf("[%s] - %s", level, message)
}

func (l *Logger) Info(format string, args ...interface{}) {
	log.Println(l.formatMessage("Info", format, args...))
}

func (l *Logger) Warning(format string, args ...interface{}) {
	log.Println(l.formatMessage("Warning", format, args...))
}

func (l *Logger) Error(format string, args ...interface{}) {
	log.Println(l.formatMessage("Error", format, args...))
}

func main() {
	cacheLogger := NewLogger("CACHE")
	cacheLogger.Warning("Key is deprecated, please use KeyGenerator")

	corsLogger := NewLogger("CORS")
	corsLogger.Warning("Both 'AllowOrigins' and 'AllowOriginsFunc' have been defined.")

	csrfLogger := NewLogger("CSRF")
	csrfLogger.Warning("TokenLookup is deprecated, please use KeyLookup")

	idempotencyLogger := NewLogger("IDEMPOTENCY")
	key := "example_key"
	err := fmt.Errorf("an error occurred")
	idempotencyLogger.Error("failed to unlock key %q: %v", key, err)

	generalLogger := NewLogger("")
	generalLogger.Warning("Parse() is deprecated, please use Load() instead.")
}
```

or we use the slog lib
https://pkg.go.dev/golang.org/x/exp/slog
the idea was from @leonklingele

PR #2437, opened 2023-04-26
feat: Add HTML minification for rendering

alwismt:
The  function has been added to minify the HTML data into a token stream, and then write the minified HTML to a buffer. This feature improves performance by reducing the size of the HTML output.

## Description
Added minify.go file with a htmlMinify function that can be used to minify HTML data. Also wrote a test case TestHtmlMinify in minify_test.go to ensure it works as expected. Added the htmlMinify function to Ctx.Render function in Ctx.go and also added an EnableHTMLMinify app config option for it. Wrote another test function Test_Ctx_RenderWithMinify in Ctx_test.go to test rendering with minification enabled.

## Type of change
New feature

## Checklist
- [x] I have added test cases to ensure my changes work as expected
- [x] I have updated the documentation accordingly
- [x] My code follows the style guidelines of the project
- [x] I have checked that there are no linting errors in my code

## Commit formatting
feat: added minify function for HTML rendering and added tests

## Changes
- Added minify.go file with htmlMinify function
- Added TestHtmlMinify test case in minify_test.go
- Added htmlMinify function to Ctx.Render in Ctx.go
- Added EnableHTMLMinify app config option for minification
- Added Test_Ctx_RenderWithMinify test function in Ctx_test.go


ReneWerner87 (changes requested):
Interesting and good feature,
Thanks for the work

Can we modify it a bit, currently the feature only looks at the render method, but could be applied to all responses which matched certain content types

I have a middleware in mind that you can add like the compress middleware and then you can configure it to enable different minification on the different routes.

Maybe we should also provide this in the contrib repository, as I don't want to bloat the core (if it's a middleware, we can vote and move the code)

alwismt:
Great, sounds like a good plan. I'll work on implementing this feature in the compress middleware so that it can minify other content types beside HTML. 

Additionally, I can create a separate middleware for minification, which can be configured to enable different minification on different routes.

Regarding the contrib repository, that's a great idea as well. We can move the code to the contrib repository if it's agreed upon by the community. Thank you for your feedback and suggestions.

ReneWerner87:
> Great, sounds like a good plan. I'll work on implementing this feature in the compress middleware so that it can minify other content types beside HTML.

here you misunderstood something, i meant the compress middleware you can use as a template
the actual compress middleware should not be adapted of course

alwismt:
Got it, I will use the compress middleware as a reference and will create a new middleware for `minification` of different content types.

PR #2441, opened 2023-04-28
:rocket:  [Feature]: Add contentType ...string to JSON

### LINKED ISSUE #2428: 🤗 [Question]: How best to return JSON with different content-type?
### Question Description

I'm playing with Fiber, and trying to get a clean solution for RFC-7807 Problem Detail responses. Which I've mostly done. The problem is the content types. These responses should have a `Content-Type` header of `application/problem+json` but Fibre is overriding this to just `application/json`.

What I've done is to write an `ErrorHandler` function that detects if my `error` is a `problem.Problem` type, and if so then handle it to return the correct response:

```go
package problem

import (
	"errors"

	"github.com/gofiber/fiber/v2"
)

func ErrorHandler(c *fiber.Ctx, err error) error {
	var problem Problem
	if errors.As(err, &problem) {
		c.Status(problem.statusCode)

		if len(problem.payload) > 0 {
			c.Set("content-type", "application/problem+json; charset=utf-8")

			return c.JSON(problem.payload)
		}

		return nil
	}

	return fiber.DefaultErrorHandler(c, err)
}
```

For the most part this works great. But the `c.JSON` call forcibly sets the `Content-Type` header, even if I've already set it.

Changing it to this does work, but feels just a bit wrong...
```go
		if len(problem.payload) > 0 {
			if err := c.JSON(problem.payload); err != nil {
				return err
			}

			c.Set("content-type", "application/problem+json; charset=utf-8")

			return nil
		}
```

Is there a better way to send a JSON response with an alternative content type? Or is this the best option?

Cheers

### Checklist:

- [X] I have checked for existing issues that describe my questions prior to opening this one.
- [X] I understand that improperly formatted questions may be closed without explanation.

cmd777:
## Description
This commit adds an optional contentType parameter to JSON, which can be used to set the Content-Type like so
```go
c.JSON(Map{
	"Name": "Grame",
	"Age":  20,
}, "application/problem+json; charset=utf-8")
```

Also, remove old info from ctx.JSON which is incorrect
Converts any **interface** or **string** to `JSON using the [goccy/go-json](https://github.com/goccy/go-json) package.`

Fixes #2428

## Type of change

## Checklist:

- [X] I have made corresponding changes to the documentation - /docs/ directory for https://docs.gofiber.io/

## Commit formatting:

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -789,14 +789,19 @@ func (c *Ctx) Is(extension string) bool {
 // Array and slice values encode as JSON arrays,
 // except that []byte encodes as a base64-encoded string,
 // and a nil slice encodes as the null JSON value.
-// This method also sets the content header to application/json.
-func (c *Ctx) JSON(data interface{}) error {
+// This method also sets the content header to application/json
+// Unless specified otherwise with contentType
+func (c *Ctx) JSON(data interface{}, contentType ...string) error {
 	raw, err := c.app.config.JSONEncoder(data)
 	if err != nil {
 		return err
 	}
 	c.fasthttp.Response.SetBodyRaw(raw)
-	c.fasthttp.Response.Header.SetContentType(MIMEApplicationJSON)
+	if len(contentType) > 0 {
+		c.fasthttp.Response.Header.SetContentType(contentType[0])
+	} else {
+		c.fasthttp.Response.Header.SetContentType(MIMEApplicationJSON)
+	}
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


ReneWerner87 (changes requested):
As i already said in the issue report, i don't like this extension of the method.
Otherwise its needed in every other method (jsonp, xml....)

cmd777:
> As i already said in the issue report, i don't like this extension of the method. Otherwise its needed in every other method (jsonp, xml....)

Yeah, fair enough. I'll close this pull request, as it's really specific to that one issue <sub>(as you said)</sub>

ReneWerner87:
But anyway, thanks for your effort.

PR #2463, opened 2023-05-16
🐛 Fix mount route positioning

### LINKED ISSUE #2460: 🐛 [Bug]: Wrong handlers execution order in some mount cases
### Bug Description

In some rare cases, fiber can wrongly build `app.treeStack`, breaking the execution chain. For example, it can call a handler from a mounted app before it calls a middleware specified at the top level earlier.

### How to Reproduce

In attached code snippet, fiber thinks that middleware printing `"i'm ignored!"` should be executed after `/methods/test` handler.

I think there are many more cases when this happens, but for now I'm able to reproduce only this one.

### Expected Behavior

Handlers execute in correct order.

### Fiber Version

2.45.0

### Code Snippet (optional)

```go
package main

import (
	"fmt"
	"github.com/gofiber/fiber/v2"
	"net/http"
)

func getApi() *fiber.App {
	var API = fiber.New()

	methods := API.Group("/methods")

	methods.Get("/test", func(c *fiber.Ctx) error {
		if c.Locals("hello") != "world" {
			panic("hello is not world")
		}
		return c.SendStatus(http.StatusOK)
	})

	return API
}

func main() {
	app := fiber.New()

		return c.Next()
	})

	app.Mount("/ws", fiber.New())

		return c.Next()
	})

	app.Get("/status", func(c *fiber.Ctx) error {
		return c.SendString("ok")
	})

		fmt.Println("i'm ignored!")
		c.Locals("hello", "world")
		return c.Next()
	})

	app.Mount("", getApi())

	test(app)
}

func test(app *fiber.App) {
	req, _ := http.NewRequest("GET", fmt.Sprintf("/methods/test"), nil)

	app.Test(req)
}
```

### Checklist:

ReneWerner87:
## Description

Fixes #2460 

## Type of change

## Checklist:

### DIFF
--- a/mount_test.go
+++ b/mount_test.go
@@ -175,7 +175,11 @@ func Test_App_Mount_RoutePositions(t *testing.T) {
 		})
 		methods := subApp2.Group("/subApp2")
 		methods.Get("/world", func(c *Ctx) error {
-			return c.SendString(c.Locals("world").(string))
+			v, ok := c.Locals("world").(string)
+			if !ok {
+				panic("unexpected data type")
+			}
+			return c.SendString(v)
 		})
 		app.Mount("", subApp2)
 	}
--- a/mount_test.go
+++ b/mount_test.go
@@ -104,7 +104,7 @@ func Test_App_Mount_Express_Behavior(t *testing.T) {
 		utils.AssertEqual(t, nil, err, "app.Test(req)")
 		body, err := io.ReadAll(resp.Body)
 		utils.AssertEqual(t, nil, err)
-		utils.AssertEqual(t, expectedBody, string(body), "Response body")
+		utils.AssertEqual(t, expectedBody, string(body), "Unexpected response body")
 	}
 
 	app := New()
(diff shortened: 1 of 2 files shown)


mount_test.go
```diff
@@ -140,6 +140,75 @@ func Test_App_Mount_Express_Behavior(t *testing.T) {
...
 
+// go test -run Test_App_Mount_RoutePositions
+func Test_App_Mount_RoutePositions(t *testing.T) {
+	t.Parallel()
+	testEndpoint := func(app *App, route, expectedBody string) {
+		resp, err := app.Test(httptest.NewRequest(MethodGet, route, nil))
+		utils.AssertEqual(t, nil, err, "app.Test(req)")
+		body, err := io.ReadAll(resp.Body)
+		utils.AssertEqual(t, nil, err)
+		utils.AssertEqual(t, expectedBody, string(body), "Unexpected response body")
```
gaby, line 151: Checking status code is missing here

gaby (approved):
Only one thing missing, but LGTM

mount_test.go
```diff
@@ -140,6 +140,75 @@ func Test_App_Mount_Express_Behavior(t *testing.T) {
 	utils.AssertEqual(t, uint32(16+9), app.routesCount)
 }
 
+// go test -run Test_App_Mount_RoutePositions
+func Test_App_Mount_RoutePositions(t *testing.T) {
+	t.Parallel()
+	testEndpoint := func(app *App, route, expectedBody string) {
+		resp, err := app.Test(httptest.NewRequest(MethodGet, route, nil))
```
leonklingele, line 147:
```suggestion
		resp, err := app.Test(httptest.NewRequest(MethodGet, route, http.NoBody))
```

This should have been caught by golangci-lint.

ReneWerner87:
@gaby @leonklingele thx for the feedback
can you look again

mount_test.go
```diff
@@ -99,12 +100,13 @@ func Test_App_Mount_Express_Behavior(t *testing.T) {
 			return c.SendString(body)
 		}
 	}
-	testEndpoint := func(app *App, route, expectedBody string) {
-		resp, err := app.Test(httptest.NewRequest(MethodGet, route, nil))
+	testEndpoint := func(app *App, route, expectedBody string, statusCode int) {
```
leonklingele, line 103:
```suggestion
	testEndpoint := func(app *App, route, expectedBody string, expectedStatusCode int) {
```

PR #2465, opened 2023-05-17
🚀 Speedup and cleanup "path" testcases and benchmarks

ReneWerner87:
i have tried to reduce the benchmark cases for routing matching 
and move the testcases in a list, so that all test and benchmark functions use the same lists instead of repeating them redundantly


ReneWerner87:
image
instead of
image

path_testcases_test.go
```diff
@@ -0,0 +1,718 @@
...
+			{
+				pattern: "/api/v1/:param<int>?",
+				testCases: []routeTestCase{
+					{url: "/api/v1/entity", params: nil, match: false},
+					{url: "/api/v1/8728382", params: []string{"8728382"}, match: true},
+					{url: "/api/v1/true", params: nil, match: false},
+					{url: "/api/v1/", params: []string{""}, match: true},
+				},
+			},
+		}...,
```
gaby, line 716: Are these golang dots or spaces? 😂

efectn: I think it's to necessary make it variadic

ReneWerner87: golang dots for the variadic function -> for the slice combination

ReneWerner87:
do you have a better name for my helper file ? 
path_testcases_test.go

path_testcases_test.go
gaby: I was looking at it on my phone, and got confused :-)

ReneWerner87:
image
lines before and after in the test file

ReneWerner87:
definitely need an approval flag if you are good with the pull request 😄

gaby (approved):
Lgtm, easier to read!

PR #2314, opened 2023-01-28
🚀 Feature: Add earlydata middleware (v2 backport)

leonklingele:
## Description

Add earlydata middleware to add support for TLS 1.3's early data ("0-RTT") feature.

This is a v2 backport of https://github.com/gofiber/fiber/pull/2270.

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

## Checklist:

- [x] For new functionalities I follow the inspiration of the express js framework and built them similar in usage
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation - https://github.com/gofiber/docs for https://docs.gofiber.io/
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] If new dependencies exist, I have checked that they are really necessary and agreed with the maintainers/community (we want to have as few dependencies as possible)
- [x] I tried to make my code as fast as possible with as few allocations as possible
- [ ] For new code I have written benchmarks so that they can be analyzed and improved

PR #2288, opened 2023-01-06
🚀 Feature: Add idempotency middleware (v2 backport)

leonklingele:
## Description

Add idempotency middleware.
Implements https://github.com/gofiber/fiber/issues/2163.

This is a v2 backport of https://github.com/gofiber/fiber/pull/2253.

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

## Checklist:

- [x] For new functionalities I follow the inspiration of the express js framework and built them similar in usage
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation - https://github.com/gofiber/docs for https://docs.gofiber.io/
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] If new dependencies exist, I have checked that they are really necessary and agreed with the maintainers/community (we want to have as few dependencies as possible)
- [x] I tried to make my code as fast as possible with as few allocations as possible
- [x] For new code I have written benchmarks so that they can be analyzed and improved

PR #2469, opened 2023-05-17
Various linter updates

leonklingele:
See the individual commit messages for details.

A data race in fasthttp was identified using the stricter config. See https://github.com/valyala/fasthttp/pull/1565.


efectn (approved):

ctx_test.go
```diff
@@ -3533,9 +3546,9 @@ func Test_Ctx_Send(t *testing.T) {
 	app := New()
 	c := app.AcquireCtx(&fasthttp.RequestCtx{})
 	defer app.ReleaseCtx(c)
-	utils.AssertEqual(t, nil, c.Send([]byte("Hello, World")))
-	utils.AssertEqual(t, nil, c.Send([]byte("Don't crash please")))
-	utils.AssertEqual(t, nil, c.Send([]byte("1337")))
+	utils.AssertEqual(t, nil, c.SendString("Hello, World"))
```
ReneWerner87, line 3549: This is the test which tested the `Send` method
this should also work with the `Send` method and not with another one

.golangci.yml
```diff
@@ -11,15 +12,29 @@ output:
   sort-results: true
 
 linters-settings:
+  depguard:
+    include-go-root: true
+    packages:
+      - flag
+      - io/ioutil
+      # - reflect # TODO: Disallow reflect package
+      - unsafe
```
ReneWerner87, line 21: the performance and speed of fiber and fasthttp relies on some trick with the unsafe packet
i am not sure if we should limit this ? or is the plan to make exceptions where this is used ?

middleware/logger/tags.go
```diff
@@ -192,7 +204,11 @@ func createTagMap(cfg *Config) map[string]LogFunc {
 		},
 		TagLatency: func(output Buffer, c *fiber.Ctx, data *Data, extraParam string) (int, error) {
 			latency := data.Stop.Sub(data.Start)
-			return output.WriteString(fmt.Sprintf("%7v", latency))
+			n, err := fmt.Fprintf(output, "%7v", latency)
+			if err != nil {
+				return 0, fmt.Errorf("failed to write: %w", err)
```
ReneWerner87, line 209: why can not we extend the error outside ?

middleware/monitor/monitor.go
```diff
@@ -84,6 +84,7 @@ func New(config ...Config) fiber.Handler {
 		}
 		if c.Get(fiber.HeaderAccept) == fiber.MIMEApplicationJSON || cfg.APIOnly {
 			mutex.Lock()
+			defer mutex.Unlock()
```
ReneWerner87, line 87: why do we close the mutext so late? after writing the status and the json? 
besides, in previous benchmarks it had turned out that the defer calls are slower

middleware/session/data.go
```diff
@@ -4,8 +4,6 @@ import (
 	"sync"
 )
 
-// go:generate msgp
```
ReneWerner87, line 7: do we really not need this anymore ?

middleware/session/store.go
```diff
@@ -43,19 +43,18 @@ func (s *Store) Get(c *fiber.Ctx) (*Session, error) {
...
 		if id, err = s.responseCookies(c); err != nil {
 			return nil, err
 		}
-	}
 
-	// If no key exist, create new one
-	if len(id) == 0 {
-		loadData = false
-		id = s.KeyGenerator()
+		// If key still doesn't exist, create new one
```
ReneWerner87, line 53: the old way of writing is flatter and offers no disadvantages to the new way of writing

please change back

mount_test.go
```diff
@@ -111,26 +111,26 @@ func Test_App_Mount_Express_Behavior(t *testing.T) {
 
 	app := New()
 	subApp := New()
+
 	// app setup
-	{
```
ReneWerner87, line 115: this kind of encapsulation i have copied from your pull requests^^ and now you remove it

Why not keep this grouping


https://github.com/gofiber/fiber/pull/2288/files#diff-654acee0bb6ef56dde4305631ed3d98769cfea139f70647ad580d5fe4737c4ec

https://github.com/gofiber/fiber/pull/2314/files#diff-5c85ef45029b22bed483baf176606b6619b80a9a470dbe7f64a3a9a98cba8161

prefork_test.go
```diff
@@ -14,12 +14,12 @@ import (
 	"github.com/gofiber/fiber/v2/utils"
 )
 
+//nolint:paralleltest // TODO: Must be run sequentially due to global testPreforkMaster var.
 func Test_App_Prefork_Child_Process(t *testing.T) {
 	// Reset test var
 	testPreforkMaster = true
 
-	setupIsChild(t)
```
ReneWerner87, line 21: why was this changed? no cleanup needed anymore? clean startup conditions would be good to avoid errors

ReneWerner87 (changes requested):

middleware/session/data.go
ReneWerner87: not really

middleware/logger/logger.go
```diff
@@ -145,16 +145,16 @@ func New(config ...Config) fiber.Handler {
...
-				fmt.Sprintf("%s |%s %3d %s| %7v | %15s |%s %-7s %s| %-"+errPaddingStr+"s %s\n",
-					timestamp.Load().(string),
-					statusColor(c.Response().StatusCode(), colors), c.Response().StatusCode(), colors.Reset,
-					data.Stop.Sub(data.Start).Round(time.Millisecond),
-					c.IP(),
-					methodColor(c.Method(), colors), c.Method(), colors.Reset,
-					c.Path(),
-					formatErr,
-				),
+			_, _ = fmt.Fprintf(
```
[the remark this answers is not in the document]
ReneWerner87: i tested it on myself -> its okay

.golangci.yml
leonklingele: I'd go with a allowlist-approach here, i.e. deny the `unsafe` package by default and require `//nolint:depguard //  Explanation why an import of the "unsafe" package is required here`.

app.go
```diff
@@ -892,10 +892,10 @@ func (app *App) Hooks() *Hooks {
 }
 
 // Test is used for internal debugging by passing a *http.Request.
-// Timeout is optional and defaults to 1s, -1 will disable it completely.
+// Timeout is optional and defaults to 10s, -1 will disable it completely.
 func (app *App) Test(req *http.Request, msTimeout ...int) (*http.Response, error) {
 	// Set timeout
-	timeout := 1000
+	timeout := int((10 * time.Second) / time.Millisecond)
```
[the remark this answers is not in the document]
leonklingele: True. I don't really know how to avoid it though. As most of the tests run concurrently now, individual tests might take longer than the previous 1 second defined here due to the Go scheduler rescheduling execution between tests.

What do you think of adding and using a new `app.TestWithDynamicTimeout(req *http.Request, msTimeout ...int)` which defaults this to a timeout of 10s as of now, and might be further increased in the future?

.golangci.yml
```diff
@@ -84,29 +120,32 @@ linters-settings:
...
       - name: max-public-structs
         disabled: true
       - name: modifies-parameter
         disabled: true
-      - name: nested-structs
-        disabled: true
       - name: package-comments
         disabled: true
+      - name: use-any
+        disabled: true # TODO: Enable once we drop support for Go 1.17
```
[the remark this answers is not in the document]
leonklingele: Yeah, that's why it's disabled :D

ctx_test.go
```diff
@@ -76,6 +77,7 @@ func Benchmark_Ctx_Accepts(b *testing.B) {
 	expectedResults := []string{".xml", "xml", "application/xml"}
 
 	for i := 0; i < len(acceptValues); i++ {
+		i := i
```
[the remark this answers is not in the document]
leonklingele: > its looking wired for other developers

Hopefully not. This is basically a requirement for a lot of loops in Go, and basically used everywhere. I'd consider this a well-known thing.

What do you think of creating a `CONTRIBUTING.md` or similar which lists other "weird" things we commonly do in the codebase?

helpers.go
```diff
@@ -153,7 +156,7 @@ func defaultString(value string, defaultValue []string) string {
 const normalizedHeaderETag = "Etag"
 
 // Generate and set ETag header to response
-func setETag(c *Ctx, weak bool) { //nolint: revive // Accepting a bool param is fine here
+func setETag(c *Ctx, weak bool) { //nolint:revive // Accepting a bool param is fine here
```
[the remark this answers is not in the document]
leonklingele: It's considered good Go-style to turn a single function accepting a boolean into two functions, i.e. in this specific case `setETagWeak(c *Ctx)` and `setETagNonWeak(c *Ctx)` as it's more descriptive and can even be optimized better by the compiler.

middleware/envvar/envvar_test.go
```diff
@@ -120,31 +103,32 @@ func TestEnvVarHandlerDefaultConfig(t *testing.T) {
...
 }
 
+//nolint:paralleltest // Test is using t.Setenv which can't be run in parallel
 func TestEnvVarHandlerSpecialValue(t *testing.T) {
-	testEnvKey := "testEnvKey"
-	fakeBase64 := "testBase64:TQ=="
-	err := os.Setenv(testEnvKey, fakeBase64)
-	utils.AssertEqual(t, nil, err)
-	defer func() {
-		err := os.Unsetenv(testEnvKey)
```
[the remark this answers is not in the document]
leonklingele: Using the `t.Setenv` func here handles everything done before but in a much shorter fashion.

middleware/envvar/envvar_test.go
```diff
@@ -6,28 +6,17 @@ import (
...
+//nolint:paralleltest // Test is using t.Setenv which can't be run in parallel
 func TestEnvVarStructWithExportVarsExcludeVars(t *testing.T) {
-	err := os.Setenv("testKey", "testEnvValue")
-	utils.AssertEqual(t, nil, err)
-	err = os.Setenv("anotherEnvKey", "anotherEnvVal")
-	utils.AssertEqual(t, nil, err)
-	err = os.Setenv("excludeKey", "excludeEnvValue")
-	utils.AssertEqual(t, nil, err)
-	defer func() {
-		err := os.Unsetenv("testKey")
```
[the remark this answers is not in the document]
leonklingele: `t.Setenv` does the cleanup automatically.

middleware/monitor/monitor.go
leonklingele: Otherwise line 97 (`return c.Status(fiber.StatusOK).JSON(data)`) has a race condition where it reads the shared `data` variable.

True, `defer`-ed functions run a tad slower, but since one of the more recent Go releases it's negligible (and will probably be reduced further in the future).

Instead of using `defer` here, we could also make a deep copy of `data` while still having the mutex locked.

middleware/session/store.go
leonklingele: It's doing the `if id == ""` check twice if `id` is non-empty in the first run, so it's speeding up the code a little bit.

mount_test.go
leonklingele: Creating a new scope only makes sense if new variables are created within it, which should not be available in the outer scope. As no such new var is created here, we can simply drop the new scope (which even costs a bit of CPU cycles if not optimized away by the compiler).

prefork_test.go
leonklingele: The teardown func was simply unsetting the env var, which is now taken care of by `t.Setenv`, thus dropping two funcs (`setupIsChild` and `teardownIsChild `) and making the code easier to read.

ReneWerner87: yes correct, but the next session in another test is contaminated with it

what if the next tests uses the variables

middleware/monitor/monitor.go
ReneWerner87: ok it can stay

.golangci.yml
ReneWerner87: ok

.golangci.yml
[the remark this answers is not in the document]
ReneWerner87: ok, sorry i read "enabled" ^^

helpers.go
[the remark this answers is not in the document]
ReneWerner87: ok understandable, thx

middleware/envvar/envvar_test.go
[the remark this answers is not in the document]
ReneWerner87: but not for other test cases(only for yourself)

middleware/envvar/envvar_test.go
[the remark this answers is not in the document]
ReneWerner87: but not for other test cases(only for yourself)

ReneWerner87:
@leonklingele can you answer the other questions from the review

leonklingele:
Will update this on Sunday or Monday.

ReneWerner87:
leon is not available for 2 weeks, lets wait with this until he is available

efectn:
hey @leonklingele any updates on this?

ReneWerner87:
@leonklingele ping

leonklingele:
Totally forgot about this! Will get to it some time next week.

ReneWerner87:
> Totally forgot about this! Will get to it some time next week.

ok thx

middleware/envvar/envvar_test.go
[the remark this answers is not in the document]
leonklingele: Has your question been resolved in the meantime? :)

leonklingele:
@ReneWerner87 Updated & rebased. PTAL :)

.github/workflows/lint_go-vet.yml
```diff
@@ -0,0 +1,24 @@
+name: lint_go-vet
+on:
+  push:
+    branches:
+      - master
+      - main
+  pull_request:
+
+permissions:
```
gaby, line 9: This perms line is not needed

.github/workflows/lint_go-vet.yml
```diff
@@ -0,0 +1,24 @@
+name: lint_go-vet
```
gaby, line 1: Rename to `govet-lint`

.github/workflows/lint_golangci-lint.yml
```diff
@@ -1,6 +1,6 @@
 # Adapted from https://github.com/golangci/golangci-lint-action/blob/b56f6f529003f1c81d4d759be6bd5f10bf9a0fa0/README.md#how-to-use
 
-name: golangci-lint
+name: lint_golangci-lint
```
gaby, line 3: Why the rename?

.github/workflows/test.yml
```diff
@@ -28,11 +29,12 @@ jobs:
...
+          cache: false # setup-go v4 caches by default
           go-version: ${{ matrix.go-version }}
 
       - name: Run Test
         uses: nick-fields/retry@v2
         with:
           max_attempts: 3
           timeout_minutes: 15
-          command: go test ./... -v -race -count=1
+          command: go test -v -race ./...
```
gaby, line 40: Why is `count=1` removed?

.github/workflows/vulncheck.yml
```diff
@@ -27,9 +27,9 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v4
         with:
+          cache: false # setup-go v4 caches by default
```
gaby, line 30: Un-neccesary comment/change

.github/workflows/lint_go-vet.yml
leonklingele: Done based on the current `linter.yml` we use for fiber: https://github.com/gofiber/fiber/blob/fe65ff8dff718db4dfb6921108371c9e248ee208/.github/workflows/linter.yml#L10-L11

.github/workflows/lint_go-vet.yml
leonklingele: Actually, I prefer to have `lint_` as a prefix and the name of the linter as the suffix. This makes it easier to spot what the job is doing.

.github/workflows/lint_golangci-lint.yml
leonklingele: See above.

.github/workflows/test.yml
leonklingele: This makes tests run concurrently again. Not only to make test-runs faster but also to spot data races.
See the relevant commit message: https://github.com/gofiber/fiber/pull/2469/commits/1e268a25039d671fb6fb84684985eedd26fa3a19

This was added in https://github.com/gofiber/fiber/commit/9effdf829a79df49c24318bf753f8332d6e75375 (the commit message doesn't match what has actually changed, maybe @ReneWerner87 can explain why this is/was needed)

.github/workflows/vulncheck.yml
leonklingele: Added so we do not use a cache for any of the `setup-go@v4` actions. See the relevant commit and https://github.com/actions/setup-go/blob/main/action.yml#L15-L17

.github/workflows/test.yml
leonklingele: I also removed the `test` job retry logic in 6de93415ef9030f930118085e705ba2532c7d12c which allows us to spot data races immediately.

leonklingele: btw, this change just detected a data race! 🥳 4c4d3475a82fb618902d2594c5f12d0c8fede7d2

ReneWerner87: this prefix was used to disable the golang cache for testing, because in the past several tests went through it

for now, would get this construct

app.go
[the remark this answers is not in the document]
leonklingele: @ReneWerner87 what do you recommend to do about this?

helpers.go
```diff
@@ -268,10 +271,11 @@ func getOffer(header string, isAccepted func(spec, offer string) bool, offers ..
 	}
 
 	for _, offer := range offers {
-		if len(offer) == 0 {
+		if offer == "" {
```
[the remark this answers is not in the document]
leonklingele: It's more idiomatic and makes it easier to spot that we are checking a string and not a slice, map, etc.

See https://go-critic.com/overview.html#emptystringtest

app_test.go
```diff
@@ -1390,7 +1388,7 @@ func Benchmark_App_ETag_Weak(b *testing.B) {
 	app := New()
 	c := app.AcquireCtx(&fasthttp.RequestCtx{})
 	defer app.ReleaseCtx(c)
-	utils.AssertEqual(b, nil, c.Send([]byte("Hello, World!")))
+	utils.AssertEqual(b, nil, c.SendString("Hello, World!"))
```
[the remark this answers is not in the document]
leonklingele: We shouldn't convert strings to byte slices if the same method is also available as a version which directly accepts the string.

gaby (approved):
Lgtm

ctx_test.go
```diff
@@ -4731,10 +4744,10 @@ func Test_Ctx_BodyStreamWriter(t *testing.T) {
 	if err := resp.Read(br); err != nil {
 		t.Fatalf("Error when reading response: %s", err)
 	}
-	body := string(resp.Body())
-	expectedBody := "body writer line 1\nbody writer line 2\n"
-	if body != expectedBody {
-		t.Fatalf("unexpected body: %q. Expecting %q", body, expectedBody)
+
+	got, want := string(resp.Body()), "body writer line 1\nbody writer line 2\n"
+	if got != want {
```
efectn, line 4749: we should use utils.AssertEqual here

ctx_test.go
```diff
@@ -5051,9 +5056,9 @@ func Test_Ctx_RepeatParserWithSameStruct(t *testing.T) {
 	defer app.ReleaseCtx(c)
 
 	type Request struct {
-		QueryParam  string `query:"query_param"`
-		HeaderParam string `reqHeader:"header_param"`
-		BodyParam   string `json:"body_param" xml:"body_param" form:"body_param"`
+		QueryParam  string `                                    query:"query_param"`
```
efectn, line 5059: Seems wrong spacing

log/default_test.go
```diff
@@ -12,7 +12,7 @@ import (
 
 const work = "work"
 
-func initDefaultLogger() {
+func init() {
```
efectn, line 15: maybe should use TestMain instead of init?

middleware/basicauth/basicauth_test.go
```diff
@@ -39,10 +41,15 @@ func Test_Middleware_BasicAuth(t *testing.T) {
 		},
 	}))
 
-	//nolint:forcetypeassert,errcheck // TODO: Do not force-type assert
 	app.Get("/testauth", func(c *fiber.Ctx) error {
-		username := c.Locals("username").(string)
-		password := c.Locals("password").(string)
+		username, ok := c.Locals("username").(string)
+		if !ok {
+			t.Fatal("missing username")
```
efectn, line 47: better to use assertequal

ctx_test.go
leonklingele: Should be part of another MR.

ctx_test.go
leonklingele: Actually, it's done intentionally so one can better differentiate the various tags being used.

middleware/basicauth/basicauth_test.go
leonklingele: Should be part of another MR.

ctx_test.go
efectn: But the line width will increase. I'd prefer old layout

.github/workflows/test.yml
ReneWerner87: > for now, would get this construct

with "-count=1"

middleware/basicauth/basicauth_test.go
efectn: > Should be part of another MR.

It can be done in this PR since it was added here

leonklingele: Sorry for the delay. How would you use assertequal here? We want to retrieve the user + pass from locals.

log/default_test.go
leonklingele: How? 🤔

middleware/basicauth/basicauth_test.go
efectn: > Sorry for the delay. How would you use assertequal here? We want to retrieve the user + pass from locals.

`utils.AssertEqual(t, true, ok)`

log/default_test.go
efectn: > How? 🤔

Nevermind, current behavior seems OK

skydiver1980 (approved):

ReneWerner87:
@leonklingele hi can you look again, last time you were unfortunately not quite finished with the merges and now some places have been added

maybe we can also split the parts into individual pull requests so that we can merge them in a timely manner

think some customizations from you from these pull requests would be very good for our code

ReneWerner87:
Closed due to conflicts and because v2 will not receive any further customizations
You are welcome to customize v3 so that these changes are implemented there

PR #2475, opened 2023-05-23
 🚀 FEATURE: add queries function

dozheiny:
This PR adds a new function called `Queries()` to the `Ctx` struct. This function returns a map of query parameters and their values.

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -1058,7 +1058,7 @@ func (c *Ctx) Query(key string, defaultValue ...string) string {
 func (c *Ctx) Queries() map[string]string {
 	m := make(map[string]string, c.Context().QueryArgs().Len())
 	c.Context().QueryArgs().VisitAll(func(key, value []byte) {
-		m[c.app.getString(key)] = c.app.getString(value)
+		m[string(key)] = string(value)
 	})
 	return m
 }
--- a/ctx.go
+++ b/ctx.go
@@ -16,6 +16,7 @@ import (
 	"mime/multipart"
 	"net"
 	"net/http"
+	"net/url"
 	"path/filepath"
 	"reflect"
 	"strconv"
@@ -1052,6 +1053,26 @@ func (c *Ctx) Protocol() string {
 // Make copies or use the Immutable setting to use the value outside the Handler.
 func (c *Ctx) Query(key string, defaultValue ...string) string {
 	return defaultString(c.app.getString(c.fasthttp.QueryArgs().Peek(key)), defaultValue)
+}
+
+// Queries returns a map of query parameters and their values.
+// It returns an error if the query string is not valid.
+func (c *Ctx) Queries() (map[string]string, error) {
+	queryRaw := c.fasthttp.QueryArgs().String()
+	queries, err := url.ParseQuery(queryRaw)
+	if err != nil {
+		return nil, err
+	}
+
+	m := make(map[string]string)
+
+	for k := range queries {
+		for v := range queries[k] {
+			m[k] = queries[k][v]
+		}
+	}
+
+	return m, nil
 }
 
 // QueryInt returns integer value of key string parameter in the url.
--- a/ctx.go
+++ b/ctx.go
@@ -1057,18 +1057,24 @@ func (c *Ctx) Query(key string, defaultValue ...string) string {
 // Queries returns a map of query parameters and their values.
 //
 // GET /?name=alex&wanna_cake=2&id=
-// Queries()["name"] = "alex"
-// Queries()["wanna_cake"] = "2"
-// Queries()["id"] = ""
+// Queries()["name"] == "alex"
+// Queries()["wanna_cake"] == "2"
+// Queries()["id"] == ""
 //
 // GET /?field1=value1&field1=value2&field2=value3
-// Queries()["field1"] = "value2"
-// Queries()["field2"] = "value3"
+// Queries()["field1"] == "value2"
+// Queries()["field2"] == "value3"
 //
 // GET /?list_a=1&list_a=2&list_a=3&list_b[]=1&list_b[]=2&list_b[]=3&list_c=1,2,3
-// Queries()["list_a"] = "3"
-// Queries()["list_b[]"] = "3"
-// Queries()["list_c"] = "1,2,3"
+// Queries()["list_a"] == "3"
+// Queries()["list_b[]"] == "3"
+// Queries()["list_c"] == "1,2,3"
+//
+// GET /api/search?filters.author.name=John&filters.category.name=Technology&filters[customer][name]=Alice&filters[status]=pending
+// Queries()["filters.author.name"] == "John"
+// Queries()["filters.category.name"] == "Technology"
+// Queries()["filters[customer][name]"] == "Alice"
+// Queries()["filters[status]"] == "pending"
 func (c *Ctx) Queries() map[string]string {
 	m := make(map[string]string, c.Context().QueryArgs().Len())
 	c.Context().QueryArgs().VisitAll(func(key, value []byte) {
--- a/ctx.go
+++ b/ctx.go
@@ -1059,6 +1059,11 @@ func (c *Ctx) Query(key string, defaultValue ...string) string {
 // It returns an error if the query string is not valid.
 func (c *Ctx) Queries() (map[string]string, error) {
 	queryRaw := c.fasthttp.QueryArgs().String()
+	queryRaw, err := url.QueryUnescape(queryRaw)
+	if err != nil {
+		return nil, err
+	}
+
 	queries, err := url.ParseQuery(queryRaw)
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse query: %w", err)
--- a/ctx.go
+++ b/ctx.go
@@ -1055,6 +1055,20 @@ func (c *Ctx) Query(key string, defaultValue ...string) string {
 }
 
 // Queries returns a map of query parameters and their values.
+//
+// GET /?name=alex&wanna_cake=2&id=
+// Queries()["name"] = "alex"
+// Queries()["wanna_cake"] = "2"
+// Queries()["id"] = ""
+//
+// GET /?field1=value1&field1=value2&field2=value3
+// Queries()["field1"] = "value2"
+// Queries()["field2"] = "value3"
+//
+// GET /?list_a=1&list_a=2&list_a=3&list_b[]=1&list_b[]=2&list_b[]=3&list_c=1,2,3
+// Queries()["list_a"] = "3"
+// Queries()["list_b[]"] = "3"
+// Queries()["list_c"] = "1,2,3"
 func (c *Ctx) Queries() map[string]string {
 	m := make(map[string]string, c.Context().QueryArgs().Len())
 	c.Context().QueryArgs().VisitAll(func(key, value []byte) {
--- a/ctx.go
+++ b/ctx.go
@@ -1072,7 +1072,7 @@ func (c *Ctx) Query(key string, defaultValue ...string) string {
 func (c *Ctx) Queries() map[string]string {
 	m := make(map[string]string, c.Context().QueryArgs().Len())
 	c.Context().QueryArgs().VisitAll(func(key, value []byte) {
-		m[string(key)] = string(value)
+		m[c.app.getString(key)] = c.app.getString(value)
 	})
 	return m
 }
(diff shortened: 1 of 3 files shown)


gaby (changes requested):

ctx_test.go
```diff
@@ -3838,6 +3838,26 @@ func Benchmark_Ctx_SendString_B(b *testing.B) {
...
 
+func Test_Ctx_Queries(t *testing.T) {
+	t.Parallel()
+	app := New()
+	c := app.AcquireCtx(&fasthttp.RequestCtx{})
+	defer app.ReleaseCtx(c)
+
+	c.Request().SetBody([]byte(``))
+	c.Request().Header.SetContentType("")
+	c.Request().URI().SetQueryString("id=1&name=tom&hobby=basketball,football&favouriteDrinks=milo,coke,pepsi&alloc=&no=1")
```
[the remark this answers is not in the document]
dozheiny: Sure I'll add this tests too.

ctx.go
```diff
@@ -1054,6 +1055,31 @@ func (c *Ctx) Query(key string, defaultValue ...string) string {
 	return defaultString(c.app.getString(c.fasthttp.QueryArgs().Peek(key)), defaultValue)
 }
 
+// Queries returns a map of query parameters and their values.
+// It returns an error if the query string is not valid.
+func (c *Ctx) Queries() (map[string]string, error) {
+	queryRaw := c.fasthttp.QueryArgs().String()
+	queryRaw, err := url.QueryUnescape(queryRaw)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse query: %w", err)
```
gaby, line 1064: This error message should be different.

ctx.go
```diff
@@ -1054,6 +1055,31 @@ func (c *Ctx) Query(key string, defaultValue ...string) string {
...
+	}
+
+	queries, err := url.ParseQuery(queryRaw)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse query: %w", err)
+	}
+
+	m := make(map[string]string)
```
gaby, line 1072: I think this can be done with

 ```go
    m := make(map[string]string)

	for k, v := range queries {
		if len(v) > 0 {
			m[k] = v[len(v)-1]
		}
	}
```

File: ctx.go [hunk 4]
```diff
@@ -1054,6 +1055,31 @@ func (c *Ctx) Query(key string, defaultValue ...string) string {
 	return defaultString(c.app.getString(c.fasthttp.QueryArgs().Peek(key)), defaultValue)
 }
 
+// Queries returns a map of query parameters and their values.
+// It returns an error if the query string is not valid.
+func (c *Ctx) Queries() (map[string]string, error) {
+	queryRaw := c.fasthttp.QueryArgs().String()
+	queryRaw, err := url.QueryUnescape(queryRaw)
+	if err != nil {
```
gaby, line 1063: If the string is already escaped this will fail. It can be tested by adding a test that sets URLEscaped values

gaby (changes requested):
Thoughts

ctx.go
dozheiny: Sure.

ctx.go
dozheiny: > If the string is already escaped this will fail. It can be tested by adding a test that sets URLEscaped values

Do you mean I do this?
```go
queryRaw := c.fasthttp.QueryArgs().String()

queryRaw, err := url.QueryUnescape(queryRaw)
if err != nil {
    queryRaw = url.QueryEscape(queryRaw)
}
```

dozheiny: I think you mean to add tests for failing the `Queries` function.
I added it too.

gaby (approved):
LGTM

efectn (commented):
Can you add benchmark

dozheiny:
Sure, I'll add it

ctx.go
ReneWerner87: why so complicated ?

```go
m := make(map[string]string, c.Context().QueryArgs().Len())
c.Context().QueryArgs().VisitAll(func(key, value []byte) {
	m[string(key)] = string(value)
})
```
don't really know if we need this functionality for the 1 line either
otherwise we would have to provide a function for each of them
-> form values, query values, route params

dozheiny: > why so complicated ?
> 
> ```go
> m := make(map[string]string, c.Context().QueryArgs().Len())
> c.Context().QueryArgs().VisitAll(func(key, value []byte) {
> 	m[string(key)] = string(value)
> })
> ```
> 
> don't really know if we need this functionality for the 1 line either otherwise we would have to provide a function for each of them -> form values, query values, route params

Your right, But this code can't parse this query encoded URL: `id%3D1%26name%3Dtom%26hobby%3Dbasketball%2Cfootball%26favouriteDrinks%3Dmilo%2Ccoke%2Cpepsi%26alloc%3D%26no%3D1`

In this problem, we should unescape these encoded queries. Would you happen to have any suggestions?

ReneWerner87: why should we parse the encodede url? the reason for encoding is that these control characters are no longer interpreted as such

ReneWerner87: correct because it is invalid to process encoded query strings functionally
when transferring via http you have to stick to the standard, if you remove the control characters by encoding, then the not possible analysis of the string is exactly what the person wanted to achieve

ReneWerner87 (changes requested):
the body of the method still needs to be corrected
encoded strings should not be taken into account, because the control characters are therefore invalid

ctx.go
dozheiny: > correct because it is invalid to process encoded query strings functionally when transferring via http you have to stick to the standard, if you remove the control characters by encoding, then the not possible analysis of the string is exactly what the person wanted to achieve

Your right.

dozheiny:
> the body of the method still needs to be corrected encoded strings should not be taken into account, because the control characters are therefore invalid

Sure.

ctx.go
```diff
@@ -1054,6 +1054,16 @@ func (c *Ctx) Query(key string, defaultValue ...string) string {
 	return defaultString(c.app.getString(c.fasthttp.QueryArgs().Peek(key)), defaultValue)
 }
 
+// Queries returns a map of query parameters and their values.
+func (c *Ctx) Queries() map[string]string {
+	m := make(map[string]string, c.Context().QueryArgs().Len())
+	c.Context().QueryArgs().VisitAll(func(key, value []byte) {
+		m[string(key)] = string(value)
```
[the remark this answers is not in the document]
dozheiny: Your right. It's faster. I commit your request change.

cmd777 (approved):
LGTM

ctx_test.go
```diff
@@ -3838,6 +3838,48 @@ func Benchmark_Ctx_SendString_B(b *testing.B) {
...
+	utils.AssertEqual(t, "1", queries["id"])
+	utils.AssertEqual(t, "tom", queries["name"])
+	utils.AssertEqual(t, "basketball,football", queries["hobby"])
+	utils.AssertEqual(t, "milo,coke,pepsi", queries["favouriteDrinks"])
+	utils.AssertEqual(t, "", queries["alloc"])
+	utils.AssertEqual(t, "1", queries["no"])
+}
+
+// go test -v  -run=^$ -bench=BenchmarkTest_Ctx_Queries -benchmem -count=4
+func BenchmarkTest_Ctx_Queries(b *testing.B) {
```
leonklingele, line 3862:
```suggestion
// go test -v  -run=^$ -bench=Benchmark_Ctx_Queries -benchmem -count=4
func Benchmark_Ctx_Queries(b *testing.B) {
```

ctx.go
```diff
@@ -1054,6 +1054,16 @@ func (c *Ctx) Query(key string, defaultValue ...string) string {
...
 }
 
+// Queries returns a map of query parameters and their values.
+func (c *Ctx) Queries() map[string]string {
+	m := make(map[string]string, c.Context().QueryArgs().Len())
+	c.Context().QueryArgs().VisitAll(func(key, value []byte) {
+		m[c.app.getString(key)] = c.app.getString(value)
+	})
+
+	return m
```
leonklingele, line 1064:
```suggestion
	})
	return m
```

leonklingele (changes requested):
@dozheiny how does this handle URLs using the query parameter key multiple times, e.g. `?field1=value1&field1=value2&field2=value3`?

And what about query parameter array, e.g. `?list_a=1&list_a=2&list_a=3&list_b[]=1&list_b[]=2&list_b[]=3&list_c=1,2,3`?

Also see this link: https://stackoverflow.com/a/9547490

cmd777:
> how does this handle URLs using the query parameter key multiple times, e.g. `?field1=value1&field1=value2&field2=value3`?

It gets the latest value, like so:
```go
c.Request().URI().SetQueryString("field1=value1&field1=value2&field2=value3")
queries := c.Queries()
t.Log(queries)

ctx_test.go:2171: map[field1:value2 field2:value3]
```
```go
c.Request().URI().SetQueryString("field1=value1&field1=value2&field1=value1&field2=value3")
queries := c.Queries()
t.Log(queries)

ctx_test.go:2171: map[field1:value1 field2:value3]
```

> And what about query parameter array, e.g. `?list_a=1&list_a=2&list_a=3&list_b[]=1&list_b[]=2&list_b[]=3&list_c=1,2,3`?

Same as previously, it will take the latest value of list_a and and list_b[], though it does not detect list_b[] as an array:

```go
c.Request().URI().SetQueryString("list_a=1&list_a=2&list_a=3&list_b[]=1&list_b[]=2&list_b[]=3&list_c=1,2,3")
queries := c.Queries()
t.Log(queries)

ctx_test.go:2171: map[list_a:3 list_b[]:3 list_c:1,2,3]
```

dozheiny:
Sure. I document these examples.

dozheiny:
@leonklingele I revert two suggestion commits because Linter gets an error.

cmd777:
>  I revert two suggestion commits because Linter gets an error.

Linter failed because of timeout, not because of the code.
https://github.com/gofiber/fiber/actions/runs/5164476070/jobs/9303393316#step:4:25
https://github.com/gofiber/fiber/actions/runs/5164478717/jobs/9303397503#step:4:25

dozheiny:
> > I revert two suggestion commits because Linter gets an error.
> 
> Linter failed because of timeout, not because of the code. https://github.com/gofiber/fiber/actions/runs/5164476070/jobs/9303393316#step:4:25 https://github.com/gofiber/fiber/actions/runs/5164478717/jobs/9303397503#step:4:25

Good to know; thanks!

ReneWerner87:
1. should we split the depth structures  and overtake some functions from here
https://github.com/gofiber/fiber/blob/c955d76f5d45a251fde2197deef59a6ec07e6aa1/ctx.go#L1133-L1141

2. why did you revert the performance optimization with the app.getString function?

ReneWerner87:
We should think about the first point, because if we release it first, a change would be a breaking change.

I assume that the users also assume that the array values are split and stored in a deep structure, because otherwise you would have to do it yourself.
With the queryparser many also assumed this

efectn:
Returning map[string][]string seems better to me

ReneWerner87:
Or map[string]interface{}

dozheiny:
> Or map[string]interface{}

If that `map[string]interface{}`, a user should convert it into a string or array of series.
Returning `map[string][]string` makes more sense.

ReneWerner87:
And what happens for these examples?

```http
GET /api/posts?filters.author.name=John&filters.category.name=Technology
```

```http
GET /api/orders?filters[customer][name]=Alice&filters[status]=pending
```

ReneWerner87:
Some more interesting examples which we should consider

```http
GET /api/search?filters[category]=electronics&filters[tags][]=apple&filters[tags][]=orange&filters[tags][]=banana
```

```http
GET /api/search?tags=apple,orange,banana
 ```

```http
GET /api/search?filters[tags]=apple,orange,banana&filters[category][name]=fruits
```

```http
GET /api/search?filters.tags=apple,orange,banana&filters.category.name=fruits
```

```http
GET /api/search?filters.tags.0=apple&filters.tags.1=orange&filters.tags.2=banana&filters.category.name=fruits
```

```http
GET /api/search?filters[tags][0]=apple&filters[tags][1]=orange&filters[tags][2]=banana&filters[category][name]=fruits
```

Can you include these examples in the test and document the expectations

ReneWerner87 (approved):

dozheiny:
@ReneWerner87 shouldn't we make output of function to `map[string][]string`?
I'm just asking because you approve PR.

ReneWerner87:
Not really, I thought about it again and thought that it is ok, as long as we describe it well in the documentation.
Then you get exactly, the key value pairs which are to be recognized also in the query url, without any postprocessing

PR #2477, opened 2023-05-26
:bug: bug: fix middleware naming and returned values of group methods

### LINKED ISSUE #2467: 🐛 [Bug]: Naming for Middlewares
### Bug Description

Currently naming on middlewares just working for last element of RequestMethods slice (it's PATCH as default). That's not expected behavior and it needs to be fixed.

```
Path: /, Method: GET, Name: 
Path: /, Method: HEAD, Name: 
Path: /, Method: POST, Name: 
Path: /, Method: PUT, Name: 
Path: /, Method: DELETE, Name: 
Path: /, Method: CONNECT, Name: 
Path: /, Method: OPTIONS, Name: 
Path: /, Method: TRACE, Name: 
Path: /, Method: PATCH, Name: test
```

### How to Reproduce

1. Add middleware and name it.
2. Naming will only apply to PATCH method.

### Expected Behavior

All methods of middleware should have the name.

```
Path: /, Method: GET Name: test
Path: /, Method: HEAD, Name: test
Path: /, Method: POST, Name: test
Path: /, Method: PUT, Name: test
Path: /, Method: DELETE, Name: test
Path: /, Method: CONNECT, Name: test
Path: /, Method: OPTIONS, Name: test
Path: /, Method: TRACE, Name: test
Path: /, Method: PATCH, Name: test
```

### Fiber Version

After naming feature introduced

### Code Snippet (optional)

```go
package main

import (
	"fmt"

	"github.com/gofiber/fiber/v2"
)

func main() {
	app := fiber.New()

		return c.Next()
	}).Name("test")

	for _, route := range app.GetRoutes() {
		fmt.Printf("Path: %s, Method: %s, Name: %s\n", route.Path, route.Method, route.Name)
	}
}
```

### Checklist:

ytsruh:
## Description
Edited app.Name() method to add a name parameter to reach route unless already assigned. This required looping over app.stack which feels ineffecient but it was the only method of accessing all of the routes that I could find

Fixes # (issue)
Resolves #2467 

## Type of change

## Checklist:

## Commit formatting:

### DIFF
--- a/app.go
+++ b/app.go
@@ -611,15 +611,15 @@ func (app *App) Name(name string) Router {
 	app.mutex.Lock()
 	for _, routes := range app.stack {
 		for _, route := range routes {
-			latestGroup := app.latestRoute.group
-			if latestGroup != nil {
-				app.latestRoute.Name = latestGroup.name + name
-			} else {
-				if route.Path == app.latestRoute.path {
-					route.Name = name
-				}
+			if route.Path == app.latestRoute.path {
+				route.Name = name
 			}
 		}
+	}
+
+	latestGroup := app.latestRoute.group
+	if latestGroup != nil {
+		app.latestRoute.Name = latestGroup.name + name
 	}
 
 	if err := app.hooks.executeOnNameHooks(*app.latestRoute); err != nil {
--- a/app.go
+++ b/app.go
@@ -609,7 +609,8 @@ func (app *App) SetTLSHandler(tlsHandler *TLSHandler) {
 // Name Assign name to specific route.
 func (app *App) Name(name string) Router {
 	app.mutex.Lock()
-	fmt.Print(app.latestRoute)
+	defer app.mutex.Unlock()
+
 	for _, routes := range app.stack {
 		for _, route := range routes {
 			if route.Path == app.latestRoute.path {
@@ -625,7 +626,6 @@ func (app *App) Name(name string) Router {
 	if err := app.hooks.executeOnNameHooks(*app.latestRoute); err != nil {
 		panic(err)
 	}
-	app.mutex.Unlock()
 
 	return app
 }
@@ -759,12 +759,16 @@ func (app *App) Patch(path string, handlers ...Handler) Router {
 
 // Add allows you to specify a HTTP method to register a route
 func (app *App) Add(method, path string, handlers ...Handler) Router {
-	return app.register(method, path, nil, handlers...)
+	app.register(method, path, nil, handlers...)
+
+	return app
 }
 
 // Static will create a file server serving static files
 func (app *App) Static(prefix, root string, config ...Static) Router {
-	return app.registerStatic(prefix, root, config...)
+	app.registerStatic(prefix, root, config...)
+
+	return app
 }
 
 // All will register the handler on all HTTP methods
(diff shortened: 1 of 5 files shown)


app.go
```diff
@@ -609,12 +609,17 @@ func (app *App) SetTLSHandler(tlsHandler *TLSHandler) {
...
 	app.mutex.Lock()
-
-	latestGroup := app.latestRoute.group
-	if latestGroup != nil {
-		app.latestRoute.Name = latestGroup.name + name
-	} else {
-		app.latestRoute.Name = name
+	for _, routes := range app.stack {
+		for _, route := range routes {
+			latestGroup := app.latestRoute.group
```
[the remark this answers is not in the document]
ytsruh: Thanks for quick review. Will get this fixed (& the lint check) and update later this afternoon.

app_test.go
```diff
@@ -1803,3 +1803,31 @@ func TestApp_GetRoutes(t *testing.T) {
 		utils.AssertEqual(t, name, route.Name)
 	}
 }
+
+func TestRouteName(t *testing.T) {
+	home, named := "home", "named"
+	app := New()
+
+	app.Use(func(c *Ctx) error {
```
ReneWerner87, line 1811: what happens if we define 2 middlewares with names one after the other?

can you add this to your test case 

and maybe one middleware which has no name

ytsruh: Thanks for quick review. Yes I have 3 test cases, two with names & one unnamed - they all pass. Without looping over the stack the names were overwriting each other so this was the only way I found to resolve this problem

ytsruh:
Right updated to pass lint rules & moved the variable outside of the loop

app_test.go
ReneWerner87: the case for the middelwares with prefix matching is a little bit different

can you add additional route declarations
```go
app.Use(func(c *Ctx) error {
    return c.Next()
}).Name("corsMW")

app.Use(func(c *Ctx) error {
    return c.Next()
}).Name("compressMW")

app.Use(func(c *Ctx) error {
    return c.Next()
})// no name
```

and check the expectation

ReneWerner87: can you include these cases in our tests

would like to prevent that the different declarations of the different routes with the same path influence each other.

ytsruh: Sure will add in for review.

But to be honest, I think you'll need to reject this PR. Looping over the stack is going to cause the paths to overwrite each other so I do think this will need a different solution.

ytsruh: Right I've added in as requested, but I'm less confident this meets expectations now so understand if you prefer to reject.

app.go
```diff
@@ -609,12 +609,17 @@ func (app *App) SetTLSHandler(tlsHandler *TLSHandler) {
...
 	app.mutex.Lock()
+	for _, routes := range app.stack {
+		for _, route := range routes {
+			if route.Path == app.latestRoute.path {
+				route.Name = name
+			}
+		}
+	}
 
 	latestGroup := app.latestRoute.group
```
efectn, line 620: Can you add test cases for group and use

ytsruh: I added a test for Group locally and it failed. Sadly, I'm not able to get the group names to work as group was always <nil>

Sorry about this

ReneWerner87:
Will check it tomorrow

ReneWerner87 (approved):

PR #2486, opened 2023-05-30
🐛fix: update getOffer to consider quality and specificity

### LINKED ISSUE #2387: ctx.Accepts() quality values and specificity
This issue is a follow on to #2386 and #2383 

              > i have reversed the loop for all accepts methods, so that the arguments are passed first and then the header values are checked
> 
> @efectn @sixcolors @derkan do you agree that this was the right thing to do ? the logic of express leading to the outputs implies this to me

@ReneWerner87 good job, this make the existing test cases behave as expected.

Reading the ctx.Accepts docs my understanding is that:

```
The Accepts function in the GoFiber context is used to check if the specified extensions or content types are acceptable based on the request's Accept HTTP header.

It takes one or more strings as input arguments representing the content types or extensions and returns the most preferred one based on the Accept header.
```

Therefore, if my understanding of the intent of the Accept function is correct, the following should hold true:

```
c.Request().Header.Set(HeaderAccept, "text/plain, application/json;q=0.5, text/html, */*;q=0.1")
	utils.AssertEqual(t, "txt", c.Accepts("json", "txt"))

	c.Request().Header.Set(HeaderAccept, "text/plain;charset=UTF-8;q=0.8,text/plain;charset=US-ASCII;q=0.4,text/plain;q=0.2,text/html")
	utils.AssertEqual(t, "text/plain;charset=US-ASCII", c.Accepts("text/plain", "text/plain;charset=US-ASCII"))
	utils.AssertEqual(t, "text/plain;charset=UTF-8", c.Accepts("text/plain;charset=US-ASCII", "text/plain;charset=UTF-8", "text/plain"))
	utils.AssertEqual(t, "text/html", c.Accepts("text/plain", "text/html"))
```

For example: It is possible that a browser could use the Accept header with the value text/plain;charset=UTF-8;q=0.8,text/plain;charset=US-ASCII;q=0.4,text/html. While text/html is typically assumed to have a default quality value of 1 if not specified, the browser could still explicitly specify it with a quality value of 1, as it is allowed by the HTTP specifications.

Furthermore, the browser could be indicating to the server that it prefers text/plain with a charset of UTF-8 over text/plain with a charset of US-ASCII, but it is still willing to accept both with different quality values. This could be useful in scenarios where the server has content in both character sets and wants to serve the content that is preferred by the client.

However, it is worth noting that the Accept header sent by the browser may vary depending on the user agent and its configuration. Therefore, it is important for the server to properly parse and interpret the Accept header to determine the preferred content type for the client.

## Background/Refs

The relevant RFCs for the HTTP/1.1 specification are RFC 7231 and RFC 7232.

Regarding the q parameter and its default value of 1, Section 5.3.1 of RFC 7231 states:

```
If no "*" is present in an Accept header field, then all
media types are acceptable, assuming that any media range
syntax is understood by the recipient.  If an Accept header
field is present, and if the server cannot send a response
which is acceptable according to the combined Accept field
value, then the server SHOULD send a 406 (not acceptable)
response.

[...]

Each media-range might be followed by one or more accept-
params, beginning with the "q" parameter for indicating a
relative quality factor.  The first "q" parameter (if any)
separates the media-range parameter(s) from the accept-
params.  Quality factors allow the user or user agent to
indicate the relative degree of preference for that media-
type, on a scale of 0 to 1.
```

Regarding the order of the MIME types in the Accept header, Section 5.3.2 of RFC 7231 states:

```
The order in which the content-codings are listed in the
header field is significant.  Typically, the first encoding
that is applied is the one listed as the first content-coding;
subsequent encodings are applied in the order listed.  The
order of charset values in the header field is significant,
whereas the order of language-tag values is not.  Specifically,
it does not mean that the first charset listed is preferred,
but rather that it represents the default charset if none is
otherwise specified.

An "Accept" header field can be present in a request to
indicate the formats that are acceptable for the response.
Field content negotiation is described in Section 5.3.  An
origin server that is capable of generating a response
message which is subject to content negotiation MAY vary the
response that is sent, based on the value of the Accept header
field supplied in the request.  For example, the server might
include a different Content-Type header field, or might
omit the content entirely.

The order in which the media type values are listed is
significant.  The first media type that is listed as acceptable
determines the format of the entity-body that is returned.  A
more precise or generally usable media type listed earlier in
the Accept header field SHOULD be given preference over those
that are listed later.
```

MDN on Quality Values https://developer.mozilla.org/en-US/docs/Glossary/Quality_values

```
The importance of a value is marked by the suffix ';q=' immediately followed by a value between 0 and 1 included, with up to three decimal digits, the highest value denoting the highest priority. When not present, the default value is 1.

[Examples](https://developer.mozilla.org/en-US/docs/Glossary/Quality_values#examples)
The following syntax

text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8

indicates the order of priority:

[...]

If there is no priority defined for the first two values, the order in the list is irrelevant. Nevertheless, with the same quality, more specific values have priority over less specific ones:

text/html;q=0.8,text/*;q=0.8,*/*;q=0.8

Value	Priority
text/html	0.8 (but totally specified)
text/*	0.8 (partially specified)
*/*	0.8 (not specified)

Some syntax, like the one of Accept, allow additional specifiers like text/html;level=1. These increase the specificity of the value. Their use is extremely rare.
```

_Originally posted by @sixcolors in https://github.com/gofiber/fiber/issues/2386#issuecomment-1485081978_

sixcolors:
## Description

Updates the getOffer helper function so that content negotiation follows https://www.rfc-editor.org/rfc/rfc9110#name-content-negotiation-fields, specifically that quality values are considered (also where q=0 type is not acceptable) and that specificity is also considered in a tie, as a final tie breaker client order is used.

Fixes #2387 

## Type of change

## Checklist:

## Commit formatting:

### DIFF
--- a/ctx_test.go
+++ b/ctx_test.go
@@ -45,7 +45,10 @@ func Test_Ctx_Accepts(t *testing.T) {
 	utils.AssertEqual(t, "", c.Accepts())
 	utils.AssertEqual(t, ".xml", c.Accepts(".xml"))
 	utils.AssertEqual(t, "", c.Accepts(".john"))
-	utils.AssertEqual(t, "application/xhtml+xml", c.Accepts("application/xml", "application/xml+rss", "application/yml", "application/xhtml+xml"))
+	utils.AssertEqual(t, "application/xhtml+xml", c.Accepts("application/xml", "application/xml+rss", "application/yml", "application/xhtml+xml"), "must use client-preferred mime type")
+
+	c.Request().Header.Set(HeaderAccept, "application/json, text/plain, */*;q=0")
+	utils.AssertEqual(t, "", c.Accepts("html"), "must treat */*;q=0 as not acceptable")
 
 	c.Request().Header.Set(HeaderAccept, "text/*, application/json")
 	utils.AssertEqual(t, "html", c.Accepts("html"))
--- a/ctx_test.go
+++ b/ctx_test.go
@@ -45,7 +45,7 @@ func Test_Ctx_Accepts(t *testing.T) {
 	utils.AssertEqual(t, "", c.Accepts())
 	utils.AssertEqual(t, ".xml", c.Accepts(".xml"))
 	utils.AssertEqual(t, "", c.Accepts(".john"))
-	utils.AssertEqual(t, "application/xhtml+xml", c.Accepts("application/xml", "application/xml+rss", "application/yml", "application/xhtml+xml"), "must use client-preferred mime type")
+	utils.AssertEqual(t, "application/xhtml+xml", c.Accepts("application/xml", "application/xml+rss", "application/yaml", "application/xhtml+xml"), "must use client-preferred mime type")
 
 	c.Request().Header.Set(HeaderAccept, "application/json, text/plain, */*;q=0")
 	utils.AssertEqual(t, "", c.Accepts("html"), "must treat */*;q=0 as not acceptable")
--- a/ctx_test.go
+++ b/ctx_test.go
@@ -45,6 +45,7 @@ func Test_Ctx_Accepts(t *testing.T) {
 	utils.AssertEqual(t, "", c.Accepts())
 	utils.AssertEqual(t, ".xml", c.Accepts(".xml"))
 	utils.AssertEqual(t, "", c.Accepts(".john"))
+	utils.AssertEqual(t, "application/xhtml+xml", c.Accepts("application/xml", "application/xml+rss", "application/yml", "application/xhtml+xml"))
 
 	c.Request().Header.Set(HeaderAccept, "text/*, application/json")
 	utils.AssertEqual(t, "html", c.Accepts("html"))
--- a/helpers.go
+++ b/helpers.go
@@ -295,7 +295,9 @@ func getOffer(header string, isAccepted func(spec, offer string) bool, offers ..
 		if factorSign := strings.IndexByte(spec, ';'); factorSign != -1 {
 			factor := utils.Trim(spec[factorSign+1:], ' ')
 			if strings.HasPrefix(factor, "q=") {
-				quality, _ = strconv.ParseFloat(factor[2:], 64)
+				if q, err := strconv.ParseFloat(factor[2:], 64); err != nil {
+					quality = q
+				}
 			}
 			spec = spec[:factorSign]
 		}
--- a/helpers.go
+++ b/helpers.go
@@ -15,6 +15,7 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
@@ -336,7 +337,14 @@ func getOffer(header string, isAccepted func(spec, offer string) bool, offers ..
 	}
 
 	// Sort accepted types by quality and specificity
-	if len(acceptedTypes) > 1 {
+	if len(acceptedTypes) > 20 {
+		sort.SliceStable(acceptedTypes, func(i, j int) bool {
+			if acceptedTypes[i].quality == acceptedTypes[j].quality {
+				return acceptedTypes[i].specificity > acceptedTypes[j].specificity
+			}
+			return acceptedTypes[i].quality > acceptedTypes[j].quality
+		})
+	} else if len(acceptedTypes) > 1 {
 		// Insertion sort
 		changes := false
 		for i := 1; i < len(acceptedTypes); i++ {
--- a/helpers.go
+++ b/helpers.go
@@ -15,7 +15,6 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
-	"sort"
 	"strconv"
 	"strings"
 	"time"
@@ -337,14 +336,7 @@ func getOffer(header string, isAccepted func(spec, offer string) bool, offers ..
 	}
 
 	// Sort accepted types by quality and specificity
-	if len(acceptedTypes) > 6 {
-		sort.SliceStable(acceptedTypes, func(i, j int) bool {
-			if acceptedTypes[i].quality == acceptedTypes[j].quality {
-				return acceptedTypes[i].specificity > acceptedTypes[j].specificity
-			}
-			return acceptedTypes[i].quality > acceptedTypes[j].quality
-		})
-	} else if len(acceptedTypes) > 1 {
+	if len(acceptedTypes) > 1 {
 		// Insertion sort
 		changes := false
 		for i := 1; i < len(acceptedTypes); i++ {
--- a/helpers.go
+++ b/helpers.go
@@ -337,16 +337,24 @@ func getOffer(header string, isAccepted func(spec, offer string) bool, offers ..
 
 	// Sort accepted types by quality and specificity
 	if len(acceptedTypes) > 1 {
-		// Insertion sort
 		for i := 1; i < len(acceptedTypes); i++ {
-			for j := i; j > 0; j-- {
-				if acceptedTypes[j-1].quality < acceptedTypes[j].quality {
-					acceptedTypes[j-1], acceptedTypes[j] = acceptedTypes[j], acceptedTypes[j-1]
-				} else if acceptedTypes[j-1].quality == acceptedTypes[j].quality && acceptedTypes[j-1].specificity < acceptedTypes[j].specificity {
-					acceptedTypes[j-1], acceptedTypes[j] = acceptedTypes[j], acceptedTypes[j-1]
+			swap := false
+			lo, hi := 0, i-1
+			for lo <= hi {
+				mid := (lo + hi) / 2
+				if acceptedTypes[i].quality > acceptedTypes[mid].quality ||
+					(acceptedTypes[i].quality == acceptedTypes[mid].quality && acceptedTypes[i].specificity > acceptedTypes[mid].specificity) {
+					lo = mid + 1
 				} else {
-					break
+					hi = mid - 1
 				}
+			}
+			for j := i; j > lo; j-- {
+				acceptedTypes[j-1], acceptedTypes[j] = acceptedTypes[j], acceptedTypes[j-1]
+				swap = true
+			}
+			if !swap {
+				break
 			}
 		}
 	}
--- a/helpers.go
+++ b/helpers.go
@@ -15,7 +15,6 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
-	"strconv"
 	"strings"
 	"time"
 	"unsafe"
@@ -298,7 +297,7 @@ func getOffer(header string, isAccepted func(spec, offer string) bool, offers ..
 		if factorSign := strings.IndexByte(spec, ';'); factorSign != -1 {
 			factor := utils.Trim(spec[factorSign+1:], ' ')
 			if strings.HasPrefix(factor, "q=") {
-				if q, err := strconv.ParseFloat(factor[2:], 64); err == nil {
+				if q, err := fasthttp.ParseUfloat([]byte(factor[2:])); err == nil {
 					quality = q
 				}
 			}
--- a/helpers.go
+++ b/helpers.go
@@ -272,13 +272,16 @@ func getOffer(header string, isAccepted func(spec, offer string) bool, offers ..
 		spec        string
 		quality     float64
 		specificity int
+		order       int
 	}
 
 	// Parse header and get accepted types with their quality and specificity
 	// See: https://www.rfc-editor.org/rfc/rfc9110#name-content-negotiation-fields
-	spec, commaPos := "", 0
+	spec, commaPos, order := "", 0, 0
 	acceptedTypes := make([]acceptedType, 0, 20)
 	for len(header) > 0 {
+		order++
+
 		// Skip spaces
 		header = utils.TrimLeft(header, ' ')
 
@@ -295,7 +298,7 @@ func getOffer(header string, isAccepted func(spec, offer string) bool, offers ..
 		if factorSign := strings.IndexByte(spec, ';'); factorSign != -1 {
 			factor := utils.Trim(spec[factorSign+1:], ' ')
 			if strings.HasPrefix(factor, "q=") {
-				if q, err := strconv.ParseFloat(factor[2:], 64); err != nil {
+				if q, err := strconv.ParseFloat(factor[2:], 64); err == nil {
 					quality = q
 				}
 			}
@@ -327,7 +330,7 @@ func getOffer(header string, isAccepted func(spec, offer string) bool, offers ..
 		}
 
 		// Add to accepted types
-		acceptedTypes = append(acceptedTypes, acceptedType{spec, quality, specificity})
+		acceptedTypes = append(acceptedTypes, acceptedType{spec, quality, specificity, order})
 
 		// Next
 		if commaPos != -1 {
@@ -337,15 +340,16 @@ func getOffer(header string, isAccepted func(spec, offer string) bool, offers ..
 		}
 	}
 
-	// Sort accepted types by quality and specificity
+	// Sort accepted types by quality and specificity, preserving order of equal elements
 	if len(acceptedTypes) > 1 {
 		for i := 1; i < len(acceptedTypes); i++ {
 			swap := false
 			lo, hi := 0, i-1
 			for lo <= hi {
 				mid := (lo + hi) / 2
-				if acceptedTypes[i].quality > acceptedTypes[mid].quality ||
-					(acceptedTypes[i].quality == acceptedTypes[mid].quality && acceptedTypes[i].specificity > acceptedTypes[mid].specificity) {
+				if acceptedTypes[i].quality < acceptedTypes[mid].quality ||
+					(acceptedTypes[i].quality == acceptedTypes[mid].quality && acceptedTypes[i].specificity < acceptedTypes[mid].specificity) ||
+					(acceptedTypes[i].quality == acceptedTypes[mid].quality && acceptedTypes[i].specificity == acceptedTypes[mid].specificity && acceptedTypes[i].order > acceptedTypes[mid].order) {
 					lo = mid + 1
 				} else {
 					hi = mid - 1
--- a/helpers.go
+++ b/helpers.go
@@ -300,7 +300,7 @@ func getOffer(header string, isAccepted func(spec, offer string) bool, offers ..
 			spec = spec[:factorSign]
 		}
 
-		if quality == 0 {
+		if quality == 0.0 {
 			// Skip this spec
 			if commaPos != -1 {
 				header = header[commaPos+1:]
--- a/helpers.go
+++ b/helpers.go
@@ -25,6 +25,16 @@ import (
 	"github.com/valyala/fasthttp"
 )
 
+// acceptType is a struct that holds the parsed value of an Accept header
+// along with quality, specificity, and order.
+// used for sorting accept headers.
+type acceptedType struct {
+	spec        string
+	quality     float64
+	specificity int
+	order       int
+}
+
 // getTLSConfig returns a net listener's tls config
 func getTLSConfig(ln net.Listener) *tls.Config {
 	// Get listener type
@@ -261,17 +271,14 @@ func acceptsOfferType(spec, offerType string) bool {
 
 // getOffer return valid offer for header negotiation
 func getOffer(header string, isAccepted func(spec, offer string) bool, offers ...string) string {
+	if isAccepted == nil {
+		return ""
+	}
 	if len(offers) == 0 {
 		return ""
-	} else if header == "" {
-		return offers[0]
 	}
-
-	type acceptedType struct {
-		spec        string
-		quality     float64
-		specificity int
-		order       int
+	if header == "" {
+		return offers[0]
 	}
 
 	// Parse header and get accepted types with their quality and specificity
@@ -339,29 +346,9 @@ func getOffer(header string, isAccepted func(spec, offer string) bool, offers ..
 		}
 	}
 
-	// Sort accepted types by quality and specificity, preserving order of equal elements
 	if len(acceptedTypes) > 1 {
-		for i := 1; i < len(acceptedTypes); i++ {
-			swap := false
-			lo, hi := 0, i-1
-			for lo <= hi {
-				mid := (lo + hi) / 2
-				if acceptedTypes[i].quality < acceptedTypes[mid].quality ||
-					(acceptedTypes[i].quality == acceptedTypes[mid].quality && acceptedTypes[i].specificity < acceptedTypes[mid].specificity) ||
-					(acceptedTypes[i].quality == acceptedTypes[mid].quality && acceptedTypes[i].specificity == acceptedTypes[mid].specificity && acceptedTypes[i].order > acceptedTypes[mid].order) {
-					lo = mid + 1
-				} else {
-					hi = mid - 1
-				}
-			}
-			for j := i; j > lo; j-- {
-				acceptedTypes[j-1], acceptedTypes[j] = acceptedTypes[j], acceptedTypes[j-1]
-				swap = true
-			}
-			if !swap {
-				break
-			}
-		}
+		// Sort accepted types by quality and specificity, preserving order of equal elements
+		sortAcceptedTypes(&acceptedTypes)
 	}
 
 	// Find the first offer that matches the accepted types
@@ -377,6 +364,35 @@ func getOffer(header string, isAccepted func(spec, offer string) bool, offers ..
 	}
 
 	return ""
+}
+
+// sortAcceptedTypes sorts accepted types by quality and specificity, preserving order of equal elements
+//
+// Parameters are not supported, they are ignored when sorting by specificity.
+//
+// See: https://www.rfc-editor.org/rfc/rfc9110#name-content-negotiation-fields
+func sortAcceptedTypes(at *[]acceptedType) {
+	if at == nil || len(*at) < 2 {
+		return
+	}
+	acceptedTypes := *at
+
+	for i := 1; i < len(acceptedTypes); i++ {
+		lo, hi := 0, i-1
+		for lo <= hi {
+			mid := (lo + hi) / 2
+			if acceptedTypes[i].quality < acceptedTypes[mid].quality ||
+				(acceptedTypes[i].quality == acceptedTypes[mid].quality && acceptedTypes[i].specificity < acceptedTypes[mid].specificity) ||
+				(acceptedTypes[i].quality == acceptedTypes[mid].quality && acceptedTypes[i].specificity == acceptedTypes[mid].specificity && acceptedTypes[i].order > acceptedTypes[mid].order) {
+				lo = mid + 1
+			} else {
+				hi = mid - 1
+			}
+		}
+		for j := i; j > lo; j-- {
+			acceptedTypes[j-1], acceptedTypes[j] = acceptedTypes[j], acceptedTypes[j-1]
+		}
+	}
 }
 
 func matchEtag(s, etag string) bool {
--- a/helpers.go
+++ b/helpers.go
@@ -301,7 +301,7 @@ func getOffer(header string, isAccepted func(spec, offer string) bool, offers ..
 		if factorSign := strings.IndexByte(spec, ';'); factorSign != -1 {
 			factor := utils.Trim(spec[factorSign+1:], ' ')
 			if strings.HasPrefix(factor, "q=") {
-				if q, err := fasthttp.ParseUfloat([]byte(factor[2:])); err == nil {
+				if q, err := fasthttp.ParseUfloat(utils.UnsafeBytes(factor[2:])); err == nil {
 					quality = q
 				}
 			}
--- a/helpers.go
+++ b/helpers.go
@@ -275,7 +275,8 @@ func getOffer(header string, isAccepted func(spec, offer string) bool, offers ..
 		specificity int
 	}
 
-	// Parse header
+	// Parse header and get accepted types with their quality and specificity
+	// See: https://www.rfc-editor.org/rfc/rfc9110#name-content-negotiation-fields
 	spec, commaPos := "", 0
 	acceptedTypes := make([]acceptedType, 0)
 	for len(header) > 0 {
@@ -300,8 +301,9 @@ func getOffer(header string, isAccepted func(spec, offer string) bool, offers ..
 			spec = spec[:factorSign]
 		}
 
+		// Skip if quality is 0.0
+		// See: https://www.rfc-editor.org/rfc/rfc9110#quality.values
 		if quality == 0.0 {
-			// Skip this spec
 			if commaPos != -1 {
 				header = header[commaPos+1:]
 			} else {
--- a/helpers.go
+++ b/helpers.go
@@ -271,9 +271,6 @@ func acceptsOfferType(spec, offerType string) bool {
 
 // getOffer return valid offer for header negotiation
 func getOffer(header string, isAccepted func(spec, offer string) bool, offers ...string) string {
-	if isAccepted == nil {
-		return ""
-	}
 	if len(offers) == 0 {
 		return ""
 	}
--- a/helpers.go
+++ b/helpers.go
@@ -300,6 +300,16 @@ func getOffer(header string, isAccepted func(spec, offer string) bool, offers ..
 			spec = spec[:factorSign]
 		}
 
+		if quality == 0 {
+			// Skip this spec
+			if commaPos != -1 {
+				header = header[commaPos+1:]
+			} else {
+				break
+			}
+			continue
+		}
+
 		// Get specificity
 		specificity := 0
 		// check for wildcard this could be a mime */* or a wildcard character *
--- a/helpers.go
+++ b/helpers.go
@@ -15,6 +15,8 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
+	"sort"
+	"strconv"
 	"strings"
 	"time"
 	"unsafe"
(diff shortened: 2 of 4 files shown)


sixcolors:
Considering client header priorities will have a performance impact. See discussion in #2387

ReneWerner87:
will check this tomorrow

helpers.go
```diff
@@ -267,29 +268,111 @@ func getOffer(header string, isAccepted func(spec, offer string) bool, offers ..
...
 
-			if commaPos != -1 {
-				header = header[commaPos+1:]
+	// Find the first offer that matches the accepted types
+	for _, acceptedType := range acceptedTypes {
+		for _, offer := range offers {
+			if len(offer) == 0 {
+				continue
+			}
+			if isAccepted(acceptedType.spec, offer) {
```
renanbastos93, line 374: Maybe we can optimize this logic. What do you think?
```suggestion
			if len(offer) != 0 && isAccepted(acceptedType.spec, offer) {
```

helpers.go
```diff
@@ -267,29 +268,111 @@ func getOffer(header string, isAccepted func(spec, offer string) bool, offers ..
...
+				return offer
```
renanbastos93, line 375: also, I have thought that. Does that make sense?
```suggestion
			if len(offer) == 0 || !isAccepted(acceptedType.spec, offer) {
				continue
			}
			return offer
```

renanbastos93 (commented):
Awesome, you did a good job, but I've considered two ideas to optimize your loop. Does that makes sense to you?

ReneWerner87:
@sixcolors looks very good 🫶

now also understand why it is slower
you have to parse all headers first and can't run through them piece by piece because you need the order
before only the first ones were parsed and then matched with the customer's input

-> so the slight performance loss is okay for me 
(you could also get rid of it by checking if the whole string contains "q=" at the beginning and then using the pass-through logic from before -> where we only parse parts)

can you still check the annotations in the PR

otherwise it would be good if we add more test scenarios with multiple quality factors and maybe a benchmark for the getOffer method, which is the core for many others 

summary from above:
- [ ] check the PR hints 
- [ ] consider the possibility to run a string without quality identifier with the old loop concept 
- [ ] add some test scenarios with strings with several qualities, for the main and getOffer method
- [] add a benchmark for the core method(getOffer) so that we can test optimizations in the future

helpers.go
```diff
@@ -267,29 +268,111 @@ func getOffer(header string, isAccepted func(spec, offer string) bool, offers ..
 		return offers[0]
 	}
 
-	for _, offer := range offers {
-		if len(offer) == 0 {
-			continue
+	type acceptedType struct {
```
ReneWerner87, line 271: declare it outside

helpers.go
```diff
@@ -267,29 +268,111 @@ func getOffer(header string, isAccepted func(spec, offer string) bool, offers ..
...
+		spec        string
+		quality     float64
+		specificity int
+		order       int
+	}
+
+	// Parse header and get accepted types with their quality and specificity
+	// See: https://www.rfc-editor.org/rfc/rfc9110#name-content-negotiation-fields
+	spec, commaPos, order := "", 0, 0
+	acceptedTypes := make([]acceptedType, 0, 20)
```
ReneWerner87, line 281: can we use syncPool to optimize the initialization?

helpers.go
```diff
@@ -267,29 +268,111 @@ func getOffer(header string, isAccepted func(spec, offer string) bool, offers ..
...
+		} else {
+			spec = utils.TrimLeft(header, ' ')
+		}
+
+		// Get quality
+		quality := 1.0
+		if factorSign := strings.IndexByte(spec, ';'); factorSign != -1 {
+			factor := utils.Trim(spec[factorSign+1:], ' ')
+			if strings.HasPrefix(factor, "q=") {
+				if q, err := strconv.ParseFloat(factor[2:], 64); err == nil {
```
ReneWerner87, line 301: can you test this

https://github.com/valyala/fasthttp/blob/f0865d4aabbbea51a81d56ab31a3de2dfc5a9b05/bytesconv.go#LL201C15-L201C15

maybe its faster

helpers.go
```diff
@@ -267,29 +268,111 @@ func getOffer(header string, isAccepted func(spec, offer string) bool, offers ..
...
+		// Next
+		if commaPos != -1 {
+			header = header[commaPos+1:]
+		} else {
+			break
+		}
+	}
+
+	// Sort accepted types by quality and specificity, preserving order of equal elements
+	if len(acceptedTypes) > 1 {
```
ReneWerner87, line 344: can we create a function from this to slim down the function body?

i know in golang it is common practice to have big functions and files, but i am a friend of the clean code concept with small parts

If it does not add allocation, then please convert to separat function
(same with the other parts, when possible)

ReneWerner87 (changes requested):

helpers.go
sixcolors: seems to be

helpers.go
sixcolors: done

sixcolors:
I also tried the sort using quick sort....

```go
func sortAcceptedTypes(acceptedTypes []acceptedType) {
	if len(acceptedTypes) < 2 {
		return
	}

	// Choose a pivot element
	pivotIndex := len(acceptedTypes) / 2
	pivot := acceptedTypes[pivotIndex]

	// Partition the slice into elements less than and greater than the pivot
	left, right := 0, len(acceptedTypes)-1
	acceptedTypes[pivotIndex], acceptedTypes[right] = acceptedTypes[right], acceptedTypes[pivotIndex]
	for i := 0; i < right; i++ {
		if acceptedTypes[i].quality < pivot.quality ||
			(acceptedTypes[i].quality == pivot.quality && acceptedTypes[i].specificity < pivot.specificity) ||
			(acceptedTypes[i].quality == pivot.quality && acceptedTypes[i].specificity == pivot.specificity && acceptedTypes[i].order > pivot.order) {
			acceptedTypes[i], acceptedTypes[left] = acceptedTypes[left], acceptedTypes[i]
			left++
		}
	}
	acceptedTypes[left], acceptedTypes[right] = acceptedTypes[right], acceptedTypes[left]

	// Recursively sort the left and right partitions
	sortAcceptedTypes(acceptedTypes[:left])
	sortAcceptedTypes(acceptedTypes[left+1:])
}
```

```console
-----------------------------------------------------------------------------------------------------------------------------------------------------------------
~/Documents/GitHub/sixcolors/fiber (2387-ctx-accepts-get-offer-q-and-specificity*) » go test -v -run=^$ -bench=Benchmark_Ctx_Accepts -benchmem -count=4
goos: darwin
goarch: arm64
Benchmark_Ctx_Accepts
Benchmark_Ctx_Accepts/run-[]string{".xml"}
Benchmark_Ctx_Accepts/run-[]string{".xml"}-12            9538138               110.2 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{".xml"}-12           10759770               110.2 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{".xml"}-12           10740316               112.0 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{".xml"}-12           10856568               111.1 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{"json",_"xml"}
Benchmark_Ctx_Accepts/run-[]string{"json",_"xml"}-12     9015339               130.6 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{"json",_"xml"}-12     9137023               130.4 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{"json",_"xml"}-12     9091927               130.6 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{"json",_"xml"}-12     9090036               130.6 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{"application/json",_"application/xml"}
Benchmark_Ctx_Accepts/run-[]string{"application/json",_"application/xml"}-12            10215805               116.2 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{"application/json",_"application/xml"}-12            10304235               115.8 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{"application/json",_"application/xml"}-12            10321711               115.9 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{"application/json",_"application/xml"}-12            10256577               115.9 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_AcceptsCharsets
Benchmark_Ctx_AcceptsCharsets-12                                                        13062492                87.74 ns/op            0 B/op          0 allocs/op
Benchmark_Ctx_AcceptsCharsets-12                                                        13360604                87.04 ns/op            0 B/op          0 allocs/op
Benchmark_Ctx_AcceptsCharsets-12                                                        13835763                88.70 ns/op            0 B/op          0 allocs/op
Benchmark_Ctx_AcceptsCharsets-12                                                        13250017                87.80 ns/op            0 B/op          0 allocs/op
Benchmark_Ctx_AcceptsEncodings
Benchmark_Ctx_AcceptsEncodings-12                                                        9873356               119.9 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_AcceptsEncodings-12                                                        9873092               119.9 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_AcceptsEncodings-12                                                        9781816               119.6 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_AcceptsEncodings-12                                                        9902232               119.6 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_AcceptsLanguages
Benchmark_Ctx_AcceptsLanguages-12                                                        5623364               213.6 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_AcceptsLanguages-12                                                        5690282               197.9 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_AcceptsLanguages-12                                                        5633530               198.9 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_AcceptsLanguages-12                                                        5958943               212.4 ns/op             0 B/op          0 allocs/op
PASS
ok      github.com/gofiber/fiber/v2     31.752s
-----------------------------------------------------------------------------------------------------------------------------------------------------------------
```

vs modified version of the binary insertion sort

```console
~/Documents/GitHub/sixcolors/fiber (2387-ctx-accepts-get-offer-q-and-specificity*) » go test -v -run=^$ -bench=Benchmark_Ctx_Accepts -benchmem -count=4
goos: darwin
goarch: arm64
Benchmark_Ctx_Accepts
Benchmark_Ctx_Accepts/run-[]string{".xml"}
Benchmark_Ctx_Accepts/run-[]string{".xml"}-12            8210182               131.7 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{".xml"}-12            9089493               131.3 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{".xml"}-12            9100646               131.2 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{".xml"}-12            9059756               131.2 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{"json",_"xml"}
Benchmark_Ctx_Accepts/run-[]string{"json",_"xml"}-12     6723105               178.3 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{"json",_"xml"}-12     6703312               178.2 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{"json",_"xml"}-12     6718827               177.7 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{"json",_"xml"}-12     6729043               178.0 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{"application/json",_"application/xml"}
Benchmark_Ctx_Accepts/run-[]string{"application/json",_"application/xml"}-12             9047026               132.0 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{"application/json",_"application/xml"}-12             9072474               132.2 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{"application/json",_"application/xml"}-12             8990721               131.9 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{"application/json",_"application/xml"}-12             8912109               132.0 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_AcceptsCharsets
Benchmark_Ctx_AcceptsCharsets-12                                                        14777224                70.21 ns/op            0 B/op          0 allocs/op
Benchmark_Ctx_AcceptsCharsets-12                                                        16863188                76.50 ns/op            0 B/op          0 allocs/op
Benchmark_Ctx_AcceptsCharsets-12                                                        15745413                73.41 ns/op            0 B/op          0 allocs/op
Benchmark_Ctx_AcceptsCharsets-12                                                        16351455                73.09 ns/op            0 B/op          0 allocs/op
Benchmark_Ctx_AcceptsEncodings
Benchmark_Ctx_AcceptsEncodings-12                                                       10800014               108.1 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_AcceptsEncodings-12                                                       11033839               108.2 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_AcceptsEncodings-12                                                       10968282               110.6 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_AcceptsEncodings-12                                                       10794322               108.5 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_AcceptsLanguages
Benchmark_Ctx_AcceptsLanguages-12                                                        7176621               172.7 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_AcceptsLanguages-12                                                        6958363               170.6 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_AcceptsLanguages-12                                                        6797910               165.7 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_AcceptsLanguages-12                                                        6938980               166.0 ns/op             0 B/op          0 allocs/op
PASS
ok      github.com/gofiber/fiber/v2     31.923s
```

helpers.go
sixcolors: to me that's more compact but less redable, should make zero performance difference since it would likley product the same binary...

helpers.go
ReneWerner87: Can you try https://github.com/gofiber/fiber/blob/master/utils/convert_s2b_new.go#L11

UnsafeBytes method instead of the other convert before you use the fasthttp method

helpers.go
sixcolors: done

helpers.go
sixcolors: It's slower and adds 2 allocs:

syncPool:

```console
~/Documents/GitHub/sixcolors/fiber (2387-ctx-accepts-get-offer-q-and-specificity*) » go test -v -run=^$ -bench=Benchmark_Ctx_Accepts -benchmem -count=4
goos: darwin
goarch: arm64
Benchmark_Ctx_Accepts
Benchmark_Ctx_Accepts/run-[]string{".xml"}
Benchmark_Ctx_Accepts/run-[]string{".xml"}-12            3662680               308.6 ns/op           920 B/op          2 allocs/op
Benchmark_Ctx_Accepts/run-[]string{".xml"}-12            3888830               307.4 ns/op           920 B/op          2 allocs/op
Benchmark_Ctx_Accepts/run-[]string{".xml"}-12            3836806               306.3 ns/op           920 B/op          2 allocs/op
Benchmark_Ctx_Accepts/run-[]string{".xml"}-12            3864616               308.3 ns/op           920 B/op          2 allocs/op
Benchmark_Ctx_Accepts/run-[]string{"json",_"xml"}
Benchmark_Ctx_Accepts/run-[]string{"json",_"xml"}-12     3338768               350.2 ns/op           920 B/op          2 allocs/op
Benchmark_Ctx_Accepts/run-[]string{"json",_"xml"}-12     3397196               351.2 ns/op           920 B/op          2 allocs/op
Benchmark_Ctx_Accepts/run-[]string{"json",_"xml"}-12     3334743               353.5 ns/op           920 B/op          2 allocs/op
Benchmark_Ctx_Accepts/run-[]string{"json",_"xml"}-12     3366334               352.8 ns/op           920 B/op          2 allocs/op
Benchmark_Ctx_Accepts/run-[]string{"application/json",_"application/xml"}
Benchmark_Ctx_Accepts/run-[]string{"application/json",_"application/xml"}-12             3876166               310.0 ns/op           920 B/op          2 allocs/op
Benchmark_Ctx_Accepts/run-[]string{"application/json",_"application/xml"}-12             3752412               307.9 ns/op           920 B/op          2 allocs/op
Benchmark_Ctx_Accepts/run-[]string{"application/json",_"application/xml"}-12             3703234               309.1 ns/op           920 B/op          2 allocs/op
Benchmark_Ctx_Accepts/run-[]string{"application/json",_"application/xml"}-12             3796650               311.3 ns/op           920 B/op          2 allocs/op
Benchmark_Ctx_AcceptsCharsets
Benchmark_Ctx_AcceptsCharsets-12                                                         4684959               251.0 ns/op           920 B/op          2 allocs/op
Benchmark_Ctx_AcceptsCharsets-12                                                         4563364               254.2 ns/op           920 B/op          2 allocs/op
Benchmark_Ctx_AcceptsCharsets-12                                                         4649942               251.3 ns/op           920 B/op          2 allocs/op
Benchmark_Ctx_AcceptsCharsets-12                                                         4701276               252.6 ns/op           920 B/op          2 allocs/op
Benchmark_Ctx_AcceptsEncodings
Benchmark_Ctx_AcceptsEncodings-12                                                        4154277               282.1 ns/op           920 B/op          2 allocs/op
Benchmark_Ctx_AcceptsEncodings-12                                                        4155589               283.1 ns/op           920 B/op          2 allocs/op
Benchmark_Ctx_AcceptsEncodings-12                                                        4226884               285.7 ns/op           920 B/op          2 allocs/op
Benchmark_Ctx_AcceptsEncodings-12                                                        4220952               284.2 ns/op           920 B/op          2 allocs/op
Benchmark_Ctx_AcceptsLanguages
Benchmark_Ctx_AcceptsLanguages-12                                                        3340310               350.9 ns/op           920 B/op          2 allocs/op
Benchmark_Ctx_AcceptsLanguages-12                                                        3328549               350.6 ns/op           920 B/op          2 allocs/op
Benchmark_Ctx_AcceptsLanguages-12                                                        3369895               353.3 ns/op           920 B/op          2 allocs/op
Benchmark_Ctx_AcceptsLanguages-12                                                        3324868               352.1 ns/op           920 B/op          2 allocs/op
PASS
ok      github.com/gofiber/fiber/v2     36.191s
```

vs `acceptedTypes := make([]acceptedType, 0, 20)`:

```console
goos: darwin
goarch: arm64
Benchmark_Ctx_Accepts
Benchmark_Ctx_Accepts/run-[]string{".xml"}
Benchmark_Ctx_Accepts/run-[]string{".xml"}-12            7756831               141.6 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{".xml"}-12            8427858               140.9 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{".xml"}-12            8356838               156.3 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{".xml"}-12            8602822               139.1 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{"json",_"xml"}
Benchmark_Ctx_Accepts/run-[]string{"json",_"xml"}-12     6509782               183.1 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{"json",_"xml"}-12     6445286               184.8 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{"json",_"xml"}-12     6444319               184.9 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{"json",_"xml"}-12     6491694               185.3 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{"application/json",_"application/xml"}
Benchmark_Ctx_Accepts/run-[]string{"application/json",_"application/xml"}-12             8612815               139.2 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{"application/json",_"application/xml"}-12             8531341               139.0 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{"application/json",_"application/xml"}-12             8657082               141.0 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Accepts/run-[]string{"application/json",_"application/xml"}-12             8553763               139.5 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_AcceptsCharsets
Benchmark_Ctx_AcceptsCharsets-12                                                        14816422                74.18 ns/op            0 B/op          0 allocs/op
Benchmark_Ctx_AcceptsCharsets-12                                                        16124913                81.23 ns/op            0 B/op          0 allocs/op
Benchmark_Ctx_AcceptsCharsets-12                                                        15506538                80.20 ns/op            0 B/op          0 allocs/op
Benchmark_Ctx_AcceptsCharsets-12                                                        15934032                75.11 ns/op            0 B/op          0 allocs/op
Benchmark_Ctx_AcceptsEncodings
Benchmark_Ctx_AcceptsEncodings-12                                                       10669336               108.5 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_AcceptsEncodings-12                                                       10658364               117.1 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_AcceptsEncodings-12                                                       10817725               108.8 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_AcceptsEncodings-12                                                       10624474               108.7 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_AcceptsLanguages
Benchmark_Ctx_AcceptsLanguages-12                                                        6412471               181.4 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_AcceptsLanguages-12                                                        6546268               180.5 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_AcceptsLanguages-12                                                        6572288               180.2 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_AcceptsLanguages-12                                                        6550489               183.7 ns/op             0 B/op          0 allocs/op
PASS
ok      github.com/gofiber/fiber/v2     32.400s
```

sixcolors:
> * [ ]  consider the possibility to run a string without quality identifier with the old loop concept

Specificity still matters even without the q value. Might not be worth the extra loops through the header to check for q and specificity...

helpers.go
sixcolors: UnsafeBytes works great!

ReneWerner87:
Can you share some benchmark numbers after these changes?
Master vs pull request

ReneWerner87 (approved):

sixcolors:
> Can you share some benchmark numbers after these changes?
> Master vs pull request

<details>
  <summary>Benchmark Compare:</summary>

| Benchmark                                                                     | Iterations Old | Time/op Old | Iterations New | Time/op New | Time Difference |
|-------------------------------------------------------------------------------|----------------|-------------|----------------|-------------|-----------------|
| Benchmark_Ctx_Accepts/run-\[]string{".xml"}                                   | 14128427       | 84.07 ns    | 8947115        | 130.6 ns    | +46.53 ns       |
| Benchmark_Ctx_Accepts/run-\[]string{"json",_"xml"}                            | 11784786       | 101.1 ns    | 6741577        | 177.8 ns    | +76.7  ns       |
| Benchmark_Ctx_Accepts/run-\[]string{"application/json",_"application/xml"}    | 15576163       | 76.06 ns    | 9009831        | 132.1 ns    | +56.04 ns       |
| Benchmark_Ctx_AcceptsCharsets                                                 | 40155767       | 29.15 ns    | 17825068       | 67.42 ns    | +38.27 ns       |
| Benchmark_Ctx_AcceptsEncodings                                                | 24537265       | 48.22 ns    | 11950063       | 100.4 ns    | +52.18 ns       |
| Benchmark_Ctx_AcceptsLanguages                                                | 39774213       | 29.28 ns    | 7444000        | 160.8 ns    | +131.52 ns      |

</details>

<details>
  <summary>Benchmark Old(master):</summary>

| Benchmark                                                                     | Iterations | Time/op  | Memory/op | Allocs/op   |
|-------------------------------------------------------------------------------|------------|----------|-----------|-------------|
| Benchmark_Ctx_Accepts/run-[]string{".xml"}                                    | 14128427   | 84.07 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_Accepts/run-[]string{".xml"}-12                                 | 14058120   | 83.88 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_Accepts/run-[]string{".xml"}-12                                 | 14375080   | 84.11 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_Accepts/run-[]string{".xml"}-12                                 | 14045703   | 84.24 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_Accepts/run-[]string{".xml"}-12                                 | 14034806   | 84.06 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_Accepts/run-[]string{"json",_"xml"}                             | 11784786   | 101.1 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_Accepts/run-[]string{"json",_"xml"}-12                          | 11851578   | 101.0 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_Accepts/run-[]string{"json",_"xml"}-12                          | 11822319   | 101.2 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_Accepts/run-[]string{"json",_"xml"}-12                          | 11654176   | 101.0 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_Accepts/run-[]string{"json",_"xml"}-12                          | 11811072   | 101.0 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_Accepts/run-[]string{"application/json",_"application/xml"}     | 15576163   | 76.06 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_Accepts/run-[]string{"application/json",_"application/xml"}-12  | 15525631   | 75.94 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_Accepts/run-[]string{"application/json",_"application/xml"}-12  | 15594296   | 76.25 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_Accepts/run-[]string{"application/json",_"application/xml"}-12  | 15636002   | 76.06 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_Accepts/run-[]string{"application/json",_"application/xml"}-12  | 15548724   | 75.98 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_AcceptsCharsets                                                 | 40155767   | 29.15 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_AcceptsCharsets-12                                              | 40543222   | 29.77 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_AcceptsCharsets-12                                              | 39439466   | 28.50 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_AcceptsCharsets-12                                              | 40780608   | 29.80 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_AcceptsCharsets-12                                              | 39859770   | 28.51 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_AcceptsEncodings                                                | 24537265   | 48.22 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_AcceptsEncodings-12                                             | 24918106   | 48.44 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_AcceptsEncodings-12                                             | 24050204   | 48.42 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_AcceptsEncodings-12                                             | 25031266   | 47.89 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_AcceptsEncodings-12                                             | 24149485   | 48.14 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_AcceptsLanguages                                                | 39774213   | 29.28 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_AcceptsLanguages-12                                             | 39904504   | 29.45 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_AcceptsLanguages-12                                             | 39634644   | 29.18 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_AcceptsLanguages-12                                             | 39830498   | 29.23 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_AcceptsLanguages-12                                             | 39727206   | 29.27 ns | 0 B/op    | 0 allocs/op |

</details>

<details>
  <summary>Benchmark New(PR):</summary>
| Benchmark                                                                    | Iterations | Time/op  | Memory/op | Allocs/op   |
|------------------------------------------------------------------------------|------------|----------|-----------|-------------|
| Benchmark_Ctx_Accepts/run-[]string{".xml"}                                   | 8947115    | 130.6 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_Accepts/run-[]string{".xml"}-12                                | 8148916    | 130.6 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_Accepts/run-[]string{".xml"}-12                                | 9242019    | 130.9 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_Accepts/run-[]string{".xml"}-12                                | 9136976    | 130.0 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_Accepts/run-[]string{".xml"}-12                                | 9260547    | 130.8 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_Accepts/run-[]string{"json",_"xml"}                            | 6741577    | 177.8 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_Accepts/run-[]string{"json",_"xml"}-12                         | 6753759    | 177.6 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_Accepts/run-[]string{"json",_"xml"}-12                         | 6774225    | 177.5 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_Accepts/run-[]string{"json",_"xml"}-12                         | 6663565    | 178.3 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_Accepts/run-[]string{"json",_"xml"}-12                         | 6774759    | 177.9 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_Accepts/run-[]string{"application/json",_"application/xml"}    | 9009831    | 132.1 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_Accepts/run-[]string{"application/json",_"application/xml"}-12 | 9149313    | 130.6 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_Accepts/run-[]string{"application/json",_"application/xml"}-12 | 8960199    | 132.8 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_Accepts/run-[]string{"application/json",_"application/xml"}-12 | 8954554    | 132.5 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_Accepts/run-[]string{"application/json",_"application/xml"}-12 | 8975256    | 132.3 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_AcceptsCharsets                                                | 17825068   | 67.42 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_AcceptsCharsets-12                                             | 17584864   | 66.64 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_AcceptsCharsets-12                                             | 17816126   | 69.00 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_AcceptsCharsets-12                                             | 17753830   | 67.45 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_AcceptsCharsets-12                                             | 11787234   | 101.3 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_AcceptsEncodings                                               | 11950063   | 100.4 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_AcceptsEncodings-12                                            | 11753605   | 100.1 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_AcceptsEncodings-12                                            | 11852026   | 102.4 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_AcceptsEncodings-12                                            | 11593240   | 102.2 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_AcceptsEncodings-12                                            | 7473263    | 161.0 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_AcceptsLanguages                                               | 7444000    | 160.8 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_AcceptsLanguages-12                                            | 7385431    | 161.1 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_AcceptsLanguages-12                                            | 7547332    | 161.5 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_AcceptsLanguages-12                                            | 7516288    | 160.4 ns | 0 B/op    | 0 allocs/op |
| Benchmark_Ctx_AcceptsLanguages-12                                            | 7516288    | 160.4 ns | 0 B/op    | 0 allocs/op |
<details>

ctx_test.go
```diff
@@ -45,6 +45,10 @@ func Test_Ctx_Accepts(t *testing.T) {
 	utils.AssertEqual(t, "", c.Accepts())
 	utils.AssertEqual(t, ".xml", c.Accepts(".xml"))
 	utils.AssertEqual(t, "", c.Accepts(".john"))
+	utils.AssertEqual(t, "application/xhtml+xml", c.Accepts("application/xml", "application/xml+rss", "application/yml", "application/xhtml+xml"), "must use client-preferred mime type")
```
gaby, line 48: According to this https://www.iana.org/assignments/media-types/media-types.xhtml it should be `application/yaml`

gaby (approved):
Small typo in the test, but overall LGTM

ReneWerner87:
@sixcolors can you do the last change with the typo

ctx_test.go
sixcolors: thanks @gaby

PR #2488, opened 2023-05-31
:bug: fix: treat case for possible timer memory leak

### LINKED ISSUE #2482: 🐛 [Bug]: Potential leak in sleep.go
### Bug Description

I notice the ctx.Done() needs a cancel function to unblock the statement of ctx.Done in https://github.com/gofiber/fiber/blob/master/internal/gopsutil/common/sleep.go#L13, since the case may block forever if the case not met.
  I was wondering if this may leak the goroutine where the ctx.Done() locates?

Maybe what I said is not right, thank you for taking a look.

### How to Reproduce

.

### Expected Behavior

.

### Fiber Version

v2.40.1

### Code Snippet (optional)

```go
package main

import "log"

func main() {
  app := fiber.New()

  // Steps to reproduce

}
```

### Checklist:

luk3skyw4lker:
## Description

It was reported in the gopsutil a possible memory leak for the Sleep function, the fix is to in case of the timer being cancelled by the context, checking if the timer isn't stopped and stoping it.

Fixes #2482 

## Type of change

## Checklist:

## Commit formatting:

### DIFF
--- a/internal/gopsutil/common/sleep.go
+++ b/internal/gopsutil/common/sleep.go
@@ -11,6 +11,9 @@ func Sleep(ctx context.Context, interval time.Duration) error {
 	var timer = time.NewTimer(interval)
 	select {
 	case <-ctx.Done():
+		if !timer.Stop() {
+			<-timer.C
+		}
 		return ctx.Err()
 	case <-timer.C:
 		return nil


gaby:
This is an upstream fix from here: shirou/gopsutil#1475

gaby (approved):
LGTM

gaby:
Thanks! @luk3skyw4lker

sixcolors (approved):
LGTM!

see: https://github.com/shirou/gopsutil/pull/1475

efectn (approved):

PR #2493, opened 2023-06-03
:rocket: Add DisableColors to set the default output format

Skyenought:
## Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change. 
Explain the *details* for making this change. What existing problem does the pull request solve?

Fixes # (issue)
https://github.com/gofiber/fiber/issues/2372

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Checklist:

- [x] For new functionalities I follow the inspiration of the express js framework and built them similar in usage
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation - /docs/ directory for https://docs.gofiber.io/
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] If new dependencies exist, I have checked that they are really necessary and agreed with the maintainers/community (we want to have as few dependencies as possible)
- [x] I tried to make my code as fast as possible with as few allocations as possible
- [x] For new code I have written benchmarks so that they can be analyzed and improved

## Commit formatting:

Use emojis on commit messages so it provides an easy way of identifying the purpose or intention of a commit. Check out the emoji cheatsheet here: https://gitmoji.carloscuesta.me/

PR #2497, opened 2023-06-05
:recycle: Refactor: Use EnableColors instead of DisableColors

cmd777:
## Description

In [this](#2493) pull request, initially, EnableColors was exported and used, [but then a new a bool variable was added](https://github.com/gofiber/fiber/pull/2493/commits/d30953786b9ce8bdf3083326606e50be805ce8fb), which I feel like is redundant, since EnableColors already exists, and works just like the added DisableColors bool

## Type of change

## Checklist:

- [X] I have made corresponding changes to the documentation - /docs/ directory for https://docs.gofiber.io/

## Commit formatting:

### DIFF
--- a/docs/api/middleware/logger.md
+++ b/docs/api/middleware/logger.md
@@ -83,7 +83,7 @@ app.Use(logger.New(logger.Config{
 
 // Disable colors when outputting to default format
 app.Use(logger.New(logger.Config{
-    DisableColors: true,
+    EnableColors: false,
 }))
 ```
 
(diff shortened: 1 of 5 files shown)


ReneWerner87 (changes requested):
By default all variables with datatype bool are false, so how do you manage the default value true when a consumer passes a custom config and does not explicitly enable coloring

Therefore we mostly use switches that are rotated when the default value is "enabled".

cmd777:
Ah, I see, thank you for the clarification! @ReneWerner87
I'll close the PR.

PR #2503, opened 2023-06-09
Disable caching when running govulncheck

gaby:
Disabling caching and using "check-latest" allows setup-go to work properly when being used with `govulncheck`.

I was waiting for the golang team to fix the official action, but it's been over 1 week to add one config option.


gaby:
Waiting for https://github.com/golang/go/issues/60704

renanbastos93 (approved):

PR #2509, opened 2023-06-17
:sparkles: feat: add liveness and readiness checks

### LINKED ISSUE #2457: 🚀 [Feature] Middleware to support Kubernetes probes
### Feature Description

Applications deployed on `Kubernetes` should provide information about their internal state with [Container Probes](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-probes).

`Fiber` could make use of a middleware to automatically manage the application availability state, so when deploying in `Kubernetes` (or any other environment) there is a built-in option for gathering the "liveness" and "readiness" information.

This could be as simple as implementing the aforementioned probes, but it can plug as well in the application's lifecycle; for instance, refusing traffic automatically when starting. etc. — but that could be phased out, if relevant, into another unit of work.

_No response_

### Code Snippet (optional)

_No response_

### Checklist:

luk3skyw4lker:
## Description

I have added some config options for probe checkers, as stated in the PR. All of them are optional and should not break any existing fiber tests or applications.

Fixes #2457 

## Type of change

## Checklist:


gaby:
@luk3skyw4lker May want to add a test for this

luk3skyw4lker:
@gaby working on it right now! I remebered the tests after the PR creation, sorry!

luk3skyw4lker:
@gaby the failing tests are in parts of the code that I haven't changed and I don't know to which extent the added code messes up with the performance benchmark. Could you check it out?

leonklingele:
I think /livez and /readyz are the more appropriate endpoints to use by default.

gaby:
According to k8s docs it should be `/healthz`  for HTTP probes. Src https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

luk3skyw4lker:
Yeah, it says healthz for the liveness probe, but it cannot be the same for the readiness probe, since they're different types of checks. I could change the liveness to healthz and the readiness to readyz, what do you guys think?

gaby:
> Yeah, it says healthz for the liveness probe, but it cannot be the same for the readiness probe, since they're different types of checks. I could change the liveness to healthz and the readiness to readyz, what do you guys think?

Sounds good to me!

luk3skyw4lker:
It's done!

leonklingele:
“The healthz endpoint is deprecated (since Kubernetes v1.16), and you should use the more specific livez and readyz endpoints instead.”

https://kubernetes.io/docs/reference/using-api/health-checks/

app.go
```diff
@@ -35,6 +35,9 @@ const Version = "2.46.0"
 // Handler defines a function to serve HTTP requests.
 type Handler = func(*Ctx) error
 
+// Probe Checker defines a function to check liveness or readiness of the application
```
leonklingele, line 38:
```suggestion
// ProbeChecker defines a function to check liveness and readiness of the application
```

app.go
```diff
@@ -390,6 +393,26 @@ type Config struct {
 	//
 	// Optional. Default: DefaultMethods
 	RequestMethods []string
+
+	// Config for liveness probe of the container engine being used
+	//
+	// Optional. Default: func(c *Ctx) bool { return true }
+	IsLive ProbeChecker
+
+	// Config for liveness probe of the container engine being used
```
leonklingele, line 402:
```suggestion
	// HTTP endpoint of the liveness probe
```

app.go
```diff
@@ -390,6 +393,26 @@ type Config struct {
...
+	//
+	// Optional. Default: /liveness
+	IsLiveEndpoint string
+
+	// Config for readiness probe of the container engine being used
+	//
+	// Optional. Default: func(c *Ctx) bool { return true }
+	IsReady ProbeChecker
+
+	// Config for readiness probe of the container engine being used
```
leonklingele, line 412:
```suggestion
	// HTTP endpoint of the readiness probe
```

app.go
```diff
@@ -390,6 +393,26 @@ type Config struct {
...
+	//
+	// Optional. Default: /readiness
```
leonklingele, line 414: Default value needs to be fixed.

gaby:
> “The healthz endpoint is deprecated (since Kubernetes v1.16), and you should use the more specific livez and readyz endpoints instead.”
> 
> [kubernetes.io/docs/reference/using-api/health-checks](https://kubernetes.io/docs/reference/using-api/health-checks/)

Good to know, haha! Thanks!

app.go
```diff
@@ -390,6 +393,26 @@ type Config struct {
...
 	RequestMethods []string
+
+	// Config for liveness probe of the container engine being used
+	//
+	// Optional. Default: func(c *Ctx) bool { return true }
+	IsLive ProbeChecker
+
+	// Config for liveness probe of the container engine being used
+	//
+	// Optional. Default: /liveness
```
leonklingele, line 404: Default value needs to be fixed.

app.go
```diff
@@ -453,8 +476,22 @@ const (
 	DefaultReadBufferSize       = 4096
 	DefaultWriteBufferSize      = 4096
 	DefaultCompressedFileSuffix = ".fiber.gz"
+	DefaultLivenessEndpoint     = "/healthz"
+	DefaultReadinessEndpoint    = "/readyz"
 )
 
+var DefaultHealthFunction = func(c *Ctx) bool { return true }
```
leonklingele, line 483: I’d not export this, or what’s the benefit of doing so?

app.go
```diff
@@ -453,8 +476,22 @@ const (
...
+
+var ProbeCheckerHandler = func(checker ProbeChecker) Handler {
```
leonklingele, line 485: I’d not export this, or what’s the benefit of doing so?

app.go
leonklingele: Also, I’d rename this to DefaultLiveFunc

docs/api/app.md
```diff
@@ -655,3 +655,26 @@ Hooks is a method to return [hooks](../guide/hooks.md) property.
...
+
+Fiber comes with an out of the box way of implementing your liveness and readiness checking, when creating a new app, you can pass your own handlers for liveness and readiness checks:
+
+```go title="Examples"
+// Create route with GET method for test:
+app := fiber.New(fiber.Config{
+  IsLive: func (c *fiber.Ctx) bool {
+    return true
+  },
+  IsLiveEndpoint: "/liveness",
```
leonklingele, line 669: Use the proper default value here.

docs/api/app.md
```diff
@@ -655,3 +655,26 @@ Hooks is a method to return [hooks](../guide/hooks.md) property.
...
+  IsReady: func (c *fiber.Ctx) bool {
+    return serviceA.Ready() && serviceB.Ready() && ...
+  }
+  IsReadyEndpoint: "/readiness",
```
leonklingele, line 673: Use the correct default value here.

leonklingele (changes requested):

app.go
luk3skyw4lker: All the other default values are exported, why not this function? Only the ProbeCheckerHandler shouldn't be exported in my opinon

ReneWerner87:
can we create this feature as a middleware? think it would be better

this middleware can then handle the different request methods and routes like the "CORS" middleware

luk3skyw4lker:
It could be done. I only took this approach beacause you said that the core middlewares shouldn't be techonology specific, so instead I have added a generic probe checker to the whole app config, as suggested by @leonklingele 

But I can refactor this into the middleware approach without any problems.

ReneWerner87:
Yes right, would not do it in the app config, then rather as middleware

luk3skyw4lker:
Alright, I'll be doing that on the course of the week, I'm more available at weekends, so probably by then I'll make a commit and ask you guys for a check. Thanks for all the suggestions!

luk3skyw4lker:
Hi everyone! If you guys could check out the new implementation I made and give any advice it would be nice. If everything is ok, I'll work on the tests and send them at this PR too.

gaby:
@luk3skyw4lker May wanna format the code :-D

efectn:
Please update docs, they need to be middleware doc

luk3skyw4lker:
@efectn yes! I have noted this after sending the first part. I will be sending the docs update together with the tests. Thanks for the heads up!

luk3skyw4lker:
I have updated the docs but I'm having trouble with the tests, I've shared my doubts into the Discord server

luk3skyw4lker:
@gaby could you check the lint error please? I have no idea what it is complaining about :/

gaby:
@luk3skyw4lker This is the error: 

```File is not `gofumpt`-ed with `-extra` (gofumpt)```

You can fix that with:

```console
go install mvdan.cc/gofumpt@latest
gofumpt -l -w .
```

Source: https://github.com/mvdan/gofumpt

gaby:
> Guys I'm sorry for not giving the PR the deserved attention, I got a little overwhelmed at work but I'm planning to get back at this ASAP.

Thanks for the update!

gaby:
@luk3skyw4lker Any updates on this? The solution to the golangci issue is posted above

luk3skyw4lker:
@gaby I'm still swamped on work, but still planning to finish this. I'll finish writing the tests ASAP and send the golangci-lint solution too

luk3skyw4lker:
@gaby now I'm only missing the tests. These ones will take a little bit of work from myself, I didn't mange to get them working since the last time we've spoke at Discord :x

app.go
```diff
@@ -453,6 +453,8 @@ const (
 	DefaultReadBufferSize       = 4096
 	DefaultWriteBufferSize      = 4096
 	DefaultCompressedFileSuffix = ".fiber.gz"
+	DefaultLivenessEndpoint     = "/livez"
+	DefaultReadinessEndpoint    = "/readyz"
```
leonklingele, line 457: This shouldn't be in here most probably.

docs/api/app.md
```diff
@@ -654,4 +654,4 @@ Hooks is a method to return [hooks](../guide/hooks.md) property.
 
 ```go title="Signature"
 func (app *App) Hooks() *Hooks
-```
+```
```
leonklingele, line 657: This change should be reverted.

docs/api/middleware/probechecker.md
```diff
@@ -0,0 +1,75 @@
...
+
+```go
+// Initializing with default config
+app.Use(probechecker.New())
+
+// Initialize with custom config
+app.Use(
+	probechecker.New(
+		IsLive: func (c *fiber.Ctx) bool {
+    	return true
```
leonklingele, line 39: This should be formatted.

middleware/probechecker/probechecker.go
```diff
@@ -0,0 +1,118 @@
...
+
+	if cfg.IsLiveEndpoint == "" {
+		cfg.IsLiveEndpoint = DefaultLivenessEndpoint
+	}
+	if cfg.IsReadyEndpoint == "" {
+		cfg.IsReadyEndpoint = DefaultReadinessEndpoint
+	}
+	if cfg.IsLive == nil {
+		cfg.IsLive = defaultLiveFunc
+	}
```
leonklingele, line 115:
```suggestion
	if cfg.IsLive == nil {
		cfg.IsLive = defaultLiveFunc
	}
	if cfg.IsLiveEndpoint == "" {
		cfg.IsLiveEndpoint = DefaultLivenessEndpoint
	}
	if cfg.IsReadyEndpoint == "" {
		cfg.IsReadyEndpoint = DefaultReadinessEndpoint
	}
```

Maintain the struct order.

middleware/probechecker/probechecker.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func defaultConfig(config ...Config) Config {
+	if len(config) < 1 {
+		return Config{
+			IsLive:          defaultLiveFunc,
+			IsLiveEndpoint:  DefaultLivenessEndpoint,
+			IsReadyEndpoint: DefaultReadinessEndpoint,
+		}
```
leonklingele, line 102:
```suggestion
		config = Config{}
```

middleware/probechecker/probechecker.go
```diff
@@ -0,0 +1,118 @@
...
+			if !ok {
+				return fmt.Errorf("routeType of %s not found in checkers", routeType)
+			}
+
+			return handler(c)
+		}
+
+		return c.Next()
+	}
+}
```
leonklingele, line 94: This all seems overly complicated. Why not do something similar to this:

```suggestion
func New(config ...Config) fiber.Handler {
	cfg := defaultConfig(config...)

	isLiveCheck := probeCheckerHandler(cfg.IsLive)
	isReadyCheck := probeCheckerHandler(cfg.IsReady)

	return func(c *fiber.Ctx) error {
		route := c.Route()
		
		if route.Method != fiber.MethodGet {
			return c.Next()
		}
		
		switch route.Path {
		case config.IsLiveEndpoint:
			return isLiveCheck(c)

		case config.IsReadyEndpoint:
			return isReadyCheck(c)
		}

		return c.Next()
	}
}
```

leonklingele (changes requested):

middleware/probechecker/probechecker.go
```diff
@@ -0,0 +1,118 @@
...
+	"fmt"
+
+	"github.com/gofiber/fiber/v2"
+)
+
+// ProbeChecker defines a function to check liveness or readiness of the application
+type ProbeChecker func(*fiber.Ctx) bool
+
+// ProbeCheckerHandler defines a function that returns a ProbeChecker
+type ProbeCheckerHandler func(ProbeChecker) fiber.Handler
```
leonklingele, line 13: What do we need this for?

tomaswarynyca:
ping @luk3skyw4lker

luk3skyw4lker:
@tomaswarynyca I'm still a mess at work, can't devote the deserved attention to the issue right now. I don't know when I'll be able to do it :/

skydiver1980 (approved):

middleware/probechecker/config.go
```diff
@@ -0,0 +1,63 @@
+package probechecker
```
gaby, line 1: How about Health, Healthcheck, Liveness

efectn: healthcheck sounds better

gaby: Yes!

docs/api/middleware/probechecker.md
```diff
@@ -0,0 +1,98 @@
...
+
+    // Config for liveness probe of the container engine being used
+    //
+    // Optional. Default: func(c *Ctx) bool { return true }
+    IsLive ProbeChecker
+
+    // HTTP endpoint of the liveness probe
+    //
+    // Optional. Default: /livez
+    IsLiveEndpoint string
```
gaby, line 68: This should be "LivenessEndpoint"

docs/api/middleware/probechecker.md
```diff
@@ -0,0 +1,98 @@
...
+
+    // Config for readiness probe of the container engine being used
+    //
+    // Optional. Default: nil
+    IsReady ProbeChecker
+
+    // HTTP endpoint of the readiness probe
+    //
+    // Optional. Default: /readyz
+    IsReadyEndpoint string
```
gaby, line 78: ReadinessEndpoint

docs/api/middleware/probechecker.md
```diff
@@ -0,0 +1,98 @@
+---
+id: probecheker
```
gaby, line 2: Suggest renaming to "healthcheck"

docs/api/middleware/probechecker.md
```diff
@@ -0,0 +1,98 @@
...
+    DefaultLivenessEndpoint  = "/livez"
+    DefaultReadinessEndpoint = "/readyz"
+)
+
+func defaultLiveFunc(*fiber.Ctx) bool { return true }
+
+// ConfigDefault is the default config
+var ConfigDefault = Config{
+    IsLive:          defaultLiveFunc,
+    IsLiveEndpoint:  DefaultLivenessEndpoint,
```
gaby, line 95: Here shows the inconsistency with the public config.

docs/api/middleware/probechecker.md
```diff
@@ -0,0 +1,98 @@
+---
+id: probecheker
+title: probecheker
+---
+
+Liveness and readiness probes middleware for [Fiber](https://github.com/gofiber/fiber) that provides two endpoints for checking the health and ready state of any HTTP application.
+
+The endpoint values default to `/livez` for liveness and `/readyz` for readiness. Both functions are optional, the liveness endpoint will return `true` right when the server is up and running but the readiness endpoint will not answer any requests if an `IsReady` function isn't provided. 
+
+The HTTP status returned to the containerized environment are: 200 OK if the checker function returns true and 503 Service Unavailable if the checker function returns false.
```
gaby, line 10: 👌

docs/api/middleware/probechecker.md
```diff
@@ -0,0 +1,98 @@
+---
+id: probecheker
+title: probecheker
+---
+
+Liveness and readiness probes middleware for [Fiber](https://github.com/gofiber/fiber) that provides two endpoints for checking the health and ready state of any HTTP application.
+
+The endpoint values default to `/livez` for liveness and `/readyz` for readiness. Both functions are optional, the liveness endpoint will return `true` right when the server is up and running but the readiness endpoint will not answer any requests if an `IsReady` function isn't provided.
```
gaby, line 8: Readiness should default to "200 Ok". If the function can execute it probably means the server us up and responding.

middleware/probechecker/probechecker_test.go
```diff
@@ -0,0 +1,110 @@
...
+	utils.AssertEqual(t, fiber.StatusNotFound, req.StatusCode)
+}
+
+func Benchmark_ProbeChecker(b *testing.B) {
+	app := fiber.New()
+
+	app.Use(New())
+
+	h := app.Handler()
+	fctx := &fasthttp.RequestCtx{}
```
gaby, line 98: I like this. We should add something like this to helper utils of Fiber to ease benchmarking stuff.

gaby (commented):
Overall looks good, some comments and suggestions.

middleware/probechecker/probechecker_test.go
efectn: > I like this. We should add something like this to helper utils of Fiber to ease benchmarking stuff.

Yeah it would be really useful! We can add something like app.Test

docs/api/middleware/probechecker.md
efectn: > Readiness should default to "200 Ok". If the function can execute it probably means the server us up and responding.

Then what will be difference between ready and live?

docs/api/middleware/probechecker.md
efectn: > This should be "LivenessEndpoint"

You mean renaming config property?

docs/api/middleware/probechecker.md
luk3skyw4lker: This could produce some false positives, since the is ready function should only return 200 OK if the server is fully ready to answer any requests and there's no way to check by default all the requirements for readiness, since it involves a lot of different checks (like connections to other servers or some sort of cleanup or starting task).

And there's a way of configuring the liveness and readiness independently with most of the containerized environments, so this shouldn't be a problem.

docs/api/middleware/probechecker.md
gaby: Yes

docs/api/middleware/probechecker.md
gaby: @luk3skyw4lker @efectn It would just be for the default. There's a use case where you dont want to check anything, just get a 200 to mark the pod/container as "green". Else we would be forcing users to define their readiness function.

middleware/probechecker/probechecker_test.go
gaby: app.Benchmark() @ReneWerner87 :-)

docs/api/middleware/probechecker.md
gaby: That way its consistent internally and public

ReneWerner87 (approved):

efectn (approved):

PR #2538, opened 2023-07-09
Update cookie.go

Hasan-Kilici:
## Description
although the performance difference is not much, there is a possibility that it will run 2-4ms faster this way


cmd777 (changes requested):
This is the v3 branch

efectn (commented):
Can you explain why this change is needed

ReneWerner87:
@Hasan-Kilici friendly ping

leonklingele:
Any benchmarks on this? I doubt it will actually improve performance.

PR #2558, opened 2023-08-05
filesystem: refactor: use `errors.Is` instead of `os.IsNotExist`

efectn:
## Description

Suggested way by https://pkg.go.dev/os#IsNotExist

Fixes # (issue)

## Type of change

## Checklist:

## Commit formatting:

### DIFF
--- a/internal/gopsutil/net/net_linux.go
+++ b/internal/gopsutil/net/net_linux.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"net"
 	"os"
 	"strconv"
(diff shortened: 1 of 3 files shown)


gaby:
So `gopsutil`, is only used by the `monitor` middleware. Thoughts on moving that to contrib and having it use upstream `gopsutil`?

efectn:
> So `gopsutil`, is only used by the `monitor` middleware. Thoughts on moving that to contrib and having it use upstream `gopsutil`?

I agree about that idea. It's bad idea to include latest gopsutil to go.mod because of it's heavy https://github.com/shirou/gopsutil/blob/master/go.mod. Also it's bad idea to vendor it due to it's unmaintainable and it's really old release, it might have some security issues.

On the other hand; it's breaking change, we should think a bit more on it. What  do you think  @ReneWerner87

ReneWerner87 (approved):

PR #2584, opened 2023-08-17
♻️ Refactor: Remove redundant nil check

Juneezee:
## Description

From the Go docs:

> "If the map is nil, the number of iterations is 0." https://go.dev/ref/spec#For_range

Therefore, an additional nil check for before the loop is unnecessary. Example: https://go.dev/play/p/Cw_wnqYzemu

## Checklist:

## Commit formatting:


efectn (approved):

ReneWerner87:
Ok thx
Good to know

PR #2586, opened 2023-08-17
✨ Adds End method to context

### LINKED ISSUE #2582: 🚀 [Feature]: Add an "End" method to flush the response stream with an empty body
### Feature Description

Since Fiber models itself to the Express API, it would be nice if Fiber provided an [end](https://expressjs.com/en/5x/api.html#res.end) method. 

This would simplify the following code from...
```go
func Example(c *fiber.Ctx) error {
    return c.Send(nil)
}
```
to:
```go
func Example(c *fiber.Ctx) error {
    return c.End()
}
```

It would also be nice for it to be invokable at the end of a context chain:

```go
func Example(c *fiber.Ctx) error {
    return c.Status(http.StatusOK).End()
}
```

While writing `.Send(nil)` over and over is not the end of the world, `.End()` would be a subjectively cleaner way to communicate an empty response body.

That said, calling `.End()` could simply be an alias for `.Send(nil)`

### Code Snippet (optional)

_No response_

### Checklist:

scandar:
## Description
Adds end method to the context [similar to express](https://expressjs.com/en/5x/api.html#res.end)
It's an alias for `c.Send(nil)`
Fixes #2582

## Type of change

## Checklist:


efectn (approved):

gaby (approved):
Lgtm

docs/api/ctx.md
```diff
@@ -1733,6 +1733,20 @@ app.Get("/not-found", func(c *fiber.Ctx) error {
 })
 ```
 
+## End
```
ReneWerner87, line 1736: Pls put it in the alphabetical order in the list of methods

ReneWerner87 (changes requested):
Small change needed

ReneWerner87:
@scandar 
```ts
const express = require('express');
const app = express();

app.get('/', (req, res) => {
  // Setting data in the response body
  res.write('Hello, ');
  res.write('World!');

  // Sending the response
  res.end();
});

app.listen(3000, () => {
  console.log('Server is listening on port 3000');
});
```
I think this is not possible or ?
how to reset the body completely and not just send it on

but i think the end method is just to say that the response can go off and not to reset everything

scandar:
@ReneWerner87 I believe it should override the body to `nil`
this is from [the express docs ](https://expressjs.com/en/5x/api.html#res.end)

> Ends the response process. This method actually comes from Node core, specifically the [response.end() method of http.ServerResponse](https://nodejs.org/api/http.html#http_response_end_data_encoding_callback).
> Use to quickly end the response without any data. If you need to respond with data, instead use methods such as [res.send()](https://expressjs.com/en/5x/api.html#res.send) and [res.json()](https://expressjs.com/en/5x/api.html#res.json).

ReneWerner87:
image
https://nodejs.org/api/http.html#http_response_end_data_encoding_callback

if you pass data to the nodejs end method, the following happens internally
image
that means it is used write and then end internally
for us this means that everything that was set before is simply terminated with end, without emptying it

i also saw examples for the write+end case 
https://github.com/nodejs/node/blob/c021b27221dbf65ea23a570d9df91bfdac4b3196/benchmark/http/end-vs-write-end.js#L36-L43

https://nodejs.org/de/docs/guides/anatomy-of-an-http-transaction#sending-response-body
image

ReneWerner87:
that means we are not allowed to do a body reset, just send it off

mattcarlotta:
@scandar Perhaps an alternative to using `.Send(nil)` would be to simply return `nil`? 
```go
func (c *Ctx) End() error {
    return nil
}
```
After reviewing the fasthttp internals, having an `.End` method wouldn't accomplish much since it doesn't include a concrete way to flush and end a Response. However it might be nicer to have a chainable End method that can be returned.

For example, currently we need to write something like this if the func returns an error:
```go
func Example(c *fiber.Ctx) error {
    // return c.Status(fiber.StatusOK).Send(nil)

    c.Status(fiber.StatusOK)
    return nil
}
```
Instead, using the code mentioned previously, the `return nil` can be chained to the context:
```go
func Example(c *fiber.Ctx) error {
    return c.Status(fiber.StatusOK).End()
}
```

ReneWerner87:
I think we have no way to explicitly trigger the content of the response to be flushed

fasthttp does this in the server flow for the request/response

for this you then create a writer for the connection and flush the content from the response to it

so end can't make the flow actually really end

Issue #2581, opened 2023-08-16
♻️  [Refactor]:  logger/middleware to Eliminate Code Duplication


sixcolors:
### Feature Description

Refactor the logger middleware in Fiber to adher to the "Don't Repeat Yourself" (DRY) principle. During recent discussions #2561, it was identified that the current implementation involves duplicated code for custom and default logger functions.

### Additional Context (optional)

Code Refactoring: The primary focus is to eliminate code duplication by consolidating the common logic shared between custom and default logger functions. This not only improves code maintainability but also ensures consistent behavior across different configurations.

### Code Snippet (optional)

```go
// Default output when no custom Format or io.Writer is given
		if cfg.Format == ConfigDefault.Format {
			// Format error if exist
			formatErr := ""
			if cfg.enableColors {
				if chainErr != nil {
					formatErr = colors.Red + " | " + chainErr.Error() + colors.Reset
				}
				_, _ = buf.WriteString( //nolint:errcheck // This will never fail
					fmt.Sprintf("%s |%s %3d %s| %7v | %15s |%s %-7s %s| %-"+errPaddingStr+"s %s\n",
						timestamp.Load().(string),
						statusColor(c.Response().StatusCode(), colors), c.Response().StatusCode(), colors.Reset,
						data.Stop.Sub(data.Start).Round(time.Millisecond),
						c.IP(),
						methodColor(c.Method(), colors), c.Method(), colors.Reset,
						c.Path(),
						formatErr,
					),
				)
			} else {
				if chainErr != nil {
					formatErr = " | " + chainErr.Error()
				}
				_, _ = buf.WriteString( //nolint:errcheck // This will never fail
					fmt.Sprintf("%s | %3d | %7v | %15s | %-7s | %-"+errPaddingStr+"s %s\n",
						timestamp.Load().(string),
						c.Response().StatusCode(),
						data.Stop.Sub(data.Start).Round(time.Millisecond),
						c.IP(),
						c.Method(),
						c.Path(),
						formatErr,
					),
				)
			}

			// Write buffer to output
			_, _ = cfg.Output.Write(buf.Bytes()) //nolint:errcheck // This will never fail

			if cfg.Done != nil {
				cfg.Done(c, buf.Bytes())
			}

			// Put buffer back to pool
			bytebufferpool.Put(buf)

			// End chain
			return nil
		}
```

### Checklist:

[...]

PR #2593, opened 2023-08-22
♻️ logger/middleware to Eliminate Code Duplication

Skyenought:
## Description

Fixes # (issue)
https://github.com/gofiber/fiber/issues/2581

## Type of change

- [x] refactor

## Checklist:

## Commit formatting:

### DIFF
--- a/middleware/logger/logger.go
+++ b/middleware/logger/logger.go
@@ -139,39 +139,32 @@ func New(config ...Config) fiber.Handler {
 
 		// Default output when no custom Format or io.Writer is given
 		if cfg.Format == ConfigDefault.Format {
-			// Format error if exist
+			// Format error if it exists
 			formatErr := ""
-			if cfg.enableColors {
-				if chainErr != nil {
+			if chainErr != nil {
+				if cfg.enableColors {
 					formatErr = colors.Red + " | " + chainErr.Error() + colors.Reset
-				}
-				_, _ = buf.WriteString( //nolint:errcheck // This will never fail
-					fmt.Sprintf("%s |%s %3d %s| %13v | %15s |%s %-7s %s| %-"+errPaddingStr+"s %s\n",
-						timestamp.Load().(string),
-						statusColor(c.Response().StatusCode(), colors), c.Response().StatusCode(), colors.Reset,
-						data.Stop.Sub(data.Start),
-						c.IP(),
-						methodColor(c.Method(), colors), c.Method(), colors.Reset,
-						c.Path(),
-						formatErr,
-					),
-				)
-			} else {
-				if chainErr != nil {
+				} else {
 					formatErr = " | " + chainErr.Error()
 				}
-				_, _ = buf.WriteString( //nolint:errcheck // This will never fail
-					fmt.Sprintf("%s | %3d | %13v | %15s | %-7s | %-"+errPaddingStr+"s %s\n",
-						timestamp.Load().(string),
-						c.Response().StatusCode(),
-						data.Stop.Sub(data.Start),
-						c.IP(),
-						c.Method(),
-						c.Path(),
-						formatErr,
-					),
-				)
 			}
+
+			// Construct the log format
+			logFormat := "%s |%s %3d %s| %7v | %15s |%s %-7s %s| %-" + errPaddingStr + "s %s\n"
+			if !cfg.enableColors {
+				logFormat = "%s | %3d | %7v | %15s | %-7s | %-" + errPaddingStr + "s %s\n"
+			}
+
+			// Write log entry to buffer
+			_, _ = buf.WriteString(fmt.Sprintf(logFormat,
+				timestamp.Load().(string),
+				statusColor(c.Response().StatusCode(), colors), c.Response().StatusCode(), colors.Reset,
+				data.Stop.Sub(data.Start).Round(time.Millisecond),
+				c.IP(),
+				methodColor(c.Method(), colors), c.Method(), colors.Reset,
+				c.Path(),
+				formatErr,
+			)) //nolint:errcheck // This will never fail
 
 			// Write buffer to output
 			_, _ = cfg.Output.Write(buf.Bytes()) //nolint:errcheck // This will never fail
--- a/middleware/logger/logger.go
+++ b/middleware/logger/logger.go
@@ -58,6 +58,10 @@ func New(config ...Config) fiber.Handler {
 		errHandler fiber.ErrorHandler
 
 		dataPool = sync.Pool{New: func() interface{} { return new(Data) }}
+
+		methodColorFmt string
+		statusCodeFmt  string
+		colorReset     string
 	)
 
 	// If colors are enabled, check terminal compatibility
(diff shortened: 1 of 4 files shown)


sixcolors:
@Skyenought why not move the colors into Tags.go and remove the entire conditional check from logger.go? Wold have the side benefit of enabling colors on custom formated logs.

middleware/logger/logger.go
```diff
@@ -139,40 +139,33 @@ func New(config ...Config) fiber.Handler {
...
 			}
 
+			// Construct the log format
+			logFormat := "%s |%s %3d %s| %7v | %15s |%s %-7s %s| %-" + errPaddingStr + "s %s\n"
+			if !cfg.enableColors {
+				logFormat = "%s | %3d | %7v | %15s | %-7s | %-" + errPaddingStr + "s %s\n"
+			}
+
+			// Write log entry to buffer
+			_, _ = buf.WriteString(fmt.Sprintf(logFormat,
```
sixcolors, line 159: add `//nolint:errcheck // This will never fail`

sixcolors, line 159: add `//nolint:errcheck // This will never fail` here and remove from line 167 to fix linter error

sixcolors:
> @Skyenought why not move the colors into tags.go and remove the entire conditional check from logger.go? Wold have the side benefit of enabling colors on custom formated logs.

Looks like TagStatus and TagMethod are already done, so that only leaves TagError which would need to be refactored as:

```go
		TagError: func(output Buffer, c *fiber.Ctx, data *Data, extraParam string) (int, error) {
			if data.ChainErr != nil {
				if cfg.enableColors {
					colors := c.App().Config().ColorScheme
					return output.WriteString(fmt.Sprintf("%s %s %s", colors.Red, data.ChainErr.Error(), colors.Reset))
				}
				return output.WriteString(data.ChainErr.Error())
			}
			return output.WriteString("-")
		},
```

sixcolors:
@ReneWerner87 @gaby @Skyenought 

I've identified an inconsistency in code formatting and error handling. While the default format is  `[${time}] ${status} - ${latency} ${method} ${path}\n` , the current implementation switches to `${time} | ${status} | ${latency} | ${ip} | ${method} | ${path}\n` (with additional logic in logger.go that adds ` " | " + chainErr.Error()` or `colors.Red + " | " + chainErr.Error() + colors.Reset` if error is != nil).

I see some issues with this:

1. Error formatting in tags.go doesn't account for spacing or seperator differences if error != nil
2. default formater doesn't actually use the format in ConfigDefault, and
3. if someone wanted to actaully use `[${time}] ${status} - ${latency} ${method} ${path}\n` as the log format might lead to unintended outcomes.

gaby:
> @Skyenought
> 
> I've identified an inconsistency in code formatting and error handling. While the default format is `[${time}] ${status} - ${latency} ${method} ${path}\n` , the current implementation [and current PR] switches to `${time} | ${status} | ${latency} | ${ip} | ${method} | ${path}\n` with "logger.go" adding " | " + chainErr.Error() or colors.Red + " | " + chainErr.Error() + colors.Reset (if error != nil) before the line break.
> 
> Several issues arise from this:
> 
> 1. The default format doesn't align with the ConfigDefault format, leading to confusion.
> 2. Opting for [${time}] ${status} - ${latency} ${method} ${path}\n is not feasible.
> 3. Custom logger formats behave inconsistently with the default due to TagError in tags.go not addressing spacing or separator differences when error != nil.
> 
> For a more coherent approach, I recommend reviewing both "tags.go" and "logger.go" logic to ensure consistent formatting and error handling.
> 
> I'd also appreciate input from @ReneWerner87 and @gaby regarding this matter.

I agree, the formatting of the logs should be consistent. Specially because some users may be parsing these logs with other tools.

Skyenought:
> > @Skyenought
> > I've identified an inconsistency in code formatting and error handling. While the default format is `[${time}] ${status} - ${latency} ${method} ${path}\n` , the current implementation [and current PR] switches to `${time} | ${status} | ${latency} | ${ip} | ${method} | ${path}\n` with "logger.go" adding " | " + chainErr.Error() or colors.Red + " | " + chainErr.Error() + colors.Reset (if error != nil) before the line break.
> > Several issues arise from this:
> > 
> > 1. The default format doesn't align with the ConfigDefault format, leading to confusion.
> > 2. Opting for [${time}] ${status} - ${latency} ${method} ${path}\n is not feasible.
> > 3. Custom logger formats behave inconsistently with the default due to TagError in tags.go not addressing spacing or separator differences when error != nil.
> > 
> > For a more coherent approach, I recommend reviewing both "tags.go" and "logger.go" logic to ensure consistent formatting and error handling.
> > I'd also appreciate input from @ReneWerner87 and @gaby regarding this matter.
> 
> I agree, the formatting of the logs should be consistent. Specially because some users may be parsing these logs with other tools.



> > @Skyenought
> > I've identified an inconsistency in code formatting and error handling. While the default format is `[${time}] ${status} - ${latency} ${method} ${path}\n` , the current implementation [and current PR] switches to `${time} | ${status} | ${latency} | ${ip} | ${method} | ${path}\n` with "logger.go" adding " | " + chainErr.Error() or colors.Red + " | " + chainErr.Error() + colors.Reset (if error != nil) before the line break.
> > Several issues arise from this:
> > 
> > 1. The default format doesn't align with the ConfigDefault format, leading to confusion.
> > 2. Opting for [${time}] ${status} - ${latency} ${method} ${path}\n is not feasible.
> > 3. Custom logger formats behave inconsistently with the default due to TagError in tags.go not addressing spacing or separator differences when error != nil.
> > 
> > For a more coherent approach, I recommend reviewing both "tags.go" and "logger.go" logic to ensure consistent formatting and error handling.
> > I'd also appreciate input from @ReneWerner87 and @gaby regarding this matter.
> 
> I agree, the formatting of the logs should be consistent. Specially because some users may be parsing these logs with other tools.

I changed `DefaultFormat` to `${time} | ${status} | ${latency} | ${method} | ${path} | ${error}\n`.

This is also the fact that the template format, `${error}`, is not always available.

sixcolors:
@Skyenought what do you think about adding an TagIfError that has optional padding?

Skyenought:
@sixcolors 
I think if you change `return output.WriteString("-")` to `return output.WriteString("")` in the last line of `TagError` and you'll get the effect you're describing, right?

sixcolors:
> @sixcolors
> I think if you change `return output.WriteString("-")` to `return output.WriteString("")` in the last line of `TagError` and you'll get the effect you're describing, right?

I was trying to suggest:

1. Introduce a TagIfError tag that conditionally displays ${iferror} only when err != nil. This approach allows for clear and concise logs while ensuring that relevant errors are included.
2. The function could mirror the formatting used in the default logger from logger.go. This consistency would maintain a familiar and user-friendly log structure.
3. To provide even more flexibility, consider enabling optional padding and separators for ${iferror}, eg ${iferror,separator=|,spacing=true}. This would allow users to customize the error presentation according to their preferences.
4. We could eliminate the default formatter code from logger.go and use the tags.go code for all format strings.
5. In practice, this would result in a cleaner and more informative log output. For instance:

${time} | ${status} | ${latency} | ${method} | ${path} | ${iferror,separator=|,spacing=true}\n

Skyenought:
> 

I see what you mean. Let me try.

ReneWerner87:
can we also have unittest for testing this format by injecting a special writer and testing our expectation with and without colors?

i also like to check the color setting inside the tag
which ensures that the format outside remains the same 
and then the coloring works not only for the standard format, but for any custom composited format

skydiver1980 (approved):

ReneWerner87:
closed by #2773

PR #2603, opened 2023-08-26
:sparkles: feat: add strings reverse and strings contain func in utils.strings file

ShyunnY:
## Description

This pr adds two commonly used string functions. I believe that in many cases string reversal and the function of determining whether the string is in the specified slice will be used. Therefore, I write this pr.

## Type of change

## Checklist:

## Commit formatting:

### DIFF
--- a/utils/strings.go
+++ b/utils/strings.go
@@ -73,3 +73,26 @@ func EqualFold(b, s string) bool {
 	}
 	return true
 }
+
+// Reverse reverse target string
+func Reverse(s string) string {
+	rs := []rune(s)
+
+	for head, tail := 0, len(rs)-1; head < tail; head, tail = head+1, tail-1 {
+		rs[head], rs[tail] = rs[tail], rs[head]
+	}
+
+	return string(rs)
+}
+
+// Contain checks if the target string is contained in the strs
+func Contain(strs []string, s string) bool {
+
+	for _, str := range strs {
+		if str == s {
+			return true
+		}
+	}
+
+	return false
+}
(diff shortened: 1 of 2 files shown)


ReneWerner87:
please also add the internal usage, otherwise we would have to move the whole thing to the utils package
https://github.com/gofiber/utils

ShyunnY:
> please also add the internal usage, otherwise we would have to move the whole thing to the utils package https://github.com/gofiber/utils

There may be codes similar to these two functions inside the fiber, but I haven't found it yet. But when writing logic/service. These two functions are used more often. @ReneWerner87

gaby:
I feel like this should go in `gofiber/utils/v2`

ShyunnY:
> I feel like this should go in `gofiber/utils/v2`



> I feel like this should go in `gofiber/utils/v2`

I do not understand. I can only see `gofiber/utils`. `gofiber/utils/v2` was not found. So should my current PR be resubmitted into `gofiber/utils`?

gaby:
> > I feel like this should go in `gofiber/utils/v2`
> 
> I do not understand. I can only see `gofiber/utils`. `gofiber/utils/v2` was not found. So should my current PR be resubmitted into `gofiber/utils`?

It's the v2 branch under https://github.com/gofiber/utils

leonklingele:
Why do we need this as part of fiber? I wouldn’t add code which is not used at all.

gaby:
@ReneWerner87 Do we move this to utils repo, or close it?

Issue #2594, opened 2023-08-23
🐛 [Bug]: `Ctx.GetReqHeaders` doesn't respect multiple fields with same name


makkes:
### Bug Description

When calling `Ctx.GetReqHeaders` on a request with this header:

```
"Accept": "text/plain",
"Accept": "application/json",
[...]
```

Then the resulting map contains only the last field value.

See the [relevant RFC section](https://datatracker.ietf.org/doc/html/rfc9110#section-5.2).

### How to Reproduce

1. Run the code snippet from below
2. Run `curl localhost:3000 -H 'Accept: text/plain' -H 'Accept: application/json'`
3. Check the logs of the app

### Expected Behavior

The logs should show the value of both `Accept` fields:

```
map[Accept:application/json,text/plain Host:localhost:3000 User-Agent:curl/8.2.1]
```

### Fiber Version

v2.48.0

### Code Snippet (optional)

```go
package main

import (
	"fmt"
	"log"

	"github.com/gofiber/fiber/v2"
)

func main() {
	app := fiber.New()

	app.Get("", func(c *fiber.Ctx) error {
		fmt.Println(c.GetReqHeaders())
		return nil
	})

	log.Fatal(app.Listen(":3000"))
}
```

### Checklist:

- [X] I agree to follow Fiber's [Code of Conduct](https://github.com/gofiber/fiber/blob/master/.github/CODE_OF_CONDUCT.md).
- [X] I have checked for existing issues that describe my problem prior to opening this one.
- [X] I understand that improperly formatted bug reports may be closed without explanation.

efectn:
I think we need to change signature to fix the bug https://github.com/gofiber/fiber/blob/master/ctx.go#L654

Also getrespheaders needs to be fixed, too.

makkes:
I don't think the signature needs to be changed. All the function would need to do is to join all the values using `, ` according to the RFC I linked above:

>  When a field name is repeated within a section, its combined field value consists of the list of corresponding field line values within that section, concatenated in order, with each field line value separated by a comma.

So a header of 

```
Accept: text/plain
Accept: application/json
```

would result in a map with key `Accept` and value `text/plain, application/json`.

efectn:
> I don't think the signature needs to be changed. All the function would need to do is to join all the values using `, ` according to the RFC I linked above:
> 
> > When a field name is repeated within a section, its combined field value consists of the list of corresponding field line values within that section, concatenated in order, with each field line value separated by a comma.
> 
[...]

PR #2628, opened 2023-09-08
return raw header map

raghavTinker:
## Description
Currently the GetHeaders method built into gofiber returns a map of headers which are formatted. There is no function which returns a map of unformatted raw headers.

Fixes # (issue)

## Type of change

## Checklist:

## Commit formatting:

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -659,6 +659,20 @@ func (c *Ctx) GetReqHeaders() map[string]string {
 
 	return headers
 }
+// The issue with GetReqHeaders is it formats keys with the first letter
+// capitalized. This function returns the raw headers as a string map.
+// The keys are not formatted by default
+func (c *Ctx) GetRawReqHeaderMap() map[string]string {
+	headerStr := string(c.Request().Header.RawHeaders())
+	headers := make(map[string]string)
+	for _, header := range strings.Split(headerStr, "\r\n") {
+		if header != "" {
+			headerSplit := strings.Split(header, ": ")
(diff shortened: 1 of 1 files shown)


ctx.go
```diff
@@ -659,6 +659,20 @@ func (c *Ctx) GetReqHeaders() map[string]string {
 
 	return headers
 }
+// The issue with GetReqHeaders is it formats keys with the first letter
```
gaby, line 662: This comment doesnt match the formatting from the other functions.

ctx.go
```diff
@@ -659,6 +659,20 @@ func (c *Ctx) GetReqHeaders() map[string]string {
...
+// capitalized. This function returns the raw headers as a string map.
+// The keys are not formatted by default
+func (c *Ctx) GetRawReqHeaderMap() map[string]string {
```
gaby, line 665: Add unit-tests for this new function

gaby (changes requested):

efectn (changes requested):
Please make return map[string][]string -> https://github.com/gofiber/fiber/issues/2594

ReneWerner87:
we already have many methods to return the headers, what benefit do the users get from this method ? don't want to bloat the framework with any method to return the headers

ReneWerner87:
we have decided not to include this functionality for the time being, as it does not offer any benefit compared to the other functionalities already available

with this we want to avoid the bloating of the framework

PR #2645, opened 2023-09-25
:sparkles: add feature parse content-type application/x-ndjson

huykn:
## Description

add feature parse content-type application/x-ndjson
ex input body json
```json
{"name":"john"}
{"name":"doe"}
```
expect return array via call method BodyParser

```go
type Demo struct {
    Name string `json:"name" xml:"name" form:"name" query:"name"`
}
c.Request().Header.SetContentType(MIMEApplicationXNDJSON)
c.Request().SetBody([]byte("{\"name\":\"john\"}\n{\"name\":\"doe\"}\n"))
c.Request().Header.SetContentLength(len(c.Body()))
var out []Demo
utils.AssertEqual(t, nil, c.BodyParser(&out))
utils.AssertEqual(t, 2, len(out))
utils.AssertEqual(t, "john", out[0].Name)
utils.AssertEqual(t, "doe", out[1].Name)
```

## Type of change

## Checklist:

## Commit formatting:

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -749,7 +749,7 @@ iploop:
 			j++
 		}
 
-		for i < j && headerValue[i] == ' ' {
+		for i < j && (headerValue[i] == ' ' || headerValue[i] == ',') {
 			i++
 		}
 
--- a/ctx.go
+++ b/ctx.go
@@ -417,8 +417,8 @@ func (c *Ctx) parseFormBody(args *fasthttp.Args, out interface{}) (map[string][]
 			return
 		}
 
-		k := string(key)
-		v := string(val)
+		k := c.app.getString(key)
+		v := c.app.getString(val)
 
 		if strings.Contains(k, "[") {
 			k, err = parseParamSquareBrackets(k)
--- a/ctx.go
+++ b/ctx.go
@@ -55,6 +55,8 @@ var (
 	decoderPoolMap = map[string]*sync.Pool{}
 	// tags is used to classify parser's pool
 	tags = []string{queryTag, bodyTag, reqHeaderTag, paramsTag}
+	// find the new line when parsing the body using method BodyParser with type application/x-ndjson
+	delimiterEnter = []byte("\n")
 )
 
 func init() {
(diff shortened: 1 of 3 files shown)


ctx.go
```diff
@@ -373,56 +375,112 @@ func decoderBuilder(parserConfig ParserConfig) interface{} {
...
+	var err error
+
+	args.VisitAll(func(key, val []byte) {
+		if err != nil {
+			return
+		}
 
-			k := c.app.getString(key)
-			v := c.app.getString(val)
+		k := string(key)
```
efectn, line 420: please use c.app.getString here

efectn (changes requested):

ctx.go
huykn: ok bro

ReneWerner87:
thanks for the feature

but i'm not sure if we should support any kind of body parsing as a framework or if it's better to use extension packages for it and focus on the general ones
image

@efectn @gaby whats your opinion

gaby:
@ReneWerner87 I feel like this should be something in `fasthttp` instead of GoFiber. It doesn't make sense to add all this custom logic to the framework `ndjson` is very useful, specially when dealing with processing logs/metrics. Is there a library that could help with this?

efectn:
> thanks for the feature
> 
> but i'm not sure if we should support any kind of body parsing as a framework or if it's better to use extension packages for it and focus on the general ones image
> 
> @efectn @gaby whats your opinion

I do agree. If we try to implement every JSON implementations, the core will get really bloated.

ReneWerner87:
cool would be a plugin system

where we could declare a parser for the content type in a map, then you could add these parsers separately and only the core parsers are included 

app.bodyParser["application/x-www-form-urlencoded"] = formDataParserClosure
app.bodyParser["multipart/form-data"] = multiformDataParserClosure
app.bodyParser["application/json"] = jsonParserClosure
app.bodyParser["application/xml>"] = xmlParserClosure
app.bodyParser["text/xml"] = xmlParserClosure
....

like express with the bodyParser

gaby:
@huykn @ReneWerner87 @efectn According to this article you can already parse ndjson using native Go. 

https://abhinavg.net/2020/07/02/parse-ndjson/

```go
decoder := json.NewDecoder(src)
for decoder.More() {
    var order Order
    if err := decoder.Decode(&order); err != nil {
        return fmt.Errorf("parse order: %w", err)
    }

    fmt.Println(order)
}
```

huykn:
> @huykn @ReneWerner87 @efectn According to this article you can already parse ndjson using native Go.
> 
> https://abhinavg.net/2020/07/02/parse-ndjson/
> 
> ```go
> decoder := json.NewDecoder(src)
> for decoder.More() {
>     var order Order
>     if err := decoder.Decode(&order); err != nil {
>         return fmt.Errorf("parse order: %w", err)
>     }
> 
>     fmt.Println(order)
> }
> ```

@gaby i think we have custom config apply for use 3rd json at: https://github.com/gofiber/fiber/blob/master/docs/guide/faster-fiber.md
if we use only json from encoding/json i think logic is not correct. i'm not sure any json 3rd support ndjson.

ReneWerner87:
I tested this and it was not working

huykn:
> I tested this and it was not working

for feature in this merge request or smt?

ReneWerner87:
Feature is working
I mean i tested the parsing of the njson content with the default parser(without extra logic) and that's not working

huykn:
> Feature is working I mean i tested the parsing of the njson content with the default parser(without extra logic) and that's not working

do you think this feature should be merged or not? or do we need to change to add parsers separately and only the core parsers are included?

ReneWerner87:
Good question, we will discuss it and then decide (unfortunately not everyone is here right now)

skydiver1980 (approved):

ReneWerner87:
sorry for the late reply
we have decided not to adopt this feature request

because we can't possibly have parsers in the core for all content types that exist
we currently only want to provide the most common ones in the core 

in v3 we can try to find a way to add parser from outside and in the code we can have a default parser for this in a separate folder or through additional packages from the contrib package

or we add functionality externally that makes it easier to write custom parsing externally

you are welcome to support us in v3 with a proposal

PR #2647, opened 2023-09-25
♻️ Refactor Ctx.Method func to improve code readability

dairlair:
## Description

I've simplified the function by breaking it down into three trivially simple steps:

Step One: We check if no arguments are provided in the `override` parameter, in which case we have no choice but to return `c.method`. If arguments are provided, we proceed to step two.

Step Two: We take the first string from `override` and validate if it's a valid HTTP Method name. If it's not, it's evident that we need to return the current method from the context. If it is valid, we move on to step three.

Now, in step three, we have a valid HTTP Method in the `override` parameter. We save this method and its code to the context and return the method's name as a string.

This pull request doesn't fix any bugs, but it improves the code's readability.

I decided to submit these changes because my younger sister (a budding programmer with a diagnosis of semantic dyslexia) asked me to explain what this function does, and I had to put in some effort to understand what was happening and explain to her that two consecutive 'if' statements are better than two nested ones. 🧡

## Type of change

- [x] Code quality improvement

## Checklist:

## Commit formatting:

:art:

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -942,7 +942,7 @@ func (c *Ctx) Method(override ...string) string {
 		// Provided override does not valid HTTP method, no override, return current method
 		return c.method
 	}
-	
+
 	c.method = method
 	c.methodINT = mINT
 	return c.method
--- a/ctx.go
+++ b/ctx.go
@@ -927,17 +927,24 @@ func (c *Ctx) Location(path string) {
 	c.setCanonical(HeaderLocation, path)
 }
 
-// Method contains a string corresponding to the HTTP method of the request: GET, POST, PUT and so on.
+// Method returns the HTTP request method for the context, optionally overridden by the provided argument.
+// If no override is given or if the provided override is not a valid HTTP method, it returns the current method from the context.
+// Otherwise, it updates the context's method and returns the overridden method as a string.
 func (c *Ctx) Method(override ...string) string {
-	if len(override) > 0 {
-		method := utils.ToUpper(override[0])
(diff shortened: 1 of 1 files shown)


ctx.go
```diff
@@ -927,17 +927,24 @@ func (c *Ctx) Location(path string) {
 	c.setCanonical(HeaderLocation, path)
 }
 
-// Method contains a string corresponding to the HTTP method of the request: GET, POST, PUT and so on.
+// Method returns the HTTP request method for the context, optionally overridden by the provided argument.
+// If no override is given or if the provided override is not a valid HTTP method, it returns the current method from the context.
+// Otherwise, it updates the context's method and returns the overridden method as a string.
```
sixcolors, line 932: The first comment line clearly conveys the functionality of the method, which is to retrieve the HTTP request method for the context while allowing for an optional override. While the subsequent lines provide additional context and detail, they might be considered somewhat redundant as they essentially restate what's already explained in the first line. By omitting these extra lines, we can maintain code clarity and simplicity, ensuring that the comment remains informative without unnecessary verbosity.

sixcolors (commented):
Great job on the PR! LGTM. Thanks for your contribution 🧼

ctx.go
sixcolors: Reading comment again, LGTM as is. Comments should stay.

sixcolors (approved):

ReneWerner87 (approved):

PR #2650, opened 2023-09-27
ctx: change signatures of GetReqHeaders and GetRespHeaders

efectn:
## Description
Fixes https://github.com/gofiber/fiber/issues/2594

## Type of change

## Checklist:

## Commit formatting:

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -651,10 +651,11 @@ func (c *Ctx) GetRespHeader(key string, defaultValue ...string) string {
 // GetReqHeaders returns the HTTP request headers.
 // Returned value is only valid within the handler. Do not store any references.
 // Make copies or use the Immutable setting instead.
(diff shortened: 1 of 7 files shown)


ctx.go
```diff
@@ -1752,7 +1754,7 @@ func (c *Ctx) SendStream(stream io.Reader, size ...int) error {
 
 // Set sets the response's HTTP header field to the specified key, value.
 func (c *Ctx) Set(key, val string) {
-	c.fasthttp.Response.Header.Set(key, val)
+	c.fasthttp.Response.Header.Add(key, val)
```
ReneWerner87, line 1757: would not miss this, why is it necessary ?

efectn: > would not miss this, why is it necessary ?

Yeah i changed it by mistake 🤦

gaby (approved):

PR #2655, opened 2023-09-30
✨ feat: add Delete method to Store struct in session middleware

KaptinLin:
## Description

This pull request introduces a new `Delete` method to the `Store` struct in the Fiber's session middleware. The `Delete` method is designed to delete a specific session id from the storage.

The motivation behind this change is to provide more control over individual session management, especially in scenarios like administrator-enforced user logout or user-initiated logout from a specific device session. With the `Delete` method, we can easily delete a specific session, providing a more efficient and flexible way to manage individual user sessions.

This pull request also includes relevant test code to ensure the correct functionality of the `Delete` method. 

## Type of change

## Checklist:

## Usage
```go
app.Post("/admin/session/:id/logout", func(c *fiber.Ctx) error {
    // Get session id from request
    sessionID := c.Params("id")

    // Delete the session
    if err := store.Delete(sessionID); err != nil {
        return c.Status(500).SendString(err.Error())
    }

    return c.SendString("Logout successful")
})
```

### DIFF
--- a/middleware/session/store.go
+++ b/middleware/session/store.go
@@ -140,3 +140,11 @@ func (s *Store) responseCookies(c *fiber.Ctx) (string, error) {
 func (s *Store) Reset() error {
 	return s.Storage.Reset()
 }
+
+// Delete deletes a session by its id
+func (s *Store) Delete(id string) error {
+	if len(id) == 0 {
+		return fmt.Errorf("session id cannot be empty")
+	}
+	return s.Storage.Delete(id)
+}
--- a/middleware/session/store.go
+++ b/middleware/session/store.go
@@ -2,6 +2,7 @@ package session
 
 import (
 	"encoding/gob"
+	"errors"
 	"fmt"
 	"sync"
 
@@ -11,6 +12,9 @@ import (
 
 	"github.com/valyala/fasthttp"
 )
+
+// ErrEmptySessionID is an error that occurs when the session ID is empty.
+var ErrEmptySessionID = errors.New("session id cannot be empty")
 
 type Store struct {
 	Config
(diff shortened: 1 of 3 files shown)


gaby (approved):

middleware/session/store.go
```diff
@@ -140,3 +140,11 @@ func (s *Store) responseCookies(c *fiber.Ctx) (string, error) {
 func (s *Store) Reset() error {
 	return s.Storage.Reset()
 }
+
+// Delete deletes a session by its id
+func (s *Store) Delete(id string) error {
+	if len(id) == 0 {
```
efectn, line 146: `id == ""` should be more idiomatic here since this is not a slice.

middleware/session/store.go
```diff
@@ -140,3 +140,11 @@ func (s *Store) responseCookies(c *fiber.Ctx) (string, error) {
...
+		return fmt.Errorf("session id cannot be empty")
```
efectn, line 147: Create new public error

middleware/session/store_test.go
```diff
@@ -85,3 +85,41 @@ func TestStore_Get(t *testing.T) {
...
+	}
+
+	// Try to get the session again
+	session, err = store.Get(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// The session ID should be different now, because the old session was deleted
+	if session.ID() == sessionID {
```
efectn, line 122: Use utils.AssertEqual here

middleware/session/store_test.go
```diff
@@ -85,3 +85,41 @@ func TestStore_Get(t *testing.T) {
...
+
+	// Delete the session
+	if err := store.Delete(sessionID); err != nil {
+		t.Fatal(err)
+	}
+
+	// Try to get the session again
+	session, err = store.Get(ctx)
+	if err != nil {
+		t.Fatal(err)
```
efectn, line 118: Here too

ReneWerner87 (approved):

ReneWerner87:
@KaptinLin pls check the review comments

efectn (approved):

Issue #2646, opened 2023-09-25
🐛 [Bug]:Incorrect log depth when use log.WithContext 


0xmkzt:
### Bug Description

When i use log.WithContext, the log depth is more deeper.

### How to Reproduce

1. Run the code snippet
2. Invoke interface `curl -i http://127.0.0.1:3000/ready`
3. You can see `...router.go:145: [Info] ready`

### Expected Behavior

Actually it should be `...main.go:14: [Info] ready`

### Fiber Version

v2.49.2

### Code Snippet (optional)

```go
package main

import (
	"github.com/gofiber/fiber/v2"
	"github.com/gofiber/fiber/v2/log"
)

func main() {
	app := fiber.New()
	app.Get("/ready", func(c *fiber.Ctx) error {
		log.Info(">>>")
		log.WithContext(c.UserContext()).Info("ready")
		log.Info("<<<")
		return c.SendString("ready")
	})
	err := app.Listen(":3000")
	if err != nil {
		log.Error(err)
		return
	}
}
```

### Checklist:

- [X] I agree to follow Fiber's [Code of Conduct](https://github.com/gofiber/fiber/blob/master/.github/CODE_OF_CONDUCT.md).
- [X] I have checked for existing issues that describe my problem prior to opening this one.
- [X] I understand that improperly formatted bug reports may be closed without explanation.

ReneWerner87:
Duplicate of https://github.com/gofiber/contrib/issues/791

ReneWerner87:
@Skyenought can you help here?

barisbll:
@ReneWerner87 if this issue is free I would like to have it as my first contribution to gofiber 😊

nikzayn:
I would like to take this up!

/assign

Skyenought:
i will fix bug in fiber soon

PR #2666, opened 2023-10-07
🩹Fix incorrect log depth when use log.WithContext

Skyenought:
## Description

Fix incorrect log depth when use log.WithContex

Fixes # (issue)
https://github.com/gofiber/fiber/issues/2646

## Type of change

## Checklist:

## Commit formatting:

### DIFF
--- a/log/default.go
+++ b/log/default.go
@@ -188,7 +188,11 @@ func (l *defaultLogger) Panicw(msg string, keysAndValues ...interface{}) {
 }
 
 func (l *defaultLogger) WithContext(_ context.Context) CommonLogger {
-	return l
+	return &defaultLogger{
+		stdlog: l.stdlog,
+		level:  l.level,
+		depth:  l.depth - 1,
+	}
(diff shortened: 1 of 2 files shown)


log/default.go
```diff
@@ -188,7 +188,11 @@ func (l *defaultLogger) Panicw(msg string, keysAndValues ...interface{}) {
 }
 
 func (l *defaultLogger) WithContext(_ context.Context) CommonLogger {
-	return l
+	return &defaultLogger{
+		stdlog: l.stdlog,
```
efectn, line 192: Are there any reason that you initialize new struct instead of adjusting the old one's depth field? Is it for usecases we use WithContext and without context together?

Skyenought: yes. The `WithContext` record is in the wrong place, so it needs to be adjusted.

efectn (approved):

ReneWerner87 (approved):

Issue #2400, opened 2023-04-02
🚀 [Feature]: Sustainably prevent data races


leonklingele:
### Feature Description

The data race "fixed" in https://github.com/gofiber/fiber/pull/2368/files [wasn't fixed at all](https://github.com/gofiber/fiber/pull/2368#issuecomment-1467891174).

Instead we should do something à la:

```go
func (s *Storage) ConnLocked(fn func(map[string]entry)) {
    s.mux.Lock()
    defer s.mux.Unlock()
    fn(s.db)
}
```

### Checklist:

- [X] I agree to follow Fiber's [Code of Conduct](https://github.com/gofiber/fiber/blob/master/.github/CODE_OF_CONDUCT.md).
- [X] I have checked for existing issues that describe my suggestion prior to opening this one.
- [X] I understand that improperly formatted feature requests may be closed without explanation.

Hrily:
Hey folks

I've raised a PR #2669 to fix this.

It also adds a unit test to check for race condition, which would otherwise fail in older implementation or without using lock.

PR #2669, opened 2023-10-08
🩹 Fix: race condition in memory storage

Hrily:
## Description

This commit updates `Conn` method of memory storage to fix race conditions.

The `Conn` method returned the reference to underlying map which could be used
for read/write and had a risk of race condition.

This commit fixes it by updating the method to take in a function to access
underlying map, which is wrapped in lock to make it thread safe.

Fixes: #2400

## Type of change

## Checklist:

## Commit formatting:

### DIFF
--- a/internal/storage/memory/memory.go
+++ b/internal/storage/memory/memory.go
@@ -137,9 +137,9 @@ func (s *Storage) gc() {
 	}
 }
 
-// Return database client
-func (s *Storage) Conn() map[string]entry {
-	s.mux.RLock()
-	defer s.mux.RUnlock()
-	return s.db
+// ConnLocked can be used to access the underlying db in a thread-safe manner.
+func (s *Storage) ConnLocked(f func(map[string]entry)) {
+	s.mux.Lock()
+	defer s.mux.Unlock()
+	f(s.db)
 }
--- a/internal/storage/memory/memory_test.go
+++ b/internal/storage/memory/memory_test.go
@@ -124,9 +124,39 @@ func Test_Storage_Memory_Close(t *testing.T) {
 	utils.AssertEqual(t, nil, testStore.Close())
 }
 
-func Test_Storage_Memory_Conn(t *testing.T) {
+func Test_Storage_Memory_ConnLocked(t *testing.T) {
 	t.Parallel()
-	utils.AssertEqual(t, true, testStore.Conn() != nil)
+	assertConn := func(db map[string]entry) {
+		utils.AssertEqual(t, true, db != nil)
+	}
+	testStore.ConnLocked(assertConn)
+}
+
+func Test_Storage_Memory_ConnLocked_Race(t *testing.T) {
+	key := "test_race"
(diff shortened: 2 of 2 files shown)


internal/storage/memory/memory.go
```diff
@@ -137,9 +137,9 @@ func (s *Storage) gc() {
 	}
 }
 
-// Return database client
-func (s *Storage) Conn() map[string]entry {
-	s.mux.RLock()
-	defer s.mux.RUnlock()
-	return s.db
+// ConnLocked can be used to access the underlying db in a thread-safe manner.
+func (s *Storage) ConnLocked(f func(map[string]entry)) {
```
gaby, line 141: This is an interface, we shouldn't rename it to avoid breaking changes.

gaby (changes requested):

internal/storage/memory/memory.go
Hrily: @gaby What interface does this implement?

I can see the earlier `Conn` method only on this struct
https://sourcegraph.com/search?q=context:global+repo:https://github.com/gofiber/fiber/+%22Conn()%22&patternType=standard&sm=1&groupBy=repo

gaby: https://github.com/gofiber/storage/blob/main/storage.go

We may need to discuss this, since it's implemented differently in multiple places 🤔

Hrily: As far as I understand, this method is only provided by `memory.Storage`, because I don't think any other storage will return a `Conn` of type  `map[string]entry`.

ReneWerner87: no this method is provided by all adapters
https://github.com/search?q=repo%3Agofiber%2Fstorage%20conn&type=code

Hrily: I see

I think we can make `Storage.db` a `sync.Map` so we can return it as it is from `Conn()`, without worrying about thread safety.

Thoughts @gaby , @ReneWerner87 ?

Hrily: A ping on this in case it was missed. @gaby @ReneWerner87

ReneWerner87: sync.Map is ok, but slower or ?

Hrily: I can run the `Benchmark_Storage_Memory` and see if it's better or worse.

ReneWerner87:
i just thought about the bug (https://github.com/gofiber/fiber/issues/2360) again
can you please first create a test case which produces the error

with the sync.map I don't know yet, is already a considerable time loss for the driver which should normally be fastest

ReneWerner87:
maybe it would also be ok if we modify the memory storage driver, as long as we don't modify the internal memory driver which is the default one

skydiver1980 (approved):

ReneWerner87:
mutext handling should take place outside or with an adjustment of the storage interface in the storage project

PR #2698, opened 2023-11-01
📚 Doc: Add additional information as to why GetReqHeaders returns a map where the values are slices of strings 

moritz157:
Added additional information as to why GetReqHeaders returns a map where the values are slices of strings (instead of a single string as one might expect)

## Description

In a (more or less) recent breaking change, the signature of [GetReqHeaders](https://docs.gofiber.io/api/ctx#getreqheaders) was changed to return a map where the values are not single strings anymore but instead slices of strings. As someone who encountered this breaking change after upgrading fiber, it was not clear why this was changed and - more importantly - what exactly would be contained in these slices. Therefore I added a short and simple explanation what one can expect to be returned by this method.

Documents the intent of #2594

## Type of change

### DIFF
--- a/docs/api/ctx.md
+++ b/docs/api/ctx.md
@@ -618,7 +618,7 @@ app.Get("/", func(c *fiber.Ctx) error {
 
 ## GetRespHeaders
 
-Returns the HTTP response headers.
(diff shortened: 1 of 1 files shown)


efectn (commented):
Can you update getrespheaders docs as well

moritz157:
sure, done

ReneWerner87 (approved):

Issue #2571, opened 2023-08-11
🚀 [Feature]: Adding Default Value Support for QueryParser


gopkg-dev:
### Feature Description

Currently, QueryParser in Fiber does not support setting default values for query parameters directly in the struct. This enhancement request suggests adding the ability to define default values for query parameters within the struct tags.

### Additional Context (optional)

https://github.com/gofiber/fiber/issues/1306

### Code Snippet (optional)

```go
type Request struct {
	Limit int `query:"limit" default:"10"`
}
var request Request
ctx.QueryParser(&request);
limit := request.Limit // 10
```

### Checklist:

- [X] I agree to follow Fiber's [Code of Conduct](https://github.com/gofiber/fiber/blob/master/.github/CODE_OF_CONDUCT.md).
- [X] I have checked for existing issues that describe my suggestion prior to opening this one.
- [X] I understand that improperly formatted feature requests may be closed without explanation.

manujgrover71:
@gopkg-dev I would like to take this up. 

Could you please assign the same to me?

manujgrover71:
As pointed out by @hi019  in the issue #1306, a [go-default package](https://github.com/mcuadros/go-defaults) with similar functionality to our current need already exists.

One potential solution is to simply incorporate the aforementioned package into QueryParser (since this package has been tried and tested and is widely used, we are confident in its accuracy and can announce that the go-default default functionality is now supported). 

@ReneWerner87 @gopkg-dev Thoughts?

ReneWerner87:
But then we can use that outside and we just mention it in the documentation for a possible solution to the problem

Since this is independent and can be used separately I would not increase the fiber code and dependencies with it 

Because this packet can then also be used for use with the body or header parser just before the structs are passed into the methods

manujgrover71:
> Because this packet can then also be used for use with the body or header parser just before the structs are passed into the methods

Yes, that makes sense.

However, considering our aim is to incorporate this functionality within struct tags, we would need to decide between utilizing or building a tag parser mechanism and using reflection to populate the fields.

As you suggested, we could advise developers to leverage an external library for this purpose (could be go-default or maybe any other). Alternatively, I'm open to the idea of developing a custom implementation if that aligns better with our goals.

I'd be interested to hear your thoughts on this or any other potential implementations you might have in mind

gaby:
The https://github.com/mcuadros/go-defaults package hasn't been updated in years, is there an alternative.

[...]

PR #2699, opened 2023-11-01
Query, Header Default Value Tag Feature

muratmirgun:
## Description

This Feature provides an approach in Go to assign default values to the fields of structs. The tagHandlers function assigns these default values based on the specific type of a field. To enhance performance, we cache which fields of the structs have default values (structCache), so we don't have to repeatedly look up these fields every time. Because Handlers Query structs not change

Fixes:
#2571 

## Type of change

## Checklist:

## Commit formatting:


ReneWerner87:
Thx

Can you share the benchmark results
How big is the impact for cases without defaults, do we have pay for that as well?

Can you put the default value handling  and the tests in a separate go file (don't like big files, I'm more of a code divider)?
If we have this built in for queryparser, it would probably make sense to build it in for the other parsers (body & header) as well.

Should we add a config setting so you can disable this handling to get the maximum performance if you don't need defaults or have your own handling?

muratmirgun:
> Thx
> 
> Can you share the benchmark results How big is the impact for cases without defaults, do we have pay for that as well?
> 
> Can you put the default value handling and the tests in a separate go file (don't like big files, I'm more of a code divider)? If we have this built in for queryparser, it would probably make sense to build it in for the other parsers (body & header) as well.
> 
> Should we add a config setting so you can disable this handling to get the maximum performance if you don't need defaults or have your own handling?

First, let me share the benchmark results. but These examples for ColdStart

Benchmark_Ctx_QueryParser
Benchmark_Ctx_QueryParser-8                                       653192              1621 ns/op             856 B/op         38 allocs/op
Benchmark_Ctx_QueryParser-8                                       751180              1611 ns/op             856 B/op         38 allocs/op
Benchmark_Ctx_QueryParser-8                                       753249              1563 ns/op             856 B/op         38 allocs/op
Benchmark_Ctx_QueryParser-8                                       771404              1646 ns/op             856 B/op         38 allocs/op

Benchmark_Ctx_QueryParserWithDefaultValues
Benchmark_Ctx_QueryParserWithDefaultValues-8             974080              1239 ns/op             720 B/op         25 allocs/op
Benchmark_Ctx_QueryParserWithDefaultValues-8             941806              1245 ns/op             720 B/op         25 allocs/op
Benchmark_Ctx_QueryParserWithDefaultValues-8             963400              1220 ns/op             720 B/op         25 allocs/op
Benchmark_Ctx_QueryParserWithDefaultValues-8             963571              1236 ns/op             720 B/op         25 allocs/op

Benchmarks for multiple Request Scenario:

Benchmark_Ctx_QueryParser
Benchmark_Ctx_QueryParser-8                                        14372             78933 ns/op           42817 B/op       1900 allocs/op
Benchmark_Ctx_QueryParser-8                                        15265             77700 ns/op           42817 B/op       1900 allocs/op
Benchmark_Ctx_QueryParser-8                                        15346             78469 ns/op           42817 B/op       1900 allocs/op
Benchmark_Ctx_QueryParser-8                                        15432             78051 ns/op           42817 B/op       1900 allocs/op

Benchmark_Ctx_QueryParserWithDefaultValues
Benchmark_Ctx_QueryParserWithDefaultValues-8              19756             60628 ns/op           36018 B/op       1250 allocs/op
Benchmark_Ctx_QueryParserWithDefaultValues-8              19581             60628 ns/op           36018 B/op       1250 allocs/op
Benchmark_Ctx_QueryParserWithDefaultValues-8              19682             60993 ns/op           36017 B/op       1250 allocs/op
Benchmark_Ctx_QueryParserWithDefaultValues-8              19599             60733 ns/op           36018 B/op       1250 allocs/op



With multiple

Apart from that, we can do it for all parsers, it will be logical and we can do them with a separate file logic. In the cold start logic, there is a small cost involuntarily at first until the first cache settles, but there will be no problem after the request comes to all the handlers that contain the default structure.

If we are going to integrate it into all parse structures, we can control this with the config for users who are worried about speed.

ReneWerner87:
Can you share the performance without and with your function for the queryparser (i.e. the times from before vs. after the adjustment)

muratmirgun:
> Can you share the performance without and with your function for the queryparser (i.e. the times from before vs. after the adjustment)

Sorry, I forgot it. 

Benchmark_Ctx_QueryParser
Benchmark_Ctx_QueryParser-8               679710              1629 ns/op             856 B/op         38 allocs/op
Benchmark_Ctx_QueryParser-8               740930              1625 ns/op             856 B/op         38 allocs/op
Benchmark_Ctx_QueryParser-8               752588              1584 ns/op             856 B/op         38 allocs/op
Benchmark_Ctx_QueryParser-8               763689              1578 ns/op             856 B/op         38 allocs/op
Benchmark_Ctx_QueryParser_Comma
Benchmark_Ctx_QueryParser_Comma-8         665384              1764 ns/op             928 B/op         44 allocs/op
Benchmark_Ctx_QueryParser_Comma-8         690904              1766 ns/op             928 B/op         44 allocs/op
Benchmark_Ctx_QueryParser_Comma-8         678898              1767 ns/op             928 B/op         44 allocs/op
Benchmark_Ctx_QueryParser_Comma-8         683109              1754 ns/op             928 B/op         44 allocs/op
PASS
ok      github.com/gofiber/fiber/v2     9.871s

gaby:
@muratmirgun Can you take a look at the ci failures

ReneWerner87:
move the functionality to the utils folder

if you can then please add the function for all parsers and add a config setting

muratmirgun:
> move the functionality to the utils folder
> 
> if you can then please add the function for all parsers and add a config setting

Okey I am working on it! 🚀

utils/default.go
```diff
@@ -0,0 +1,94 @@
...
+				field.SetFloat(f)
+			}
+		}
+	case reflect.Bool:
+		if !field.Bool() {
+			if b, err := strconv.ParseBool(tagValue); err == nil {
+				field.SetBool(b)
+			}
+		}
+	case reflect.Slice:
```
efectn, line 34: Can you add support for single and multiple pointers here?

muratmirgun: Okey, I will add it newt commit.

ReneWerner87:
just change the PR type from draft in the normal one, when you are finished with the PR

muratmirgun:
> just change the PR type from draft in the normal one, when you are finished with the PR

Yes I know thanks for remember 💯

utils/default.go
```diff
@@ -0,0 +1,103 @@
...
+					intPtr := reflect.New(elemType.Elem())
+					intPtr.Elem().SetInt(intVal)
+					val = intPtr
+				}
+			}
+		case reflect.String:
+			val = reflect.ValueOf(item)
+		case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+			if intVal, err := strconv.ParseInt(item, 10, 64); err == nil {
+				val = reflect.ValueOf(int64(intVal))
```
efectn, line 64: Do we really need int64() here

muratmirgun: I delete it

muratmirgun:
I saw linter results and I am working on it @efectn

efectn (approved):
Mostly seems OK to me. Perhaps we can add more features in the future.

muratmirgun:
@efectn I don't have merge access you can merge

muratmirgun:
I have posted the fix code for int slice parser function you can check it. now test codes will not give error

ReneWerner87:
> @efectn I don't have merge access you can merge

Will look again tomorrow and merge then

ReneWerner87:
@muratmirgun can you please look at the lint errors

2. comments
- why do we set the default values after the filling process? it should be easier beforehand and then there is no chance that something will be overwritten by faulty code
- can you add the lines for the `BodyParser`, `ParamParser` and `CookieParser`, it should only be a small block which is added + tests

muratmirgun:
@ReneWerner87 Yes, I will work on the bugs and look at the additions you mentioned

muratmirgun:
@ReneWerner87 Sorry for delay. I've been a little busy because of my job change, so I will update the PR shortly.

skydiver1980 (approved):

ReneWerner87:
@muratmirgun did you have time to look again?

muratmirgun:
> @muratmirgun did you have time to look again?

Hi, I'll work on it today

utils/default.go
```diff
@@ -0,0 +1,124 @@
...
+		if val.IsValid() {
+
+			slice = reflect.Append(slice, val)
+		}
+	}
+
+	field.Set(slice)
+}
+
+var structCache = make(map[reflect.Type][]reflect.StructField)
```
rohitChaku, line 96: I was wondering if `structCache` here could cause race conditions on concurrent access?Maybe we should add a RW mutex?

rohitChaku:
Hello, I was checking the implementation of `default` values in gin, and it seems pretty neat. I was wondering is we could borrow some constructs from there? I'm primarily referring to the syntax as the logic is pretty similar [form_mapping.go](https://github.com/gin-gonic/gin/blob/master/binding/form_mapping.go#L167C10-L167C10).

I'm relatively new to Golang and open source, so I want to ensure my suggestion aligns with the community's spirit. If there's any concern or if this isn't appropriate, please let me know, and I'll promptly remove my comment. Thanks.

ReneWerner87:
@muratmirgun can you check and consider the last comments, thanks for your work, any help is appreciated

ReneWerner87:
- [ ] pls check the other parsers https://github.com/gofiber/fiber/pull/2699#issuecomment-1805345603
- [ ] can we check https://github.com/gofiber/fiber/pull/2699#discussion_r1422480231 again
- [ ] one last lint error https://github.com/gofiber/fiber/actions/runs/7310291693/job/20152420546?pr=2699

utils/default.go
ReneWerner87: fixed with https://github.com/gofiber/fiber/pull/2699/commits/ad89ced1a32fb033133ef1afe3c219a5cfeea702

ReneWerner87 (approved):

muratmirgun:
Hi, I am sorry for being offline lately. I have had some health problems and I have seen your articles and now I am looking at the example given. @ReneWerner87

ReneWerner87:
> Hi, I am sorry for being offline lately. I have had some health problems and I have seen your articles and now I am looking at the example given. @ReneWerner87

no problem, the pull request should be almost ready i have done the last steps with the other maintainers

health always comes first of course, hope you are well again

ReneWerner87:
image
image

muratmirgun:
> > Hi, I am sorry for being offline lately. I have had some health problems and I have seen your articles and now I am looking at the example given. @ReneWerner87
> 
> no problem, the pull request should be almost ready i have done the last steps with the other maintainers
> 
> health always comes first of course, hope you are well again

thank you very much 🙏 we will move forward again in the next issues🚀

ctx.go
```diff
@@ -410,28 +408,37 @@ func (c *Ctx) BodyParser(out interface{}) error {
...
+			utils.SetDefaultValues(out)
+		}
+
+		switch {
+		case strings.HasPrefix(ctype, MIMETextXML), strings.HasPrefix(ctype, MIMEApplicationXML):
+			if err := xml.Unmarshal(c.Body(), out); err != nil {
+				return fmt.Errorf("failed to unmarshal: %w", err)
+			}
+			return nil
+		case strings.HasSuffix(ctype, "json"):
```
gaby, line 437: Why was `json` moved down here? It used to be the first case

utils/default.go
```diff
@@ -0,0 +1,131 @@
...
+		if val.IsValid() {
+			slice = reflect.Append(slice, val)
+		}
+	}
+
+	field.Set(slice)
+}
+
+var (
+	mu          sync.Mutex
```
gaby, line 97: Can this be `RWMutex` ?

efectn: > Can this be `RWMutex` ?

It can. Then we should not have to be waiting for writing lock

ReneWerner87: yes you are right

ReneWerner87:
@muratmirgun maybe you can help, i found out 2 important things through the review

something is not thread safe -> added parallel test execution and analyzed the problems, unfortunately I can't find the right solution
if i try to correct it i get into a deadlock when running the tests because something is wrong with the mutex hirachy

the defaults cannot be applied deeply, which is what you would expect if you use the function from the utils directly, for example
-> I have created a test that proves this and helps to correct the problem

furthermore, i noticed that the performance could be optimized even further, instead of parsing and converting the value from the default tag every time, you could keep it in the same way or create a slice with clousures, which is later only run through with the element and sets the values
therefore the [go-defaults](https://github.com/mcuadros/go-defaults/tree/master) also used the [filler](https://github.com/mcuadros/go-defaults/blob/master/filler.go)

however, the biggest problem is that the functionality is not thread safe

muratmirgun:
I saw your last commits and new test cases I am looking for your comment maybe we can make a mechanism where we can separate the default parts. I've started working on it. @ReneWerner87

muratmirgun:
I am looking the last changes from this pr for sync

ReneWerner87:
another good library
https://github.com/creasty/defaults/tree/master

https://github.com/gofiber/fiber/issues/2571#issuecomment-1684924938
I'm asking myself again whether we should really integrate this, as the performance could suffer and you can simply use it from outside

```go
import "github.com/creasty/defaults"
```
and
```go
defaults.Set(myStruct)
```
is enough to do it outside

muratmirgun:
that's a good question, I made a structure like this at first with the aim of less dependency because a performance problem on the library side can affect us very seriously. we must be in control. if I can't solve the thread safe problem, I will also examine the structures in this repo on the side. Because

ReneWerner87:
we exclude this feature from the current release for now
if you find a good solution we can put it in the next version or v3

by the way, it is planned to be the last v2 release
but maybe we can still use it as an after feature

ReneWerner87:
@muratmirgun can you work on the last notes? in the v3 branch or should we rather just add documentation and recommend other packages?

efectn:
Hi @muratmirgun the PR is now closed as default feature was added in the last versions of Gorilla schema and it's been just updated for Fiber. Anyway, thanks for your contribution!

PR #2194, opened 2022-11-04
Add URL prefix to pprof middleware

gmlewis:
## Description

In systems with multiple ingress endpoints all sharing the same load balancer and base URL, it is common to place servers behind a URL prefix. To facilitate this, this PR allows the pprof middleware to define a URL Prefix during configuration so that the endpoints are served on an appropriate URL.

## Type of change

Please delete options that are not relevant.

- [ ] New feature (non-breaking change which adds functionality)
- [ ] This change requires a documentation update

## Checklist:

- [x] For new functionalities I follow the inspiration of the express js framework and built them similar in usage
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation - https://github.com/gofiber/docs for https://docs.gofiber.io/
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] If new dependencies exist, I have checked that they are really necessary and agreed with the maintainers/community (we want to have as few dependencies as possible)
- [x] I tried to make my code as fast as possible with as few allocations as possible
- [ ] For new code I have written benchmarks so that they can be analyzed and improved

## Commit formatting:

Use emojis on commit messages so it provides an easy way of identifying the purpose or intention of a commit. Check out the emoji cheatsheet here: https://gitmoji.carloscuesta.me/

PR #2709, opened 2023-11-06
⚡ middleware/pprof: improve performance

database64128:
## Description

Concatenate the custom and fixed prefixes beforehand, so the trimmed path can be switched on against constant strings.

Here's a simple benchmark:

```go
package pprof

import (
	"crypto/rand"
	"errors"
	"strings"
	"testing"
	"unsafe"
)

var errBadPath = errors.New("bad path")

func doSlow(prefix, path string) error {
	if len(path) < 12 || !strings.HasPrefix(path, prefix+"/debug/pprof") {
		return errBadPath
	}
	// Switch to original path without stripped slashes
	switch path {
	case prefix + "/debug/pprof/":
		return errBadPath
	case prefix + "/debug/pprof/cmdline":
		return errBadPath
	case prefix + "/debug/pprof/profile":
		return errBadPath
	case prefix + "/debug/pprof/symbol":
		return errBadPath
	case prefix + "/debug/pprof/trace":
		return errBadPath
	case prefix + "/debug/pprof/allocs":
		return errBadPath
	case prefix + "/debug/pprof/block":
		return errBadPath
	case prefix + "/debug/pprof/goroutine":
		return errBadPath
	case prefix + "/debug/pprof/heap":
		return errBadPath
	case prefix + "/debug/pprof/mutex":
		return errBadPath
	case prefix + "/debug/pprof/threadcreate":
		return errBadPath
	default:
		return nil
	}
}

func newSlowBenchFunc(prefix, path string) func(b *testing.B) {
	return func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			if err := doSlow(prefix, path); err != nil {
				b.Error(err)
			}
		}
	}
}

func doFast(prefix, path string) error {
	path, found := strings.CutPrefix(path, prefix)
	if !found {
		return errBadPath
	}
	// Switch to original path without stripped slashes
	switch path {
	case "", "/":
		return errBadPath
	case "/cmdline":
		return errBadPath
	case "/profile":
		return errBadPath
	case "/symbol":
		return errBadPath
	case "/trace":
		return errBadPath
	case "/allocs":
		return errBadPath
	case "/block":
		return errBadPath
	case "/goroutine":
		return errBadPath
	case "/heap":
		return errBadPath
	case "/mutex":
		return errBadPath
	case "/threadcreate":
		return errBadPath
	default:
		return nil
	}

}

func newFastBenchFunc(prefix, path string) func(b *testing.B) {
	prefix += "/debug/pprof"
	return func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			if err := doFast(prefix, path); err != nil {
				b.Error(err)
			}
		}
	}
}

func BenchmarkPprof(b *testing.B) {
	buf := make([]byte, 16)
	if _, err := rand.Read(buf); err != nil {
		b.Fatal(err)
	}
	prefix := unsafe.String(unsafe.SliceData(buf), len(buf))
	path := prefix + "/debug/pprof/nonexistent"

	b.Run("Slow", newSlowBenchFunc(prefix, path))
	b.Run("Fast", newFastBenchFunc(prefix, path))
}
```

```
goos: linux
goarch: amd64
pkg: github.com/gofiber/fiber/v2/middleware/pprof
cpu: 13th Gen Intel(R) Core(TM) i9-13900K
BenchmarkPprof/Slow-32         	 4912642	       246.3 ns/op	     480 B/op	      10 allocs/op
BenchmarkPprof/Fast-32         	411908472	         2.913 ns/op	       0 B/op	       0 allocs/op
PASS
```

To the maintainers: These slow patterns introduced by #2194 were quite obvious, and it would've been nice if they were called out when the PR was being reviewed. Thanks!

## Type of change

## Checklist:

## Commit formatting:

### DIFF
--- a/middleware/pprof/pprof.go
+++ b/middleware/pprof/pprof.go
@@ -29,6 +29,9 @@ func New(config ...Config) fiber.Handler {
 		pprofThreadcreate = fasthttpadaptor.NewFastHTTPHandlerFunc(pprof.Handler("threadcreate").ServeHTTP)
 	)
 
+	// Construct actual prefix
+	prefix := cfg.Prefix + "/debug/pprof"
+
 	// Return new handler
 	return func(c *fiber.Ctx) error {
 		// Don't execute middleware if Next returns true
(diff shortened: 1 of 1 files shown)


efectn (approved):
Nice catch

ReneWerner87:
We need to support go1. 17

middleware/pprof/pprof.go
```diff
@@ -38,43 +41,53 @@ func New(config ...Config) fiber.Handler {
...
 
 			return c.Redirect(path, fiber.StatusFound)
 		}
 		return nil
 	}
 }
+
+// cutPrefix is a copy of [strings.CutPrefix] added in Go 1.20.
+// Remove this function when we drop support for Go 1.19.
+func cutPrefix(s, prefix string) (after string, found bool) {
```
database64128, line 88: The `nonamedreturns` linter does not like its signature. Should I drop the names to make the linter happy, or add `//nolint:nonamedreturns` so it stays the same with its equivalent in std?

ReneWerner87: @leonklingele what do you think? 
golang is also using named returns https://cs.opensource.google/go/go/+/refs/tags/go1.21.3:src/strings/strings.go;l=1289

ReneWerner87 (approved):

middleware/pprof/pprof.go
database64128: Went ahead with `//nolint:nonamedreturns`.

database64128: Updated the comment to make the linter happy.

gaby (approved):
LGTM!

PR #2716, opened 2023-11-10
test(limiter): fix intermittent failures

nickajacks1:
The limiter middleware unit tests are failing due to a race between the storage garbage collector and the unit test itself. The sliding window limiter tracks requests using memory storage. In several of the unit tests, this storage expiry ends up being 4 seconds. The test waits for 4 seconds, then sends a request, expecting it to succeed. However, the unit test occasionally wakes up before the storage GC kicks in. As an effect of the very coarse timer (using seconds as units), the middleware correctly rejects the request, causing the test to fail.

Update the sleep to 4.5 seconds. This will not slow down the execution of the test suite, as these tests run in parallel with a separate 9 second long test.

I'm not 100% sure this solves the issue, and ideally we'd be able to run tests without time.Sleep.

## Description

## Type of change

- [x] Unit Tests

## Checklist:

### DIFF
--- a/middleware/limiter/limiter_test.go
+++ b/middleware/limiter/limiter_test.go
@@ -218,7 +218,7 @@ func Test_Limiter_Sliding_Window_No_Skip_Choices(t *testing.T) {
 	utils.AssertEqual(t, nil, err)
 	utils.AssertEqual(t, 429, resp.StatusCode)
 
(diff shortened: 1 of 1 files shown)


nickajacks1:
CI failed on CORS and Session middleware tests

ReneWerner87 (approved):

PR #2720, opened 2023-11-11
✅ test: fix failing csrf test

nickajacks1:
## Description

A test validating that expired tokens fail was hitting a race condition with garbage collection. Sometimes, an assertion that expects memory storage GC to have triggered happens too quickly, causing the assertion to fail. Give the GC a little bit more time to process before asserting.

## Type of change

## Checklist:


nickajacks1:
CI failed on Test_Client_Agent_Dest and Session middleware tests. There were some races reported logs.

ReneWerner87 (approved):

gaby (approved):

middleware/csrf/csrf_test.go
```diff
@@ -186,7 +186,7 @@ func Test_CSRF_ExpiredToken(t *testing.T) {
 	utils.AssertEqual(t, 200, ctx.Response.StatusCode())
 
 	// Wait for the token to expire
-	time.Sleep(1 * time.Second)
```
sixcolors, line 189: Why not just `1250*time.Millisecond`

sixcolors (changes requested):

middleware/csrf/csrf_test.go
```diff
@@ -186,7 +186,7 @@ func Test_CSRF_ExpiredToken(t *testing.T) {
...
+	time.Sleep(1*time.Second + 250*time.Millisecond)
```
sixcolors, line 189: Why not just 1250*time.Millisecond

nickajacks1: Good point

sixcolors (approved):

PR #2730, opened 2023-11-18
Changing default log output

### LINKED ISSUE #2729: 🐛 [Bug]: Logger middleware default Format isn't the format that's used
### Bug Description

> I'm filing this as a bug since I think the behaviour is counter-intuitive.

The logger middleware has a configuration with a default format: https://github.com/gofiber/fiber/blob/2374cad3cd109d3bf33f5863e6eb14c32de1882e/middleware/logger/config.go#L89

However, when the logger proceeds to print this, it ignores the actual format: https://github.com/gofiber/fiber/blob/2374cad3cd109d3bf33f5863e6eb14c32de1882e/middleware/logger/logger.go#L141-L174

Despite it checking for if it should use the default format, it then uses a different format string in the `fmt.Sprintf()` calls.

### How to Reproduce

* Add the logging middleware without overriding the `Format`
* Do a request and observe the log output

### Expected Behavior

Logged line should use the `ConfigDefault.Format` for output, as shown in the configuration: https://docs.gofiber.io/api/middleware/logger#default-config

### Fiber Version

v2.50.0

### Code Snippet (optional)

_No response_

### Checklist:

brunodmartins:
## Description

The default log Format described on [here](https://github.com/brunodmartins/fiber/blob/1b060cb150c6eab80f71e51120b05da555b43c27/middleware/logger/config.go#L89) is not respected on the Log middleware, as described on the related issue.

Looking at the beginning of the middleware, before it starts it completes the configuration filling the missing options with the default ones, thus, if a user do not override the Format parameter, the default one shall be used instead. This configuration is later on loaded on the `templateChain` and `logFunChain`.

```golang
templateChain, logFunChain, err := buildLogFuncChain(&cfg, createTagMap(&cfg))
```

Thus, even if a user do not provide a Format, the middleware shall build a log func chain and could use it later. 

To **avoid breaking** client logs, I modified the expected default log format, to be equal as the one that has been used on the middleware. 

image

Fixes #2729

## Type of change

## Checklist:

## Commit formatting:


middleware/logger/logger.go
```diff
@@ -85,9 +85,6 @@ func New(config ...Config) fiber.Handler {
 			return c.Next()
 		}
 
-		// Alias colors
-		colors := c.App().Config().ColorScheme
```
gaby, line 89: Why was this removed?

gaby (changes requested):
We need a unit-test to validate the logs are in the expected output format

middleware/logger/logger.go
brunodmartins: Because this variable was not used anymore

brunodmartins:
Ouch! It's was passing here! I will check it again

brunodmartins:
@gaby the test that failed is not related to my change: 

--- FAIL: Test_CSRF_ExpiredToken (1.00s)

Could this be another problem?

ReneWerner87:
> @gaby the test that failed is not related to my change:
> 
> --- FAIL: Test_CSRF_ExpiredToken (1.00s)
> 
> Could this be another problem?

yes, there are some tests with race conditions -> we are on it

skydiver1980 (approved):

middleware/logger/logger.go
```diff
@@ -137,56 +134,6 @@ func New(config ...Config) fiber.Handler {
 		// Get new buffer
 		buf := bytebufferpool.Get()
 
-		// Default output when no custom Format or io.Writer is given
-		if cfg.Format == ConfigDefault.Format {
```
ReneWerner87, line 141: not sure, is this part no longer needed ? where is this code now covered ?

gaby:
@brunodmartins Can you fix the merge conflicts?

brunodmartins:
@gaby I solved the merge commit with my changes, as I commented on the PR description: to do not break current client logs format, I kept the one that as been used due to the bug

middleware/logger/logger.go
ReneWerner87: i checked it again -> its fine

ReneWerner87:
i tested the change
image

without colors:
image

default with colors:
image

with special format:
image

LGTM

gaby (approved):

middleware/logger/logger.go
ReneWerner87: thanks, this old colorization is therefore superfluous, since the enabledColors are queried in the methods for the tags

ReneWerner87 (approved):

ReneWerner87:
@brunodmartins thank you for cleaning up, now we have one less big block to maintain
and sorry for the long wait

brunodmartins:
Thanks for the help on this contribution =)

efectn (approved):

Issue #2684, opened 2023-10-19
📝  [Proposal]: Standardization of Middleware Context Key Prevent Collisions (Breaking Change)


sixcolors:
## Issue Description

The middleware in GoFiber sets string values to `ctx.Locals` using non-standardized default context keys, and since these defaults are non-zero values this forces their existence. This behavior can lead to conflicts and confusion for users. Furthermore, the specificity of these context keys, i.e., their purpose and potential conflicts, should be addressed. 

Additionally, the presence of dots or dashes in context keys can cause issues when using `PassLocalsToViews`, which injects key-value pairs into various template engines. Even template engines that correctly handle dots and dashes may cause fiber to feel hacked together and not polished.

Below are some examples of middleware and their non-standardized default context keys:

- BasicAuth:
  - ContextUsername: "username"
  - ContextPassword: "password"

- CSRF:
  - SessionKey: "fiber.csrf.token"
  - HandlerContextKey: "fiber.csrf.handler"

- Keyauth:
  - ContextKey: "token"

- RequestID:
  - ContextKey: "requestid"

- Paseto:
  - ContextKey: "auth-token"

- JWT:
  - ContextKey: "user"

## Key Specificity and Conflicts

The default context keys provided by the middleware lack specificity, making it unclear what each key is for. This lack of clarity can lead to conflicts when multiple middleware components use similar or identical context keys for different purposes. Additionally, these keys are used to set string values to `ctx.Locals`, which can lead to unintended side effects if conflicts occur.

### Key Specificity Example

For instance, it's not immediately clear what "ContextUsername" and "ContextPassword" refer to in the BasicAuth middleware, and what kind of string values they set in `ctx.Locals`. Providing more specific key names can enhance user understanding.

### Potential Conflicts

Additionally, the non-standardized default context keys across middleware components can result in unintended conflicts when setting values in `ctx.Locals`. If multiple middleware components use the same key name for different purposes, it can lead to unexpected behavior and errors in `ctx.Locals`.

### Dots and Dashes in Context Keys

The use of dots or dashes in context keys, while generally accepted in many scenarios, can pose challenges when using `PassLocalsToViews`. This middleware injects key-value pairs into various template engines, including HTML, Django, Pug, Mustache, etc. These template engines may not handle dots and dashes correctly or may make it cumbersome to access keys with such characters.

## Proposed Solution

[...]

PR #2369, opened 2023-03-13
:bug: requestid.Config.ContextKey is interface{}

benjajaja:
## Description

https://github.com/gofiber/fiber/issues/2143 allows to use any as key for c.Locals(key, v). The RequestID middleware should be consistent with that.
This allows to use a custom type for context keys, like for other keys. The RequestID is likely read back somewhere else along other context keys.

Fixes #2356

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] This change requires a documentation update

## Checklist:

- [x] I have performed a self-review of my own code
- [x] I have made corresponding changes to the documentation - /docs/ directory for https://docs.gofiber.io/
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

## Commit formatting:

Use emojis on commit messages so it provides an easy way of identifying the purpose or intention of a commit. Check out the emoji cheatsheet here: https://gitmoji.carloscuesta.me/

PR #2731, opened 2023-11-18
fix(requestid): fix condition for setting default value of ContextKey…

gopkg-dev:
… in configDefault function

The condition for setting the default value of ContextKey in the configDefault function has been fixed. Previously, it only checked if cfg.ContextKey was an empty string, but now it also checks if it is nil. This ensures that the default value is set correctly even if cfg.ContextKey is not explicitly provided.

## Description

Fixes # (issue)

:warning: **For changes specific to v3, please switch to the [v3 Pull Request Template](?template=v3-changes.md).**

## Type of change

## Checklist:

## Commit formatting:

### DIFF
--- a/middleware/requestid/config.go
+++ b/middleware/requestid/config.go
@@ -57,7 +57,7 @@ func configDefault(config ...Config) Config {
 	if cfg.Generator == nil {
 		cfg.Generator = ConfigDefault.Generator
 	}
-	if cfg.ContextKey == "" {
+	if cfg.ContextKey == nil || cfg.ContextKey == "" {
 		cfg.ContextKey = ConfigDefault.ContextKey
 	}
 	return cfg


middleware/requestid/config.go
```diff
@@ -57,7 +57,7 @@ func configDefault(config ...Config) Config {
 	if cfg.Generator == nil {
 		cfg.Generator = ConfigDefault.Generator
 	}
-	if cfg.ContextKey == "" {
+	if cfg.ContextKey == nil || cfg.ContextKey == "" {
```
nickajacks1, line 60: ContextKey will only ever be "" if the user explicitly sets it to that value. In that case, overriding it to be "requestid" might be surprising.
Though I suppose changing it to accept "" now might break someone's code out there.

ReneWerner87:
thanks for the fix
image

due to the underlying locals customization of fasthttp, we have changed the type to interface to also allow numbers etc, this customization is correct
image
image

middleware/requestid/config.go
ReneWerner87: should be fine, check this https://github.com/gofiber/fiber/pull/2731#issuecomment-1823236098

ReneWerner87:
@gaby @efectn @gopkg-dev @Z3NTL3 @nickajacks1 @sixcolors 
should we rather revert this pull request instead of this customization? meanwhile my opinion has changed and normally there is no reason for other types instead of strings if this should be a key
https://github.com/gofiber/fiber/pull/2369

@benjajaja sorry for that, but i think it would be better to go back to the previous state

nickajacks1:
I feel like I've seen context keys as types/global vars before, something like
```go
type contextKey int
const (
    contextKeyRequestId contextKey = iota
    contextKeySession
    // etc...
)
```
For this PR, if it was me, I'd probably keep it as an interface, do a nil check and leave it at that.
But even so, it's probably perfectly valid for Fiber to only support strings if that's a desired direction.
Cf. #2684

sixcolors:
Will review tonight or tomorrow night.

Z3NTL3:
I have looked on GoDoc and I've seen that ``ContextKey`` is an ``inteface{}`` type and not a string. So this change would be valid.

But there is one more thing, I've looked into the middleware code. And I've seen there is no appropiate type checking. Because an ``interface{}`` type in Go can be anything, when the users passes an object thats very off, like a ``struct`` or ``map`` it will put this type as a key to ``c.Locals`` which will result in unexpected behaviour and leak something bad off. There should be a check with ``reflect`` package to check if the given ContextKey is a ``string`` or ``int``, anything else should set configuration to default.


> https://github.com/valyala/fasthttp/blob/master/userdata.go#L14
This check here for setting, it's not enoguh in my opinion, Fiber uses this of fasthttp under the hood for ``c.Locals``

Z3NTL3:
> I have looked on GoDoc and I've seen that `ContextKey` is an `inteface{}` type and not a string. So this change would be valid.
> 
> But there is one more thing, I've looked into the middleware code. And I've seen there is no appropiate type checking. Because an `interface{}` type in Go can be anything, when the users passes an object thats very off, like a `struct` or `map` it will put this object as a key to `c.Locals` which will result in unexpected behaviour and leak something bad off. There should be a check with `reflect` package to check if the given ContextKey is a `string` or `int`, anything else should set configuration to default.
> 
> > https://github.com/valyala/fasthttp/blob/master/userdata.go#L14
> > This check here for setting, it's not enoguh in my opinion, Fiber uses this of fasthttp under the hood for `c.Locals`

@ReneWerner87

ReneWerner87:
@Z3NTL3 whats your opinion on this https://github.com/gofiber/fiber/pull/2731#issuecomment-1823245097

Z3NTL3:
> @Z3NTL3 whats your opinion on this [#2731 (comment)](https://github.com/gofiber/fiber/pull/2731#issuecomment-1823245097)

revert

sixcolors:
Revert. String zero value not set table is behavior elsewhere too. 

Follow on discussion can happen and a universal approach can be adopted when agreed upon.

skydiver1980 (approved):

gopkg-dev:
If the `ContextKey` parameter is not passed, it defaults to `nil`. This can cause the default configuration to lose its effect and prevent the use of the default `ContextKey` value. Therefore, I believe this is a bug. Currently, I am using it as follows:

```go
package main

import (
	"log"

	"github.com/gofiber/fiber/v2"
	"github.com/gofiber/fiber/v2/middleware/requestid"
)

func main() {
	app := fiber.New()

	// Current implementation
	app.Use(requestid.New(requestid.Config{
		ContextKey: requestid.ConfigDefault.ContextKey,
		Generator:  util.NewXID,
	}))
	
	// Expected behavior
	app.Use(requestid.New(requestid.Config{
		Generator:  util.NewXID,
	}))

}
```

Z3NTL3:
@gopkg-dev 

When viewing the GoDoc we can see ContextKey is an ``interface{}``, which has to be a string for the following reason:

I've looked into the middleware source. And I've seen there is no appropiate type checking. Because an ``interface{} type in Go can be anything, when the users passes an object thats very off, like a struct or map it will put this object as a key to c.Locals which will result in unexpected behaviour and leak something bad off. There should be a check with reflect package to check if the given ContextKey is a string or int, anything else should set configuration to default.

**The reason for this is - Explained**
> https://github.com/valyala/fasthttp/blob/master/userdata.go#L14
This is used by c.Locals (also the middleware for setting ContextKey)
It will leak something bad in the memory when the key is anything else than int or string, so passing an interface{} is very much illegal

So its better to revert it back to be a string instead of an interface, to avoid overloads in different validating parts. In that matter, your PR becomes sadly said invalid. Idk what the maintainers will do.

Z3NTL3:
@gopkg-dev @ReneWerner87 

**``ContextKey`` is set using:**
> ``https://github.com/valyala/fasthttp/blob/master/userdata.go#L14``.

When we examine this method from ``fasthttp`` which is also being used by``c.Locals`` under the hood 

We can see that this method converts anything to a byte representation. So when we allow an interface the user may pas anything. When a struct or map is passed for example then it will be represented like:

```
ContextKey: string= "struct{Test bool} {Test true}"
``` 
So this is the ContextKey, a string including the byte representation of the struct

Thats so illegal, the performance will tear down so hard. Some very illegal behaviour.

For those reasons, reverting the ``ContextKey`` back to be a ``string`` is valid.

efectn:
> @gopkg-dev @ReneWerner87
> 
> **`ContextKey` is set using:**
> 
> > `https://github.com/valyala/fasthttp/blob/master/userdata.go#L14`.
> 
> When we examine this method from `fasthttp` which is also being used by`c.Locals` under the hood
> 
> We can see that this method converts anything to a byte representation. So when we allow an interface the user may pas anything. When a struct or map is passed for example then it will be represented like:
> 
> ```
> In request Context object representation:
> ContextKey: string= "struct{Test bool} {Test true}"
> ```
> 
> So this is the ContextKey, a string including the byte representation of the struct
> 
> Thats so illegal, the performance will tear down so hard. Some very illegal behaviour.
> 
> For those reasons, reverting the `ContextKey` back to be a `string` is valid.

I do agree. We should keep parameters/fields as type-safe as possible.

ReneWerner87:
closed because of #2742

benjajaja:
> @gopkg-dev @ReneWerner87
> 
> **`ContextKey` is set using:**
> 
> > `https://github.com/valyala/fasthttp/blob/master/userdata.go#L14`.
> 
> When we examine this method from `fasthttp` which is also being used by`c.Locals` under the hood
> 
> We can see that this method converts anything to a byte representation. 
> [...]

This is not correct, at all: `*userData.Set(k,v)` ONLY converts keys of type `[]byte` to `string` (because comparing `[]byte`s behind `any` would panic on [line 22](https://github.com/valyala/fasthttp/blob/master/userdata.go#L22)).

The std package `context` [says](https://pkg.go.dev/context#Context) the following about keys:

> A key can be any type that supports equality; packages should define keys as an unexported type to avoid collisions.

So context keys should never be restricted to strings by libraries, and both fiber and fasthttp correctly follow this and allow `any` for their abstractions over `context.Context`.

---

The requestid middleware cannot not be an exception to the `any` key rule. If you make this exception, you disallow the key type that is recommended by the standard library: `type key int`.

---

The revert should be reverted, but the condition should be changed to a simple `nil` check, and the default value should not be of an unexported type.

ReneWerner87:
Just because the methodic which is used inside supports all types, this does not have to be used outside for the key
How useful is it, for example, to use a struct as a key, which is also possible there? 
Keys are usually strings or ints, but in this case strings make more sense and if numbers are to be used, this can also be done as a string number 
By allowing only one type from the outside, we prevent possible errors or incorrect use

benjajaja:
@ReneWerner87 this is not the place to discuss what type a context key should be, because this is specific to the requestid middleware and not e.g. the `c.Locals` API. The requestid middleware must be consistent with `c.Locals` and the rest of fiber.

Re: context key type, please read the standard library documentation: https://pkg.go.dev/context#Context. Also check the official blog post on the matter: https://go.dev/blog/context. But I insist, this is not the place to discuss this - if you _really_ want to force context keys back to `string`s, then please make a PR that changes `c.Locals` API and all other middlewares.

PR #2734, opened 2023-11-18
🚨 Test: fix race condition in parallel tests

nickajacks1:
## Description

Tests that call SetParserDecoder were causing a race condition with other tests that read from decoderPoolMap. Fix by making the offending tests not run in parallel.
This often manifested as CI failures.

To reproduce, run ` go test -race -v . -count=1 -run=Test_Ctx_QueryParser`. It will fail virtually every time.

## Type of change

## Checklist:


renanbastos93 (approved):
LGTM

PR #2742, opened 2023-11-27
Revert ":bug: requestid.Config.ContextKey is interface{} (#2369)"

ReneWerner87:
we have decided to reverse this change from #2369

https://github.com/gofiber/fiber/pull/2731

Fixes #2732

PR #2750, opened 2023-12-02
Revert revert requestid contextkey

### LINKED ISSUE #2356: 🐛 [Bug]: requestId middleware still only accepts `string` for `ContextKey` config value, should be `any` like `c.Locals`.
### Bug Description

#2143 allows to use `any` as key for `c.Locals(key, v)`. The RequestID middleware should be consistent with that.
This allows to use a custom type for context keys, like for other keys. The RequestID is likely read back somewhere else along other context keys.

### How to Reproduce

See https://docs.gofiber.io/api/middleware/requestid#config

### Expected Behavior

https://docs.gofiber.io/api/middleware/requestid#config should have `ContextKey interface{}`

### Fiber Version

v2.42.0

### Code Snippet (optional)

_No response_

### Checklist:

benjajaja:
## Description

Revert the revert of #2742, fix condition check, add a unit test that asserts that the condition is correct, and add a comment about why the default value is `string` instead of the recommended private-type-style.

See also the initial discussion and my latest [comment](https://github.com/gofiber/fiber/pull/2731#issuecomment-1837084401) in #2731 - there are several incorrect assumptions about how the context works or should work, which lead to the commit reversal.

Fixes #2356 

## Specific case example

* I am a developer of some app that uses fiber, and other libraries, which use `context.Context`.
* I follow the [standard go library documentation's recommendation](https://pkg.go.dev/context#Context) on context keys and use unexported types.
* I cannot use the `requestid` middleware because #2742 has constrained the key to be a `string`.

## Type of change

## Checklist:

### DIFF
--- a/docs/api/middleware/requestid.md
+++ b/docs/api/middleware/requestid.md
@@ -45,7 +45,7 @@ app.Use(requestid.New(requestid.Config{
 | Next       | `func(*fiber.Ctx) bool` | Next defines a function to skip this middleware when returned true.                               | `nil`          |
 | Header     | `string`                | Header is the header key where to get/set the unique request ID.                                  | "X-Request-ID" |
 | Generator  | `func() string`         | Generator defines a function to generate the unique identifier.                                   | utils.UUID     |
-| ContextKey | `string`                | ContextKey defines the key used when storing the request ID in the locals for a specific request. | "requestid"    |
+| ContextKey | `interface{}`           | ContextKey defines the key used when storing the request ID in the locals for a specific request. | "requestid"    |
 
 ## Default Config
 The default config uses a fast UUID generator which will expose the number of
--- a/middleware/requestid/config.go
+++ b/middleware/requestid/config.go
@@ -26,7 +26,7 @@ type Config struct {
 	// the locals for a specific request.
 	//
 	// Optional. Default: requestid
-	ContextKey string
+	ContextKey interface{}
 }
 
 // ConfigDefault is the default config
--- a/middleware/requestid/config.go
+++ b/middleware/requestid/config.go
@@ -24,8 +24,10 @@ type Config struct {
 
 	// ContextKey defines the key used when storing the request ID in
 	// the locals for a specific request.
+	// Should be a private type instead of string, but too many apps probably
+	// rely on this exact value.
 	//
-	// Optional. Default: requestid
+	// Optional. Default: "requestid"
 	ContextKey interface{}
 }
 
@@ -57,7 +59,7 @@ func configDefault(config ...Config) Config {
 	if cfg.Generator == nil {
 		cfg.Generator = ConfigDefault.Generator
 	}
-	if cfg.ContextKey == "" {
+	if cfg.ContextKey == nil {
 		cfg.ContextKey = ConfigDefault.ContextKey
 	}
 	return cfg
--- a/middleware/requestid/requestid_test.go
+++ b/middleware/requestid/requestid_test.go
@@ -55,20 +55,21 @@ func Test_RequestID_Next(t *testing.T) {
 func Test_RequestID_Locals(t *testing.T) {
 	t.Parallel()
 	reqID := "ThisIsARequestId"
-	ctxKey := "ThisIsAContextKey"
+	type ContextKey int
+	const requestContextKey ContextKey = iota
 
 	app := fiber.New()
 	app.Use(New(Config{
 		Generator: func() string {
 			return reqID
 		},
-		ContextKey: ctxKey,
+		ContextKey: requestContextKey,
 	}))
 
 	var ctxVal string
 
 	app.Use(func(c *fiber.Ctx) error {
-		ctxVal = c.Locals(ctxKey).(string) //nolint:forcetypeassert,errcheck // We always store a string in here
+		ctxVal = c.Locals(requestContextKey).(string) //nolint:forcetypeassert,errcheck // We always store a string in here
 		return c.Next()
 	})
 
--- a/middleware/requestid/requestid_test.go
+++ b/middleware/requestid/requestid_test.go
@@ -77,3 +77,27 @@ func Test_RequestID_Locals(t *testing.T) {
 	utils.AssertEqual(t, nil, err)
 	utils.AssertEqual(t, reqID, ctxVal)
 }
+
+// go test -run Test_RequestID_DefaultKey
+func Test_RequestID_DefaultKey(t *testing.T) {
+	t.Parallel()
+	reqID := "ThisIsARequestId"
+
+	app := fiber.New()
+	app.Use(New(Config{
+		Generator: func() string {
+			return reqID
+		},
+	}))
+
+	var ctxVal string
+
+	app.Use(func(c *fiber.Ctx) error {
+		ctxVal = c.Locals("requestid").(string) //nolint:forcetypeassert,errcheck // We always store a string in here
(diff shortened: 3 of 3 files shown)


ReneWerner87:
Hi, we have decided against other types for the context key. 
From our point of view, there is no big real world benefit to allow other types for the context key and only adds complexity and errors without any real benefit. 

String is sufficient as a type for all conceivable real cases that you might want to use in an application.

benjajaja:
We're obviously not on the same page. Let's try to work this out.

# This PR only aligns the requestid middleware back to fiber

The rest of fiber accepts `any` for context keys:

https://pkg.go.dev/github.com/gofiber/fiber/v2#Ctx.Locals
https://github.com/gofiber/fiber/blob/master/ctx.go#L970

# Context keys should be strings 

Please read the standard library documentation: https://pkg.go.dev/context#Context. Also check the official blog post on the matter: https://go.dev/blog/context. However, this PR is not the place to discuss this - if you _really_ want to force context keys back to `string`s, then please make a PR that changes `c.Locals` API to take `string` instead of `any`, and argue your case there.

nickajacks1:
IMO it's probably best to follow the standard Go guidance of using `any`.
It neither precludes fiber nor users from using strings, and it's less surprising for people used to working with Contexts.

ReneWerner87:
Hello again,

I'd like to further explain our decision to explicitly set the data type as `string`.

While it's true that the underlying context in GoFiber allows any data type (`interface{}`), and the `locals` also adhere to this flexible design, it's not always necessary or advantageous to pass this level of flexibility on to the end user of the middleware. Our approach follows the "[Principle of Least Astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment)," which aims to create an API that behaves in a way most users would naturally expect.

By specifying the data type as `string` in the configuration, we aim to achieve several important objectives:

1. **Clarity and Predictability**: It provides the user with a clear understanding of the expected data type (a simple `string`). This reduces confusion and potential errors that might arise from misuse of the type flexibility offered by `interface{}`.

2. **Simplicity and Reduced Complexity**: Although GoFiber's core is powerful and flexible, exposing too much of this flexibility can be overwhelming or confusing for users. Limiting the data type to `string` simplifies the user experience, making the middleware more approachable.

3. **Avoiding Unnecessary Overhead**: Using a specific type like `string` can also help optimize performance, as the system does not need to constantly check and adapt to different data types.

For instance, consider a scenario where a user mistakenly passes a complex object as the `contextKey` when a simple `string` identifier would suffice. This could lead to unnecessary processing and potential errors, which are avoided by enforcing a `string` type.

Currently, there have been no reported downsides to this approach in terms of functionality. It doesn't render the middleware non-functional, nor does the alternative of an unspecified type (`interface{}`) offer any significant real-world benefits in this context. Therefore, we prefer to keep the type specification as `string` for its clarity, simplicity, and alignment with the principles of sound software design.

I hope this clarifies our design choice. If you have any further questions or suggestions, feel free to share them.

benjajaja:
> Hello again,
> 
> I'd like to further explain our decision to explicitly set the data type as `string`.
> 
> While it's true that the underlying context in GoFiber allows any data type (`interface{}`), and the `Locals` also adhere to this flexible design, it's not always necessary or advantageous to pass this level of flexibility on to the end user of the middleware. Our approach follows the "[Principle of Least Astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment)," which aims to create an API that behaves in a way most users would naturally expect.

You are breaking the principle you are quoting by _constraining_ the context keys of middlewares to be `string` only. That is:

* I CAN use [`c.Locals()`](https://pkg.go.dev/github.com/gofiber/fiber/v2#Ctx.Locals) with an unexported type.
* I surprisingly CANNOT use an unexported type for the requestid middleware.
   _To be clear: this contradicts both the general go guidelines on context keys, AND fiber's own context wrapper!_

This PR changes this to make it **less** astonishing:

* I CAN use an unexported type for the requestid middleware.
* I CAN also use a `string` if I want.
* The default value still is `requestid` of type `string`.


> By specifying the data type as `string` in the configuration, we aim to achieve several important objectives:
> 
>     1. **Clarity and Predictability**: It provides the user with a clear understanding of the expected data type (a simple `string`). This reduces confusion and potential errors that might arise from misuse of the type flexibility offered by `interface{}`.
> 
>     2. **Simplicity and Reduced Complexity**: Although GoFiber's core is powerful and flexible, exposing too much of this flexibility can be overwhelming or confusing for users. Limiting the data type to `string` simplifies the user experience, making the middleware more approachable.
> 
>     3. **Avoiding Unnecessary Overhead**: Using a specific type like `string` can also help optimize performance, as the system does not need to constantly check and adapt to different data types.
> 

All three points are moot:

1. It is unclear why **only the middlewares** are constrained to `string`, but not `c.Locals` otherwise.
2. Strings may be more simple, but are open to collisions, have you read https://pkg.go.dev/context#Context?
3. There is absolutely no overhead, on the contrary, a type alias to `int` would have less overhead than `string` in any case.

> For instance, consider a scenario where a user mistakenly passes a complex object as the `ContextKey` when a simple `string` identifier would suffice. This could lead to unnecessary processing and potential errors, which are avoided by enforcing a `string` type.

Then go ahead an make a PR to change `c.Locals` to only take `string` as key, then. Taking this out here is absurd.
 
> Currently, there have been no reported downsides to this approach in terms of functionality. It doesn't render the middleware non-functional, nor does the alternative of an unspecified type (`interface{}`) offer any significant real-world benefits in this context. Therefore, we prefer to keep the type specification as `string` for its clarity, simplicity, and alignment with the principles of sound software design.

Yes there is, per my description of this PR:

* I am a developer of some app that uses fiber, and other libraries, which use `context.Context`.
* I follow the [standard go library documentation's recommendation](https://pkg.go.dev/context#Context) on context keys and use unexported types.
* I cannot use the `requestid` middleware because #2742 has constrained the key to be a `string`.

Additionally, a user that doesn't change the key, or changed it to a `string` is completely unaffected by this PR.

sixcolors (changes requested):
I suggest we open an issue or discussion where we can collaborate and work towards a solution that ensures consistency across the project.

sixcolors:
> IMO it's probably best to follow the standard Go guidance of using `any`. It neither precludes fiber nor users from using strings, and it's less surprising for people used to working with Contexts.

GoFiber still supports go 1.17.... Approach?

benjajaja:
@sixcolors the universal patch is open here #2751.

ReneWerner87:
> @ReneWerner87 who exactly is "we"?

-> the maintainer team + some of the contributors from the community

Can you please provide a (few) short examples of real world cases so that i can understand the more precise need for the change

ReneWerner87:
okay, thanks for that
you have convinced me a little more

however, the probability of a collision is low

however, the security aspect that pjebs brings up here, this could be considered when building wrapper applications or plugins, for example, where the outer applications do not have the possibility of modification

image

https://github.com/valyala/fasthttp/issues/1385

we will consider these aspects again and discuss the advantages and disadvantages of the current version compared to the new version

thanks for the insights and your effort

nickajacks1:
> GoFiber still supports go 1.17.... What approach would you recommend?

`any` is just an alias for `interface{}`

sixcolors:
Recomend we close this PR and work on #2751 as it includes this change.

ReneWerner87:
We merged #2751
No need for this PR

PR #2760, opened 2023-12-12
🔥 feat(utils): add IsZeroValue

sixcolors:
## Description

Add utils function to support [change](https://github.com/gofiber/fiber/pull/2751#issuecomment-1852086176) to #2751

Adds IsZeroValue function to utils pkg. Which reports whether x is the zero value for its type. For basic types and error, it uses a fast path without reflection. For other types, it uses reflection which is slower.

## Type of change

## Checklist:

### DIFF
--- a/utils/zero.go
+++ b/utils/zero.go
@@ -6,7 +6,7 @@ import (
 
 // IsZeroValue reports whether x is the zero value for its type.
 //
-// For basic types, error and slices of basic types, it uses a fast path
+// For basic types and error it uses a fast path
 // that does not use reflection.
 //
 // For all other types, it uses reflection.
--- a/utils/zero.go
+++ b/utils/zero.go
@@ -1,0 +1,88 @@
+package utils
+
+import (
+	"reflect"
+)
+
+// IsZeroValue reports whether x is the zero value for its type.
+//
+// For basic types and slices of basic types, it uses a fast path without reflection.
+// For other types, it uses reflection.
+func IsZeroValue(x interface{}) bool {
+	// Fast path for basic types
+	switch v := x.(type) {
+	case nil:
+		return true
+	case bool:
+		return !v
+	case int:
+		return v == 0
+	case int8:
+		return v == 0
+	case int16:
+		return v == 0
+	case int32:
+		return v == 0
+	case int64:
+		return v == 0
+	case uint:
+		return v == 0
+	case uint8:
+		return v == 0
+	case uint16:
+		return v == 0
+	case uint32:
+		return v == 0
+	case uint64:
+		return v == 0
+	case uintptr:
+		return v == 0
+	case float32:
+		return v == 0.0
+	case float64:
+		return v == 0.0
+	case complex64:
+		return v == 0+0i
+	case complex128:
+		return v == 0+0i
+	case string:
+		return v == ""
+	case []bool:
+		return len(v) == 0
+	case []int:
+		return len(v) == 0
+	case []int8:
+		return len(v) == 0
+	case []int16:
+		return len(v) == 0
+	case []int32:
+		return len(v) == 0
+	case []int64:
+		return len(v) == 0
+	case []uint:
+		return len(v) == 0
+	case []uint8:
+		return len(v) == 0
+	case []uint16:
+		return len(v) == 0
+	case []uint32:
+		return len(v) == 0
+	case []uint64:
+		return len(v) == 0
+	case []uintptr:
+		return len(v) == 0
+	case []float32:
+		return len(v) == 0
+	case []float64:
+		return len(v) == 0
+	case []complex64:
+		return len(v) == 0
+	case []complex128:
+		return len(v) == 0
+	case []string:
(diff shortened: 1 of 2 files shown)


sixcolors:
basic types based on https://go.dev/tour/basics/11

utils/zero.go
```diff
@@ -0,0 +1,88 @@
...
+		return len(v) == 0
+	case []complex64:
+		return len(v) == 0
+	case []complex128:
+		return len(v) == 0
+	case []string:
+		return len(v) == 0
+	default:
+		// Slow path using reflection
+		return reflect.DeepEqual(x, reflect.Zero(reflect.TypeOf(x)).Interface())
```
efectn, line 86:
```go
type test string

func main() {
	x := []test{}
	fmt.Print(IsZeroValue(x))
}
```
will print false due to DeepEqual is pretty strict. Maybe this part should be adjusted?

sixcolors:
If we were keen to add maps to the fast path we could make a generator, but that would add a lot of code, so I did not do it.

Generator code for reference:

```go
package main

import (
    "fmt"
    "os"
    "text/template"
)

type Type struct {
    Key   string
    Value string
}

func main() {
    types := []string{
        "bool", "int", "int8", "int16", "int32", "int64", "uint", "uint8", "uint16", "uint32", "uint64", "uintptr", "float32", "float64", "complex64", "complex128", "string", "error",
    }

    tmpl := template.Must(template.New("").Parse(`
    case map[{{.Key}}]{{.Value}}:
        return len(v.(map[{{.Key}}]{{.Value}})) == 0
    `))

    for _, keyType := range types {
        for _, valueType := range types {
            t := Type{Key: keyType, Value: valueType}
            err := tmpl.Execute(os.Stdout, t)
            if err != nil {
                fmt.Println("executing template:", err)
            }
        }
    }
}
```

utils/zero.go
efectn: maybe this check is OK?
```go
val := reflect.ValueOf(x)
if val.Kind() == reflect.Slice && val.Len() == 0 {
	return true
}
```

efectn: > maybe this check is OK above this part?
> 
> ```go
> val := reflect.ValueOf(x)
> if val.Kind() == reflect.Slice && val.Len() == 0 {
> 	return true
> }
> ```

sixcolors: AFAIK in Go, a zero value for a slice is nil. An empty slice (like `x := []test{}` in your example) is not considered a zero value; it's just an empty slice.

sixcolors: I already considered empty slices zero in the fast path. So that would have to be the same... I'll fix.

efectn: > AFAIK in Go, a zero value for a slice is nil. An empty slice (like `x := []test{}` in your example) is not considered a zero value; it's just an empty slice.

Then we need to adjust switch-case part, right?

sixcolors: Which behaviour would you consider correct?

sixcolors: [cdb2885](https://github.com/gofiber/fiber/pull/2760/commits/cdb288542b077ecf5bb764d1d0033ba659c7a1dc) handles edge cases such as `x:= []test{}` and improves test/bench.

If we are going to consider those cases ZeroValues, otherwise we can revert to using DeepEqual.

efectn: > > > AFAIK in Go, a zero value for a slice is nil. An empty slice (like `x := []test{}` in your example) is not considered a zero value; it's just an empty slice.
> > 
> > 
> > Then we need to adjust switch-case part, right?
> 
> Correct.
> 
> [cdb2885](https://github.com/gofiber/fiber/pull/2760/commits/cdb288542b077ecf5bb764d1d0033ba659c7a1dc) handles edge cases such as `x:= []test{}` and improves test/bench.
> 
> If we are going to consider those cases ZeroValues we can keep it, otherwise we can revert to using DeepEqual and remove the zero length slice cases.

That's OK to me but not sure `[]string{}` etc. is zero value according to Go. Should we either make switch case behavior same as reflect's zerovalue or make zero length slices zero value?

sixcolors: https://github.com/gofiber/fiber/pull/2760/commits/c228dbd3c4c1fcc05c4508e568e31594bafa3446 aligns with conventions by removing slice cases and updating tests.

gaby:
@sixcolors Probably worth comparing your implementation with this one: https://github.com/duke-git/lancet/blob/main/validator/validator.go#L390

Unit-Tests here: https://github.com/duke-git/lancet/blob/main/validator/validator_test.go#L413

Lancet is used by over +1k repos, so it's worth comparing.

sixcolors:
Set to draft while considering if related change is appropriate or not

sixcolors:
As @nickajacks1 pointed out https://github.com/gofiber/fiber/pull/2751#issuecomment-1852451213 a IsZeroValue check for ContdxtKeys would cause issues for some types that could be valid keys. So I’m closing this PR.

Issue #2745, opened 2023-11-30
📝 [v3 Proposal]: Parity for res.format and Ctx.Format


nickajacks1:
### Feature Proposal Description

The current API for `Ctx.Format` does not align with `res.format` from Express.  While they both perform content negotiation, the usage is different (see below for a comparison). This may be confusing for Gophers coming from Express who expect to be able to add their own handlers.

I propose reworking `Ctx.Format` to take a mapping of accept types to handlers.  We can move the existing `Ctx.Format` functionality to a separate method (e.g., `Ctx.AutoFormat`)

### Alignment with Express API

In Express, `res.format` is used to perform content negotiation and call a different handler based on the result.  The handlers are fully user-defined aside from the default handler which will send a 406 HTML response.
From https://expressjs.com/en/api.html#res.format
```javascript
res.format({
  'text/plain': function () {
    res.send('hey')
  },

  'text/html': function () {
    res.send('<p>hey</p>')
  },

  'application/json': function () {
    res.send({ message: 'hey' })
  },

  default: function () {
    // log the request and respond with 406
    res.status(406).send('Not Acceptable')
  }
})
```

Fiber's [Ctx.Format](https://github.com/gofiber/fiber/blob/v3-beta/ctx.go#L376) has automatic handlers for a fixed set of Accept headers. The usage is different; it hides details of the request's Accept header and the outgoing Content Type from the implementer.
```go
data := SomeData{ FirstName: "Jane", LastName: "Doe" }
c.Format(data)
```

See below for proposed syntax.

### HTTP RFC Standards Compliance

[RFC-9110](https://www.rfc-editor.org/rfc/rfc9110#name-content-negotiation) does not specify how servers should reply to requests with respect to content - it simply gives various suggestions and algorithms.  `res.format` follows the [proactive negotiation](https://www.rfc-editor.org/rfc/rfc9110#name-proactive-negotiation) algorithm.

### API Stability

The new `Ctx.Format` should be in line with `res.format`, meaning it should have a stable API.  The main risk for API stability is the data structure in which the handlers are passed, as that may have performance considerations (allocations, etc).

### Feature Examples

```go
package main

import (
	"fmt"
	"time"

	"github.com/gofiber/fiber/v2"
	"github.com/gofiber/fiber/v2/middleware/logger"
)

type IceCream struct {
	Flavor string `json:"flavor" xml:"flavor"`
	Color  string `json:"color" xml:"color"`
}

func (ic IceCream) String() string {
[...]

PR #2766, opened 2023-12-17
🔥 v3: update Ctx.Format to match Express's res.format

nickajacks1:
## Description

While the existing Ctx.Format provides a concise convenience method for basic content negotiation on simple structures, res.format allows developers to set their own custom handlers for each content type.

Docs: moved the existing Ctx.Format docs to a section called AutoFormat and added docs for the new Ctx.Format.

I'm taking suggestions for better names than `Fmt` for the mediaType-handler struct...was just worried about making it overly long since people will need to type it over and over.

<details>
<summary>Benchmarks</summary>

Performs ever so slightly better than AutoFormat (formerly known as Format)

```
Benchmark_Ctx_Format/with_arg_allocation-8               2578868               464.5 ns/op            96 B/op          1 allocs/op
Benchmark_Ctx_Format/with_arg_allocation-8               2577880               466.9 ns/op            96 B/op          1 allocs/op
Benchmark_Ctx_Format/with_arg_allocation-8               2596575               462.0 ns/op            96 B/op          1 allocs/op
Benchmark_Ctx_Format/with_arg_allocation-8               2564602               460.6 ns/op            96 B/op          1 allocs/op
Benchmark_Ctx_Format/pre-allocated_args-8                2983128               407.3 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Format/pre-allocated_args-8                3031767               400.3 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Format/pre-allocated_args-8                3045684               397.7 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Format/pre-allocated_args-8                2986521               407.1 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Format/text/plain-8                       12827778                92.41 ns/op            0 B/op          0 allocs/op
Benchmark_Ctx_Format/text/plain-8                       13453144                93.66 ns/op            0 B/op          0 allocs/op
Benchmark_Ctx_Format/text/plain-8                       13534003                86.67 ns/op            0 B/op          0 allocs/op
Benchmark_Ctx_Format/text/plain-8                       13168622                88.81 ns/op            0 B/op          0 allocs/op
Benchmark_Ctx_Format/json-8                              9903925               128.8 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Format/json-8                              9780688               124.1 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Format/json-8                              9499648               128.1 ns/op             0 B/op          0 allocs/op
Benchmark_Ctx_Format/json-8                              9510656               122.2 ns/op             0 B/op          0 allocs/op
```

</details>

Related to #2745

## Migration

To make their code compatible with Fiber v3, users need only replace their uses of `c.Format` to `c.AutoFormat`.
To make full use of the new `c.Format`, users should note where in their application they are performing manual content negotiation and consider if `c.Format` will simplify their code.

## Changes Introduced

- [x] Changelog/What's New: Include a summary of the additions for the upcoming release notes.
- [x] Migration Guide: If necessary, provide a guide or steps for users to migrate their existing code to accommodate these changes.

## Type of Change

## Checklist

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -8,6 +8,7 @@ import (
 	"bytes"
 	"context"
 	"crypto/tls"
+	"errors"
 	"fmt"
 	"io"
 	"mime/multipart"
@@ -370,10 +371,60 @@ func (c *DefaultCtx) Response() *fasthttp.Response {
 	return &c.fasthttp.Response
 }
 
+type Fmt struct {
+	MediaType string
+	Handler   func(Ctx) error
+}
+
 // Format performs content-negotiation on the Accept HTTP header.
+// It uses Accepts to select a proper format and calls the matching
+// user-provided handler function.
+// If no accepted format is found, and a format with MediaType "default" is given,
+// that default handler is called. If no format is found and no default is given,
+// StatusNotAcceptable is sent.
+func (c *DefaultCtx) Format(handlers ...Fmt) error {
+	if len(handlers) == 0 {
+		panic("Format requires at least one handler")
+	}
+
+	if c.Get(HeaderAccept) == "" {
+		c.Response().Header.SetContentType(handlers[0].MediaType)
+		return handlers[0].Handler(c)
+	}
+
+	types := make([]string, 0, 8)
+	var defaultHandler Handler
+	for _, h := range handlers {
+		if h.MediaType == "default" {
+			defaultHandler = h.Handler
+			continue
+		}
+		types = append(types, h.MediaType)
+	}
+	accept := c.Accepts(types...)
+
+	if accept == "" {
+		if defaultHandler == nil {
+			return c.SendStatus(StatusNotAcceptable)
+		}
+		return defaultHandler(c)
+	}
+
+	for _, h := range handlers {
+		if h.MediaType == accept {
+			c.Response().Header.SetContentType(h.MediaType)
+			return h.Handler(c)
+		}
+	}
+
+	// unreachable code
+	panic(errors.New("fiber: an Accept was found but no handler was called - please file a bug report"))
+}
+
+// AutoFormat performs content-negotiation on the Accept HTTP header.
 // It uses Accepts to select a proper format.
 // If the header is not specified or there is no proper format, text/plain is used.
-func (c *DefaultCtx) Format(body any) error {
+func (c *DefaultCtx) AutoFormat(body any) error {
 	// Get accepted content type
 	accept := c.Accepts("html", "json", "txt", "xml")
 	// Set accepted content type
--- a/ctx.go
+++ b/ctx.go
@@ -8,7 +8,6 @@ import (
 	"bytes"
 	"context"
 	"crypto/tls"
-	"errors"
 	"fmt"
 	"io"
 	"mime/multipart"
@@ -100,6 +99,12 @@ type Cookie struct {
 type Views interface {
 	Load() error
 	Render(io.Writer, string, any, ...string) error
+}
+
+// ResFmt associates a Content Type to a fiber.Handler for c.Format
+type ResFmt struct {
+	MediaType string
+	Handler   func(Ctx) error
 }
 
 // Accepts checks if the specified extensions or content types are acceptable.
@@ -371,27 +376,28 @@ func (c *DefaultCtx) Response() *fasthttp.Response {
 	return &c.fasthttp.Response
 }
 
-type Fmt struct {
-	MediaType string
-	Handler   func(Ctx) error
-}
-
 // Format performs content-negotiation on the Accept HTTP header.
 // It uses Accepts to select a proper format and calls the matching
 // user-provided handler function.
 // If no accepted format is found, and a format with MediaType "default" is given,
 // that default handler is called. If no format is found and no default is given,
 // StatusNotAcceptable is sent.
-func (c *DefaultCtx) Format(handlers ...Fmt) error {
+func (c *DefaultCtx) Format(handlers ...ResFmt) error {
 	if len(handlers) == 0 {
-		panic("Format requires at least one handler")
+		return ErrNoHandlers
 	}
+
+	c.Vary(HeaderAccept)
 
 	if c.Get(HeaderAccept) == "" {
 		c.Response().Header.SetContentType(handlers[0].MediaType)
 		return handlers[0].Handler(c)
 	}
 
+	// Using an int literal as the slice capacity allows for the slice to be
+	// allocated on the stack. The number was chosen arbitrarily as an
+	// approximation of the maximum number of content types a user might handle.
+	// If the user goes over, it just causes allocations, so it's not a problem.
 	types := make([]string, 0, 8)
 	var defaultHandler Handler
 	for _, h := range handlers {
@@ -417,8 +423,7 @@ func (c *DefaultCtx) Format(handlers ...Fmt) error {
 		}
 	}
 
-	// unreachable code
-	panic(errors.New("fiber: an Accept was found but no handler was called - please file a bug report"))
+	return fmt.Errorf("%w: format: an Accept was found but no handler was called", errUnreachable)
 }
 
 // AutoFormat performs content-negotiation on the Accept HTTP header.
--- a/ctx.go
+++ b/ctx.go
@@ -428,6 +428,8 @@ func (c *DefaultCtx) Format(handlers ...ResFmt) error {
 
 // AutoFormat performs content-negotiation on the Accept HTTP header.
 // It uses Accepts to select a proper format.
+// The supported content types are text/html, text/plain, application/json, and application/xml.
+// For more flexible content negotiation, use Format.
 // If the header is not specified or there is no proper format, text/plain is used.
 func (c *DefaultCtx) AutoFormat(body any) error {
 	// Get accepted content type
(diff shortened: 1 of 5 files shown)


sixcolors:
Thanks @nickajacks1 quick glance on mobile looks good. I will review tomorrow.

ctx.go
```diff
@@ -370,10 +371,60 @@ func (c *DefaultCtx) Response() *fasthttp.Response {
...
+
 // Format performs content-negotiation on the Accept HTTP header.
+// It uses Accepts to select a proper format and calls the matching
+// user-provided handler function.
+// If no accepted format is found, and a format with MediaType "default" is given,
+// that default handler is called. If no format is found and no default is given,
+// StatusNotAcceptable is sent.
+func (c *DefaultCtx) Format(handlers ...Fmt) error {
+	if len(handlers) == 0 {
+		panic("Format requires at least one handler")
```
efectn, line 387: We should avoid using panic in handlers

ctx.go
```diff
@@ -370,10 +371,60 @@ func (c *DefaultCtx) Response() *fasthttp.Response {
...
+	}
+
+	if c.Get(HeaderAccept) == "" {
+		c.Response().Header.SetContentType(handlers[0].MediaType)
+		return handlers[0].Handler(c)
+	}
+
+	types := make([]string, 0, 8)
```
efectn, line 395: Why was capacity set as 8?

dammy001: errors.New should be used right? @efectn

ctx.go
dammy001: `errors.New` should be used right? @efectn

efectn: > `errors.New` should be used right? @efectn

Yes create new error variable

ctx.go
```diff
@@ -370,10 +371,60 @@ func (c *DefaultCtx) Response() *fasthttp.Response {
 	return &c.fasthttp.Response
 }
 
+type Fmt struct {
```
efectn, line 374: Please add some comments

efectn: And not sure `Fmt` is good choice as name. Perhaps choose more explanatory name

ctx.go
sixcolors: vote for sentinel errors

ctx.go
sixcolors: My guess to avoid runtime allocs if it's under 8, which is probably a reasonable limit

ctx.go
sixcolors: `ResFmt`?

ctx.go
```diff
@@ -370,10 +371,60 @@ func (c *DefaultCtx) Response() *fasthttp.Response {
...
+
+	for _, h := range handlers {
+		if h.MediaType == accept {
+			c.Response().Header.SetContentType(h.MediaType)
+			return h.Handler(c)
+		}
+	}
+
+	// unreachable code
+	panic(errors.New("fiber: an Accept was found but no handler was called - please file a bug report"))
```
sixcolors, line 421:
```suggestion
return fmt.Errorf("fiber: an Accept was found but no handler was called")
```

docs/api/ctx.md
```diff
@@ -184,6 +184,35 @@ app.Get("/", func(c *fiber.Ctx) error {
...
+## AutoFormat
+
+Performs content-negotiation on the [Accept](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept) HTTP header. It uses [Accepts](ctx.md#accepts) to select a proper format.
+
+:::info
+If the header is **not** specified or there is **no** proper format, **text/plain** is used.
+:::
+
+```go title="Signature"
+func (c *Ctx) AutoFormat(body interface{}) error
```
sixcolors, line 196:
```suggestion
func (c *Ctx) AutoFormat(content interface{}) error {
```

ctx.go
```diff
@@ -370,10 +371,60 @@ func (c *DefaultCtx) Response() *fasthttp.Response {
...
+}
+
+// AutoFormat performs content-negotiation on the Accept HTTP header.
 // It uses Accepts to select a proper format.
 // If the header is not specified or there is no proper format, text/plain is used.
-func (c *DefaultCtx) Format(body any) error {
+func (c *DefaultCtx) AutoFormat(body any) error {
```
sixcolors, line 427:
```suggestion
func (c *DefaultCtx) AutoFormat(content interface{}) error {
```

ctx_interface.go
```diff
@@ -90,9 +90,17 @@ type Ctx interface {
...
+	// If no accepted format is found, and a format with MediaType "default" is given,
+	// that default handler is called. If no format is found and no default is given,
+	// StatusNotAcceptable is sent.
+	Format(handlers ...Fmt) error
+
+	// AutoFormat performs content-negotiation on the Accept HTTP header.
 	// It uses Accepts to select a proper format.
 	// If the header is not specified or there is no proper format, text/plain is used.
-	Format(body any) error
+	AutoFormat(body any) error
```
sixcolors, line 103:
```suggestion
AutoFormat(content interface{}) error
```

ctx.go
sixcolors: ^ although a sentinel error would be better

sixcolors (approved):

ctx.go
efectn: > My guess to avoid runtime allocs if it's under 8, which is probably a reasonable limit

Yeah probably. It would be really nice if he adds a comment.

nickajacks1: using a static size allows stack allocation, and 8 is an approximation of how many content types a user might use. The number was selected completely arbitrarily.  Even though it's stack allocated, making the slice bigger makes it slower.  I'll add this info as a comment as suggested

ctx.go
nickajacks1: this is for v3 which uses any over interface{} in several places. The CI only tests 1.20 and 1.21 right now, so I assumed 1.20 was the min version

ctx.go
nickajacks1: You're right, an error (visible as a 5xx response) is better. I will do the same at the unreachable code on line 421.

ctx.go
nickajacks1: Yes this needs comments and a better name.  ResFmt seems better to me.  What do you think @efectn

ctx.go
sixcolors: @ReneWerner87 what is the min go for v3?

@nickajacks1 `interface{}` is used elsewhere in this file, so simple consistancy was part of my comment.

ctx_interface.go
ReneWerner87: not needed we use the last golang version

ctx.go
ReneWerner87: the latest at the time of the v3 release -> so >= 1.21

ReneWerner87: so "any" is ok

ReneWerner87:
thanks really cool improvement

ctx.go
nickajacks1: Hmm, should this condition return an error or nil?  As a user I'd probably assume no handlers means "do nothing successfully", but Express returns a 500 (a bit ungracefully).

efectn: > Hmm, should this condition return an error or nil? As a user I'd probably assume no handlers means "do nothing successfully", but Express returns a 500 (a bit ungracefully).

Yeah i'd also return nil instead of error. What do you think @sixcolors @gaby @ReneWerner87

ReneWerner87: would do it like express, if nothing is configured, it should be an error and it is pointless to use this methodic

gaby: Yup, return error. Nil is too ambigious.

docs/api/ctx.md
```diff
@@ -184,6 +184,35 @@ app.Get("/", func(c *fiber.Ctx) error {
...
+  c.AutoFormat("Hello, World!")
+  // => Hello, World!
+
+  // Accept: text/html
+  c.AutoFormat("Hello, World!")
+  // => <p>Hello, World!</p>
+
+  // Accept: application/json
+  c.AutoFormat("Hello, World!")
+  // => "Hello, World!"
```
gaby, line 211: Is this the right output?

gaby (commented):
Not sure if I missed it, but there's no test using a json object?

docs/api/ctx.md
nickajacks1: yes, it's a bare string value, which is valid json; note the double quotes.  Perhaps the example could be more illustrative.

gaby: Yeah, the officisl express docs use a simple json with key/value pair.

nickajacks1:
> Not sure if I missed it, but there's no test using a json object?

Do you mean for Format or AutoFormat?  For Format, there's not a lot of significance on specifically _which_ content types are tested since you'll be calling `c.JSON` yourself.

For AutoFormat there's this:
```go
	c.Request().Header.Set(HeaderAccept, MIMEApplicationJSON)
	err = c.AutoFormat("Hello, World!")
	require.NoError(t, err)
	require.Equal(t, `"Hello, World!"`, string(c.Response().Body()))
```
But maybe it would be nice to add something that takes advantage of struct tags to get better use-case coverage. I can add that tomorrow.

ctx.go
sixcolors: having no handlers seems like a error, what is an intentional use case?

PR #2771, opened 2023-12-22
🩹 Fix: CORS middleware should use the defined AllowedOriginsFunc config when AllowedOrigins is empty

### LINKED ISSUE #2770: 🐛 [Bug]: CORS middleware should ignore config AllowedOrigins when AllowedOriginsFunc is defined
### Bug Description

I tried to set the CORS config `AllowedOriginsFunc` to define my custom logic for `Origin` header handling. The response always returned `*` in the `Access-Control-Allow-Origin` header. I expected the header to contain the origin allowed by the `AllowedOriginFunc` config I defined before. 

I checked the code and found that the `AllowedOrigins` config was set to `*` when it was empty, which IMO is not required if the `AllowedOriginsFunc` config is already defined. When the user sets the `AllowedOriginsFunc` then they should know that the `AllowedOrigins` config will not be used or ignored. 

I think this is a bug and should be fixed. I can make time for it. Let me know if you think otherwise. Thank you.

### How to Reproduce

1. Set CORS middleware with **only** `AllowedOriginsFunc` config defined.
2. The preflight (OPTIONS) request will always return `Access-Control-Allow-Origin` header with value `*`.

### Expected Behavior

The middleware should return `Access-Control-Allow-Origin` header with a value containing the origin allowed by `AllowedOriginsFunc` config instead of `*`.

### Fiber Version

2.49.2

### Code Snippet (optional)

```go
package main

import "github.com/gofiber/fiber/v2/middleware/cors"
import "log"
import "strings"

func main() {
  app := fiber.New()

  // Steps to reproduce
  app.Use(cors.New(cors.Config{
    AllowOriginsFunc: func(origin string) bool {
      // Complicated logic here
      return strings.HasSuffix(origin, "example.com:13131")
    },
  }))

}
```

### Checklist:

muhammadkholidb:
## Description

Current CORS middleware always returns the header `Access-Control-Allow-Origin` with the value of `*` when `AllowOriginsFunc` is defined and the function does not return `true`. It was due to the middleware always setting the default origin value with `*` whenever the `AllowOrigins` config is empty. So whenever the request origin is not allowed by the function (returns false), the middleware will return `*` in the `Access-Control-Allow-Origin` header which makes the function useless.

This PR sets the default origin value of `*` only when both `AllowOrigins` and `AllowOriginsFunc` configs are not defined so that it will not return the `Access-Control-Allow-Origin` header with the value of `*` when the function does not return `true`. Instead, it will return an empty origin when the function returns false, which means the request is rejected by CORS.

Current use cases
| AllowedOrigins | AllowedOriginsFunc | Request Origin | Response Origin |  | 
| --- | --- | --- | --- | --- |
| http://aaa.com |  nil | http://aaa.com | http://aaa.com | ✅ |
| http://aaa.com |  nil | http://bbb.com | empty | ✅ |
| http://aaa.com | return **true** | http://aaa.com | http://aaa.com | ✅ |
| http://aaa.com | return **true** | http://bbb.com | http://bbb.com | ✅ |
| http://aaa.com | return **false** | http://aaa.com | http://aaa.com | ✅ |
| http://aaa.com | return **false** | http://bbb.com | empty | ✅ |
| empty | nil | http://aaa.com | `*` | ✅ |
| empty | return **true** | http://aaa.com | http://aaa.com | ✅ |
| empty | return **false** | http://bbb.com | `*` | ❌ |

New use cases
| AllowedOrigins | AllowedOriginsFunc | Request Origin | Response Origin |  |
| --- | --- | --- | --- | --- |
| http://aaa.com |  nil | http://aaa.com | http://aaa.com | ✅ |
| http://aaa.com |  nil | http://bbb.com | empty | ✅ |
| http://aaa.com | return **true** | http://aaa.com | http://aaa.com | ✅ |
| http://aaa.com | return **true** | http://bbb.com | http://bbb.com | ✅ |
| http://aaa.com | return **false** | http://aaa.com | http://aaa.com | ✅ |
| http://aaa.com | return **false** | http://bbb.com | empty | ✅ |
| empty | nil | http://aaa.com | `*` | ✅ |
| empty | return **true** | http://aaa.com | http://aaa.com | ✅ |
| empty | return **false** | http://bbb.com | empty | ✅ |

Fixes #2770

## Type of change

## Checklist:

## Commit formatting:

Use emojis on commit messages so it provides an easy way of identifying the purpose or intention of a commit. Check out the emoji cheatsheet here: [CONTRIBUTING.md](https://github.com/gofiber/fiber/blob/master/.github/CONTRIBUTING.md#pull-requests-or-commits)

### DIFF
--- a/middleware/cors/cors.go
+++ b/middleware/cors/cors.go
@@ -94,13 +94,14 @@ func New(config ...Config) fiber.Handler {
 		if cfg.AllowMethods == "" {
 			cfg.AllowMethods = ConfigDefault.AllowMethods
 		}
-		if cfg.AllowOrigins == "" {
+		// When none of the AllowOrigins or AllowOriginsFunc config was defined, set the default AllowOrigins value with "*"
+		if cfg.AllowOrigins == "" && cfg.AllowOriginsFunc == nil {
 			cfg.AllowOrigins = ConfigDefault.AllowOrigins
 		}
 	}
 
 	// Warning logs if both AllowOrigins and AllowOriginsFunc are set
-	if cfg.AllowOrigins != ConfigDefault.AllowOrigins && cfg.AllowOriginsFunc != nil {
+	if cfg.AllowOrigins != "" && cfg.AllowOriginsFunc != nil {
 		log.Warn("[CORS] Both 'AllowOrigins' and 'AllowOriginsFunc' have been defined.")
 	}
 
@@ -145,7 +146,7 @@ func New(config ...Config) fiber.Handler {
 		// Run AllowOriginsFunc if the logic for
 		// handling the value in 'AllowOrigins' does
 		// not result in allowOrigin being set.
-		if (allowOrigin == "" || allowOrigin == ConfigDefault.AllowOrigins) && cfg.AllowOriginsFunc != nil {
+		if allowOrigin == "" && cfg.AllowOriginsFunc != nil {
 			if cfg.AllowOriginsFunc(origin) {
 				allowOrigin = origin
 			}
--- a/middleware/cors/cors_test.go
+++ b/middleware/cors/cors_test.go
@@ -331,9 +331,9 @@ func Test_CORS_AllowOriginsFunc(t *testing.T) {
 	// Perform request
 	handler(ctx)
 
-	// Allow-Origin header should be "*" because http://google.com does not satisfy 'strings.Contains(origin, "example-2")'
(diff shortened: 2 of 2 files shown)


efectn (approved):

ReneWerner87 (approved):

ReneWerner87:
thx for the effort

PR #2773, opened 2023-12-22
♻️ logger/middleware colorize logger error message #2593

ReneWerner87:
## Description

Color for the error tag of the logger and error tag by default in default format

### DIFF
--- a/docs/api/middleware/logger.md
+++ b/docs/api/middleware/logger.md
@@ -92,27 +92,27 @@ app.Use(logger.New(logger.Config{
 
 ### Config
 
-| Property         | Type                       | Description                                                                                                                      | Default                                                |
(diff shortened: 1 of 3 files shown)


ReneWerner87:
if you have any ideas to prevent the error from taking up space in the format if there is none, please feel free to add it to the branch

sixcolors:
@ReneWerner87 why not align with [morgan](https://expressjs.com/en/resources/middleware/morgan.html)?

sixcolors (approved):
LGTM!

sixcolors:
> LGTM!

Although I'm not 💯 on error being in the default log format.

gaby:
I think for v2 its fine, for v3 we can have a discussion with a path forward

gaby (approved):

PR #2782, opened 2024-01-03
add ulid routing Constraint

alfadhela:
## Description

Add new constraint for `ulid` using `github.com/oklog/ulid/v2` pkg.

## Type of change

## Checklist:

## Commit formatting:

Use emojis on commit messages so it provides an easy way of identifying the purpose or intention of a commit. Check out the emoji cheatsheet here: [CONTRIBUTING.md](https://github.com/gofiber/fiber/blob/master/.github/CONTRIBUTING.md#pull-requests-or-commits)

### DIFF
--- a/docs/guide/routing.md
+++ b/docs/guide/routing.md
@@ -162,7 +162,8 @@ Constraints aren't validation for parameters. If constraint aren't valid for par
 | range(min,max)    | :age<range(18,120)\>                 | 91 (Integer value must be at least 18 but no more than 120)                                 |
 | alpha             | :name<alpha\>                        | Rick (String must consist of one or more alphabetical characters, a-z and case-insensitive) |
 | datetime          | :dob<datetime(2006\\\\-01\\\\-02)\>  | 2005-11-01                                                                                  |
-| regex(expression) | :date<regex(\\d{4}-\\d{2}-\\d{2})\> | 2022-08-27 (Must match regular expression)                                                  |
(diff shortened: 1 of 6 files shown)


ReneWerner87:
image
hi, as already mentioned here, we want to include as few dependencies as possible in the core

thanks for the effort, but another solution to minimize this external dependency for everyone would be better for the core
e.g. about the possibility to add other contrains from outside

ReneWerner87:
you are welcome to work on it in the beta branch for version 3

we don't want new feature requests for v2 at the moment, because we want to stop the development there for now

alfadhela:
No problem but is it acceptable if write regexp for ulid instead of using `oklog/ulid`pkg.

ReneWerner87:
well, then the exam would only be half right
since there are many possibilities for contrains, but we shouldn't cover them all within the core, i think you should make them expandable

we can't possibly include all patterns, such as uuidv3, uuidv4, ulid etc.

to solve the problem on your side/project, however, you could use a regex contrain in your route

alfadhela:
It's a solution.
If I can suggest to make route constraint plugable similar to `go-playground/validator`. 
I'll try to create a PR if it's acceptable.

alfadhela:
Hi @ReneWerner87,
After trying to add `regex(expression)` I faced a new issue, in path all constraints will be lower-case. the solution is to add `config.CaseSensitive = true` to `fiber.App()` which is not ideal.

PR #2813, opened 2024-01-26
Addition of Locals Function with Go Generics as an Alternative to c.Locals

ryanbekhen:
Introduced a new function, Locals, that utilizes Go's generics to handle and retrieve type-specific local values within a request context. This enhancement provides more accurate data type control within the context. Included are tests for generic and custom struct use-cases to ensure the function performs as expected.

## Description

This pull request presents an alternative version of the Locals function using Go's generics. This new implementation of Locals function provides enhanced type safety and allows for more specific data type handling within the request context.

Key features of this implementation:

- It utilizes the Go's generics feature allowing it to handle varying data types dynamically. It takes a generic parameter [V any].
- The function tries to convert the input context to a DefaultCtx pointer and will throw an error if this conversion fails.
- The implementation uses the generic V for defining the type of value being retrieved from the context.
- It ensures that the returned value matches the type defined by the generic parameter.

This alternative Locals function provides an optional method for type-specific context manipulation. It brings enhanced type safety and data type specificity to the table which results in more resilient code.

>Noteworthy: This code requires at least Go 1.18 for its use of generics.

## Changes Introduced

## Type of Change

## Checklist

## Commit Formatting

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -817,6 +817,21 @@ func (c *DefaultCtx) Locals(key any, value ...any) any {
 	return value[0]
 }
 
+// Locals function utilizing Go's generics feature.
+// This function allows for manipulating and retrieving local values within a request context with a more specific data type.
+func Locals[V any](c Ctx, key any, value ...any) V {
+	ctx, ok := c.(*DefaultCtx)
+	if !ok {
+		panic(fmt.Errorf("failed to type-assert to *DefaultCtx"))
+	}
+	var v V
+	v, ok = ctx.Locals(key, value...).(V)
+	if !ok {
+		panic(fmt.Errorf("failed to type-assert to %T", v))
+	}
+	return v
+}
+
 // Location sets the response Location HTTP header to the specified path parameter.
 func (c *DefaultCtx) Location(path string) {
 	c.setCanonical(HeaderLocation, path)
--- a/ctx.go
+++ b/ctx.go
@@ -819,11 +819,16 @@ func (c *DefaultCtx) Locals(key any, value ...any) any {
 
 // Locals function utilizing Go's generics feature.
 // This function allows for manipulating and retrieving local values within a request context with a more specific data type.
-func Locals[V any](c Ctx, key any, value ...any) V {
-	v, ok := c.Locals(key, value...).(V)
+func Locals[V any](c Ctx, key any, value ...V) V {
+	var v V
+	var ok bool
+	if len(value) == 0 {
+		v, ok = c.Locals(key).(V)
+	} else {
+		v, ok = c.Locals(key, value[0]).(V)
+	}
 	if !ok {
-		// return zero value of type V
-		return v
+		return v // return zero of type V
 	}
 	return v
 }
--- a/ctx.go
+++ b/ctx.go
@@ -820,14 +820,10 @@ func (c *DefaultCtx) Locals(key any, value ...any) any {
 // Locals function utilizing Go's generics feature.
 // This function allows for manipulating and retrieving local values within a request context with a more specific data type.
 func Locals[V any](c Ctx, key any, value ...any) V {
-	ctx, ok := c.(*DefaultCtx)
+	v, ok := c.Locals(key, value...).(V)
 	if !ok {
-		panic(fmt.Errorf("failed to type-assert to *DefaultCtx"))
-	}
-	var v V
-	v, ok = ctx.Locals(key, value...).(V)
-	if !ok {
-		panic(fmt.Errorf("failed to type-assert to %T", v))
+		// return zero value of type V
+		return v
 	}
 	return v
 }
--- a/docs/api/ctx.md
+++ b/docs/api/ctx.md
@@ -1009,16 +1009,16 @@ func Locals[V any](c Ctx, key any, value ...any) V
 
 ```go title="Example"
 app.Use(func(c Ctx) error {
-    fiber.Locals[string](c, "john", "doe")
-    fiber.Locals[int](c, "age", 18)
-    fiber.Locals[bool](c, "isHuman", true)
-    return c.Next()
+  fiber.Locals[string](c, "john", "doe")
+  fiber.Locals[int](c, "age", 18)
+  fiber.Locals[bool](c, "isHuman", true)
+  return c.Next()
 })
 app.Get("/test", func(c Ctx) error {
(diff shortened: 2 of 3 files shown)


docs/api/ctx.md
```diff
@@ -1000,6 +1000,34 @@ app.Get("/admin", func(c fiber.Ctx) error {
...
+})
+app.Get("/test", func(c Ctx) error {
+	fiber.Locals[string](c, "john")     // "doe"
+    fiber.Locals[int](c, "age")         // 18
+    fiber.Locals[bool](c, "isHuman")    // true
+return nil
+})
+````
+
+Keep in mind, you will need Go 1.18 or later to take advantage of generics. Also, be aware of the types you pass into
```
efectn, line 1025: No need to write `Keep in mind, you will need Go 1.18 or later to take advantage of generics. ` v3 will support only latest 2 versions of Go

ryanbekhen: Thx, i'll remove it

efectn (approved):

ctx.go
```diff
@@ -817,6 +817,21 @@ func (c *DefaultCtx) Locals(key any, value ...any) any {
 	return value[0]
 }
 
+// Locals function utilizing Go's generics feature.
+// This function allows for manipulating and retrieving local values within a request context with a more specific data type.
+func Locals[V any](c Ctx, key any, value ...any) V {
+	ctx, ok := c.(*DefaultCtx)
```
nickajacks1, line 823: what is the need for the type assertion?

ctx.go
```diff
@@ -817,6 +817,21 @@ func (c *DefaultCtx) Locals(key any, value ...any) any {
...
+	if !ok {
+		panic(fmt.Errorf("failed to type-assert to *DefaultCtx"))
+	}
+	var v V
+	v, ok = ctx.Locals(key, value...).(V)
+	if !ok {
+		panic(fmt.Errorf("failed to type-assert to %T", v))
```
nickajacks1, line 830: would it be better to return an error instead of panicking?

ctx.go
```diff
@@ -817,6 +817,21 @@ func (c *DefaultCtx) Locals(key any, value ...any) any {
 	return value[0]
 }
 
+// Locals function utilizing Go's generics feature.
+// This function allows for manipulating and retrieving local values within a request context with a more specific data type.
+func Locals[V any](c Ctx, key any, value ...any) V {
```
nickajacks1, line 822: should value be of type V? that way you can call local without the type param and it increases type safety a bit.

`fiber.Locals(c, "age", 50)`

docs/api/ctx.md
```diff
@@ -1000,6 +1000,31 @@ app.Get("/admin", func(c fiber.Ctx) error {
...
+    fiber.Locals[string](c, "john", "doe")
+    fiber.Locals[int](c, "age", 18)
+    fiber.Locals[bool](c, "isHuman", true)
+    return c.Next()
+})
+app.Get("/test", func(c Ctx) error {
+	fiber.Locals[string](c, "john")     // "doe"
+    fiber.Locals[int](c, "age")         // 18
+    fiber.Locals[bool](c, "isHuman")    // true
+return nil
```
nickajacks1, line 1021: sorry can you fix the indent here? 1018 uses a tab instead of spaces and 1021 isn't indented. thx

ctx.go
ryanbekhen: I realize we shouldn't use a type assertion here because we don't need access to private functions.

ctx.go
ryanbekhen: ok i'll change it

docs/api/ctx.md
ryanbekhen: ok i'll change it

ctx.go
ryanbekhen: Maybe I think it could be changed with the default of its data type, like this:
```go
v, ok = c.Locals(key, value...).(V)
  if !ok {
    return v
  }
return v
```
What do you think?

ryanbekhen: What do you think, if we return default value of variable and Locals Generic change to only get value from local not for set the value?

`c.Locals("name", "john")`
`fiber.Locals(c, "name")` // john
`fiber.Locals[string](c, "unknown") // <empty string>
`fiber.Locals[string](c, "unknown", "default") // default

so, we don't need for return an error, maybe like Query Generic?

ryanbekhen: @nickajacks1 @efectn

ryanbekhen: maybe like this:
image

nickajacks1:
```go
user := fiber.Locals(c, "user")
if user == User{} { /* stuff */ }

// or

user := fiber.Locals(c, "user", DefaultUser)
if user == DefaultUser { /* stuff */ }
```
vs
```go
user, err := fiber.Locals(c, "user")
if err != nil { /* stuff */ }
```
I think you're right that returning the zero value makes more sense given that there's a default value parameter.  Returning an error is a bit redundant with the default return value.

efectn: Yeah returning zero-value seems better to me

what do you thing @gofiber/maintainers

ctx.go
ReneWerner87: yes good point

ctx.go
ReneWerner87: pls update this

ctx.go
ReneWerner87: yes zero value please and without defaultValue
the method should also be usable as a setter and getter like the other method, so in any case I would expect that and no other pattern
so value should be the value you set and not defaultValue, otherwise we would have to rename the method

ReneWerner87 (changes requested):
some little adjustements needed

ryanbekhen:
@ReneWerner87 I've done the push

ctx.go
ReneWerner87: @ryanbekhen can you look at this

```suggestion
func Locals[V any](c Ctx, key any, value ...V) V {
```

ryanbekhen: @ReneWerner87 do you want to change it to `func Locals[V any](c Ctx, key any, value ...V) V`

ryanbekhen:
@ReneWerner87 I've done the push, pls check again.

Issue #2816, opened 2024-01-27
🧹 [Maintenance]: Update golangci-lint


nickajacks1:
### Maintenance Task Description

As of writing, the latest version of golangci-lint is v1.55.2. This issue encompasses updating the github actions to use at least this version of golangci-lint along with updating the .golanci.yml config to enable or disable new rules and linters as agreed upon by the fiber team.

- [gosmopolitan](https://github.com/xen0n/gosmopolitan) checks for i18n usage, including time.Local
- [mirror](https://github.com/butuzov/mirror) checks for allocs in string/[]byte conversions
- [tagalign](https://github.com/4meepo/tagalign) checks for struct tag formatting
- [gochecksumtype](https://github.com/alecthomas/go-check-sumtype) checks for exhaustive type switches. Somewhat complex, check out the examples.
- [inamedparam](https://github.com/macabu/inamedparam) requires named params for interface functions (can exclude single param funcs)
- [perfsprint](https://github.com/catenacyber/perfsprint) checks faster alternatives to fmt.Sprintf
- [testifylint](https://github.com/Antonboom/testifylint) checks idiomatic testify usage
- [zerologlint](https://github.com/ykadowak/zerologlint) zerolog usage
- [sloglint](https://github.com/go-simpler/sloglint) slog usage
- [protogetter](https://github.com/ghostiam/protogetter) protobuf usage

In addition to the new linters, there are several new linter rules for existing linters, many of which raise errors in the current codebase. These will likely need to be inspected individually.

### Impact on the Project

Updating the linter benefits the project by keeping up with modern Go practices, maintaining code consistency, and reducing the probability of bugs and API issues.

### Checklist:

- [X] I have confirmed that this maintenance task is currently not being addressed.
- [X] I understand that this task will be evaluated by the maintainers and prioritized accordingly.
- [X] I am available to provide further information if needed.

efectn:
We should add support for `fieldalignment` linter too.

PR #2817, opened 2024-01-28
ci: Update golangci-lint to v1.55.2

nickajacks1:
## Description

Update golangci-lint to the latest version.
New linters and failing rules were commented out for visibility and tagged with TODO comments.

Related to #2816

## Type of Change

## Checklist

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -39,18 +39,20 @@ linters-settings:
     extra-rules: true
 
   gosec:
+    excludes:
+      - G104
     config:
       global:
         audit: true
   
   depguard:
-    include-go-root: true
-    packages:
-      - flag
-      - io/ioutil
-    packages-with-error-message:
-      - flag: '`flag` package is only allowed in main.go'
-      - io/ioutil: '`io/ioutil` package is deprecated, use the `io` and `os` package instead'
+    rules:
+      main:
+        deny:
+          - pkg: flag
+            desc: '`flag` package is only allowed in main.go'
+          - pkg: io/ioutil
+            desc: '`io/ioutil` package is deprecated, use the `io` and `os` package instead'
 
   govet:
     check-shadowing: true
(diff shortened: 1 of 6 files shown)


nickajacks1:
In addition to the new linters mentioned in #2816, several rules and existing linters had a large number of new issues, so I just disabled them for now.

Disabled rules and linters:
- gosec: G104 is now raised for _ assignments (at least when audit: true is set).  Exclude G104 until unchecked errors are ignored (there are a lot).
- depguard: the config format has changed
- revive
  - comment-spacings:  spaces after the // in a comment - there are a few instances of this, including "decorated" comments
  - exported,disableStutterCheck:  some functions, eg csrf.CsrfFromHeader are advised to be changed to csrf.FromHeader. Updating these would be breaking changes, so they should be fixed before v3 is released, and they should be noted in any migration guides.
  - unchecked-type-assertion:  overlaps with errcheck
- goconst: this seems to be more strict now.  A few of the new rules might be a little awkward to fix. we'll see.
- wastedassign: this seems to be more strict now.

Maybe we can drop errcheck in favor of revive's unchecked-type-assertion and gosec's G104? Or the other way around?

ReneWerner87:
> Maybe we can drop errcheck in favor of revive's unchecked-type-assertion and gosec's G104? Or the other way around?

as long as a check is on, that would be okay, the important thing is that the error returns are always checked

.golangci.yml
```diff
@@ -93,13 +95,17 @@ linters-settings:
         disabled: true
       - name: cognitive-complexity
         disabled: true
+      - name: comment-spacings
+        disabled: true # TODO https://github.com/gofiber/fiber/issues/2816
```
gaby, line 99: Is this still TODO? Or are we handling this in a diff PR?

gaby: Nevermind :-)

gaby (approved):
LGTM

ReneWerner87 (approved):

PR #2830, opened 2024-02-04
fix(logger): print to stderr if log fails for default format

nickajacks1:
## Description

We log to stderr if logging fails when a custom format is used, but not for the default format. This change addresses this inconsistency.

## Type of Change

## Checklist

### DIFF
--- a/middleware/logger/default_logger.go
+++ b/middleware/logger/default_logger.go
@@ -2,6 +2,7 @@ package logger
 
 import (
 	"fmt"
+	"io"
 	"os"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


nickajacks1:
One other thing I noticed was that the log write is only protected by a mutex lock when the non-default format is used.  Perhaps we should be locking in both places...?

efectn:
> One other thing I noticed was that the log write is only protected by a mutex lock when the non-default format is used. Perhaps we should be locking in both places...?

There should not be issue when writing logs to a file using POSIX `write()` or stdout. But there are probably some edge cases that may cause data races. So i'd be OK to add mutex both places.

nickajacks1:
Thought about the locking some more. It's probably not a good idea for fiber to add locking. Since os.File operations are goroutine safe, and since I'm sure >99% of cases will be logging to os.File, it's probably better to have users implement locking in their Write methods.  Either way, I'll follow up with a separate PR later.

ReneWerner87 (approved):

PR #2842, opened 2024-02-10
Bump golangci-lint to v1.56.1

gaby:
- Bump golangci-lint to v1.56.1


ReneWerner87:
@gaby 
image

hm and this
image
is it a flaky test?

gaby:
@ReneWerner87 golangci-lint changed quite a few things in v1.56.0, I plan on fixing those. The test failing is a different issue.

nickajacks1:
> is it a flaky test?
Yes, Session is one of the intermittently failing tests.

And just to say it out loud, 1.56 added https://github.com/jjti/go-spancheck which has to do with opentelemetry, so we don't need it right now.

app_test.go
```diff
@@ -866,7 +866,7 @@ func Test_App_ShutdownWithContext(t *testing.T) {
 	ln := fasthttputil.NewInmemoryListener()
 
 	go func() {
-		require.NoError(t, app.Listener(ln))
+		require.NoError(t, app.Listener(ln)) //nolint:testifylint // Needed for test
```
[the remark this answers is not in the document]
gaby: @nickajacks1 After running it locally there is +40 instances of this same error from testifylint. Disabling this check is a better fix than adding `nolint` in so many places.

gaby:
This was way more changes that I thought. For some reason golangci-lint stops checking after N amount of errors. So i had to constantly be running it. Turns out there was over +100 warnings

.golangci.yml
```diff
@@ -11,6 +11,10 @@ output:
   sort-results: true
 
 linters-settings:
+  testifylint:
+    enable-all: true
+    disable:
+      - go-require
```
gaby, line 17: This will allow us to use `require` inside a goroutine during tests

efectn (approved):

middleware/adaptor/adaptor.go
```diff
@@ -56,7 +56,7 @@ func CopyContextToFiberContext(context any, requestContext *fasthttp.RequestCtx)
 				CopyContextToFiberContext(reflectValue.Interface(), requestContext)
 			} else if reflectField.Name == "key" {
 				lastKey = reflectValue.Interface()
-			} else if lastKey != nil && reflectField.Name == "val" {
+			} else if lastKey != nil && reflectField.Name == "val" { //nolint:revive // We need these checks here
```
nickajacks1, line 59: should be able to avoid the nolint by inverting the outer `if` to return early
```go
if contextKeys.Kind() != reflect.Struct {
	return
}
```

nickajacks1 (approved):
one small comment, thanks for fixing all of these

sixcolors (approved):

nickajacks1 (approved):

Issue #2758, opened 2023-12-11
🚀 [Feature]: Implement Generic APIs for Enhanced Variable Type


dozheiny:
### Feature Description

Some APIs like [ctx.QueryBool](https://pkg.go.dev/github.com/gofiber/fiber/v2#Ctx.QueryBool) or [ctx.QueryFloat](https://pkg.go.dev/github.com/gofiber/fiber/v2#Ctx.QueryFloat) convert given variables. Why don't add APIs that let users convert the needed variable types?

For example, Query convertor API can be implemented like this *(Since we can't use generics in methods)*:
```go
func QueryFunc[T any](c *Ctx, key string, convertor func(string)(T, error), defaultValue ...T) (T, error)
```

Usage:
```go
package main

import (
    "github.com/gofiber/fiber/v2"
    "github.com/google/uuid"
    "log"
)

func main() {
	app := fiber.New()

	// [GET] /?id=01827964-1320-47b4-b2fa-c67fa9c39bed
	app.Get("/", func(c *fiber.Ctx) error {

		id, err := fiber.QueryFunc(c, "id", uuid.Parse)
       		if err != nil {
			log.Fatal(err)
			}

       		// some functionality
    })

    app.Listen(":3000")
}
```

We can implement other APIs like [ctx.Params](https://pkg.go.dev/github.com/gofiber/fiber/v2#Ctx.Params) and [ctx.Get](https://pkg.go.dev/github.com/gofiber/fiber/v2#Ctx.Get) functions.

### Checklist:

- [X] I agree to follow Fiber's [Code of Conduct](https://github.com/gofiber/fiber/blob/master/.github/CODE_OF_CONDUCT.md).
- [X] I have checked for existing issues that describe my suggestion prior to opening this one.
- [X] I understand that improperly formatted feature requests may be closed without explanation.

ReneWerner87:
We could do it for v3 not for v2

brunodmartins:
@ReneWerner87 I would like to work on this issue. Can I work on it? In case so, which branch should I use to fork?

nickajacks1:
@brunodmartins v3-beta is the latest v3 development branch

ReneWerner87:
Reopened for Param and Get function
@brunodmartins Can you abstract these functionalities and also the queryType and the sub methods so that we have the same for the other 2 topic methods

dozheiny:
@ReneWerner87 I would like to do this task. If it's possible

[...]

PR #2850, opened 2024-02-11
🔥 [V3] Params Generic function

dozheiny:
## Description
This PR implements a new type of Params function that can handle different value types of Parameters function(#2758).

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -972,6 +972,31 @@ func (c *DefaultCtx) Params(key string, defaultValue ...string) string {
 	return defaultString("", defaultValue)
 }
 
+// Params is used to get the route parameters.
+// This function is generic and can handle differnet route paramaters type values.
+//
+// Example:
+//
+// http://example.com/user/:user -> http://example.com/user/john
+// Parmas[string](c, "user") -> returns john
+//
+// http://example.com/id/:id -> http://example.com/user/114
+// Params[int](c, "id") ->  returns 114 as integer.
+//
+// http://example.com/id/:number -> http://example.com/id/john
+// Params[string](c, "number", 0) -> returns 0 because can't parse 'john' as integer.
+func Params[V GenericType](c Ctx, key string, defaultValue ...V) V {
+	ctx, ok := c.(*DefaultCtx)
+	if !ok {
+		panic(fmt.Errorf("failed to type-assert to *DefaultCtx"))
+	}
+
+	var v V
+	p := ctx.Params(key)
+
+	return genericParseType[V](ctx, p, v, defaultValue...)
+}
+
 // ParamsInt is used to get an integer from the route parameters
 // it defaults to zero if the parameter is not found or if the
 // parameter cannot be converted to an integer
@@ -1107,7 +1132,7 @@ func (c *DefaultCtx) Queries() map[string]string {
 //	name := Query[string](c, "search") // Returns "john"
 //	age := Query[int](c, "age") // Returns 8
 //	unknown := Query[string](c, "unknown", "default") // Returns "default" since the query parameter "unknown" is not found
-func Query[V QueryType](c Ctx, key string, defaultValue ...V) V {
+func Query[V GenericType](c Ctx, key string, defaultValue ...V) V {
 	ctx, ok := c.(*DefaultCtx)
 	if !ok {
 		panic(fmt.Errorf("failed to type-assert to *DefaultCtx"))
@@ -1115,68 +1140,26 @@ func Query[V QueryType](c Ctx, key string, defaultValue ...V) V {
 	var v V
 	q := ctx.app.getString(ctx.fasthttp.QueryArgs().Peek(key))
 
-	switch any(v).(type) {
-	case int:
-		return queryParseInt[V](q, 32, func(i int64) V { return assertValueType[V, int](int(i)) }, defaultValue...)
-	case int8:
-		return queryParseInt[V](q, 8, func(i int64) V { return assertValueType[V, int8](int8(i)) }, defaultValue...)
-	case int16:
-		return queryParseInt[V](q, 16, func(i int64) V { return assertValueType[V, int16](int16(i)) }, defaultValue...)
-	case int32:
-		return queryParseInt[V](q, 32, func(i int64) V { return assertValueType[V, int32](int32(i)) }, defaultValue...)
-	case int64:
-		return queryParseInt[V](q, 64, func(i int64) V { return assertValueType[V, int64](i) }, defaultValue...)
-	case uint:
-		return queryParseUint[V](q, 32, func(i uint64) V { return assertValueType[V, uint](uint(i)) }, defaultValue...)
-	case uint8:
-		return queryParseUint[V](q, 8, func(i uint64) V { return assertValueType[V, uint8](uint8(i)) }, defaultValue...)
-	case uint16:
-		return queryParseUint[V](q, 16, func(i uint64) V { return assertValueType[V, uint16](uint16(i)) }, defaultValue...)
-	case uint32:
-		return queryParseUint[V](q, 32, func(i uint64) V { return assertValueType[V, uint32](uint32(i)) }, defaultValue...)
-	case uint64:
-		return queryParseUint[V](q, 64, func(i uint64) V { return assertValueType[V, uint64](i) }, defaultValue...)
-	case float32:
-		return queryParseFloat[V](q, 32, func(i float64) V { return assertValueType[V, float32](float32(i)) }, defaultValue...)
-	case float64:
-		return queryParseFloat[V](q, 64, func(i float64) V { return assertValueType[V, float64](i) }, defaultValue...)
-	case bool:
-		return queryParseBool[V](q, func(b bool) V { return assertValueType[V, bool](b) }, defaultValue...)
-	case string:
-		if q == "" && len(defaultValue) > 0 {
-			return defaultValue[0]
-		}
-		return assertValueType[V, string](q)
-	case []byte:
-		if q == "" && len(defaultValue) > 0 {
-			return defaultValue[0]
-		}
-		return assertValueType[V, []byte](ctx.app.getBytes(q))
-	default:
-		if len(defaultValue) > 0 {
-			return defaultValue[0]
-		}
-		return v
-	}
+	return genericParseType[V](ctx, q, v)
 }
 
-type QueryType interface {
-	QueryTypeInteger | QueryTypeFloat | bool | string | []byte
+type GenericType interface {
+	GenericTypeInteger | GenericTypeFloat | bool | string | []byte
 }
 
-type QueryTypeInteger interface {
-	QueryTypeIntegerSigned | QueryTypeIntegerUnsigned
+type GenericTypeInteger interface {
+	GenericTypeIntegerSigned | GenericTypeIntegerUnsigned
 }
 
-type QueryTypeIntegerSigned interface {
+type GenericTypeIntegerSigned interface {
 	int | int8 | int16 | int32 | int64
 }
 
-type QueryTypeIntegerUnsigned interface {
+type GenericTypeIntegerUnsigned interface {
 	uint | uint8 | uint16 | uint32 | uint64
 }
 
-type QueryTypeFloat interface {
+type GenericTypeFloat interface {
 	float32 | float64
 }
 
--- a/ctx_test.go
+++ b/ctx_test.go
@@ -2294,6 +2294,126 @@ func Benchmark_Ctx_Params(b *testing.B) {
 	require.Equal(b, "awesome", res)
 }
 
+// go test -run Test_Params
+func Test_Params(t *testing.T) {
+	t.Parallel()
+
+	app := New()
+
+	app.Get("/test/:string/:number/:bool/:float", func(c Ctx) error {
+		// int
+		require.Equal(t, 8, Params[int](c, "number"))
+		require.Equal(t, 8, Params[int](c, "number", 1))
+		require.Equal(t, 8, Params[int](c, "number", 1, 2))
+		require.Equal(t, 0, Params[int](c, "string"))
+		require.Equal(t, 1, Params[int](c, "string", 1))
+		require.Equal(t, 1, Params[int](c, "string", 1, 2))
+
+		// int8
+		require.Equal(t, int8(8), Params[int8](c, "number"))
+		require.Equal(t, int8(8), Params[int8](c, "number", 1))
+		require.Equal(t, int8(8), Params[int8](c, "number", 1, 2))
+		require.Equal(t, int8(0), Params[int8](c, "string"))
+		require.Equal(t, int8(1), Params[int8](c, "string", 1))
+		require.Equal(t, int8(1), Params[int8](c, "string", 1, 2))
+
+		// int16
+		require.Equal(t, int16(8), Params[int16](c, "number"))
+		require.Equal(t, int16(8), Params[int16](c, "number", 1))
+		require.Equal(t, int16(8), Params[int16](c, "number", 1, 2))
+		require.Equal(t, int16(0), Params[int16](c, "string"))
+		require.Equal(t, int16(1), Params[int16](c, "string", 1))
+		require.Equal(t, int16(1), Params[int16](c, "string", 1, 2))
+
+		// int32
+		require.Equal(t, int32(8), Params[int32](c, "number"))
+		require.Equal(t, int32(8), Params[int32](c, "number", 1))
+		require.Equal(t, int32(8), Params[int32](c, "number", 1, 2))
+		require.Equal(t, int32(0), Params[int32](c, "string"))
+		require.Equal(t, int32(1), Params[int32](c, "string", 1))
+		require.Equal(t, int32(1), Params[int32](c, "string", 1, 2))
+
+		// int64
+		require.Equal(t, int64(8), Params[int64](c, "number"))
+		require.Equal(t, int64(8), Params[int64](c, "number", 1))
+		require.Equal(t, int64(8), Params[int64](c, "number", 1, 2))
+		require.Equal(t, int64(0), Params[int64](c, "string"))
+		require.Equal(t, int64(1), Params[int64](c, "string", 1))
+		require.Equal(t, int64(1), Params[int64](c, "string", 1, 2))
+
+		// uint
+		require.Equal(t, uint(8), Params[uint](c, "number"))
+		require.Equal(t, uint(8), Params[uint](c, "number", 1))
+		require.Equal(t, uint(8), Params[uint](c, "number", 1, 2))
+		require.Equal(t, uint(0), Params[uint](c, "string"))
+		require.Equal(t, uint(1), Params[uint](c, "string", 1))
+		require.Equal(t, uint(1), Params[uint](c, "string", 1, 2))
+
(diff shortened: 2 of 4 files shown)


gaby:
What's the benchmark difference between using this and the previous implementation?

dozheiny:
> What's the benchmark difference between using this and the previous implementation?

I'll implement this.

ctx_test.go
```diff
@@ -2294,6 +2294,305 @@ func Benchmark_Ctx_Params(b *testing.B) {
...
+		require.InEpsilon(t, float32(3.1415), Params[float32](c, "float", float32(1.234)), epsilon)
+		require.InDelta(t, float32(0), Params[float32](c, "string"), 0)
+		require.InEpsilon(t, float32(1.234), Params[float32](c, "string", float32(1.234)), epsilon)
+
+		// float64
+		require.InEpsilon(t, float64(3.1415), Params[float64](c, "float"), epsilon)
+		require.InEpsilon(t, float64(3.1415), Params[float64](c, "float", float64(1.234)), epsilon)
+		require.InDelta(t, float64(0), Params[float64](c, "string"), 0)
+		require.InEpsilon(t, float64(1.234), Params[float64](c, "string", float64(1.234)), epsilon)
+		return nil
```
sixcolors, line 2409:
```suggestion
		// default
		require.Equal(t, "doe", Params[string](c, "fake", "doe"))
		require.Equal(t, 0, Params[int](c, "fake"))
		require.Equal(t, 1, Params[int](c, "fake", 1))

		return nil
```

ctx_test.go
```diff
@@ -2294,6 +2294,305 @@ func Benchmark_Ctx_Params(b *testing.B) {
...
+	})
+
+	resp, err := app.Test(httptest.NewRequest(MethodGet, "/test/john/8/true/3.1415", nil))
+	require.NoError(t, err, "app.Test(req)")
+	require.Equal(t, StatusOK, resp.StatusCode, "Status code")
+}
+
```
sixcolors, line 2416:
```suggestion

    // go test -run Test_Params_TypeAssertFail
func Test_Params_TypeAssertFail(t *testing.T) {
	type customCtx struct {
		Ctx
	}

	app := New()
	c := &customCtx{
		Ctx: app.NewCtx(&fasthttp.RequestCtx{}),
	}

	defer func() {
		if r := recover(); r == nil {
			t.Errorf("The code did not panic")
		}
	}()

	Params[int](c, "key")
}

```

sixcolors: and similar for `func Query[V GenericType](c Ctx, key string, defaultValue ...V) V {`

sixcolors (approved):
otherwise LGTM

gaby:
Wait for #2842 to get merged. I suspect a lot of warnings will pop up

dozheiny:
> Wait for #2842 to get merged. I suspect a lot of warnings will pop up

Ok. I'll wait.

utils.go
```diff
@@ -25,22 +25,68 @@ func queryParseDefault[V QueryType](err error, parser func() V, defaultValue ...
...
 
-func queryParseBool[V QueryType](q string, parser func(bool) V, defaultValue ...V) V {
-	result, err := strconv.ParseBool(q)
-	return queryParseDefault[V](err, func() V { return parser(result) }, defaultValue...)
+func genericParseBool[V GenericType](str string, parser func(bool) V, defaultValue ...V) V {
+	result, err := strconv.ParseBool(str)
+	return genericParseDefault[V](err, func() V { return parser(result) }, defaultValue...)
+}
+
+func genericParseType[V GenericType](ctx *DefaultCtx, str string, v V, defaultValue ...V) V {
```
ReneWerner87, line 48: maybe we should export this function

you can also add the converter function
```go
Convert[T any](value string, convertor func(string) (T, error), defaultValue ...T) (*T, error) { 
	converted, err := convertor(value)
	if err != nil {
		if len(defaultValue) > 0 {
			return &defaultValue[0], nil
		}

		return nil, fmt.Errorf("failed to convert: %w", err)
	}

	return &converted, nil
}
```

ReneWerner87 (changes requested):
thx, some adjustments needed

utils.go
dozheiny: I plan to add this function in the following PR when implementing the `GetGeneric` function.
Should I add the `GetGeneric` function to this PR, too?

docs/api/ctx.md
```diff
@@ -1197,6 +1197,34 @@ app.Get("/v1/*/shop/*", func(c fiber.Ctx) error {
...
+
+The generic Params function supports returning the following data types based on V genericType:
+- Integer: int, int8, int16, int32, int64
+- Unsigned integer: uint, uint8, uint16, uint32, uint64
+- Floating-point numbers: float32, float64
+- Boolean: bool
+- String: string
+- Byte array: []byte
+
 ## ParamsInt
```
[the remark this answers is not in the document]
dozheiny: Users may have problems when migrating from v2 to v3. How about I use the generic Params function in `ParamsInt` and other similar functions? I also added a DEPRECATED flag in the comment section.

utils.go
ReneWerner87: yes would be ok for me

docs/api/ctx.md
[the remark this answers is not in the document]
dozheiny: I thought about it. ParamsInt returns an error if can't parse it. So if we rewrite the function with new Params, it returns an empty error always. so I'll remove it.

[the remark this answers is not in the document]
ReneWerner87: the aim was to replace these functions with the generics
so let's stick to it

utils.go
ReneWerner87: sorry i meant we can add it right here in this pull request to minimize the intermediate steps

you just need to change the description and headline of the pull request

ReneWerner87:
@dozheiny please tag me if you have any questions or are done with the pull request customizations

utils.go
dozheiny: Sure. I'm gonna change the pull request headline and description.

ReneWerner87:
> @ReneWerner87 tests go wrong because the `Benchmark_genericParse` function shouldn't receive parameters. Should I use the `b.Run` child on every benchmark? How about to use `genericParseType` directly?

image
yes pls that´s the plan

ReneWerner87:
@dozheiny what does it look like? do you need anything else?

dozheiny:
Thanks, @ReneWerner87 . Sorry for being late. I'll update the PR in the next 24 hours.

sixcolors:
#2854 may affect how we deal with this. @nickajacks1 please review.

dozheiny:
I handled `genericParseType` without the pointer of DefaultCtx, the only way it is needed is when the Generic type is the array of bytes.
Should I return to the previous method?

dozheiny:
@ReneWerner87 Should I use `t.Run` and `b.Run` as a child? Tests can run because test functions receive parameters.

ReneWerner87:
> @ReneWerner87 Should I use `t.Run` and `b.Run` as a child? Tests can run because test functions receive parameters.

yes, child cases are good for both

ctx.go
```diff
@@ -1761,3 +1732,17 @@ func (c *DefaultCtx) Bind() *Bind {
 	}
 	return c.bind
 }
+
+// Converts a string value to a specified type, handling errors and optional default values.
+func Convert[T any](value string, convertor func(string) (T, error), defaultValue ...T) (*T, error) {
```
efectn, line 1737: Are there any reason we return pointer here?

nickajacks1 (commented):
I'm starting to wonder if there's enough value in this new generics API to warrant implementation and maintenance. For one, it's solving the same problem as v3's Bind, meaning there are now three ways to get query parameters.

It's also not actually taking advantage of generics.  We're still eventually taking the type parameter, casting it to `any`, and doing a type switch. In effect, we're doing our own hand-rolled dynamic dispatch.

There are ways to avoid the type switch, but that would involve using custom types like this

```go
type Parser[v any] interface {
	Parse(s string, defaultRet ...v) v
}

type fUint16 uint16

func (fUint16) Parse(s string, defaultRet ...uint16) uint16 {
	return ParseUint16(s, defaultRet...)
}


func ParseUint16(s string, defaultRet ...uint16) uint16 {
	v, err := strconv.ParseUint(s, 10, 16)
	if err != nil {
		if len(defaultRet) == 0 {
			return 0
		}
		return defaultRet[0]
	}
	return uint16(v)
}

var Uint16 fUint16

func QueryZ[V Parser[T], T any](str string, v V, defaultRet ...T) T {
	return v.Parse(str, defaultRet...)
}

func QueryY[V Parser[T], T any](str string, defaultRet ...T) T {
	var v V
	return v.Parse(str, defaultRet...)
}

func QueryX[T any](str string, f func(string, ...T) T, defaults ...T) T {
	return f(str, defaults...)
}

// ...

func Test_Thingies(t *testing.T) {
	require.Equal(t, uint16(23), QueryZ("-10", Uint16, 23))
	require.Equal(t, uint16(10), QueryZ("10", Uint16, 23))
	require.Equal(t, uint16(23), QueryY[fUint16]("-10", 23))
	require.Equal(t, uint16(10), QueryY[fUint16]("10", 23))
	require.Equal(t, uint16(23), QueryX("-10", ParseUint16, 23))
	require.Equal(t, uint16(10), QueryX("10", ParseUint16, 23))
}

```

But then we're getting even closer to overlapping with Bind.  I kind of worry about having two separate ways of doing the same thing....What do others think?

ctx.go
```diff
@@ -955,40 +967,47 @@ func (c *DefaultCtx) OriginalURL() string {
...
+// http://example.com/id/:id -> http://example.com/user/114
+// Params[int](c, "id") ->  returns 114 as integer.
+//
+// http://example.com/id/:number -> http://example.com/id/john
+// Params[string](c, "number", 0) -> returns 0 because can't parse 'john' as integer.
+func Params[V GenericType](c Ctx, key string, defaultValue ...V) V {
+	ctx, ok := c.(*DefaultCtx)
+	if !ok {
+		panic(fmt.Errorf("failed to type-assert to *DefaultCtx"))
+	}
```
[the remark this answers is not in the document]
efectn: > Won't forced assertion defeat the purpose of the interface concept? Or is it a requirement that implementers of Ctx struct-compose a DefaultCtx? @efectn Could this be changed to use c.Route() and c.App()?

Yes exactly it needs to be avoided

ctx.go
ReneWerner87: @dozheiny

ReneWerner87:
> I'm starting to wonder if there's enough value in this new generics API to warrant implementation and maintenance. For one, it's solving the same problem as v3's Bind, meaning there are now three ways to get query parameters.
> 
> It's also not actually taking advantage of generics. We're still eventually taking the type parameter, casting it to `any`, and doing a type switch. In effect, we're doing our own hand-rolled dynamic dispatch.
> 
> There are ways to avoid the type switch, but that would involve using custom types like this
> 
> ```go
> type Parser[v any] interface {
> 	Parse(s string, defaultRet ...v) v
> }
> 
> type fUint16 uint16
> 
> func (fUint16) Parse(s string, defaultRet ...uint16) uint16 {
> 	return ParseUint16(s, defaultRet...)
> }
> 
> 
> func ParseUint16(s string, defaultRet ...uint16) uint16 {
> 	v, err := strconv.ParseUint(s, 10, 16)
> 	if err != nil {
> 		if len(defaultRet) == 0 {
> 			return 0
> 		}
> 		return defaultRet[0]
> 	}
> 	return uint16(v)
> }
> 
> var Uint16 fUint16
> 
> func QueryZ[V Parser[T], T any](str string, v V, defaultRet ...T) T {
> 	return v.Parse(str, defaultRet...)
> }
> 
> func QueryY[V Parser[T], T any](str string, defaultRet ...T) T {
> 	var v V
> 	return v.Parse(str, defaultRet...)
> }
> 
> func QueryX[T any](str string, f func(string, ...T) T, defaults ...T) T {
> 	return f(str, defaults...)
> }
> 
> // ...
> 
> func Test_Thingies(t *testing.T) {
> 	require.Equal(t, uint16(23), QueryZ("-10", Uint16, 23))
> 	require.Equal(t, uint16(10), QueryZ("10", Uint16, 23))
> 	require.Equal(t, uint16(23), QueryY[fUint16]("-10", 23))
> 	require.Equal(t, uint16(10), QueryY[fUint16]("10", 23))
> 	require.Equal(t, uint16(23), QueryX("-10", ParseUint16, 23))
> 	require.Equal(t, uint16(10), QueryX("10", ParseUint16, 23))
> }
> ```
> 
> But then we're getting even closer to overlapping with Bind. I kind of worry about having two separate ways of doing the same thing....What do others think?

@nickajacks1 bindings always need structs to bind, but this functionality holds the bare value 
i would support both, as in some cases it is not worth initializing an extra struct for only one value

ctx.go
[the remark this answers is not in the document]
ReneWerner87: @dozheiny could you do this

ReneWerner87 (approved):

ReneWerner87:
@coderabbitai review

ReneWerner87:
@coderabbitai review

nickajacks1:
> > I'm starting to wonder if there's enough value in this new generics API to warrant implementation and maintenance. For one, it's solving the same problem as v3's Bind, meaning there are now three ways to get query parameters.
> > It's also not actually taking advantage of generics. We're still eventually taking the type parameter, casting it to `any`, and doing a type switch. In effect, we're doing our own hand-rolled dynamic dispatch.
> > There are ways to avoid the type switch, but that would involve using custom types like this
> > ....
> > But then we're getting even closer to overlapping with Bind. I kind of worry about having two separate ways of doing the same thing....What do others think?
> 
> @nickajacks1 bindings always need structs to bind, but this functionality holds the bare value i would support both, as in some cases it is not worth initializing an extra struct for only one value

that makes sense, thanks

ctx.go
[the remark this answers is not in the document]
dozheiny: Sure

ctx.go
dozheiny: @efectn I made this pointer because `Convert` returns nil when reaching to error.
but you are right. I think it's better to avoid nil pointers.

ctx_test.go
```diff
@@ -4953,3 +4456,1109 @@ func Benchmark_Ctx_GetReqHeaders(b *testing.B) {
...
+	for _, test := range int8s {
+		var v int8
+		b.Run("benchmark_genericParseTypeInt8s", func(b *testing.B) {
+			b.ReportAllocs()
+			b.ResetTimer()
+
+			res := genericParseType(test.str, v)
+			require.Equal(b, test.value, res)
+		})
```
nickajacks1, line 5186: benchmarks need to have a loop. either of these:

```
	b.Run("thing1", func(b *testing.B) {
		for n := 0; n < b.N; n++ {
			// do something
		}
	})
	b.RunParallel(func(pb *testing.PB) {
		for pb.Next() {
			// do something
		}
	})

```

nickajacks1 (changes requested):

ReneWerner87:
@dozheiny can you take a look at the last comments
we need to make sure we plan this perfectly for v3, hence the many loops unfortunately

dozheiny:
> @dozheiny can you take a look at the last comments we need to make sure we plan this perfectly for v3, hence the many loops unfortunately

yeah sure.

utils.go
```diff
@@ -25,22 +25,68 @@ func queryParseDefault[V QueryType](err error, parser func() V, defaultValue ...
...
-	return queryParseDefault[V](err, func() V { return parser(result) }, defaultValue...)
+func genericParseBool[V GenericType](str string, parser func(bool) V, defaultValue ...V) V {
+	result, err := strconv.ParseBool(str)
+	return genericParseDefault[V](err, func() V { return parser(result) }, defaultValue...)
+}
+
+func genericParseType[V GenericType](str string, v V, defaultValue ...V) V {
+	switch any(v).(type) {
+	case int:
+		return genericParseInt[V](str, 32, func(i int64) V { return assertValueType[V, int](int(i)) }, defaultValue...)
```
nickajacks1, line 51: From strconv.ParseInt:
> // The bitSize argument specifies the integer type
// that the result must fit into. Bit sizes 0, 8, 16, 32, and 64
// correspond to int, int8, int16, int32, and int64.
// If bitSize is below 0 or above 64, an error is returned.

So that 32 should be changed to a 0

dozheiny: thanks

dozheiny:
Wow, only two pipelines failed; this is progress!

ctx.go
```diff
@@ -955,40 +962,46 @@ func (c *DefaultCtx) OriginalURL() string {
...
+	var v V
+	for i := range c.Route().Params {
+		if len(key) != len(c.Route().Params[i]) {
 			continue
 		}
-		if c.route.Params[i] == key || (!c.app.config.CaseSensitive && utils.EqualFold(c.route.Params[i], key)) {
+		if c.Route().Params[i] == key || (!c.App().config.CaseSensitive && utils.EqualFold(c.Route().Params[i], key)) {
 			// in case values are not here
-			if len(c.values) <= i || len(c.values[i]) == 0 {
+			if len(c.Bind().ctx.values) <= i || len(c.Bind().ctx.values[i]) == 0 {
```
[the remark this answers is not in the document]
dozheiny: I think it's a better way. I'm thinking about converting Ctx to DefaultCtx but we should handle the error(perhaps panic way).
@ReneWerner87 what do you think?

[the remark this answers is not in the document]
ReneWerner87: Yeah we can do it
@efectn any concerns about this?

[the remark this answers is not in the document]
efectn: > Yeah we can do it @efectn any concerns about this?

Sure it can be done

nickajacks1:
I'll try to do a final look over this weekend

docs/api/ctx.md
```diff
@@ -2202,3 +2203,24 @@ app.Get("/", func(c fiber.Ctx) error {
...
+## Convert 
+Converts a string value to a specified type, handling errors and optional default values.
+You can combine this function with other methods and functions.
+
+```go title="Signature"
+func Convert[T any](value string, convertor func(string) (T, error), defaultValue ...T) (*T, error)
+```
+
+```go title="Example"
+  fiber.Convert("true", strconv.ParseBool) // true, nil
```
nickajacks1, line 2216: This doc section should probably explain why someone would use fiber.Convert instead of the conversion function directly.

nickajacks1 (commented):
Nothing in particular stands out

efectn (approved):

docs/api/ctx.md
ReneWerner87: @coderabbitai can you suggest a good improvement for these documentation and also consider this comment


> This doc section should probably explain why someone would use fiber.Convert instead of the conversion function directly.

ReneWerner87: @dozheiny can you use it to improve the documentation so that we can merge directly afterwards ?

dozheiny: > @dozheiny can you use it to improve the documentation so that we can merge directly afterwards ?

sure

ReneWerner87 (approved):

Issue #2860, opened 2024-02-16
🐛 [Bug]: Healthcheck middleware doesn't work with group


ulasakdeniz:
### Bug Description

Hi, I have a problem with running healthcheck middleware with a group:

```go
package main

import (
	"github.com/gofiber/fiber/v2"
	"github.com/gofiber/fiber/v2/middleware/compress"
	"github.com/gofiber/fiber/v2/middleware/healthcheck"
	"github.com/gofiber/fiber/v2/middleware/recover"
)

func main() {
	app := fiber.New()

	app.Use(recover.New())
	app.Use(compress.New())

	group := app.Group("/app")
	group.Use(healthcheck.New())

	group.Get("/hello", func(c *fiber.Ctx) error {
		return c.SendString("Hello, World 👋!")
	})

	err := app.Listen(":5001")
	if err != nil {
		panic(err)
	}
}
```

- `localhost:5001/app/livez` doesn't work:
   ```bash
   curl localhost:5001/app/livez
   Cannot GET /app/livez
   ```
- `localhost:5001/app/hello` works

### How to Reproduce

1. Run the main file above
2. Run curl localhost:5001/app/livez
3. It returns not found.

### Expected Behavior

The endpoint should return HTTP 200

### Fiber Version

v2.52.0

### Checklist:

- [X] I agree to follow Fiber's [Code of Conduct](https://github.com/gofiber/fiber/blob/master/.github/CODE_OF_CONDUCT.md).
- [X] I have checked for existing issues that describe my problem prior to opening this one.
- [X] I understand that improperly formatted bug reports may be closed without explanation.

gaby:
@ulasakdeniz What happens if you do this instead:

```go
group.Use("/livez", healthcheck.New())
```

gaby:
`group.Use()` is expecting multiple params, and the middleware only returns a Fiber.Handler()

https://github.com/gofiber/fiber/blob/v2/group.go#L62


https://github.com/gofiber/fiber/blob/v2/middleware/healthcheck/healthcheck.go#L13

@ReneWerner87 Is this a limitation, bug, or possible feature we need to add?

gaby:
@efectn Forgot to tag you :-)

ulasakdeniz:
> @ulasakdeniz What happens if you do this instead:
> 
> ```go
> group.Use("/livez", healthcheck.New())
> ```

Hi @gaby thanks for your answer. That also doesn't work.

[...]

PR #2863, opened 2024-02-18
fix: healthcheck middleware not working with route group

luk3skyw4lker:
Fixing PR #2860 

I have changed the checking method of routes to use a regex which checks if the route ends with the specified liveness or readiness route and included tests for the strict routing mode, all passing with no issues.

As far as the benchmark, here are the results with the changes in this PR:
image

And without the changes in this PR:
image

### DIFF
--- a/middleware/healthcheck/healthcheck.go
+++ b/middleware/healthcheck/healthcheck.go
@@ -1,8 +1,7 @@
 package healthcheck
 
 import (
-	"fmt"
-	"regexp"
+	"strings"
 
 	"github.com/gofiber/fiber/v2"
 )
@@ -30,11 +29,6 @@ func healthCheckerHandler(checker HealthChecker) fiber.Handler {
 func New(config ...Config) fiber.Handler {
 	cfg := defaultConfig(config...)
 
-	baseRegexp := `\%s$`
-
-	readinessIdentifier := regexp.MustCompile(fmt.Sprintf(baseRegexp, cfg.ReadinessEndpoint))
-	livenessIdentifier := regexp.MustCompile(fmt.Sprintf(baseRegexp, cfg.LivenessEndpoint))
-
 	isLiveHandler := healthCheckerHandler(cfg.LivenessProbe)
 	isReadyHandler := healthCheckerHandler(cfg.ReadinessProbe)
 
@@ -48,11 +42,14 @@ func New(config ...Config) fiber.Handler {
 			return c.Next()
 		}
 
-		if readinessIdentifier.Match([]byte(c.Path())) {
-			return isReadyHandler(c)
-		}
+		// Find the start of the last segment
+		lastSlashIndex := strings.LastIndex(c.Path(), "/")
+		lastSegment := c.Path()[lastSlashIndex:] // +1 to skip the slash itself
 
-		if livenessIdentifier.Match([]byte(c.Path())) {
+		// Direct string comparison for the last segment of the path
+		if lastSegment == cfg.ReadinessEndpoint {
+			return isReadyHandler(c)
+		} else if lastSegment == cfg.LivenessEndpoint {
 			return isLiveHandler(c)
 		}
 
--- a/middleware/healthcheck/healthcheck.go
+++ b/middleware/healthcheck/healthcheck.go
@@ -1,9 +1,8 @@
 package healthcheck
 
 import (
-	"strings"
-
 	"github.com/gofiber/fiber/v2"
+	"github.com/gofiber/fiber/v2/utils"
 )
 
 // HealthChecker defines a function to check liveness or readiness of the application
@@ -42,15 +41,18 @@ func New(config ...Config) fiber.Handler {
 			return c.Next()
 		}
 
-		// Find the start of the last segment
-		lastSlashIndex := strings.LastIndex(c.Path(), "/")
-		lastSegment := c.Path()[lastSlashIndex:] // +1 to skip the slash itself
-
-		// Direct string comparison for the last segment of the path
-		if lastSegment == cfg.ReadinessEndpoint {
-			return isReadyHandler(c)
-		} else if lastSegment == cfg.LivenessEndpoint {
-			return isLiveHandler(c)
+		prefixCount := len(utils.TrimRight(c.Route().Path, '/'))
+		if len(c.Path()) >= prefixCount {
+			checkPath := c.Path()[prefixCount:]
+			if !c.App().Config().StrictRouting {
+				checkPath = utils.TrimRight(checkPath, '/')
+			}
+			switch checkPath {
+			case cfg.ReadinessEndpoint:
+				return isReadyHandler(c)
+			case cfg.LivenessEndpoint:
+				return isLiveHandler(c)
+			}
 		}
 
 		return c.Next()
--- a/middleware/healthcheck/healthcheck.go
+++ b/middleware/healthcheck/healthcheck.go
@@ -48,7 +48,7 @@ func New(config ...Config) fiber.Handler {
 			if !c.App().Config().StrictRouting {
 				checkPathTrimmed = utils.TrimRight(checkPath, '/')
 			}
-			switch true {
+			switch {
 			case checkPath == cfg.ReadinessEndpoint || checkPathTrimmed == cfg.ReadinessEndpoint:
 				return isReadyHandler(c)
 			case checkPath == cfg.LivenessEndpoint || checkPathTrimmed == cfg.LivenessEndpoint:
--- a/middleware/healthcheck/healthcheck.go
+++ b/middleware/healthcheck/healthcheck.go
@@ -1,6 +1,9 @@
 package healthcheck
 
 import (
+	"fmt"
+	"regexp"
+
 	"github.com/gofiber/fiber/v2"
 )
 
@@ -27,6 +30,11 @@ func healthCheckerHandler(checker HealthChecker) fiber.Handler {
 func New(config ...Config) fiber.Handler {
 	cfg := defaultConfig(config...)
 
+	baseRegexp := `\%s$`
+
+	readinessIdentifier := regexp.MustCompile(fmt.Sprintf(baseRegexp, cfg.ReadinessEndpoint))
+	livenessIdentifier := regexp.MustCompile(fmt.Sprintf(baseRegexp, cfg.LivenessEndpoint))
+
 	isLiveHandler := healthCheckerHandler(cfg.LivenessProbe)
 	isReadyHandler := healthCheckerHandler(cfg.ReadinessProbe)
 
@@ -40,10 +48,11 @@ func New(config ...Config) fiber.Handler {
 			return c.Next()
 		}
 
-		switch c.Path() {
-		case cfg.ReadinessEndpoint:
+		if readinessIdentifier.Match([]byte(c.Path())) {
 			return isReadyHandler(c)
-		case cfg.LivenessEndpoint:
+		}
+
+		if livenessIdentifier.Match([]byte(c.Path())) {
 			return isLiveHandler(c)
 		}
 
--- a/middleware/healthcheck/healthcheck.go
+++ b/middleware/healthcheck/healthcheck.go
@@ -44,13 +44,14 @@ func New(config ...Config) fiber.Handler {
 		prefixCount := len(utils.TrimRight(c.Route().Path, '/'))
 		if len(c.Path()) >= prefixCount {
 			checkPath := c.Path()[prefixCount:]
+			checkPathTrimmed := checkPath
 			if !c.App().Config().StrictRouting {
-				checkPath = utils.TrimRight(checkPath, '/')
+				checkPathTrimmed = utils.TrimRight(checkPath, '/')
 			}
-			switch checkPath {
-			case cfg.ReadinessEndpoint:
+			switch true {
+			case checkPath == cfg.ReadinessEndpoint || checkPathTrimmed == cfg.ReadinessEndpoint:
 				return isReadyHandler(c)
-			case cfg.LivenessEndpoint:
+			case checkPath == cfg.LivenessEndpoint || checkPathTrimmed == cfg.LivenessEndpoint:
 				return isLiveHandler(c)
 			}
 		}
--- a/middleware/healthcheck/healthcheck_test.go
+++ b/middleware/healthcheck/healthcheck_test.go
@@ -33,8 +33,8 @@ func Test_HealthCheck_Group_Default(t *testing.T) {
 
 	app := fiber.New()
 	app.Group("/v1", New())
-	v2 := app.Group("/v2/")
-	customer := v2.Group("/customer/")
+	v2Group := app.Group("/v2/")
+	customer := v2Group.Group("/customer/")
 	customer.Use(New())
 
 	req, err := app.Test(httptest.NewRequest(fiber.MethodGet, "/v1/readyz", nil))
--- a/middleware/healthcheck/healthcheck_test.go
+++ b/middleware/healthcheck/healthcheck_test.go
@@ -10,6 +10,50 @@ import (
 	"github.com/valyala/fasthttp"
 )
 
+func Test_HealthCheck_Strict_Routing_Default(t *testing.T) {
+	t.Parallel()
+
+	app := fiber.New(fiber.Config{
+		StrictRouting: true,
+	})
+
+	app.Use(New())
+
+	req, err := app.Test(httptest.NewRequest(fiber.MethodGet, "/v1/readyz", nil))
+	utils.AssertEqual(t, nil, err)
+	utils.AssertEqual(t, fiber.StatusOK, req.StatusCode)
+
+	req, err = app.Test(httptest.NewRequest(fiber.MethodGet, "/v1/livez", nil))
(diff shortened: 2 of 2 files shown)


luk3skyw4lker:
@ReneWerner87 @gaby could you guys take a look at this for me? I don't think I have messed up the benchmark that badly but maybe something else can be worked out.

gaby:
@luk3skyw4lker I can see regex making performance drop substantially

gaby:
Yeah, this definitely needs a different approach between Sprintf + Regex it will make the middleware super slow.

luk3skyw4lker:
@gaby the only other thing I could think of is splitting the route by slashes and checking the last route to see if it's either one of the set up routes. I don't see any other way out of this

ReneWerner87:
https://github.com/gofiber/fiber/issues/2860#issuecomment-1949132294

```go
prefixCount := len(utils.TrimRight(c.Route().Path, '/'))
if len(c.Path()) >= prefixCount {
	switch c.Path()[prefixCount:] {
	case cfg.ReadinessEndpoint:
		return isReadyHandler(c)
	case cfg.LivenessEndpoint:
		return isLiveHandler(c)
	}
}
```
Have you ever tested this?

gaby:
> [#2860 (comment)](https://github.com/gofiber/fiber/issues/2860#issuecomment-1949132294)
> 
> ```go
> prefixCount := len(utils.TrimRight(c.Route().Path, '/'))
> if len(c.Path()) >= prefixCount {
> 	switch c.Path()[prefixCount:] {
> 	case cfg.ReadinessEndpoint:
> 		return isReadyHandler(c)
> 	case cfg.LivenessEndpoint:
> 		return isLiveHandler(c)
> 	}
> }
> ```
> 
> Have you ever tested this?

I was about to comment something similar:

```go
        // Extract the last segment of the path
        pathSegments := strings.Split(c.Path(), "/")
        lastSegment := pathSegments[len(pathSegments)-1]

        // Direct string comparison for the last segment of the path
        if lastSegment == strings.TrimPrefix(cfg.ReadinessEndpoint, "/") {
            return isReadyHandler(c)
        } else if lastSegment == strings.TrimPrefix(cfg.LivenessEndpoint, "/") {
            return isLiveHandler(c)
        }
```

luk3skyw4lker:
@ReneWerner87 yours fail in the strict routing test but @gaby solution passed it no prob, benchmark results:

image

luk3skyw4lker:
It was kind of what I wanted to do at first, but maybe I misunderstood @ReneWerner87 saying he wanted a different approach than that

gaby:
> @ReneWerner87 yours fail in the strict routing test but @gaby solution passed it no prob, benchmark results:
> 
> image

Hmmm that still drops performance in half. The ns/op went from 100 to 200 ns/op

ReneWerner87:
then let's find a way without allocations, have an idea

luk3skyw4lker:
I managed to extract 7 ns/op of @gaby solution by using switch case, let me see if I can do more

gaby:
@luk3skyw4lker Try this:

```go
// Find the start of the last segment
lastSlashIndex := strings.LastIndex(c.Path(), "/")
lastSegment := c.Path()[lastSlashIndex+1:] // +1 to skip the slash itself

// Direct string comparison for the last segment of the path
if lastSegment == cfg.ReadinessEndpoint {
    return isReadyHandler(c)
} else if lastSegment == cfg.LivenessEndpoint {
    return isLiveHandler(c)
}
```

luk3skyw4lker:
@gaby without the +1 looks better, because then we wouldn't have to trim the already set up config to remove the slashes, passed it fine and it looks very good for me:
image

ReneWerner87:
> @luk3skyw4lker Try this:
> 
> ```go
> // Find the start of the last segment
> lastSlashIndex := strings.LastIndex(c.Path(), "/")
> lastSegment := c.Path()[lastSlashIndex+1:] // +1 to skip the slash itself
> 
> // Direct string comparison for the last segment of the path
> if lastSegment == cfg.ReadinessEndpoint {
>     return isReadyHandler(c)
> } else if lastSegment == cfg.LivenessEndpoint {
>     return isLiveHandler(c)
> }
> ```
> 
> Try with and without "+1"

same idea 
```go

	cfg := defaultConfig(config...)

	isLiveHandler := healthCheckerHandler(cfg.LivenessProbe)
	isReadyHandler := healthCheckerHandler(cfg.ReadinessProbe)

		if cfg.Next != nil && cfg.Next(c) {
			return c.Next()
		}

			return c.Next()
		}

		// Extract the last segment of the path
		slashPos := strings.LastIndex(c.Path(), "/")
		lastSegment := c.Path()[slashPos+1:]

		// Direct string comparison for the last segment of the path
		if lastSegment == utils.TrimLeft(cfg.ReadinessEndpoint, '/') {
			return isReadyHandler(c)
		} else if lastSegment == utils.TrimLeft(cfg.LivenessEndpoint, '/') {
			return isLiveHandler(c)
		}

		return c.Next()
	}
}
```

gaby:
> @gaby without the +1 looks better, because then we wouldn't have to trim the already set up config to remove the slashes, passed it fine and it looks very good for me: image

Awesome, is that using the Switch ? + lastindex?

ReneWerner87:
why is this test passing?
image

luk3skyw4lker:
@gaby using if and no +lastIndex

luk3skyw4lker:
Let me commit so you guys can check it out

luk3skyw4lker:
Using switch in @gaby solution makes it go back to around 190 ns/op

gaby:
> Using switch in @gaby solution makes it go back to around 190 ns/op

Wow!

luk3skyw4lker:
@ReneWerner87 I think I wrote it badly, let me check it again

ReneWerner87:
> @ReneWerner87 I think I wrote it badly, let me check it again

yes but this is a good test, we should leave something like this in, because this path should not match

luk3skyw4lker:
@ReneWerner87 I have just added a test for livez with a final slash (/livez/) and it is indeed not matching. I'll commit it right now so you can check it out

gaby:
@luk3skyw4lker We definitely need a test that checks both `/livez` and `/livez/`.

luk3skyw4lker:
@gaby I have just added this case, checking for /livez/ and /readyz/

ReneWerner87:
we have to check the pre-route as well, otherwise every suffix would match the path

please also add negative tests

let me create some briefly

luk3skyw4lker:
@ReneWerner87 that's something I have talked about in the issue, every route with /livez and /readyz will be matched in case of using the middleware globally.

Also, if a user pass a nested path as a endpoint, the actual structure would fail in running the middleware, since we're only checking the suffix

luk3skyw4lker:
In this case the nested path would work actually, but only if the user placed the middleware into a route group

ReneWerner87:
i added some more cases
https://github.com/gofiber/fiber/pull/2863/commits/1b1bc46432d0907e44579ff87d42ee87c4c6ac96

so you can work test driven

luk3skyw4lker:
Alright, I'll be checking out ways to improve the route matching globally, thanks for all the help! @ReneWerner87 and @gaby 

I'll see if I can work something out by today, if not, I'll leave an update here

ReneWerner87:
@luk3skyw4lker can you benchmark and verify my last commit
https://github.com/gofiber/fiber/pull/2863/commits/0b92ab881af80d13fbe0b526b9d18a06969fc803

luk3skyw4lker:
@ReneWerner87 all works and with good benchmark results:
image

ReneWerner87:
do you( @efectn @gaby @luk3skyw4lker ) have any comments or other test cases you would like to add?

luk3skyw4lker:
@ReneWerner87 I have no other ideas, I think that's it, LGTM

luk3skyw4lker:
actually, I think it would be nice to add a nested route test, like if a user passes a /probe/livez route in the liveness or readiness endpoint, what do you think?

luk3skyw4lker:
@ReneWerner87 

We have two options for this:

1. Improve the route matching and support nested routes in the config of the middleware
2. Add in the middleware docs instructions saying that nested routes are not allowed, only one level routes

I think that improving the route matching part will be quite hard to do, but maybe if we go to another PR to do this it could be possible.

luk3skyw4lker:
This should be the test case for this:
```go
func Test_HealthCheck_Custom_Nested(t *testing.T) {
	t.Parallel()

	app := fiber.New()

	c1 := make(chan struct{}, 1)
	go func() {
		time.Sleep(1 * time.Second)
		c1 <- struct{}{}
	}()

	app.Use(New(Config{
		LivenessProbe: func(c *fiber.Ctx) bool {
			return true
		},
		LivenessEndpoint: "/probe/live",
		ReadinessProbe: func(c *fiber.Ctx) bool {
			select {
			case <-c1:
				return true
			default:
				return false
			}
		},
		ReadinessEndpoint: "/probe/ready",
	}))

	shouldWork(t, app, "/probe/livez")
	shouldWork(t, app, "/probe/readyz")
	shouldNotWork(t, app, "/probe/livez/")
	shouldNotWork(t, app, "/probe/readyz/")
	shouldNotWork(t, app, "/livez")
	shouldNotWork(t, app, "/readyz")
	shouldNotWork(t, app, "/readyz/")
	shouldNotWork(t, app, "/livez/")
}
```

efectn:
> do you( @efectn @gaby @luk3skyw4lker ) have any comments or other test cases you would like to add?

Will check the code tomorrow

ReneWerner87:
have added some cases
unfortunately i realized that this kind of matching is not perfect, because normally we should allow all router possibilities
f.e.
```go
app.Use("*", New())
app.Use("/bla/:param/", New())
...
```
I would solve this in version 3

ReneWerner87:
in v3 i will do something like this

```go
	app.Use("/livez", NewLivenessCheck(Config{
		Next: func(c *fiber.Ctx) bool {
			return true // somthing
		},
	}))
	app.Use(DefaultLivenessEndpoint, NewLivenessCheck(Config{
		Next: func(c *fiber.Ctx) bool {
			return true // somthing
		},
	}))
```

routing stuff should be done by the router and not the middleware itself

luk3skyw4lker:
@ReneWerner87 that would be perfect! I thought about doing that too inside the middleware, but analyzing the framework structure made me realize that it would be complicated.

If you need any help in v3 I could chip in too, it wouldn't be too much. I have more free time now that I have solved some work related problems.

ReneWerner87:
Can you pls benchmark the last changes? I got totally different numbers

ReneWerner87 (approved):

luk3skyw4lker:
@ReneWerner87 I runned it twice and it looks fine:
image
image

middleware/healthcheck/healthcheck_test.go
```diff
@@ -10,6 +10,50 @@ import (
...
+func Test_HealthCheck_Strict_Routing_Default(t *testing.T) {
+	t.Parallel()
+
+	app := fiber.New(fiber.Config{
+		StrictRouting: true,
+	})
+
+	app.Use(New())
+
+	req, err := app.Test(httptest.NewRequest(fiber.MethodGet, "/v1/readyz", nil))
```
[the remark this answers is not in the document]
luk3skyw4lker: This is a leftover comment from a previous unfinished review right? Because lines 19 to 22 don't look like that anymore

luk3skyw4lker:
@ReneWerner87 those helpers could be moved to a helpers package after, they could be used by tests all around the framework. Maybe this could be an improvement to v3 too, for a better developer experience while contributing.

middleware/healthcheck/healthcheck_test.go
```diff
@@ -1,28 +1,120 @@
...
+	v2Group := app.Group("/v2/")
+	customer := v2Group.Group("/customer/")
+	customer.Use(New())
+
+	v3Group := app.Group("/v3/")
+	v3Group.Group("/todos/", New(Config{ReadinessEndpoint: "/readyz/", LivenessEndpoint: "/livez/"}))
+
+	shouldGiveOK(t, app, "/v1/readyz")
+	shouldGiveOK(t, app, "/v1/livez")
+	shouldGiveOK(t, app, "/v1/readyz/")
```
gaby, line 61: Why is this OK?

middleware/healthcheck/healthcheck_test.go
```diff
@@ -1,28 +1,120 @@
...
+	shouldGiveOK(t, app, "/v1/livez/")
```
gaby, line 62: Same here

middleware/healthcheck/healthcheck_test.go
```diff
@@ -1,28 +1,120 @@
...
+	shouldGiveOK(t, app, "/v2/customer/readyz")
+	shouldGiveOK(t, app, "/v2/customer/livez")
+	shouldGiveOK(t, app, "/v2/customer/readyz/")
```
gaby, line 65: Same here

middleware/healthcheck/healthcheck_test.go
```diff
@@ -1,28 +1,120 @@
...
+	shouldGiveOK(t, app, "/v2/customer/livez/")
```
gaby, line 66: Here too

gaby (commented):
Shouldnt these be "Not found"?

middleware/healthcheck/healthcheck.go
```diff
@@ -40,11 +41,19 @@ func New(config ...Config) fiber.Handler {
...
-		case cfg.ReadinessEndpoint:
-			return isReadyHandler(c)
-		case cfg.LivenessEndpoint:
-			return isLiveHandler(c)
+		prefixCount := len(utils.TrimRight(c.Route().Path, '/'))
+		if len(c.Path()) >= prefixCount {
+			checkPath := c.Path()[prefixCount:]
+			checkPathTrimmed := checkPath
+			if !c.App().Config().StrictRouting {
+				checkPathTrimmed = utils.TrimRight(checkPath, '/')
```
gaby, line 49: For Strict Routing we are calling TrimRight twice. Once here and once in line 44

middleware/healthcheck/healthcheck.go
```diff
@@ -40,11 +41,19 @@ func New(config ...Config) fiber.Handler {
 			return c.Next()
 		}
 
-		switch c.Path() {
-		case cfg.ReadinessEndpoint:
-			return isReadyHandler(c)
-		case cfg.LivenessEndpoint:
-			return isLiveHandler(c)
+		prefixCount := len(utils.TrimRight(c.Route().Path, '/'))
```
gaby, line 44: Why do we use c.Route().Path here but c.Path() below?

ReneWerner87: because we want to subtract the current route path from the request route

middleware/healthcheck/healthcheck.go
ReneWerner87: right, because we want to subtract the slash once from the request path and once from the route path

middleware/healthcheck/healthcheck_test.go
ReneWerner87: as long as strict routing is not activated, you can create the request url with or without slash, both are valid

middleware/healthcheck/healthcheck_test.go
ReneWerner87: as long as strict routing is not activated, you can create the request url with or without slash, both are valid

middleware/healthcheck/healthcheck_test.go
ReneWerner87: as long as strict routing is not activated, you can create the request url with or without slash, both are valid

middleware/healthcheck/healthcheck_test.go
ReneWerner87: as long as strict routing is not activated, you can create the request url with or without slash, both are valid

ReneWerner87:
@gaby can you check my last comments and resolve the conversations

gaby:
> @gaby can you check my last comments and resolve the conversations

Looks good 💪

gaby (approved):

middleware/healthcheck/healthcheck.go
gaby: I see it now, i was confused for a second

ReneWerner87:
> in v3 i will do something like this
> 
> ```go
> 	app.Use("/livez", NewLivenessCheck(Config{
> 		Next: func(c *fiber.Ctx) bool {
> 			return true // somthing
> 		},
> 	}))
> 	app.Use(DefaultLivenessEndpoint, NewLivenessCheck(Config{
> 		Next: func(c *fiber.Ctx) bool {
> 			return true // somthing
> 		},
> 	}))
> ```
> 
> routing stuff should be done by the router and not the middleware itself

@luk3skyw4lker if you want, you can support with a PR for this change
Target is "v2ToV3Merge" branch (merge is not completely done, but the base is the pr code https://github.com/gofiber/fiber/pull/2864 )

luk3skyw4lker:
@ReneWerner87 Will do! I'll be checking it out this week and see if I can open a PR already. Thanks for the suggestion and heads up!

Issue #2854, opened 2024-02-12
📝 [v3 Proposal]: Koa Style `ctx.Request` and `ctx.Response` Objects


nickajacks1:
### Feature Proposal Description

Currently, the Ctx object handles almost all of the work that is performed in a Handler function. This can potentially lead to confusion about what parts of the API interact with the request and what parts interact with the response (eg `Cookie` vs `Cookies`).

Previously, we decided against eliminating the Ctx object entirely (#2764), but at the tail end of the discussion in Discord, it was brought up that we could avoid almost all breaking changes by following what Koa does: put the core implementation in Request and Response objects and have a generic Ctx API which provides convenient aliases for the request+response APIs.

In other words, the existing Ctx object stays exactly as-is, but ctx.Response() and ctx.Request() are reworked to hold more of the core functionality. As alluded to above, this allows users to more easily reason about what parts of the API touch the request vs response.

I am creating this proposal to attempt to take that idea and arrive at a concrete conclusion for whether we would like to implement this or not for v3 and beyond.

More reading: https://koajs.com/

### Alignment with Express API

While inspired by Koa, the approach of splitting Ctx into distinct Request and Response objects brings Fiber closer to Express's res and req centered APIs.

### HTTP RFC Standards Compliance

The proposal is not relevant to HTTP RFCs.

### API Stability

This would be a breaking change, as it would change `Ctx.Request()` and `Ctx.Response()` to no longer directly return the underlying fasthttp Request and Response objects.  Consequently, it will need to be implemented during a major version bump.
The only way around this would be to make the new Request and Response objects compose the fasthttp objects, but that might prove to be a clumsy solution.

Having a clear separation between Request and Response APIs allows for a more idiomatic design. With the current Ctx-only design, we already have examples of collisions between request-based and response-based actions.  For example, `Ctx.Get` and `Ctx.GetResponseHeader`.

### Feature Examples

```go
package main

import (
	"github.com/gofiber/fiber/v3"
)

func main() {
	app := fiber.New()

	// This syntax is still valid under the proposed changes
	app.Get("/old-stuff", func(c fiber.Ctx) error {
		if c.Accepts("text/plain") == "" {
			return c.SendStatus(fiber.StatusNotAcceptable)
		}

		name := c.Cookies("name")

		cookie := new(fiber.Cookie)
		cookie.Name = "userId"
		cookie.Value = "1234"
		c.Cookie(cookie)

		// ...

		return c.SendString("Hello " + name)
	})

[...]

PR #2892, opened 2024-03-03
🧹 v3 (Maintenance): Updates to Client and Proxy Unit-Tests

gaby:
## Description

- Remove usage of `t.Parallel()` for client and proxy middleware code.
- Create helper function for running `app.Listen()` that checks error and panics in case of failure.
- Remove usage of `require.NoError()` in client, and proxy code. This is no longer recommended since it triggers race conditions.
- When running locally every 4-5 tries the `Test_Listen_Graceful_Shutdown` test will fail because it expects HTTP 200 but gets a 0. This only happens when running ALL the tests together. When running only the `listen_test.go` file this doesn't happen (+10 tries).

## Type of change

Please delete options that are not relevant.

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Enhancement (improvement to existing features and functionality)
- [ ] Documentation update (changes to documentation)
- [ ] Performance improvement (non-breaking change which improves efficiency)
- [x] Code consistency (non-breaking change which improves code reliability and robustness)

PR #1986, opened 2022-07-27
✨ v3 (feature): client refactor

wangjq4214:
# Aims

> **Please provide enough information so that others can review your pull request:**

Move `client.go` to `client/` to make project structure more tidy.
Re-create client by inspiring by [go-resty/resty](https://github.com/go-resty/resty) or [axios/axios](https://github.com/axios/axios).

# Details

> **Explain the *details* for making this change. What existing problem does the pull request solve?**

This PR is related to #1906.

## TODO List

- [X] Core request process, hook and plugin mechanism
- [X] URL and baseurl merge mechanism
- [X] Header setting
- [X] Query parameter setting
- [X] User agent setting
- [X] Cookie setting
- [x] Path parameter setting
- [x] Form data support
- [x] JSON body, XML body, Raw body, Files body
- [x] Refer setting
- [x] Unmarshal
- [x] Chain API (processing)
- [x] API like axios
- [x] Timeout
- [x] Response save to file or io.Writer
- [x] TLS support
- [x] Proxy
- [x] Retry (https://github.com/gofiber/fiber/pull/1972) (Thanks for it, I'll continue this work)
- [x] Throttle
- [x] Extending the axios-like API options
- [x] Redirect
- [x] Improve test coverage (95% up, 96.7% now)
- [x] Add benchmarks
- [x] Cookie jar 
- [x] Logger (A simple impl, may be improve)
- [x] ~~Trace~~ (Need PR for fasthttp)
- [ ] Add API docs and examples

## Summary by CodeRabbit

- **New Features**
	- Introduced a comprehensive `Client` struct in the Fiber Client for enhanced request handling, including support for custom headers, parameters, cookies, timeouts, user agents, TLS configurations, and more.
	- Added robust HTTP request and response handling capabilities, including methods for GET, POST, PUT, DELETE, etc., with extensive configuration options.
	- Implemented new test functions for validating HTTP response attributes, ensuring reliability in response parsing and saving.

- **Refactor**
	- Refined core client functionalities for improved request and response processing, including error handling, retries, and timeouts.

PR #2894, opened 2024-03-03
🔥 Feature: Add Req and Res API

nickajacks1:
## CURRENT STATUS

- [x] Response
  - [x] Ctx.Request => Ctx.Req
  - [x] Complete fiber.Request API
  - [ ] Create Request Interface with a DefaultRequest implementation 
  - [x] Create Doc Page
- [x] Response API
  - [x] Ctx.Response => Ctx.Res
  - [x] Complete fiber.Response API
  - [ ] Create Response Interface with a DefaultResponse implementation
  - [x] Create Doc Page
## Description

Split the existing Ctx API into two separate APIs for Requests and Responses. Having distinct APIs for Request and Response types will allow developers to more easily write self-documenting code by avoiding ambiguous methods such as `Ctx.Cookie` and `Ctx.Cookies`. Much of the existing Ctx API simply calls the corresponding Req or Res API. For example, `Ctx.Get` calls `Request.Get` and `Ctx.Set` calls `Response.Set`.

Because Express defines only res and req APIs with no context API, this change will make it easier to manage Fiber's adherence to Express's APIs.  In addition, it will allow us to avoid needing to create certain one-off methods.  For example, `Ctx.GetRespHeader` can now become `Ctx.Res().Get`

Although the majority of the Ctx API is unchanged, `Ctx.Request` and `Ctx.Response` have been removed in favor of `Ctx.Req` and `Ctx.Res`.  Now, instead of a raw `fasthttp.Request` or `fasthttp.Response`, we return `fiber.Request` and `fiber.Response` objects, which implement their respective APIs.  As such, this is a breaking change for users who access `Ctx.Request` and `Ctx.Response` direct.

Inspired by [Koa](https://koajs.com/) and fasthttp, both of which have individual request/response objects and a single context object whose methods are simply aliases for corresponding methods on the req/res APIs.

## Deprecation:

* `Ctx.GetRespHeader` is replaced by `Ctx.Res().Get`

Related to #2854

## Changes introduced

## Type of change

## Checklist

- [ ] Conducted a self-review of the code and provided comments for complex or critical parts.
- [ ] Ensured that new and existing unit tests pass locally with the changes.


client.go
```diff
@@ -19,14 +19,16 @@ import (
 	"github.com/valyala/fasthttp"
 )
 
+// XXX TODO FIXME had to comment out the below types
```
gaby, line 22: I wouldnt worry about client.go until #1986 is mergef

gaby:
@nickajacks1 Just so you are aware, the tests are very unstable/flaky right now in `main`. This is because we no longer retry failed tests multiple times on the CI. This is being fixing in #2892

nickajacks1:
3/9: Moved some more methods from Ctx to Request. Rebased.

nickajacks1:
3/17: Rebase. Moved more trivial methods to Request.

nickajacks1:
Hi @gofiber/maintainers 

Sorry this is taking so long.
I've reached the first major milestone - the Request API is fully moved over to the Request type.
I'd like to get a health check - does the overall direction still seem OK?  If so, I suspect that moving the Response API over will be comparatively quick.
None of the unit tests have been moved over.  IMO that's quite a bit of work and shouldn't hold up v3 since it's not public facing and can be done at any time.

nickajacks1:
Even though Express has Locals as part of the [res API](https://expressjs.com/en/api.html#res.locals), I feel like it makes more sense to keep it as part of Ctx since it doesn't inherently have anything to do with the response (or request for that matter). I'm going to leave in Ctx for now.

ReneWerner87:
ok, sounds reasonable we can always adapt it later

nickajacks1:
Ok, the Response API is fully moved over.

Some remaining work:
I didn't move Redirect to the Response API yet. The Redirect API has some dependencies on DefaultCtx that requires more thought. Moving Redirect to Res won't break the API in any way, so we can take our time thinking about this.

Next week I'll start on either the docs or the interface part.

nickajacks1:
I was thinking, is it actually that useful to make Request and Response interfaces if Ctx is already an interface and can thus be expanded instead of Request/Response? If we say that custom extensions should just go through Ctx, that might simplify both the API and implementation a bit.
Thoughts @gofiber/maintainers ?

sixcolors:
> I was thinking, is it actually that useful to make Request and Response interfaces if Ctx is already an interface and can thus be expanded instead of Request/Response? If we say that custom extensions should just go through Ctx, that might simplify both the API and implementation a bit. Thoughts @gofiber/maintainers ?

I am interested in learning more about the approach of expanding the `Ctx` interface for custom extensions, instead of introducing separate `Request` and `Response` interfaces. Could you please explain how this approach works and how it simplifies the API for developers?

In addition, can you help us by merging the changes added to the main branch, and updating your PR to use the new Bind() interface?

Thank you, great work so far, and I look forward to your reply!

nickajacks1:
> I am interested in learning more about the approach of expanding the Ctx interface for custom extensions, instead of introducing separate Request and Response interfaces. Could you please explain how this approach works and how it simplifies the API for developers?

The two primary points I have in mind are:

1. Any extension of behavior can already be handled by using a custom Ctx since the split implementation is at its core a kind of syntactic sugar. I have not been able to think of a use case for Req/Res interfaces that can't be handled with the Ctx interface.
2. Since the `Ctx` and `Request`+`Response` APIs have overlap, it may be less clear to developers which they should implement. If they want to update `SendString` to tee out some logs, should they do it by making a custom `Ctx` or a custom `Respose`? I feel like having _either_ `Ctx` _or_ `Request`+`Response` be interfaces may be more straightforward (and I lean toward `Ctx`). This is more what I meant when I implied that having all three of `Ctx`, `Request`, and `Response` be interfaces could make the API more complex.

> In addition, can you help us by merging the changes added to the main branch, and updating your PR to use the new Bind() interface?

ye. will also probably be able to mark the PR ready for review soon.

nickajacks1:
Marking as ready for review. If it is decided to make Request and Response interfaces, I will update
Regarding the docs, I think it might be better to get this in first and then write up the docs.

nickajacks1:
Finished docs in time, so I added them to this PR after all.
I'd like some criticism on how I updated the docs. I took the approach of keeping everything on a single page and using subsections for Request and Response methods.

I'm not sure what the best way to document the whole "`c.Send()` is the same as `c.Res().Send()`" thing is. I basically just sprinkled both forms in the examples, but I wonder if that would be confusing to newcomers.
I put a big green Tip box at the top to highlight it: 
chrome_8rQ4swzyeG
I didn't individually list the Ctx methods that simply call a Req/Res method since I felt that would get overly verbose.

gaby:
@nickajacks1 How about a table that shows both styles?

efectn:
> > I am interested in learning more about the approach of expanding the Ctx interface for custom extensions, instead of introducing separate Request and Response interfaces. Could you please explain how this approach works and how it simplifies the API for developers?
> 
> The two primary points I have in mind are:
> 
> 1. Any extension of behavior can already be handled by using a custom Ctx since the split implementation is at its core a kind of syntactic sugar. I have not been able to think of a use case for Req/Res interfaces that can't be handled with the Ctx interface.
> 2. Since the `Ctx` and `Request`+`Response` APIs have overlap, it may be less clear to developers which they should implement. If they want to update `SendString` to tee out some logs, should they do it by making a custom `Ctx` or a custom `Respose`? I feel like having _either_ `Ctx` _or_ `Request`+`Response` be interfaces may be more straightforward (and I lean toward `Ctx`). This is more what I meant when I implied that having all three of `Ctx`, `Request`, and `Response` be interfaces could make the API more complex.
> 

I'm not sure about making Req and Resp not an interface. We're still able to update Ctx methods but it won't affect response for example. And it may cause some confusions as well. For example, if i've updated SendString, it means i'll have different behaving resp and ctx methods

nickajacks1:
@efectn 
> For example, if i've updated SendString, it means i'll have different behaving resp and ctx methods

Yeah, but this would be a problem whether or not Response is an interface. If response is an interface and you only override Ctx.SendString, you still have different behavior between Response and Ctx. Gonna have to think about that one some more since we'll want to address that either way.

Changing the subject a little...If Ctx implements both Request and Response, there's a corner case trap developers might hit if both Request and Response share method signatures, such as Get:
```go
printResponseContentType := func(r Response) {
	fmt.Println(r.Get("Content-Type"))
}
printResponseContentType(c) // this prints the request's header, not the response's
```
The only ways I've thought of to get around this so far are to explicitly add methods to Request and Response that Ctx does not implement OR enforce that Request and Response do not have overlapping methods (which eliminates one of the benefits of splitting the APIs).  I'll brainstorm for clean ways to deal with it, but feel free to spitball ideas.

nickajacks1:
RE: overriding a method in Ctx but not Req/Res, resulting in differing behavior between e.g., `c.Get` and `c.Req().Get`, I fear the only real options to prevent/mitigate this that I could come up with may be to 
- Establish a clear direction for how we want the usage of interfaces to go and provide a detailed guide for interface usage (which we should make anyway) that lists ways you can shoot yourself in the foot. This unfortunately still leaves the door open for people to not read the guide and still shoot themselves in the foot.
- Do not provide aliases in Ctx, which seems heavy handed and unpopular.
- Do not make Ctx an interface, which may not be an option at this stage.

I imagine the last two options are non-starters, so I will focus on the first.
If the user overrides the Req or Res method instead of the Ctx method, everything is hunky dory. If they override Ctx methods that are also provided by Res or Res, we hit the case where c.Get and c.Req().Get do different things. This means that the undesirable case is only avoidable by making Req and Res interfaces. A relatively safe rule of thumb is probably that users should **prefer to override at the Req/Res level if possible, falling back to Ctx if the situation requires it**.

efectn:
> RE: overriding a method in Ctx but not Req/Res, resulting in differing behavior between e.g., `c.Get` and `c.Req().Get`, I fear the only real options to prevent/mitigate this that I could come up with may be to
> 
> * Establish a clear direction for how we want the usage of interfaces to go and provide a detailed guide for interface usage (which we should make anyway) that lists ways you can shoot yourself in the foot. This unfortunately still leaves the door open for people to not read the guide and still shoot themselves in the foot.
> * Do not provide aliases in Ctx, which seems heavy handed and unpopular.
> * Do not make Ctx an interface, which may not be an option at this stage.
> 
> I imagine the last two options are non-starters, so I will focus on the first. If the user overrides the Req or Res method instead of the Ctx method, everything is hunky dory. If they override Ctx methods that are also provided by Res or Res, we hit the case where c.Get and c.Req().Get do different things. This means that the undesirable case is only avoidable by making Req and Res interfaces. A relatively safe rule of thumb is probably that users should **prefer to override at the Req/Res level if possible, falling back to Ctx if the situation requires it**.

Maybe it's better if we don't make Req and Resp an interface

app_test.go
```diff
@@ -981,8 +981,8 @@ func Test_App_Static_Custom_CacheControl(t *testing.T) {
 	app := New()
 
 	app.Static("/", "./.github", Static{ModifyResponse: func(c Ctx) error {
-		if strings.Contains(c.GetRespHeader("Content-Type"), "text/html") {
-			c.Response().Header.Set("Cache-Control", "no-cache, no-store, must-revalidate")
+		if strings.Contains(c.Res().Get("Content-Type"), "text/html") {
+			c.Set("Cache-Control", "no-cache, no-store, must-revalidate")
```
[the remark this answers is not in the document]
renanbastos93: Here we could use constant `MIMETextHTML`

efectn:
Hi @nickajacks1 are there any update?

ReneWerner87:
definitely thanks for the PR

[ ] - an update with the current master code would still be good
[ ] - documentation markdown files

what if you leave the functionality in the ctx struct and the request response structs only refer to those of the ctx

then you could extend the functionality with the custom ctx and also inject your own request and response classes
I don't think the interface for request and response is bad at all

then you would only need a setResponse and setRequest function in the context

and the subclasses then get the context injected

with this behavior we would be able to logically separate the functionalities for the users in the 2 directions without moving the code and in my opinion we would have the best of both worlds

ReneWerner87:
@coderabbitai what do you think about it and how should we implement it
please give us your opinion on several implementation options with advantages and disadvantages so that we can compare and apply the best possible concept

nickajacks1:
@ReneWerner87  what do you think the usage should be like for customizing the request and response?

I'm thinking about the following cases:
1. customizing only the Ctx
2. customizing only the Request or Response
3. customizing each of the Ctx, Request, and Response

Where do you see `setResponse` and `setRequest` being called?

ReneWerner87:
Will try to answer tomorrow

nickajacks1:
I've updated the PR with the more conservative approach that @ReneWerner87 suggested that keeps the implementation in Ctx.

ReneWerner87:
> I've updated the PR with the more conservative approach that @ReneWerner87 suggested that keeps the implementation in Ctx.

thank you, is there anything else open here apart from the review?

nickajacks1:
I was thinking about how to best update the docs. Right now the Ctx page has every method in alphabetical order. Perhaps it could be reordered a bit, something roughly like this

```text
# Ctx

## Request
These methods operate on the incoming request blah blah

### Accepts
blah

...

## Response
These methods operate on the outgoing request blah blah

### Append
blah
```

Also, I was thinking that a future improvement could be to maybe auto generate req.go and res.go. I handwrote them, but it might not be that hard to write a tiny program to generate them based on some tag in the comments of Ctx methods or something. That could help prevent req and res from getting out of sync with Ctx.

Other than those two things, in my opinion, there isn't much of a point of writing unit tests for the entire Req and Res APIs. Maybe just a couple that call some basic methods so that the Ctx.Req() and Ctx.Res() methods have greater than 0 coverage?

ctx.go
```diff
@@ -1463,6 +1465,18 @@ func (c *DefaultCtx) renderExtensions(bind any) {
 	}
 }
 
+// Res returns a convenience type whose API is limited to operations
```
gaby, line 1468: Should say `Req` instead of `Res`

gaby (commented):
One small comment, we should probably add unit-tests that use these interfaces.

nickajacks1:
@ReneWerner87  this could probably be merged whenever

ReneWerner87:
Ok i will check it today, sorry lost focus last time

ReneWerner87 (approved):

ReneWerner87:
@nickajacks1 looks good to me, but can you please provide some more tests for the use of req/resp

ReneWerner87:
- [ ] I had seen that you had customized the current ones, but I would find it better if you added additional ones
so that the use with the direct use of req/res is tested and if you use it directly from the context object

- [ ] performance technically, do we still need to create pools for the objects ? to minimize the allocations in the process and then reset them when clearing the context ?

grivera64:
Thanks for working on this PR! It will be quite useful for distinguishing between Request and Response-based methods in `Ctx`.

I have an idea that could help with this bulletpoint from above:

> - [ ] performance technically, do we still need to create pools for the objects ? to minimize the allocations in the process and then reset them when clearing the context ?
> 
> 

From reviewing the PR, could we theoretically use an interface for `Req` and `Res` instead of a dedicated struct to avoid allocations/clearing altogether (aside from the already done allocation/clearing for the `DefaultCtx` struct) for performance?

I'm not entirely sure if it’s idiomatic to do this in Go, but we could set up `Ctx`, `Req`, and `Res` like in the following example:

```go
// All Req and Res-based methods
type Ctx interface {
  Queries() map[string]string
  SendString(body string) error
  // …
}

// Req - Request-based Ctx methods
type Req interface {
  Queries() map[string]string
  // …
}

// Res - Response-based Ctx methods
type Res interface {
  SendString(body string) error
  // …
}
```

We would need to look more into it, but this could potentially be autogenerated in the same way that `Ctx` is currently generated by [ifacemaker](https://github.com/vburenin/ifacemaker). Then, all we would need is to rewrite [`(*DefaultCtx).Req()`](https://github.com/gofiber/fiber/blob/2b891e830b93ce6e7c2893f861ca0e788a381b2b/ctx.go#L1470) and [`(*DefaultCtx).Res()`](https://github.com/gofiber/fiber/blob/2b891e830b93ce6e7c2893f861ca0e788a381b2b/ctx.go#L1476) as:

```go
func (c *DefaultCtx) Req() Req {
  return c
}

func (c *DefaultCtx) Res() Res {
  return c
}
```

If we do this, it should not allocate any new memory but rather cast the `Ctx` interface to a sub-interface for `Req` or `Res`. This would also eliminate the need to create any new methods for Req and Res directly when a new feature is added to `Ctx`.

If this idea wouldn't work well with this PR, I agree that we could also just add a pool for low memory footprint. I believe that and adding tests for the old and new v3 `Ctx` methods (e.g. [`c.SendStreamWriter()`](https://github.com/gofiber/fiber/blob/main/ctx.go#L1694)) would be all that is left before this is ready to merge.

Thanks again for the PR!

ReneWerner87:
@grivera64  in the first iteration i would like to go with this concept and after that you can show the new concept via pr or issue report
in the end, nothing should change in the usage, so that there is no breaking change, even if we do it in two steps

ReneWerner87:
@nickajacks1 are you working on the last comments ?

nickajacks1:
@ReneWerner87 Yes, I will work on this on Friday. Can you give a brief example of what kind of tests you would like to see

grivera64:
> @grivera64 in the first iteration i would like to go with this concept and after that you can show the new concept via pr or issue report in the end, nothing should change in the usage, so that there is no breaking change, even if we do it in two steps

For sure, sounds good to me! After this is merged, I can create an issue/PR with the new concept so we can discuss more about it.

nickajacks1:
I just realized how many tests there are 😁 
I thought of some ways to do this

1. Duplicate all tests for Req and Res
2. Create smaller smoke tests for Req and Res that are more compact/not as thorough
3. Somehow templatize the tests so that they end up in a format like 
```go
func testAccepts(t *testing.T, r Req) {}
func TestAccepts(t *testing.T) {
    app := New()
    t.Run("ctx", func(t *testing.T) { testAccepts(t, c) })
    t.Run("req", func(t *testing.T) { testAccepts(t, c.Req()) })
}
```

what do folks think? 3 sounds potentially nice but would have a huge impact to the tests and be a lot of work and potentially increased cognitive overhead. 1 seems overkill and prone to subtle drift unless the duplicated tests were generated somehow. 2 seems reasonable to me.

By the way, even with these changes, there are no additional allocations (I checked `Benchmark_Ctx_AcquireReleaseFlow`). That said, I kind of like @grivera64 's idea of simply having req and res be views onto the Ctx interface rather than bespoke objects.

ReneWerner87:
Iam god with idea 2 and 3

Btw we also need to somehow duplizieren the docs 

- [ ] Updated the documentation in the /docs/ directory for Fiber's documentation.

nickajacks1:
Update: rearranged docs. Added some methods I missed.
image

gaby:
@nickajacks1 What is left to-do here?

nickajacks1:
@gaby 
Only thing left is smoke-test type unit tests for the new interfaces. IMO this could be merged without the tests, but I'm still working on getting that part done.

gaby:
@nickajacks1 agree my browser freezes when I open this PR 😂😂😂

ReneWerner87:
> definitely thanks for the PR
> 
> * [x]  an update with the current master code would still be good
> * [ ]  documentation markdown files

documentation is also missing

nickajacks1:
@ReneWerner87 I updated ctx.md and reordered the methods into Request and Response subsections, let me know what you think

ReneWerner87:
https://github.com/gofiber/fiber/pull/2894/files#diff-eed7ea7ab33c80ec8dfbd395f659cc34525831b8553a4c8737064a9b8d8e0e34


think we should not reduce the ctx.md, because you can also use this function directly
the use via the request and response object is additional

what do you think ? @gaby @efectn

ReneWerner87:
premerged, you are welcome to open a new PR for the other things

PR #2904, opened 2024-03-08
Fix CSRF Referer port verification on HTTPS

sebytza23:
## Description

This commit introduces a fix for an issue encountered during development with localhost servers and clients using HTTPS. Previously, the application failed to pass CSRF token verification due to strict hostname matching. This was particularly problematic for development environments and infrastructures where both server and client are on the same system but use HTTPS.

Related to # (issue) 

## Changes introduced

The fix includes the addition of a new configuration option for CSRF verification. This option, when enabled, relaxes hostname matching requirements, allowing for successful CSRF token verification under development conditions and in same-server infrastructures using HTTPS. This enhancement ensures that developers facing this issue can proceed without hindrance, improving the development experience and workflow efficiency.

## Type of change

## Checklist


ReneWerner87:
maybe we should put the whole check in one function and not a part which can be controlled by an extra configuration setting
similar to the allowOriginFn in Cors
@sixcolors what do you think ?

sixcolors (changes requested):
Recommend against accepting this PR. @gofiber/maintainers a possible solution that maintains best practice for CSRF protection while addressing the users issue could be to mirror https://docs.djangoproject.com/en/5.0/ref/settings/#std-setting-CSRF_TRUSTED_ORIGINS

ReneWerner87:
okay that's what i thought, if it's a safety concern to leave space here, we shouldn't do that either

thanks @sixcolors for your assessment

sebytza23:
I am still curious, for me, that I use an https protocol, I don't have any way how to add a trusted origin over the csrf middleware, some people use fiber for servers only, and using it on a dedicated server, where the client is hosted on the same machine (https://mydomain.com) and the server is hosted as well here (https://mydomain.com:5555).
The request over csrf will fail, because of the port. We need a way to add a trusted origin. Could be a good thing, after all, gorilla already has this option.

sebytza23:
I could make a PR with this implementation(CSRF-AllowedOrigins), if needed.

sixcolors:
> I am still curious, for me, that I use an https protocol, I don't have any way how to add a trusted origin over the csrf middleware, some people use fiber for servers only, and using it on a dedicated server, where the client is hosted on the same machine (https://mydomain.com) and the server is hosted as well here (https://mydomain.com:5555). The request over csrf will fail, because of the port. We need a way to add a trusted origin. Could be a good thing, after all, gorilla already has this option.

Your using a non-standard https port to separate the backend and front-end in production?

sixcolors:
Closed as we will address this via #2909

PR #2919, opened 2024-03-14
session id can be specified within the function

bndrmrtn:
## Description

This feature provides a way to get a session provided by an ID within a function argument. This could be useful if the ID is stored in a different place than the default methods.

Fixes # (issue)

## Changes introduced

It's just a small change that accepts developers to gather the ID of the session from somewhere else.

## Commit formatting

🩹 Fix: Improved the session middleware with a function that allow developers to get session by an ID they specified.

## Summary by CodeRabbit

- **Refactor**
	- Improved session handling by introducing a method to get or create sessions by ID, enhancing the efficiency of session management.

### DIFF
--- a/middleware/session/store.go
+++ b/middleware/session/store.go
@@ -42,10 +42,13 @@ func (*Store) RegisterType(i any) {
 
 // Get will get/create a session
 func (s *Store) Get(c fiber.Ctx) (*Session, error) {
+	return s.GetByID(c, s.getSessionID(c))
+}
+
+// GetByID will get/create a session
+func (s *Store) GetByID(c fiber.Ctx, id string) (*Session, error) {
 	var fresh bool
 	loadData := true
-
-	id := s.getSessionID(c)
 
 	if len(id) == 0 {
 		fresh = true


gaby:
@bndrmrtn Your change is affecting indirectly the coverage of the session storage. See here https://app.codecov.io/gh/gofiber/fiber/pull/2919/indirect-changes

sixcolors:
Should we need a ctx for looking up a session by ID?

bndrmrtn:
> Should we need a ctx for looking up a session by ID?

I mean it not. But the ctx is passed to sess.ctx.

ReneWerner87 (changes requested):
not sure about this customization

you should move this part into the `Get` method
https://github.com/gofiber/fiber/blob/9555bffcb237068b9f44e991c0dbd5eebdf7eaca/middleware/session/store.go#L50-L65

as @sixcolors has already said, the injecting of the ctx is strange
then you can only use this within the request handler and so it offers no advantage over the `Get` method

if this functionality is really needed, you have to create a completely separate function and also make sure that the sessionObjects can be released back into the pool

ReneWerner87:
@bndrmrtn thanks for the effort

unfortunately we should implement the feature a little differently
i also see the great benefit of controlling the sessions in an admin interface or making them deletable in each case

nevertheless, we should make this possible without entering the current context
therefore some functionalities should be made independent of the context, by splitting them up 
in any case, somehow split it up and make it reusable, so that no context has to be used where it doesn't actually belong

Translated with DeepL.com (free version)

PR #2946, opened 2024-03-28
Performance optimization

ReneWerner87:


gaby:
@ReneWerner87 Why are there so many commits? 😂

PR #3016, opened 2024-05-28
feat!(middleware/session): re-write session middleware with handler

### LINKED ISSUE #2741: 🐛 [Bug]: Inconsistency in Session and CSRF Middleware Handling of Timeout and Expiration
## Inconsistency in Session and CSRF Middleware Handling of Timeout and Expiration

The current behaviour of both the session and CSRF middlewares introduces a notable inconsistency in handling expiration, sometimes as a timeout, deviating from established standards such as those outlined in [NIST Special Publication 800-63B](https://pages.nist.gov/800-63-3/sp800-63b.html#sec7).

TD;DR for NIST guidelines:

Expiration refers to the overall validity period of a session and may be extended by reauthentication, while timeout is associated with inactivity and requires user activity within a defined period to prevent session termination.

### Session Middleware

The session middleware currently has a predetermined endpoint for session expiration. However, there is an unexpected side-effect when `sess.Save()` is called, which extends the session as if it were a timeout. This behaviour is not intuitive and needs to be reviewed to comply with NIST SP 800-63B guidelines. The guidelines make a clear distinction between timeout and expiration, and the session middleware should be updated accordingly to follow these guidelines. Alternatively, developers should have some mechanism to control which actions extend a session expiration and which do not.

### CSRF Middleware

The CSRF middleware treats expiration as a timeout and extends it with every request. This behaviour aligns with a specific use case but deviates from the principles outlined in the NIST documentation, which separates Timeout and Expiration. 

After every request, if this middleware is used with a session, the `createOrExtendTokenInStorage()` function is called. This function, in turn, calls `sessionManager.setRaw()` which calls `sess.Save()`, extending the session as if it too had a timeout and not an expiration.

### Recommendation

We should evaluate how expiration is handled in middlewares, eliminate inconsistencies, and align with best practices.

1. **Session Middleware:**

- [ ] Evaluate the Session middleware and consider adopting a consistent and transparent mechanism for session management, adhering to the guidance provided by NIST on distinguishing between timeout and expiration.

2. **CSRF Middleware:**

- [ ] Evaluate the CSRF middleware to align its behaviour with the recognized standards outlined in NIST documentation. Ensure that the middleware distinguishes between Timeout and Expiration, providing a secure and predictable approach to CSRF token management.
- [ ] Evaluate how using this middleware with the Session middleware affects expiry and timeout in sessions and consider any side effects.

3. **Documentation:**

- [ ] Update the middleware documentation to provide developers with clear guidance on how session expiration, timeout, and CSRF token management work.

### Additional Considerations

- [ ] Engage the community for feedback on the proposed changes and their alignment with NIST guidelines to ensure comprehensive coverage of developer concerns and use cases.
- [ ] Conduct thorough testing to validate that any modifications made to the middleware do not introduce regressions or disrupt existing applications.

### How to Reproduce

Use middleware

### Expected Behavior

Expiration and Timeout act is described in issue description

### Fiber Version

<= 2.51.0

### Code Snippet (optional)

_No response_

### Checklist:

sixcolors:
# Feature: Re-write Session Middleware with Handler

## Summary

This pull request introduces significant changes to the session middleware, enhancing its functionality and improving its robustness. The primary focus is on the re-write of the session middleware with a new handler.

## Status

- [x] Initial PR for concept
- [x] Optimize Performance
- [x] Migrate to Middleware Style (`session.New()` returns middleware handler)
- [x] Add new tests for the middleware handler to cover various scenarios including error handling, session retrieval, and data manipulation.
- [x] Improve test coverage.
- [x] Update Documentation
- [x] Update Migration Guide
- [x] Update CSRF Middleware
- [x] Update CSRF Middleware Tests to cover Middleware-style sessions.

### Middleware Changes

- **Re-write session middleware**: The session middleware has been re-written to use a handler for better session management.
- **New Middleware File**: Added `middleware/session/middleware.go` to define the session middleware and its configuration.
- **Enhanced Session Management**:
  - Introduced a middleware pool for efficient session handling.
  - Added methods for setting, getting, and deleting session data within the middleware.
  - Improved error handling during session operations.

### CSRF Session Manager Changes

- **Improved Session Handling**:
  - Modified `getRaw`, `setRaw`, and `delRaw` methods to use the session from the context first before falling back to the store.
  - Removed logging dependency (`log`) from the session manager.

### Configuration Changes

- **Config Struct Update**:
  - Renamed `Expiration` to `IdleTimeout` to better reflect the session idle duration.
  - Updated default values and their usage in the configuration.

### Session Struct Changes

- **Renamed Fields**:
  - `exp` is now `idleTimeout` to indicate the idle timeout duration for sessions.
- **Session Methods Update**:
  - Updated methods to use `idleTimeout` instead of `exp`.
  - Improved the `Save` method to handle idle timeout correctly.

## Changes From This PR

This is an new middleware approach. Changes have been made to:
- Update `session.New` to return the `Middleware` handler instead of the old `session.Store` struct.

A suggested standard use for the session might look something like this after these modifications:

```go
func main() {
	app := fiber.New()

	// Middleware
	app.Use(logger.New())
	app.Use(limiter.New())
	sessionMiddleware, sessionStore := session.NewWithStore()
	app.Use(sessionMiddleware)
	 app.Use(csrf.New(
		csrf.Config{
			Store: sessionStore,
		},
	))

	// Routes
		ses := session.FromContext(c)
		if ses == nil {
			return c.Status(fiber.StatusInternalServerError).JSON(fiber.Map{
				"message": "Internal Server Error",
			})
		}
		user, ok := sess.Get("user").(User)
		if !ok {
			return c.Redirect("/login")
		}

		return c.SendString(fmt.Sprintf("Welcome %s!", user.Username))

	})

	app.Get("/session", func(c *fiber.Ctx) error {
		ses := session.FromContext(c)
		if ses == nil {
			return c.Status(fiber.StatusInternalServerError).JSON(fiber.Map{
				"message": "Internal Server Error",
			})
		}
		// In this proposal, we will return the session ID
		// which we will always have, because we always have a session
		// even if it's empty. It's generated on the first request.
		_, sessionHasUser := sess.Get("user").(User)
		return c.JSON(fiber.Map{
			"session":        sess.ID(),
			"sessionHasUser": sessionHasUser,
		})
	})

	app.Get("/login", func(c *fiber.Ctx) error {
		return c.SendFile("login.html")
	})

	app.Post("/auth/login", func(c *fiber.Ctx) error {
		type LoginInput struct {
			Username string `json:"username"`
			Password string `json:"password"`
		}
		var input LoginInput
		if err := c.BodyParser(&input); err != nil {
			return err
		}
		// Check username and password
		usr, err := repo.Login(input.Username, input.Password)
		if err != nil {
			return c.Status(fiber.StatusUnauthorized).JSON(fiber.Map{
				"message": "Unauthorized",
			})
		}
		// Create session
		ses := session.FromContext(c)
		if ses == nil {
			return c.Status(fiber.StatusInternalServerError).JSON(fiber.Map{
				"message": "Internal Server Error",
			})
		}
		// Since this is a login, we should regenerate the session
		sess.Regenerate()
		// Set user in session
		sess.Set("user", usr)
		// Return message
		return c.JSON(fiber.Map{
			"message": "Login successful",
		})
		// Save session is automatic when middleware handler is used
		// it will happen after the handler has returned.
	})

	app.Post("/auth/logout", func(c *fiber.Ctx) error {
		ses := session.FromContext(c)
		if ses == nil {
			return c.Status(fiber.StatusInternalServerError).JSON(fiber.Map{
				"message": "Internal Server Error",
			})
		}
		// Destroy session
		sess.Destroy()
		// a new session will be created automatically on the next request
		// by the session middleware
		return c.JSON(fiber.Map{
			"message": "Logout successful",
		})
	})

	app.Get("/auth/timeout", func(c *fiber.Ctx) error {
		ses := session.FromContext(c)
		if ses == nil {
			return c.Status(fiber.StatusInternalServerError).JSON(fiber.Map{
				"message": "Internal Server Error",
			})
		}
		// Calling any route where the session middleware is
		// used will extend the idle timeout.
		return c.JSON(fiber.Map{
			"message":     "Session extended",
			"idleTimeout": sess.IdleTimeout(),
		})
	})

	// Start server
	app.Listen(":3000")
}
```

## Testing

- Ensure all existing tests pass.
- Add new tests for the middleware handler to cover various scenarios including error handling, session retrieval, and data manipulation.

## Notes

- This update introduces breaking changes due to the renaming of configuration fields and changes in session handling methods. Ensure to update the dependent code accordingly.
- The re-write aims to improve performance and maintainability of the session management logic.

Please review the changes and provide feedback or approval for merging.

------

Fixes #2741


sixcolors:
## Known Issues

### Storage Adapter UPSERT Behavior

There is an issue with the way Fiber's storage adapters handle setting keys. The current operation for setting a key is an UPSERT, which means it will insert a new key if it does not exist or update it if it does. This can lead to issues in scenarios with simultaneous requests altering the same session. For example:

- One request destroys the session.
- Another request makes an inconsequential change to the session.

If the session destruction completes first, the inconsequential change could recreate the session, potentially preventing actions like logging out and posing security risks.

Since this PR changes the session behavior to save and update on every request, it amplifies the problem if not addressed.

sixcolors:
See: https://github.com/gofiber/recipes/tree/v3-csrf-with-session-PR-3016-update for usage example of this PR in current state.

gaby:
Those new sequence diagrams are awesome

middleware/csrf/session_manager.go
```diff
@@ -27,11 +27,22 @@ func newSessionManager(s *session.Store, k string) *sessionManager {
 
 // get token from session
 func (m *sessionManager) getRaw(c fiber.Ctx, key string, raw []byte) []byte {
```
gaby, line 29: Do we need the Mutex Locks here too?

middleware/session/middleware.go
```diff
@@ -0,0 +1,188 @@
...
+		return &Middleware{}
+	},
+}
+
+// Session is a middleware to manage session state
+//
+// Session middleware manages common session state between requests.
+// This middleware is dependent on the session store, which is responsible for
+// storing the session data.
+func NewMiddleware(config MiddlewareConfig) fiber.Handler {
```
efectn, line 53: I think it's better if we use New() here. Maybe we can rename current New to NewStore, NewStorage etc.

middleware/csrf/session_manager.go
```diff
@@ -27,11 +27,22 @@ func newSessionManager(s *session.Store, k string) *sessionManager {
...
+
+	if sess != nil {
+		token, ok = sess.Get(m.key).(Token)
+	} else {
+		// Try to get the session from the store
+		storeSess, err := m.session.Get(c)
+		if err != nil {
+			// Handle error
+			return nil
+		}
```
renanbastos93, line 42: sounds not so good. Why do you ignore error?

middleware/csrf/session_manager.go
```diff
@@ -44,25 +55,39 @@ func (m *sessionManager) getRaw(c fiber.Ctx, key string, raw []byte) []byte {
...
+	if sess != nil {
+		// the key is crucial in crsf and sometimes a reference to another value which can be reused later(pool/unsafe values concept), so a copy is made here
+		sess.Set(m.key, &Token{key, raw, time.Now().Add(exp)})
+	} else {
+		// Try to get the session from the store
+		storeSess, err := m.session.Get(c)
+		if err != nil {
+			// Handle error
+			return
+		}
```
renanbastos93, line 68: same here about ignore error

middleware/csrf/session_manager.go
```diff
@@ -44,25 +55,39 @@ func (m *sessionManager) getRaw(c fiber.Ctx, key string, raw []byte) []byte {
...
-	sess.Delete(m.key)
-	if err := sess.Save(); err != nil {
-		log.Warn("csrf: failed to save session: ", err)
+	sess := session.FromContext(c)
+	if sess != nil {
+		sess.Delete(m.key)
+	} else {
+		// Try to get the session from the store
+		storeSess, err := m.session.Get(c)
+		if err != nil {
```
renanbastos93, line 84: 😢

middleware/session/middleware.go
renanbastos93: I believe that uses only e `New` or `NewSession`

middleware/session/middleware.go
```diff
@@ -0,0 +1,188 @@
...
+		}
+
+		// release the middleware back to the pool
+		releaseMiddleware(m)
+
+		return stackErr
+	}
+}
+
+var ErrTypeAssertionFailed = errors.New("failed to type-assert to *Middleware")
```
renanbastos93, line 99: I have been using global variables at the top of the file. Maybe we need to follow this example:

```go
package main

type A interface{}
type B struct{}
const C = "foo"
var D = "bar"

func something() {}
func main() {}
```

This way, we can help the compiler optimize the binary for people using Fiber.

middleware/session/middleware.go
```diff
@@ -0,0 +1,188 @@
...
+
+// acquireMiddleware returns a new Middleware from the pool
+func acquireMiddleware() *Middleware {
+	middleware, ok := middlewarePool.Get().(*Middleware)
+	if !ok {
+		panic(fmt.Errorf("%w", ErrTypeAssertionFailed))
```
renanbastos93, line 105: Here you are generating unnecessary processing.
```suggestion
		panic(ErrTypeAssertionFailed.Error())
```

renanbastos93 (commented):
Wow, you've been doing a good job. I commented on a few details. Thanks for contributing here.

middleware/session/middleware.go
sixcolors: How should we handle the interaction between Session and Middleware? Here's the potential issue: if the middleware is in use and the session for the same user is retrieved from the store, any changes made and saved in the handler would be overwritten by the middleware.

We could add a check in the store. If the middleware is in the context, it could pass that session of `Store.Get(c)` and `Session.Save()`, and it could `noop` if it's the middleware session.

middleware/csrf/session_manager.go
sixcolors: The CSRF middleware lacks a mechanism to handle that error beyond failing the CSRF check. A refactor of the CSRF middleware would be necessary to address this issue.

middleware/csrf/session_manager.go
sixcolors: Yep... Unfortunalty `middleware/csrf/storage_manager.go` also ignored errors in the same way eg. `_ = m.storage.Set(key, raw, exp) //nolint:errcheck // TODO: Do not ignore error`

middleware/session/middleware.go
sixcolors: It's a good convention, but I don't think it changes the compiled binary. I'll move it.

middleware/csrf/session_manager.go
sixcolors: The mu mutex in the Middleware struct is used to ensure thread-safety when modifying the Session and hasChanged fields of the Middleware instance. This is necessary because the Middleware instance could potentially be accessed by multiple goroutines concurrently.

However, since the Session struct itself already ensures thread-safety in its methods (i.e., it uses its own mutexes to protect its data), then the mu mutex in the Middleware struct might be redundant for operations on the Session.

But, the mu mutex is still necessary for the Session and hasChanged field. If this field is accessed concurrently without a mutex, it could lead to race conditions.

I will update.

middleware/session/config.go
```diff
@@ -88,8 +88,8 @@ func configDefault(config ...Config) Config {
 	cfg := config[0]
 
 	// Set default values
-	if int(cfg.Expiration.Seconds()) <= 0 {
-		cfg.Expiration = ConfigDefault.Expiration
+	if int(cfg.IdleTimeout.Seconds()) <= 0 {
+		cfg.IdleTimeout = ConfigDefault.IdleTimeout
```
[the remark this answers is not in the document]
sixcolors: TODO: There should probably be a way to allow for not having an idle timeout.

middleware/session/middleware.go
renanbastos93: @sixcolors it makes sense to me. I believe that follow with your point.

middleware/csrf/session_manager.go
renanbastos93: It's okay, can you open this issue for us?

middleware/csrf/session_manager.go
renanbastos93: let's go to solve together

middleware/session/middleware.go
renanbastos93: of course, thanks.

middleware/csrf/session_manager.go
sixcolors: @coderabbitai open an issue to improve error handling in the middleware/csrf/session_manager.go and middleware/csrf/storage_manager.go files as discussed here.

middleware/csrf/session_manager.go
sixcolors: The "safe" option appears to be failing the CSRF, so it's likely the correct result already. We could incorporate logging to improve things.

middleware/csrf/session_manager.go
renanbastos93: @sixcolors https://github.com/gofiber/fiber/pull/3021

sixcolors: > @sixcolors #3021

I’ll take a look tomorrow

ReneWerner87:
@efectn @gaby can you recheck
@sixcolors can you maybe refresh with the master

gaby:
👍 For these changes, pending the tasks from the list and updating the test to cover the lines missing in coverage report.

ReneWerner87 (approved):

ReneWerner87:
@sixcolors looks very good

ReneWerner87:
@sixcolors 

- [ ] please add examples in the documentation and update the current api descriptions of the session middleware

Which tasks are still open?

ReneWerner87:
ok, there was a task list above^^

sixcolors:
@coderabbitai full review

middleware/csrf/csrf_test.go
```diff
@@ -156,6 +156,61 @@
...
+	require.NotEmpty(t, sessionID)
+
+	// Use the CSRF token and session_id
+	ctx.Request.Reset()
+	ctx.Response.Reset()
+	ctx.Request.Header.SetMethod(fiber.MethodPost)
+	ctx.Request.Header.Set(HeaderName, csrfToken)
+	ctx.Request.Header.SetCookie(ConfigDefault.CookieName, csrfToken)
+	ctx.Request.Header.SetCookie("session_id", sessionID)
+	h(ctx)
```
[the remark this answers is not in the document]
sixcolors: This is not correct. The session middleware is created without custom config in this test.

middleware/session/config.go
```diff
@@ -5,60 +5,91 @@ import (
...
 	Storage fiber.Storage
 
+	// Next defines a function to skip this middleware when it returns true.
+	// Optional. Default: nil
+	Next func(c fiber.Ctx) bool
+
+	// Store defines the session store.
+	//
+	// Required.
+	Store *Store
```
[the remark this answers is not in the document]
sixcolors: A session *Store is backed with fiber.Storage, they are different.

middleware/session/config.go
```diff
@@ -5,60 +5,91 @@
...
```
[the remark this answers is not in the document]
sixcolors: newStore, New and NewWithStore handle ensuring that a Store is present.

middleware/session/config.go
```diff
@@ -67,28 +98,57 @@
...
+//   - c: The Fiber context.
+//   - err: The error to handle.
+//
+// Usage:
+//
+//	DefaultErrorHandler(c, err)
+func DefaultErrorHandler(c *fiber.Ctx, err error) {
+	log.Errorf("session: %v", err)
+	if c != nil {
+		if sendErr := (*c).SendStatus(fiber.StatusInternalServerError); sendErr != nil {
```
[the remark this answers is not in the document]
sixcolors: not so `c.SendStatus undefined (type *fiber.Ctx is pointer to interface, not interface)`

middleware/session/middleware.go
```diff
@@ -0,0 +1,299 @@
...
+type Middleware struct {
+	Session   *Session
+	ctx       *fiber.Ctx
+	config    Config
+	mu        sync.RWMutex
+	destroyed bool
+}
+
+// Context key for session middleware lookup.
+const key = 0
```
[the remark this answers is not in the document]
sixcolors: I'll do this instead:

type contextKey int
const key contextKey = 0

middleware/session/store.go
```diff
@@ -26,7 +30,7 @@
 }
 
 // New creates a new session store with the provided configuration.
-func New(config ...Config) *Store {
+func newStore(config ...Config) *Store {
```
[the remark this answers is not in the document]
sixcolors: It's exported now.

middleware/session/middleware_test.go
```diff
@@ -0,0 +1,469 @@
...
+				expired = true
+				break
+			}
+		}
+	}
+	require.True(t, expired, "Session cookie should be expired")
+
+	// Sleep so that the session expires
+	time.Sleep(1 * time.Second)
+
```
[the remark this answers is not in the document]
sixcolors: it's checking for a session idletimeout to expire.

middleware/csrf/csrf_test.go
```diff
@@ -156,6 +156,61 @@ func Test_CSRF_WithSession(t *testing.T) {
 	}
 }
 
+// go test -run Test_CSRF_WithSession_Middleware
+func Test_CSRF_WithSession_Middleware(t *testing.T) {
+	t.Parallel()
+	app := fiber.New()
+
+	// session mw
+	smh, sstore := session.NewWithStore()
```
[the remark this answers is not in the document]
sixcolors: default is fine here.

middleware/session/middleware_test.go
```diff
@@ -0,0 +1,469 @@
...
+	)
+
+	app := fiber.New()
+
+	app.Use(New(Config{
+		Next: func(_ fiber.Ctx) bool {
+			muNext.RLock()
+			defer muNext.RUnlock()
+			return doNext
+		},
```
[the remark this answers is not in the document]
sixcolors: it already does this:

	muNext.Lock()
	doNext = true
	muNext.Unlock()

sixcolors:
@coderabbitai review

sixcolors:
Note: Session Middleware does not impliment a absolute timeout and instead impliments an idle timeout for session duration. The example https://github.com/gofiber/recipes/tree/v3-csrf-with-session-PR-3016-update/csrf-with-session demonstrates how a user can impliment this functionality themselves using a custom middleware.

ReneWerner87:
Now you can switch the state from draft to ready or? Will check the last changes tomorrow
Big thx for the work and unification of this mw

sixcolors:
> Now you can switch the state from draft to ready or? Will check the last changes tomorrow Big thx for the work and unification of this mw

I'm considering two additional changes:

1. **Absolute Timeout Mechanism**: I'm considering weather or not to integrate an absolute timeout, similar to the implementation mentioned in the PR comments. This would complement the idle timeout functionality and ensure sessions expire after a fixed duration, regardless of activity.

2. **Changing Storage Key**: I’m also thinking about updating the storage key type from `string` to `any` for added flexibility. This would allow developers to use unexported key types as session keys if needed.

Additionally, I plan to update a more feature-complete example that demonstrates these changes and uses the new PR. I’ll likely mark the PR as ready within the next few days.

sixcolors:
@gofiber/maintainers I've implemented the following changes:

1. **Absolute Timeout Mechanism**: Sessions now support an absolute timeout, ensuring they expire after a fixed duration, as discussed in the PR comments. To avoid modifying `fiber.Storage`, this has been achieved using a key/value pair stored within the session itself.

2. **Storage Key Change**: The session storage now supports `any` type for keys, providing greater flexibility. I’ve also updated the CSRF middleware to use an un-exported key type when storing tokens in the session.

I would appreciate it if reviewers could take a close look to ensure the implementation is both robust and aligned with best practices. Your feedback will be invaluable in finalizing these changes.

efectn:
> @gofiber/maintainers
> 
> I've made the following changes:
> 
> 1. **Absolute Timeout Mechanism**: Sessions now include support for an absolute timeout, ensuring they expire after a fixed duration, as outlined in the PR comments. To avoid modifying `fiber.Storage`, this has been implemented by storing a key/value pair directly in the session.
> 2. **Flexible Storage Key**: The session storage now supports `any` type for keys, offering greater flexibility. I’ve also updated the CSRF middleware to use an un-exported key type when storing tokens in session storage.
> 3. **IdleTimeout Consistency**: The expiration behavior in the CSRF middleware was functioning as an IdleTimeout, so I’ve renamed it for consistency with the updated Session middleware. (Note: I did not implement an AbsoluteTimeout option for the CSRF middleware. If this is a desired feature, we can open a separate issue and PR to address it.)
> 
> I’d appreciate it if reviewers could review these changes to ensure the implementation is robust and adheres to best practices. Your feedback will be instrumental in finalizing these improvements.

I will leave a review next days, thanks for your efforts

docs/middleware/session.md
```diff
@@ -2,142 +2,453 @@
...
+    Next              func(c *fiber.Ctx) bool
+    Store             *Store
+    ErrorHandler      func(*fiber.Ctx, error)
+    KeyGenerator      func() string
+    KeyLookup         string
+    CookieDomain      string
+    CookiePath        string
+    CookieSameSite    string
+    IdleTimeout       time.Duration
+    AbsoluteTimeout   time.Duration
```
efectn, line 139: An example in session.md docs would be cool to demonstrate difference between absolute and idle timeouts

efectn: Like what's new docs

middleware/session/data_test.go
```diff
@@ -0,0 +1,204 @@
...
+	t.Run("Concurrent access", func(t *testing.T) {
+		t.Parallel()
+		d := acquireData()
+		defer dataPool.Put(d)
+		defer d.Reset()
+		d.Set("key1", "value1")
+		d.Set("key2", "value2")
+		d.Set("key3", "value3")
+
+		done := make(chan bool)
```
efectn, line 124: wouldn't buffered channel be more appropriate here?

middleware/session/middleware.go
```diff
@@ -0,0 +1,304 @@
...
+	"sync"
+
+	"github.com/gofiber/fiber/v3"
+	"github.com/gofiber/fiber/v3/log"
+)
+
+// Middleware holds session data and configuration.
+type Middleware struct {
+	Session   *Session
+	ctx       *fiber.Ctx
```
efectn, line 16: is pointer necessary here?i think the interface itself can be treated as a pointer

middleware/session/middleware.go
```diff
@@ -0,0 +1,304 @@
...
+//   - *Middleware: The middleware object if found, otherwise nil.
+//
+// Usage:
+//
+//	m := session.FromContext(c)
+func FromContext(c fiber.Ctx) *Middleware {
+	m, ok := c.Locals(middlewareContextKey).(*Middleware)
+	if !ok {
+		// TODO: since this may be called we may not want to log this except in debug mode?
+		log.Warn("session: Session middleware not registered. See https://docs.gofiber.io/middleware/session")
```
efectn, line 183: I think the log is useful. Anyway, it might be still worth to add a config field to enable/disable logging

middleware/session/session.go
```diff
@@ -12,39 +12,78 @@ import (
...
-	data       *data         // key value data
-	byteBuffer *bytes.Buffer // byte buffer for the en- and decode
-	id         string        // session id
-	exp        time.Duration // expiration of this session
-	mu         sync.RWMutex  // Mutex to protect non-data fields
-	fresh      bool          // if new session
+	ctx         fiber.Ctx     // fiber context
+	config      *Store        // store configuration
+	data        *data         // key value data
+	byteBuffer  *bytes.Buffer // byte buffer for encoding/decoding
```
efectn, line 20: can we use bytebufferpool instead?

sixcolors: *bytebufferpool.ByteBuffer does not implement io.Reader, so I would have to add a wrapper, however Session never releases the bytes.Buffer and simple calls Reset when it puts it in the Session sync.Pool, so I'm not sure it would be of significant advantage.

middleware/session/middleware.go
sixcolors: log removed, logic for this: middleware that supports both legacy (Store) session API and Middleware API will need to check this and the log will be misleading and spammy.

middleware/session/middleware.go
sixcolors: Thanks for the feedback! You're right that interfaces are often treated like pointers. However, in this case, using `*fiber.Ctx` was intentional. AFAIK Since `fiber.Ctx` is a struct, passing it by value would create copies, which could lead to unnecessary overhead and inconsistent state. Using a pointer ensures we're working with the same instance, especially important when reusing objects from `sync.Pool`.

Let me know if you think there's a better way! Happy to discuss further.

sixcolors: Copilot says:

```
Using *fiber.Ctx is better in this context. Here’s why:

Performance: Passing a pointer (*fiber.Ctx) is more efficient than passing a value (fiber.Ctx) because it avoids copying the entire fiber.Ctx struct, which can be large.
Mutability: Using a pointer allows the middleware to modify the fiber.Ctx object directly, which is often necessary in middleware to set response headers, status codes, etc.
Consistency: Fiber's context (fiber.Ctx) is designed to be used as a pointer throughout the framework, ensuring consistency and expected behavior.
Therefore, the current usage of ctx *fiber.Ctx is appropriate.
```

sixcolors: Considering the changes in the rest of v3 I have made the modification here too.

sixcolors:
@coderabbitai review

middleware/session/session.go
efectn: > *bytebufferpool.ByteBuffer does not implement io.Reader, so I would have to add a wrapper, however Session never releases the bytes.Buffer and simple calls Reset when it puts it in the Session sync.Pool, so I'm not sure it would be of significant advantage.

Yeah but it's a little bit weird since it's a field of a struct. I think separate pool would be cleaner than making it a struct property

middleware/session/store.go
```diff
@@ -89,16 +161,38 @@
 	// Decode session data if found
 	if rawData != nil {
 		sess.data.Lock()
-		defer sess.data.Unlock()
-		if err := sess.decodeSessionData(rawData); err != nil {
+		err := sess.decodeSessionData(rawData)
+		sess.data.Unlock()
+		if err != nil {
+			sess.mu.Unlock()
```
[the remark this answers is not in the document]
sixcolors: sure write me a unit for this!

middleware/session/session.go
```diff
@@ -168,37 +263,57 @@
...
+	// If the session is being used in the handler, it should not be saved
+	if m, ok := s.ctx.Locals(middlewareContextKey).(*Middleware); ok {
+		if m.Session == s {
+			// Session is in use, so we do nothing and return
+			return nil
+		}
+	}
+
+	return s.saveSession()
+}
```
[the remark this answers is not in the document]
sixcolors: saveSession either returns nil or an error, so this logic would not be necessary.

middleware/session/session.go
sixcolors: I removed the code from the Session struct and added a byteBufferPool for encoding and decoding operations. Since not every session will be stored (e.g. new sessions), this change saves time and memory. Tests comparing the old commit with the new one show approximately a 2.17% performance improvement using this method. Thanks for the review!

sixcolors:
Just a reminder: The change to the `New()` function is a breaking change. It was made to emphasize that the Middleware handler style API is the preferred way to handle sessions. The legacy approach `NewStore()` is still available for users who do not wish to modify their code to use sessions in a new way. What do you think?

efectn (approved):

middleware/session/store.go
```diff
@@ -123,14 +219,113 @@
...
+			if err := sess.Destroy(); err != nil {
+				sess.Release()
+				log.Errorf("failed to destroy session: %v", err)
+			}
+			return nil, ErrSessionIDNotFoundInStore
+		}
+	}
+
+	return sess, nil
+}
```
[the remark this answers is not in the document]
sixcolors: return nil, fmt.Errorf("failed to decode session data: %w", decodeErr) is correct here.

middleware/session/session.go
```diff
@@ -12,39 +12,78 @@
...
 	s := sessionPool.Get().(*Session) //nolint:forcetypeassert,errcheck // We store nothing else in the pool
 	if s.data == nil {
 		s.data = acquireData()
 	}
-	if s.byteBuffer == nil {
-		s.byteBuffer = new(bytes.Buffer)
-	}
 	s.fresh = true
 	return s
 }
```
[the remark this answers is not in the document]
sixcolors: We store nothing else in the pool

middleware/session/store.go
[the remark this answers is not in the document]
sixcolors: We use testify require, so I wrote a unit using that. This suggestion isn't required anymore.

middleware/session/data.go
```diff
@@ -8,57 +8,120 @@
 //
 //go:generate msgp -o=data_msgp.go -tests=true -unexported
 type data struct {
-	Data         map[string]any
+	Data         map[any]any
```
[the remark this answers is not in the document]
gaby: Is this a real concern? @ReneWerner87 @efectn

middleware/session/data.go
```diff
@@ -8,57 +8,120 @@
...
+// Usage:
+//
+//	value := d.Get("key")
+func (d *data) Get(key any) any {
 	d.RLock()
-	v := d.Data[key]
-	d.RUnlock()
-	return v
+	defer d.RUnlock()
+	return d.Data[key]
```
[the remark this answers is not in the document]
gaby: @efectn @ReneWerner87 same here

middleware/session/store_test.go
```diff
@@ -116,3 +116,102 @@ func Test_Store_DeleteSession(t *testing.T) {
 	// The session ID should be different now, because the old session was deleted
 	require.NotEqual(t, sessionID, session.ID())
 }
+
+func TestStore_Get_SessionAlreadyLoaded(t *testing.T) {
+	// Create a new Fiber app
+	app := fiber.New()
+
+	// Create a new context
+	ctx := app.AcquireCtx(&fasthttp.RequestCtx{})
```
gaby, line 125: Missing defer for releasectx

middleware/session/store_test.go
```diff
@@ -116,3 +116,102 @@ func Test_Store_DeleteSession(t *testing.T) {
...
+		sess, err := store.GetByID("non-existent-session-id")
+		require.Error(t, err)
+		require.Nil(t, sess)
+		require.Equal(t, ErrSessionIDNotFoundInStore, err)
+	})
+
+	t.Run("valid session ID", func(t *testing.T) {
+		t.Parallel()
+		// Create a new session
+		ctx := fiber.New().AcquireCtx(&fasthttp.RequestCtx{})
```
gaby, line 182: Missing context release

middleware/session/store_test.go
```diff
@@ -73,7 +73,7 @@ func Test_Store_Get(t *testing.T) {
 	t.Run("session should be re-generated if it is invalid", func(t *testing.T) {
 		t.Parallel()
 		// session store
-		store := New()
+		store := NewStore()
 		// fiber context
```
gaby, line 77: Missing ctx release

gaby (commented):
Overall looks fine, few concerns with the "any" key for maps and the timeout/idle of CSRF.

docs/whats_new.md
```diff
@@ -314,9 +315,19 @@ Added support for specifying Key length when using `encryptcookie.GenerateKey(le
 
 ### Session
 
-:::caution
-DRAFT section
-:::
+The Session middleware has undergone key changes in v3 to improve functionality and flexibility. While v2 methods remain available for backward compatibility, we now recommend using the new middleware handler for session management.
```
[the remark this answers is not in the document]
ReneWerner87: @sixcolors @efectn what do you think?

middleware/session/data.go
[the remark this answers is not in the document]
ReneWerner87: shouldn't be a problem, the ai is right, but the user has to be able to think a bit too
this was probably done to be able to use these keys which are not exported

middleware/session/data.go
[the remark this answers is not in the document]
ReneWerner87: shouldn't be a problem, the ai is right, but the user has to be able to think a bit too
this was probably done to be able to use these keys which are not exported

docs/whats_new.md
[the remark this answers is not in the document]
ReneWerner87: i asked him in juli about this
image

@gaby

middleware/session/store_test.go
ReneWerner87: solved

middleware/session/store_test.go
ReneWerner87: solved

middleware/session/store_test.go
ReneWerner87: solved

docs/whats_new.md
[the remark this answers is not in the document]
efectn: Yeah i agree we can keep the old way

middleware/session/data.go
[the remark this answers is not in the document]
sixcolors: Yes. The change was to support unexported keys.

docs/middleware/csrf.md
```diff
@@ -34,7 +34,7 @@ app.Use(csrf.New(csrf.Config{
     KeyLookup:      "header:X-Csrf-Token",
     CookieName:     "csrf_",
     CookieSameSite: "Lax",
-    Expiration:     1 * time.Hour,
+    IdleTimeout:    30 * time.Minute,
```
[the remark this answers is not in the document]
sixcolors: Is using with sessions the argument makes sense. However CSRF can be used without our session mw. And then you would want a way for the CSRF key to expire.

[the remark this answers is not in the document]
ReneWerner87: @gaby we have to try to improve this in a next PR, via process change or documentation description
as already mentioned in the direct chat, this is the time how long the storage keeps this token for a session

Issue #2945, opened 2024-03-28
📝 [Proposal]: Allow disabling cache for ctx.SendFile()


bpatyi:
### Question Description

hello. Is there any way to turn off the ctx.SendFile cache? In our project, we further distribute files to devices in chunks, which can change every second. If I understand the code correctly, it is currently fixed to cache for 10 seconds, which causes it to arrive incorrectly. Thanks for your help.

### Checklist:

- [X] I agree to follow Fiber's [Code of Conduct](https://github.com/gofiber/fiber/blob/master/.github/CODE_OF_CONDUCT.md).
- [X] I have checked for existing issues that describe my questions prior to opening this one.
- [X] I understand that improperly formatted questions may be closed without explanation.

gaby:
@bpatyi This would need to be added as a feature in `v3`. The filesystem middleware problably needs the same feature.

gaby:
While we are at it, there's no unit-tests for the `compress` param.

https://github.com/gofiber/fiber/blob/main/ctx_test.go

PR #3017, opened 2024-05-29
:sparkles: v3 (feature): add configuration support to c.SendFile()

efectn:
## Description

This PR refactors SendFile functionality to make it work with different routes and different configurations.

Fixes https://github.com/gofiber/fiber/issues/2945

## Changes introduced

## Type of change

## Commit formatting


ctx.go
```diff
@@ -1398,47 +1399,99 @@ func (c *DefaultCtx) Send(body []byte) error {
...
+	}
 
 	// Keep original path for mutable params
 	c.pathOriginal = utils.CopyString(c.pathOriginal)
+
 	// Disable compression
-	if len(compress) == 0 || !compress[0] {
-		// https://github.com/valyala/fasthttp/blob/7cc6f4c513f9e0d3686142e0a1a5aa2f76b3194a/fs.go#L55
-		c.fasthttp.Request.Header.Del(HeaderAcceptEncoding)
+	if cfg.Compress {
```
ReneWerner87, line 1489: not sure about this

think the complication is activated by default in the fs handler and must be deactivated and not activated

ctx.go
```diff
@@ -1398,47 +1399,99 @@ func (c *DefaultCtx) Send(body []byte) error {
...
 
 // SendFile transfers the file from the given path.
 // The file is not compressed by default, enable this by passing a 'true' argument
 // Sets the Content-Type response HTTP header field based on the filenames extension.
-func (c *DefaultCtx) SendFile(file string, compress ...bool) error {
+func (c *DefaultCtx) SendFile(file string, config ...SendFile) error {
 	// Save the filename, we will need it in the error message if the file isn't found
 	filename := file
 
+	route := c.Route()
```
ReneWerner87, line 1444: not sure if we should store this data in the route itself, if then abstracted 
what if you have multiple sendFiles in a route ? 
shouldn't we create a map with the filenames and config settings as a key and store it underneath ?

ReneWerner87 (changes requested):
is still a bit to clarify

ctx.go
```diff
@@ -1398,47 +1399,99 @@ func (c *DefaultCtx) Send(body []byte) error {
...
 
 	// Keep original path for mutable params
 	c.pathOriginal = utils.CopyString(c.pathOriginal)
+
 	// Disable compression
-	if len(compress) == 0 || !compress[0] {
-		// https://github.com/valyala/fasthttp/blob/7cc6f4c513f9e0d3686142e0a1a5aa2f76b3194a/fs.go#L55
-		c.fasthttp.Request.Header.Del(HeaderAcceptEncoding)
+	if cfg.Compress {
+		c.fasthttp.Request.Header.Set(HeaderAcceptEncoding, "gzip")
```
gaby, line 1490: Why is gzip hardcoded here? This should be configurable

ctx.go
efectn: > not sure about this
> 
> think the complication is activated by default in the fs handler and must be deactivated and not activated

I think it's necessary to set encoding header to allow compression https://github.com/valyala/fasthttp/blob/7cc6f4c513f9e0d3686142e0a1a5aa2f76b3194a/fs.go#L64 otherwise it doesn't compress the served file

ctx.go
efectn: > Why is gzip hardcoded here? This should be configurable

I will

ctx.go
efectn: > not sure if we should store this data in the route itself, if then abstracted what if you have multiple sendFiles in a route ? shouldn't we create a map with the filenames and config settings as a key and store it underneath ?

The filename itself is not an issue since we don't keep it anywhere. 

Where can we store FS other than routes? I don't think it's a good idea to store them with filename since the filename can be changed according to the request

ReneWerner87: @efectn we can create a separate map and use the filename and the config as key
we don't want to extend the route for the feature, otherwise we can't abstract it later and it has several responsibilities that should be separated.

ctx.go
ReneWerner87: @efectn if not already present, we can then have a unittest for the compression case and the non-compression case, which checks the response encoding and the response itself to see if it is compressed (by simply checking the content length).

ctx.go
```diff
@@ -1414,48 +1415,160 @@ func (c *DefaultCtx) Send(body []byte) error {
 	return nil
 }
 
-var (
-	sendFileOnce    sync.Once
-	sendFileFS      *fasthttp.FS
-	sendFileHandler fasthttp.RequestHandler
-)
+// SendFile defines configuration options when to transfer file with SendFileWithConfig.
+type SendFile struct {
```
ReneWerner87, line 1419: pls move the struct to the top or in a different file togehter with the other logic

app.go
```diff
@@ -123,6 +123,9 @@ type App struct {
 	configured Config
 	// customConstraints is a list of external constraints
 	customConstraints []CustomConstraint
+	// sendfiles stores configurations for handling ctx.SendFile operations
+	sendfiles      []*sendFileStore
+	sendfilesMutex sync.RWMutex
```
gaby, line 128: add comment for `sendfilesMutex`

gaby (changes requested):

docs/api/ctx.md
```diff
@@ -1690,19 +1690,62 @@ app.Get("/", func(c fiber.Ctx) error {
 Transfers the file from the given path. Sets the [Content-Type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type) response HTTP header field based on the **filenames** extension.
 
 :::caution
-Method doesn´t use **gzipping** by default, set it to **true** to enable.
+Compression is not enabled by default. You need to pass config and enable `Compress` option.
```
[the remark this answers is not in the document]
efectn: You are right it's not needed

docs/api/ctx.md
```diff
@@ -1689,20 +1689,59 @@ app.Get("/", func(c fiber.Ctx) error {
...
+  // that is set on the file response. MaxAge is defined in seconds.
+  //
+	// Optional. Default value 0.
+  MaxAge int `json:"max_age"`
+}
+```
 
 ```go title="Signature" title="Signature"
-func (c Ctx) SendFile(file string, compress ...bool) error
+func (c Ctx) SendFile(file string, config ...SendFile) error
```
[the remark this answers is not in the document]
efectn: This is not needed. SendFİle already returns nil if there is no error

app.go
```diff
@@ -123,6 +123,10 @@ type App struct {
 	configured Config
 	// customConstraints is a list of external constraints
 	customConstraints []CustomConstraint
+	// sendfiles stores configurations for handling ctx.SendFile operations
+	sendfiles []*sendFileStore
```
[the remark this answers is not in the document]
efectn: It's available above the ctx.SendFile method

ctx_test.go
```diff
@@ -2970,19 +2971,239 @@ func Test_Ctx_SendFile(t *testing.T) {
...
+			resp, err := app.Test(req, 10*time.Second)
+
+			require.NoError(t, err, "app.Test(req)")
+			require.Equal(t, 200, resp.StatusCode, "Status code")
+			require.NotEqual(t, "58726", resp.Header.Get(HeaderContentLength))
+		})
+	}
+}
+
+func Test_Ctx_SendFile_CompressGzipBody(t *testing.T) {
```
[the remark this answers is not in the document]
efectn: done

ctx.go
```diff
@@ -1414,48 +1493,87 @@ func (c *DefaultCtx) Send(body []byte) error {
...
+		c.app.sendfilesMutex.Unlock()
+	}
 
 	// Keep original path for mutable params
 	c.pathOriginal = utils.CopyString(c.pathOriginal)
-	// Disable compression
-	if len(compress) == 0 || !compress[0] {
+
+	// Delete the Accept-Encoding header if compression is disabled
+	if !cfg.Compress {
```
gaby, line 1570: I know this is very old line of code, but why are we removing the header?

gaby: Shouldnt fasthttp itself be checling the header and not serving compressed content if Compress is disabled in fasthttp.FS ?

ctx_test.go
```diff
@@ -27,8 +28,10 @@ import (
 	"text/template"
 	"time"
 
+	"github.com/andybalholm/brotli"
 	"github.com/gofiber/fiber/v3/internal/storage/memory"
 	"github.com/gofiber/utils/v2"
+	"github.com/klauspost/compress/zstd"
```
gaby, line 34: Why did this get added?

ctx_test.go
```diff
@@ -2970,19 +2973,259 @@ func Test_Ctx_SendFile(t *testing.T) {
...
+		})
+		require.NoError(t, err)
+
+		return bytes.NewReader(c.Response().Body())
+	}
+
+	t.Run("gzip", func(t *testing.T) {
+		t.Parallel()
+
+		gz, err := gzip.NewReader(sendFileBodyReader("gzip"))
```
gaby, line 3096: We dont need all these readers/writers. Just use the built-in functions from fasthttp that compress data.

```go
	t.Run("Gzip", func(t *testing.T) {
		t.Parallel()
		compressedBody := fasthttp.AppendGzipBytes(nil, reqBody)
		require.NotEqual(t, reqBody, compressedBody)
		testCompressedBody(t, compressedBody, "gzip")
	})
```

See here: https://github.com/gofiber/fiber/blob/main/bind_test.go#L838

ctx_test.go
gaby: If you run go mod tidy, these will get added to go.mod outside of fasthttp which is not ideal. See below.

gaby (changes requested):
Overall LGTM, just the tests are too verbose. We should use the fasthttp AppendBytes functions for compressing data during these, that will reduce boilerplate on our side.

ctx.go
efectn: I'm not sure but fasthttp also seems to add it to servceuncompressed strAcceptEncodinghttps://github.com/valyala/fasthttp/blob/c7576cc10cabfc9c993317a2d3f8355497bea156/fs.go#L55

gaby (approved):
LGTM

ReneWerner87 (approved):

PR #3027, opened 2024-06-08
🔥 Feature: Add support for multiple keys in the KeyAuth middleware

dave-gray101:
## Description

I'm working on a project that must accept api keys from multiple possible sources, such as different header names. I've implemented this by slightly modifying the existing keyauth middleware, and I'd like to contribute it upstream. 

## Changes introduced

This is a fairly minimal enhancement to the existing `keyauth` middleware. I have extracted the existing `extractor` assignment code into a private function, and added a new `[]string` parameter to Config named `AdditionalKeyLookups`. Any values set here are interpreted just like `KeyLookup`, and are checked in order if there is no API key found at `KeyLookup`

## Type of change

- [X] Enhancement (improvement to existing features and functionality)


middleware/keyauth/keyauth.go
```diff
@@ -19,23 +19,37 @@ const (
...
-	case query:
-		extractor = keyFromQuery(parts[1])
-	case form:
-		extractor = keyFromForm(parts[1])
-	case param:
-		extractor = keyFromParam(parts[1])
-	case cookie:
-		extractor = keyFromCookie(parts[1])
+
+	parts := strings.Split(cfg.KeyLookup, "|")
```
gaby, line 31: I don't agree with this change, it's not standard practice. No other framework does something like this.

dave-gray101: @gaby - 

Do you mind elaborating slightly? Is your concern the entire concept of "multiple key lookup sources", or the way I tried to reuse the existing KeyLookup parameter? I'll admit it's a half-baked idea of mine rather than copying an example from Express - but I don't believe that they ship a standard keyauth middleware like this one to copy the API from.

If its the former, I'll just drop this PR entirely :)
If its the latter, I'll revise things to introduce a new parameter rather than overload the existing one. The only reason I didn't do this up front is that it implies we'll need to intelligently handle the case where a user provides _both_ parameters if I were to add say, a second parameter to Config named "KeyLookups" of type []string - in that case, I'm personally leaning to defaulting to the single "keylookup" parameter for backwards compatibility

dave-gray101: I realized the simple and best solution was to leave KeyLookup entirely alone, and add an "Additional" field. PR updated. Let me know if this addresses your concerns, @gaby

docs/api/middleware/keyauth.md
```diff
@@ -219,6 +219,7 @@ curl --header "Authorization: Bearer my-super-secret-key"  http://localhost:3000
 | SuccessHandler | `fiber.Handler`                          | SuccessHandler defines a function which is executed for a valid key.                                 | `nil`                         |
 | ErrorHandler   | `fiber.ErrorHandler`                     | ErrorHandler defines a function which is executed for an invalid key.                                | `401 Invalid or expired key`  |
 | KeyLookup      | `string`                                 | KeyLookup is a string in the form of "`<source>:<name>`" that is used to extract key from the request. | "header:Authorization"        |
+| AdditionalKeyLookups | `[]string` | If additional fallback sources of keys are required, they can be specified here in order of precedence | []string{} (empty) |
```
gaby, line 222: I'd probably name this `ExtraKeyLookup`, `ExtraKeyLookups`,  or something like that?

docs/api/middleware/keyauth.md
gaby, line 1: On line 240 the new config key is missing

gaby (commented):
Looks good, just two minor comments

middleware/keyauth/keyauth.go
gaby: Agree, much better 💪

docs/api/middleware/keyauth.md
dave-gray101: `FallbackKeyLookups` was the best I could come up with - indicates clearly that they are used _after_ `KeyLookup` in all cases.

gaby:
@coderabbitai full review

docs/api/middleware/keyauth.md
```diff
@@ -219,6 +219,7 @@
 | SuccessHandler | `fiber.Handler`                          | SuccessHandler defines a function which is executed for a valid key.                                 | `nil`                         |
 | ErrorHandler   | `fiber.ErrorHandler`                     | ErrorHandler defines a function which is executed for an invalid key.                                | `401 Invalid or expired key`  |
 | KeyLookup      | `string`                                 | KeyLookup is a string in the form of "`<source>:<name>`" that is used to extract key from the request. | "header:Authorization"        |
+| FallbackKeyLookups | `[]string` | If additional fallback sources of keys are required, they can be specified here in order of precedence | []string{} (empty) |
```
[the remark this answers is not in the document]
dave-gray101: This seems to be a bit of context window confusion on the part of the bot - the concept of pipe delimited lookups is no more.

gaby (approved):

gaby:
@dave-gray101 I just realized this is on the v2 branch, we are not adding any new features for v2. The target branch would have to be changed to v3

dave-gray101:
> @dave-gray101 I just realized this is on the v2 branch, we are not adding any new features for v2. The target branch would have to be changed to v3

Ah, that's unfortunate. I had dropped a question in the discord a few days ago regarding that, and didn't get a reply and figured I'd just create it rather than wait :D 

I based this branch off of v2, mostly as a guess. I'll try changing the target here, but it may be cleaner to re-apply off v3.

PR #3044, opened 2024-06-26
test(middleware/session): Remove extra release and acquire ctx calls in session_test.go

sixcolors:
Remove extra release and aquire ctx calls in session_test.go so Test_Session_Types passes in -race


sixcolors:
@coderabbitai no new stucts introduced in PR, re-review.

ReneWerner87 (approved):

PR #3050, opened 2024-06-29
fix(middleware/session): mutex for thread safety

sixcolors:
This pull request adds a mutex to the session middleware to ensure thread safety when accessing session data. The mutex is used to protect non-data fields and prevent race conditions. This improves the reliability and stability of the session middleware.

### DIFF
--- a/middleware/session/session_test.go
+++ b/middleware/session/session_test.go
@@ -676,6 +676,86 @@ func Benchmark_Session(b *testing.B) {
 	})
 }
 
+// go test -v -run=^$ -bench=Benchmark_Session_Parallel -benchmem -count=4
+func Benchmark_Session_Parallel(b *testing.B) {
+	b.Run("default", func(b *testing.B) {
+		app, store := fiber.New(), New()
+		b.ReportAllocs()
+		b.ResetTimer()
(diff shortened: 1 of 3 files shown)


middleware/session/session_test.go
```diff
@@ -673,3 +675,150 @@ func Benchmark_Session(b *testing.B) {
 		utils.AssertEqual(b, nil, err)
 	})
 }
+
+// go test -v -run=^$ -bench=Benchmark_Session_Asserted_Parallel -benchmem -count=4
+func Benchmark_Session_Asserted_Parallel(b *testing.B) {
```
gaby, line 680: Add another parallel benchmark without asserts

sixcolors: added all benchmark variations from v3.

ReneWerner87:
@sixcolors i don't see any mutex adjustments?

sixcolors:
> @sixcolors i don't see any mutex adjustments?

seems I medded up the merge, give me a couple minutes to fix.

sixcolors:
@ReneWerner87 ready

ReneWerner87 (approved):

PR #3075, opened 2024-07-16
🐛 [Bug]: cache middleware: runtime error: index out of range [0] with length 0

### LINKED ISSUE #3072: 🐛 [Bug]: cache middleware: runtime error: index out of range [0] with length 0
### Bug Description

Use cache middleware to cache home page "/".
Use CacheInvalidator to skip cache, got error: runtime error: index out of range [0] with length 0.

### How to Reproduce

1.use cache middleware to cache home page "/".
2.do not visit "/" in browser, so there will not be any cache of it.
3.visit "/login" page to login, then CacheInvalidator would return true now.
4.visit "/".
5.got error.

### Expected Behavior

When restart the program, not every page is cached, when user logged in visit no cache page, it should be open normal.

### Fiber Version

v3.0.0-beta.3

### Code Snippet (optional)

```go
var WebCache_Home = cache.New(cache.Config{
	CacheInvalidator: auth.ValidateLogin,
	Expiration: 10 * time.Minute,
	KeyGenerator: func(c fiber.Ctx) string {
		cacheKey := "homeLogout"
		if auth.ValidateLogin(c){
			cacheKey = "homeLogin"
		}
		return cacheKey
	},
	StoreResponseHeaders: true,
	MaxBytes:             1 * 1024 * 1024,
})

app.Use("/:p<maxLen(0)>?", cache.WebCache_Home)
app.Get("/:p<maxLen(0)>?", user.HomeHandler)
```

### Checklist:

brunodmartins:
# Description

Fixes #3072 

The bug happens due to some scenarios together: 

- The cache middleware always create a pointer `item` even though the cache entry was not found. 
- The application can define a `CacheInvalidator` that commands to expire the cache **even though** it doesn't not have any entry

In case the application defines a `MaxBytes` size for the cache, it will try to remove it from the heap, resulting in a runtime error due to the heap not having even initialized.

To fix the issue, I made the following changes:

- The cache will return `nil` in case it does not found the entry. 
- The expiration flow will not be triggered if no cache entry was found, this will avoid unexpected behaviors.

## Changes introduced

## Type of change

## Checklist

## Commit formatting

### DIFF
--- a/middleware/cache/manager_test.go
+++ b/middleware/cache/manager_test.go
@@ -1,0 +1,26 @@
+package cache
+
+import (
+	"testing"
+	"time"
+
+	"github.com/gofiber/utils/v2"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_manager_get(t *testing.T) {
+	t.Parallel()
+	cacheManager := newManager(nil)
+	t.Run("Item not found in cache", func(t *testing.T) {
+		t.Parallel()
+		assert.Nil(t, cacheManager.get(utils.UUID()))
+	})
+	t.Run("Item found in cache", func(t *testing.T) {
+		t.Parallel()
+		id := utils.UUID()
+		cacheItem := cacheManager.acquire()
+		cacheItem.body = []byte("test-body")
+		cacheManager.set(id, cacheItem, 10*time.Second)
+		assert.NotNil(t, cacheManager.get(id))
+	})
+}
--- a/middleware/cache/cache.go
+++ b/middleware/cache/cache.go
@@ -117,46 +117,49 @@ func New(config ...Config) fiber.Handler {
 		// Get timestamp
 		ts := atomic.LoadUint64(&timestamp)
 
-		// Invalidate cache if requested
-		if cfg.CacheInvalidator != nil && cfg.CacheInvalidator(c) && e != nil {
-			e.exp = ts - 1
-		}
-
-		// Check if entry is expired
-		if e.exp != 0 && ts >= e.exp {
-			deleteKey(key)
-			if cfg.MaxBytes > 0 {
-				_, size := heap.remove(e.heapidx)
-				storedBytes -= size
-			}
-		} else if e.exp != 0 && !hasRequestDirective(c, noCache) {
-			// Separate body value to avoid msgp serialization
-			// We can store raw bytes with Storage 👍
-			if cfg.Storage != nil {
-				e.body = manager.getRaw(key + "_body")
-			}
-			// Set response headers from cache
-			c.Response().SetBodyRaw(e.body)
-			c.Response().SetStatusCode(e.status)
-			c.Response().Header.SetContentTypeBytes(e.ctype)
-			if len(e.cencoding) > 0 {
-				c.Response().Header.SetBytesV(fiber.HeaderContentEncoding, e.cencoding)
-			}
-			for k, v := range e.headers {
-				c.Response().Header.SetBytesV(k, v)
-			}
-			// Set Cache-Control header if enabled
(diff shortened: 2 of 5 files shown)


ReneWerner87:
@brunodmartins thanks for the fix, can you please create another unittests for the case from the bug ticket so that we have this case covered forever

this unittest should lead to errors in the master and go through successfully in a pull request if you would transfer it for testing purposes

brunodmartins:
> @brunodmartins thanks for the fix, can you please create another unittests for the case from the bug ticket so that we have this case covered forever
> 
> this unittest should lead to errors in the master and go through successfully in a pull request if you would transfer it for testing purposes

Sure! I pushed just to don't lose the job on my machine. I am coding the tests right now on some way it makes sense to exist, and not just to pass the coverage! I believe on the next 2 days I will open the PR for review.

brunodmartins:
@ReneWerner87 I have added the tests to validate the issue. I tested it before the fix and they were falling. Despite that, I added new tests due to coverage failure on the parts that were moved to the conditional block to be executed only in case the cache entry exist.

I need some help to validate why the tests are failing on Windows machines (PR actions) =(.

brunodmartins:
@gaby , I have checked your point that the `cache.go` should have some mutex to avoid Race conditions, but the calls to manipulate the cache are already secured by a mutex at the middleware level (a.k.a the caller). Could you do a double check? In case we really need to move the mutex more inside, I can try to work on it at another PR =)

brunodmartins:
Thanks again for the help and patience

ReneWerner87:
> @ReneWerner87 I have added the tests to validate the issue. I tested it before the fix and they were falling. Despite that, I added new tests due to coverage failure on the parts that were moved to the conditional block to be executed only in case the cache entry exist.
> 
> I need some help to validate why the tests are failing on Windows machines (PR actions) =(.

image

you should not use the time as body, this varies greatly in the github machines used for the actions

rather use an incremental counter that counts up when the handler is reached

ReneWerner87:
@brunodmartins just seen that the other test cases also use times as returns, but in the nano seconds range

can you do me a favor and rework them too, so that they use a variable that just counts up and gives a different result if there is no cache hit?

brunodmartins:
@ReneWerner87 thanks for the help! I was just that! I fixed the whole test file and it worked =)

sixcolors:
> @brunodmartins just seen that the other test cases also use times as returns, but in the nano seconds range image
> 
> can you do me a favor and rework them too, so that they use a variable that just counts up and gives a different result if there is no cache hit?
> 
> all things with timing cause flakyness in our tests

"Timer resolution varies depending on the Go runtime, the operating system and the underlying hardware. On Unix, the resolution is approximately 1ms. On Windows, the default resolution is approximately 16ms, but a higher resolution may be requested using [golang.org/x/sys/windows.TimeBeginPeriod](https://pkg.go.dev/golang.org/x/sys/windows#TimeBeginPeriod)."
https://pkg.go.dev/time#hdr-Timer_Resolution

ReneWerner87:
@gofiber/maintainers can you re-review

ReneWerner87 (approved):

middleware/cache/manager_test.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"testing"
+	"time"
+
+	"github.com/gofiber/utils/v2"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_manager_get(t *testing.T) {
```
ReneWerner87, line 11: @brunodmartins pls add t.Parallel() in parent and child tests

brunodmartins: Done =)

PR #3105, opened 2024-08-20
v3: Improve path validation in Static Middleware

ashermyers:
# Description

The code uses filepath.Abs to resolve both the root directory and the requested path to their absolute forms. This improves security by further preventing advanced directory traversal attacks.

## Changes introduced

Absolute Path Resolution

## Type of change
- [ X ] Enhancement (improvement to existing features and functionality)


ReneWerner87:
@ashermyers thx for the adjustment
can you pls provide some test cases for the functionallity extension

ReneWerner87:
@ashermyers can you add some test cases code you have changed

gaby:
@coderabbitai full review

efectn:
@ashermyers can you check @ReneWerner87's comment

ReneWerner87:
i will continue on this

ReneWerner87:
> i will continue on this

@gaby is on it

gaby:
My change from `\\` to `/` broke windows tests as expected.

ReneWerner87:
https://github.com/valyala/fasthttp/pull/1698 is related

MegaManSec:
Note this introduces a vulnerability in systems which use a proxy in front of the server.

If your fiber server is behind a proxy, and the proxy disallows access to an endpoint, say `/secret-file`, and does not perform the same normalization that this PR includes, somebody can access `/normal-file%2F..%2Fsecret-file`, which fiber will treat as `/secret-file`, while the proxy will allow access. This is even more likely with %5C (backslash) instead of %2F.

If this is introduced into fiber, it should be disabled by default and IMO show a clear warning sign if enabling it. I am not sure how this PR defends against traversal attacks, as it seems to actually enable them (consider the usage of `encodeURIComponent()` by client-side applications: they will encode `%2F` and %5C` from `/` and `
` respectively, but now fiber will decode them.)

gaby:
@MegaManSec Yeah, we are aware that's why it hasn't been merged.

There's an upstream PR in fasthttp that needs to be merged first. I plan on adding extensive unit-tests to cover these cases.

ReneWerner87:
> @MegaManSec Yeah, we are aware that's why it hasn't been merged.
> 
> There's an upstream PR in fasthttp that needs to be merged first. I plan on adding extensive unit-tests to cover these cases.

@gaby ping

PR #3132, opened 2024-09-14
🩹 Fix: static server in sub app with mount (#3104)

### LINKED ISSUE #3104: 🤗 [Question]: Static server in sub app does not work
### Question Description

I am trying to set app a static server in a sub app, but it does not seem to work.
The problem can be reproduced with the snippet below.

Project struture:
```bash
tree .
#.
#├── go.mod
#├── go.sum
#├── main.go
#└── public
#    └── test.txt
```

Tests:

```bash
cat public/test.txt
# test

curl http://127.0.0.1:3000/
# Hello, world!%

curl http://127.0.0.1:3000/public/test.txt
# test

curl http://127.0.0.1:3000/subApp
# Hello from subApp%

curl http://127.0.0.1:3000/subApp/public/test.txt
# Cannot GET /subApp/public/test.txt%
```

### Code Snippet (optional)

```go
package main

import (
	"github.com/gofiber/fiber/v2"
	"github.com/gofiber/fiber/v2/middleware/logger"
)

func main() {
	app := fiber.New()
	logger := logger.New()
	app.Use(logger)

	app.Static("/public", "./public")

		return c.SendString("Hello, world!")
	})

	subApp := fiber.New()
	subApp.Use(logger)

	subApp.Static("/public", "./public")

	subApp.Get("/", func(c *fiber.Ctx) error {
		return c.SendString("Hello from subApp")
	})

	app.Mount("/subApp", subApp)
	app.Listen(":3000")
}
```

### Checklist:

- [X] I have checked for existing issues that describe my questions prior to opening this one.
- [X] I understand that improperly formatted questions may be closed without explanation.

yinheli:
# Description

This commit fixes an issue with static file serving in sub-app when using the Mount method. The problem was that static routes in sub-apps were not being properly handled when the sub-app was mounted to a parent app.

Key changes:

1. Modified the `registerStatic` function in `router.go` to handle mounted sub-apps correctly.
2. Updated the path handling logic to account for the mount prefix when serving static files.
3. Added a new test case in `router_test.go` to verify the correct behavior of static file serving in mounted sub-apps.

Fixes #3104 

## Changes introduced

- [x] Examples:
  ```go
  parentApp := fiber.New()
  subApp := fiber.New()

  subApp.Static("/public", "./public")
  parentApp.Mount("/sub", subApp)

  // Now, "/sub/public/*" will correctly serve static files from the "./public" directory
  ```

## Type of change

## Checklist


router_test.go
```diff
@@ -471,6 +471,37 @@ func Test_Route_Static_HasPrefix(t *testing.T) {
...
+
+	body, err = io.ReadAll(resp.Body)
+	utils.AssertEqual(t, nil, err, "app.Test(req)")
+	utils.AssertEqual(t, true, strings.Contains(app.getString(body), "color"))
+}
+
+func Test_Route_Static_SubApp(t *testing.T) {
+	t.Parallel()
+
+	dir := "./.github/testdata/fs/css"
```
gaby, line 490: Either add a nolint or move it to a variable.

yinheli: Got. Thanks! I'll try it.

router_test.go
```diff
@@ -471,6 +471,37 @@ func Test_Route_Static_HasPrefix(t *testing.T) {
...
+	app := New()
+
+	subApp := New()
+	subApp.Static("/css", dir)
+
+	app.Mount("/sub", subApp)
```
efectn, line 496: Can you add a testcase for nested mounting

gaby:
@yinheli Does this issue affect `v3` too? If so, can you do a Pull Request against `main` for v3. Thanks!

router_test.go
yinheli: Nested mounting is not working. I'll try to fix it.

router.go
```diff
@@ -357,6 +358,12 @@ func (app *App) registerStatic(prefix, root string, config ...Static) {
 		IndexNames:           []string{"index.html"},
 		PathRewrite: func(fctx *fasthttp.RequestCtx) []byte {
 			path := fctx.Path()
+			mountPath := app.FullMountPath()
+			if n := len(mountPath); n > 0 {
+				if bytes.Equal(path[:n], utils.UnsafeBytes(mountPath)) {
```
efectn, line 363: You don't need two-level if here. You can write:
```go
if n := len(mountPath); n > 0 && bytes.Equal(path[:n], utils.UnsafeBytes(mountPath)) {
...
}
```

mount.go
```diff
@@ -23,6 +23,8 @@ type mountFields struct {
 	subAppsProcessed sync.Once
 	// Prefix of app if it was mounted
 	mountPath string
+	// Parent app of the current app
```
efectn, line 26: @ReneWerner87 do you think we should use full-mount path for mounting? I think this might be a bug related to MountPath

yinheli: MountPath can only see that of subApp. Currently, the app cannot know how the upper layer is mounted. So I introduce a reference to Parent here. Do you have any good suggestions?

My current thinking is that when PathRewrite in registerStatic can replace the mounted part of the path.

ReneWerner87: @efectn @yinheli yes it should have something to do with the path
don't know if we necessarily need the complete parent reference for this

in the bug you said that the path is public/public, then we should start there

efectn: > @efectn @yinheli yes it should have something to do with the path don't know if we necessarily need the complete parent reference for this
> 
> in the bug you said that the path is public/public, then we should start there

Instead, perhaps we can do it during the path setting https://github.com/gofiber/fiber/blob/main/mount.go#L52

yinheli: I have investigated this. However, here only the path of the subapp can be obtained. And `registerStatic` initializes the fileHandler when registering. So I chose the current implementation and added FullMountPath. I think that is easy to implement and has a lower cost.

efectn: > I have investigated this. However, here only the path of the subapp can be obtained. And `registerStatic` initializes the fileHandler when registering. So I chose the current implementation and added FullMountPath. I think that is easy to implement and has a lower cost.

Ok i will check it tomorrow

yinheli: Appreciate

efectn: Does the problem occur with v3 @yinheli ? I've just given a try with v3, and seems to work well:

```go
func TestStaticMount(t *testing.T) {
	t.Parallel()
	app := fiber.New()
	sub := fiber.New()
	sub2 := fiber.New()

	sub2.Get("/doe*", New("../../.github"))

	sub.Use("/sub2", sub2)
	app.Use("/sub", sub)

	resp, err := app.Test(httptest.NewRequest(fiber.MethodGet, "/sub/sub2/doe/release.yml", nil))
	require.NoError(t, err, "app.Test(req)")

	reqq, err := io.ReadAll(resp.Body)
	require.NoError(t, err)

	require.Equal(t, 200, resp.StatusCode, "Status code")
	require.Equal(t, uint32(1), app.HandlersCount())
	require.Equal(t, " ... ", string(reqq))
}
```

yinheli: I haven't investigated yet. I'm a bit busy during this period.

ReneWerner87:
https://github.com/gofiber/fiber/issues/3442#issuecomment-2876327394

ReneWerner87:
@yinheli thanks for the effort, but since it does not solve the basic problem that the order of the routes is wrong, because the routes get a wrong initial position, i would close the pull request and provide one myself in the next days

https://github.com/gofiber/fiber/issues/3442#issuecomment-2876327394

PR #3153, opened 2024-10-02
:fire: Feature: Add AllLogger to Config

### LINKED ISSUE #2737: 🚀 [Feature]: middleware/logger: use Logger interface
### Feature Description

With the introduction of `Logger` interface (https://github.com/gofiber/fiber/pull/2499), it is easier now to plug my own log library into Fiber. With that said, `logger` middleware still uses `fmt.Sprintf()`.

It would be nice if `logger` middleware is extended to accept logging via `Logger` interface. This creates an easier method to have structure log for requests/responses without the need of making a custom external middleware. Judging from https://github.com/gofiber/fiber/issues/1293#issuecomment-829210674, if we want to keep the current `logger` middleware as simple as it is, making a new logger middleware would also be reasonable.

_No response_

### Code Snippet (optional)

_No response_

### Checklist:

haochunchang:
# Description

This adds a new field for the middleware/logger to allow users to pass in a logger with fiber logger interface.
To use the fiber logger interface, users need to define how the logger will be used in `LoggerFunc`.
Otherwise, it falls back to use original `fmt.Sprintf` logger instance.

Fixes #2737

## Changes introduced

## Type of change

## Checklist

## Commit formatting


ReneWerner87:
thx for the PR, we will have a look

can you check the pre hints from our review ai 
hopefully these make sense, otherwise we simply ignore them or have to fine-tune them

haochunchang:
Hi, I have addressed a few comments from the coderabbitai.
Please have a look when you have time. Thanks!

ReneWerner87:
Thx will do it tomorrow

middleware/logger/logger_test.go
```diff
@@ -181,6 +182,37 @@ func Test_Logger_ErrorTimeZone(t *testing.T) {
...
+	customLoggerFunc := func(_ fiber.Ctx, data *Data, cfg Config) error {
+		cfg.Logger.SetOutput(cfg.Output)
+		cfg.Logger.SetFlags(0)
+		cfg.Logger.Error(data.ChainErr.Error())
+		return nil
+	}
+
+	app.Use(New(Config{
+		Output:     buf,
+		Logger:     fiberlog.DefaultLogger(),
```
ReneWerner87, line 202: please add a second test where you inject another logger and show that it receives the data

middleware/logger/config.go
```diff
@@ -43,6 +44,13 @@ type Config struct {
 	// Optional. Default: defaultLogger
 	LoggerFunc func(c fiber.Ctx, data *Data, cfg Config) error
 
+	// Logger allows the use of a custom logger that implements the AllLogger interface.
+	// This field can be used in the LoggerFunc to do the logging.
+	// If you don't define LoggerFunc, it'll use default logger of Fiber without using this field.
+	//
+	// Optional. Default: nil
+	Logger fiberlog.AllLogger
```
ReneWerner87, line 52: where is this used in the middleware code ?

docs/middleware/logger.md
```diff
@@ -108,6 +121,8 @@ Writing to os.File is goroutine-safe, but if you are using a custom Output that
...
 | Output           | `io.Writer`                | Output is a writer where logs are written.                                                                                       | `os.Stdout`                                                           |
+| LoggerFunc           | `func(c fiber.Ctx, data *Data, cfg Config) error`  | You can use custom loggers with Fiber by using this field. This field is really useful if you're using Zerolog, Zap, Logrus, apex/log etc. If you don't define anything for this field, it'll use the default logger of Fiber.  | `see default_logger.go defaultLoggerInstance`                                                           |
+| Logger           | `fiberlog.AllLogger`  | Logger allows the use of a custom logger that implements the AllLogger interface. This field can be used in the LoggerFunc to do the logging. If you don't define LoggerFunc, this field will be ignored. |  `nil`                                                        |
```
ReneWerner87, line 125: please add the default value of this new property to the default config

ReneWerner87 (changes requested):
@haochunchang 
thank you for the effort, have not made any comments

middleware/logger/config.go
ReneWerner87: @haochunchang @efectn 
I think I understand, instead of the buffer in the default logger function another logger is used

three remarks
1. would not add logger to the config if it is only used in the custom variant of the consumer, then it can also use the logger instance from outside, because the logger is not relevant for the existing default logging code
2. if we get the logger instance, we should also use it in the default code, regardless of whether the custom logger function is set or not
3. what is the difference between the logger and the output property? both define the target
maybe you can use output for this?

what do you think about this?

haochunchang: I agree that we should also use the logger instance in the default code.

But I haven’t figured out how to use the methods of the AllLogger interface to achieve the same functionality of the original default logger. Especially in the part: https://github.com/gofiber/fiber/blob/c86c3c0a30a87eba3ada305fc70162ea8c875a54/middleware/logger/default_logger.go#L109-L120

efectn:
I think we can add a wrapper instead of messing with new config field and a function. Hereby, fiber logger interface can be used with the middleware by adjusting `Output` field. what do you think @haochunchang ?

```go
type dumbLogger struct {
	logger log.AllLogger
	level  log.Level
}

func (l *dumbLogger) Write(p []byte) (n int, err error) {
	switch l.level {
	case log.LevelDebug:
		l.logger.Debug(string(p))
	case log.LevelError:
		l.logger.Error(string(p))
	case log.LevelFatal:
		l.logger.Fatal(string(p))
	case log.LevelInfo:
		l.logger.Info(string(p))
	case log.LevelTrace:
		l.logger.Trace(string(p))
	case log.LevelWarn:
		l.logger.Warn(string(p))
	}

	return len(p), nil
}

func LoggerToWriter(logger log.AllLogger, level log.Level) io.Writer {
	return &dumbLogger{
		logger: logger,
		level:  level,
	}
}
```

haochunchang:
Cool. Then we can use this wrapper like the following:
```go
cfg.Output = LoggerToWriter(logger, level)
```
Is that what you mean?

efectn:
> Cool. Then we can use this wrapper like the following:
> 
> ```go
> cfg.Output = LoggerToWriter(logger, level)
> ```
> 
> Is that what you mean?

Yes it is

haochunchang:
I think using the wrapper method @efectn suggested can prevent disrupting the original Config and default logger 👍 

If using the wrapper method, then we only need to add an example into the documentation to provide users an example of using fiberlog interface.

I also added a unit test and benchmark tests for using the wrapper method.

ReneWerner87:
Wrapper method sounds good
If we can put it directly into the code of the logger middleware that would be good
And then tests and examples for the documentation

docs/middleware/logger.md
```diff
@@ -88,6 +88,33 @@ app.Use(logger.New(logger.Config{
...
+    }
+	return len(p), nil
+}
+
+func LoggerToWriter(customLogger fiberlog.AllLogger, level fiberlog.Level) io.Writer {
+	return &dumbLogger{
+        logger: customLogger,
+        level: level,
+    }
+}
```
ReneWerner87, line 113: @haochunchang can we put that or a better generic wrapper in the code so that not every dev has to create this struct? or would that be too undynamic? 

what do you think? @gaby @efectn

ReneWerner87: maybe we should recommend something like this ?

```go
func LoggerToWriter(customLogger fiberlog.AllLogger, level fiberlog.Level) io.Writer {
	return &struct {
		Write func(p []byte) (n int, err error)
	}{
		Write: func(p []byte) (n int, err error) {
			switch level {
			case fiberlog.LevelDebug:
				customLogger.Debug(string(p))
			case fiberlog.LevelError:
				customLogger.Error(string(p))
			}
			return len(p), nil
		},
	}
}
```
then we really don´t need extra code in the codebase

haochunchang: I think we can provide the `LoggerToWriter` in the example to show how to use fiberlog interface in the middleware.

ReneWerner87: ok @haochunchang pls adjust it a little bit

haochunchang: I’ve tried to condensed into a single function wrapper that return io.Writer but it seems it still need to declare the `Write` method in order to satisfy the `io.Writer` interface.

ReneWerner87: @haochunchang yeah, but in my example its there
so can you update the PR with this code snippet ?

haochunchang: Do you mean adding it into the example?
Cause I failed to adopt this into the unit tests :sweat_smile:

ReneWerner87: @haochunchang 
yeah add it in the readme in a new sub section

docs/middleware/logger.md
```diff
@@ -88,6 +88,27 @@ app.Use(logger.New(logger.Config{
...
+func LoggerToWriter(customLogger fiberlog.AllLogger, level fiberlog.Level) io.Writer {
+	return &struct {
+		Write func(p []byte) (n int, err error)
+	}{
+		Write: func(p []byte) (n int, err error) {
+			switch level {
+			case fiberlog.LevelDebug:
+				customLogger.Debug(string(p))
+			case fiberlog.LevelError:
+				customLogger.Error(string(p))
```
efectn, line 102: Can you add support for other logging levels too?

log/log.go
```diff
@@ -56,6 +56,7 @@ type CommonLogger interface {
 type ControlLogger interface {
 	SetLevel(level Level)
 	SetOutput(w io.Writer)
+	SetFlags(flags int)
```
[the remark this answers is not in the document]
efectn: Can you update log docs @haochunchang for SetFlags

middleware/logger/logger_test.go
```diff
@@ -181,6 +182,45 @@ func Test_Logger_ErrorTimeZone(t *testing.T) {
...
+	}))
+
+	app.Get("/", func(_ fiber.Ctx) error {
+		return errors.New("some random error")
+	})
+
+	resp, err := app.Test(httptest.NewRequest(fiber.MethodGet, "/", nil))
+	require.NoError(t, err)
+	require.Equal(t, fiber.StatusInternalServerError, resp.StatusCode)
+	require.Equal(t, "[Error] some random error\n", buf.String())
```
efectn, line 221: Can you also add a testcase for colorized log output using LoggerToWriter?

docs/middleware/logger.md
ReneWerner87: @coderabbitai pls prepare this change

log/log.go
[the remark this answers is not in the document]
ReneWerner87: @coderabbitai pls prepare all the changes from the hints

ReneWerner87:
@haochunchang can you look at the last notes so we can merge it then

efectn (approved):

log/default.go
```diff
@@ -210,6 +210,11 @@ func (l *defaultLogger) SetOutput(writer io.Writer) {
 	l.stdlog.SetOutput(writer)
 }
 
+// GetLoggerInstance returns the logger instance. It can be used to adjust the logger configurations in case of need.
+func (l *defaultLogger) GetLoggerInstance() any {
```
gaby, line 214: We should rename this to `GetLogger()` or `Logger()` ?

log/log.go
```diff
@@ -54,15 +54,25 @@ type CommonLogger interface {
...
+	// SetLevel sets logging level.
+	//
+	// Available levels: Trace, Debug, Info, Warn, Error, Fatal, Panic.
 	SetLevel(level Level)
+
+	// SetOutput sets the logger output.
 	SetOutput(w io.Writer)
+
+	// GetLoggerInstance returns the logger instance. It can be used to adjust the logger configurations in case of need.
+	GetLoggerInstance() any
```
gaby, line 66: Rename this to `Logger() any`

log/default.go
gaby: Probably `Logger()`  to match `ControlLogger`

middleware/logger/utils.go
```diff
@@ -37,3 +41,48 @@ func statusColor(code int, colors fiber.Colors) string {
...
 }
+
+type customLoggerWriter struct {
+	loggerInstance fiberlog.AllLogger
+	level          fiberlog.Level
+}
+
+func (cl *customLoggerWriter) Write(p []byte) (int, error) {
+	switch cl.level {
+	case fiberlog.LevelInfo:
```
gaby, line 52: This order is wrong, it should be by `severity`:

- Trace
- Debug
- Info
- Warning
- Error
- Fatal

middleware/logger/utils.go
```diff
@@ -37,3 +41,48 @@ func statusColor(code int, colors fiber.Colors) string {
...
+	}
+
+	return len(p), nil
+}
+
+// LoggerToWriter is a helper function that returns an io.Writer that writes to a custom logger.
+// You can integrate 3rd party loggers such as zerolog, logrus, etc. to logger middleware using this function.
+//
+// Valid levels: fiberlog.LevelInfo, fiberlog.LevelTrace, fiberlog.LevelWarn, fiberlog.LevelDebug, fiberlog.LevelError
+func LoggerToWriter(customLogger fiberlog.AllLogger, level fiberlog.Level) io.Writer {
```
gaby, line 73: Could we name the first param `logger` ?

log/log.go
```diff
@@ -54,15 +54,25 @@ type CommonLogger interface {
 
 // ControlLogger provides methods to config a logger.
 type ControlLogger interface {
```
gaby, line 56: Not a fan of `ControlLogger` but can't think of anything else to name this 😆

gaby (changes requested):

middleware/logger/utils.go
ReneWerner87: shouldn't matter, but we can still customize it

middleware/logger/utils.go
ReneWerner87: `logger` would be okay for me

log/log.go
ReneWerner87: @coderabbitai give good suggestions for this interface by thinking about how to use it

ReneWerner87: @gaby @efectn what do you think about the given names ?

docs/middleware/logger.md
```diff
@@ -108,6 +108,8 @@ Writing to os.File is goroutine-safe, but if you are using a custom Output that
...
 | TimeInterval     | `time.Duration`            | TimeInterval is the delay before the timestamp is updated.                                                                       | `500 * time.Millisecond`                                              |
 | Output           | `io.Writer`                | Output is a writer where logs are written.                                                                                       | `os.Stdout`                                                           |
+| LoggerFunc           | `func(c fiber.Ctx, data *Data, cfg Config) error`  | You can use custom loggers with Fiber by using this field. This field is really useful if you're using Zerolog, Zap, Logrus, apex/log etc. If you don't define anything for this field, it'll use default logger of Fiber.  | `see default_logger.go defaultLoggerInstance`                                                           |
+| Logger           | `fiberlog.AllLogger`  |   |                                                           |
```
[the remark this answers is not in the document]
efectn: > Can you add an example in this .MD file about using `zerolog` with this function.

We don't have a package in contrib just as we did for zap, so i haven't added an example for it yet

ReneWerner87 (approved):

ReneWerner87 (approved):

log/log.go
ReneWerner87: `ConfigurableLogger` is our choice

PR #3162, opened 2024-10-10
🔥 feat: Improve and Optimize ShutdownWithContext Func

JIeJaitt:
# Description

Optimize `ShutdownWithContext ` function and elegant shutdown flow of fiber, streamline `Listen` structure

## Changes introduced

- By holding the mutex lock throughout the operation, any modifications to the app's state are protected. defer ensures that the hook function is executed only after it is unlocked. At this time, no other goroutine will access the shared resource at the same time, ensuring concurrency safety.
- defer ensures that the hook function will be executed regardless of whether an error occurs in the function. It also avoids execution order issues and further ensures concurrency safety.
- Using executeOnPreShutdownHooks and executeOnPostShutdownHooks Instead of OnShutdownSuccess and OnShutdownError

## Type of change

## Checklist


app_test.go
```diff
@@ -860,45 +860,61 @@ func Test_App_ShutdownWithContext(t *testing.T) {
...
+    go func() {
+        conn, err := ln.Dial()
+        assert.NoError(t, err)
+        _, err = conn.Write([]byte("GET / HTTP/1.1\r\nHost: example.com\r\n\r\n"))
+        assert.NoError(t, err)
+        close(clientDone)
+    }()
+	
+	<-clientDone
+    time.Sleep(100 * time.Millisecond)
```
efectn, line 893: why do we need a sleep here? i think it's arbitrary since we already wait for clientDone channel

app.go
```diff
@@ -841,16 +841,18 @@ func (app *App) ShutdownWithTimeout(timeout time.Duration) error {
...
 	app.mutex.Lock()
 	defer app.mutex.Unlock()
+
 	if app.server == nil {
 		return ErrNotRunning
 	}
+
+	// Execute shutdown hooks in a deferred function
+	if app.hooks != nil {
+		defer app.hooks.executeOnShutdownHooks()
```
efectn, line 853: It would be nice to update the docs to explain shutdown & hook execution order

app_test.go
JIeJaitt: > why do we need a sleep here? i think it's arbitrary since we already wait for clientDone channel

This sleep is to ensure that our server has started processing the request. Although we have confirmed that the client has sent the request through <-clientDone, this means that the request has been sent to the server's listening port, and there is no guarantee that the server's processing coroutine has been sent. Start processing the request, so I think sleep 100 * Millisecond is required here as well.

app.go
JIeJaitt: > It would be nice to update the docs to explain shutdown & hook execution order

done

gaby:
@JIeJaitt The Makefile in the root of the repo, has a the commands to run tests/lint, etc

JIeJaitt:
> @JIeJaitt The Makefile in the root of the repo, has a the commands to run tests/lint, etc

Thanks, I'll try to fix any go-lint issues I have

JIeJaitt:
@gaby Hello, I have completed the repair of lint related issues
联想截图_20241013194726

docs/api/fiber.md
```diff
@@ -205,7 +205,7 @@ Shutdown gracefully shuts down the server without interrupting any active connec
 
 ShutdownWithTimeout will forcefully close any active connections after the timeout expires.
 
-ShutdownWithContext shuts down the server including by force if the context's deadline is exceeded.
+ShutdownWithContext shuts down the server including by force if the context's deadline is exceeded.Shutdown hooks will still be executed, even if an error occurs during the shutdown process, as they are deferred to ensure cleanup happens regardless of errors.
```
gaby, line 208: Missing space before "Shutdown

app_test.go
```diff
@@ -860,45 +861,60 @@ func Test_App_ShutdownWithContext(t *testing.T) {
...
 		ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
 		defer cancel()
 		shutdownErr <- app.ShutdownWithContext(ctx)
 	}()
 
 	select {
-	case <-time.After(5 * time.Second):
-		t.Fatal("idle connections not closed on shutdown")
+	case <-time.After(2 * time.Second):
+		t.Fatal("shutdown did not complete in time")
```
gaby, line 906: Is it "in time" or "on time"

go.mod
```diff
@@ -21,6 +21,7 @@ require (
 	github.com/philhofer/fwd v1.1.3-0.20240612014219-fbbf4953d986 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
 	github.com/valyala/tcplisten v1.0.0 // indirect
+	golang.org/dl v0.0.0-20241001165935-bedb0f791d00 // indirect
```
gaby, line 24: Why is this here?

go.sum
```diff
@@ -29,6 +29,8 @@ github.com/valyala/fasthttp v1.56.0 h1:bEZdJev/6LCBlpdORfrLu/WOZXXxvrUQSiyniuaoW
 github.com/valyala/fasthttp v1.56.0/go.mod h1:sReBt3XZVnudxuLOx4J/fMrJVorWRiWY2koQKgABiVI=
 github.com/valyala/tcplisten v1.0.0 h1:rBHj/Xf+E1tRGZyWIWwJDiRY0zc1Js+CV5DqwacVSA8=
 github.com/valyala/tcplisten v1.0.0/go.mod h1:T0xQ8SeCZGxckz9qRXTfG43PvQ/mcWh7FwZEA7Ioqkc=
+golang.org/dl v0.0.0-20241001165935-bedb0f791d00 h1:OX0WPBB1pQPZy1SL0+q5C/VuuM6e1wv6uEuB9iyBi/I=
```
gaby, line 32: And these?

gaby (changes requested):

go.mod
JIeJaitt: This is an official multi-version control function of the go language, because the Go version I use locally is higher than the go version in fiber's go mod. In order to keep the version consistent, I rely on this method to use it locally in the same way as the fiber framework. Go language version, you can see this to get more information: https://pkg.go.dev/golang.org/dl and https://pkg.go.dev/golang.org/dl@v0.0.0-20241001165935-bedb0f791d00/go1.22.6 . It will not have any impact on the function of the fiber

go.sum
JIeJaitt: Same as my answer above

docs/api/fiber.md
JIeJaitt: Thank you, I have solved this problem

app_test.go
JIeJaitt: I prefer to use it `in time`, because timely means to complete a certain goal within a temporary interval, while `on time` means to complete a certain goal at a certain moment. Such as, I will have breakfast at 8 o'clock tomorrow

ReneWerner87 (approved):

ReneWerner87:
false positive for the benchmark because of the same naming
https://github.com/benchmark-action/github-action-benchmark/issues/264 


@gaby can you check your last comments again to see if it is ready to merge

ReneWerner87:
@JIeJaitt can you make another comment in the whats new markdown?

JIeJaitt:
> @JIeJaitt can you make another comment in the whats new markdown?

No problem, I will add it in what new later

ReneWerner87:
@efectn @gaby pls check this PR

app_test.go
```diff
@@ -860,45 +861,60 @@ func Test_App_ShutdownWithContext(t *testing.T) {
...
-		_, err = conn.Write([]byte("GET / HTTP/1.1\r\nHost: google.com\r\n\r\n"))
+		_, err = conn.Write([]byte("GET / HTTP/1.1\r\nHost: example.com\r\n\r\n"))
 		assert.NoError(t, err)
+		close(clientDone)
 	}()
 
-	time.Sleep(1 * time.Second)
+	<-clientDone
+	// Sleep to ensure the server has started processing the request
+	time.Sleep(100 * time.Millisecond)
```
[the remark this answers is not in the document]
efectn: We are already doing it with clientDone channel

docs/api/fiber.md
```diff
@@ -205,7 +205,7 @@ Shutdown gracefully shuts down the server without interrupting any active connec
 
 ShutdownWithTimeout will forcefully close any active connections after the timeout expires.
 
-ShutdownWithContext shuts down the server including by force if the context's deadline is exceeded.
+ShutdownWithContext shuts down the server including by force if the context's deadline is exceeded. Shutdown hooks will still be executed, even if an error occurs during the shutdown process, as they are deferred to ensure cleanup happens regardless of errors.
```
efectn, line 208: Not sure about this behavior. I think we should only execute them during successful shutdown. Maybe we can OnShutdownError hook. What do you think @gaby @ReneWerner87

gaby: I'm not sure either. Also not sure about the defer of the hook

efectn: I would prefer to add Post Pre hook

JIeJaitt: The add Post Pre hook is a good idea, but how do I tell which hooks are pre closed and which are post closed? @efectn

efectn: You can add new hook called PostShutdown, PreShutdown and replace the old one with them

ReneWerner87: sound good, @JIeJaitt can you do the suggested change

JIeJaitt: This means making a change below the `type Hooks struct` to add these two fields that mean `PreShutdownHandler` and `PostShutdownHandler` respectively , if I understand it correctly I'll try to implement it @ReneWerner87 @efectn @gaby

efectn: > This means making a change below the `type Hooks struct` to add these two fields that mean `PreShutdownHandler` and `PostShutdownHandler` respectively , if I understand it correctly I'll try to implement it @ReneWerner87 @efectn @gaby

Yes it is. You can also add err parameter to PostShutdown hook, so that we can also get rid of OnShutdownError, OnShutdownSuccess properties of ListenerConfig.

JIeJaitt: > > This means making a change below the `type Hooks struct` to add these two fields that mean `PreShutdownHandler` and `PostShutdownHandler` respectively , if I understand it correctly I'll try to implement it @ReneWerner87 @efectn @gaby
> 
> Yes it is. You can also add err parameter to PostShutdown hook, so that we can also get rid of OnShutdownError, OnShutdownSuccess properties of ListenerConfig.

OK, I will press on with this idea to realize it again in my free time, by the way what is the v3 deadline as I am busy with my final exams these days.

efectn: > > > This means making a change below the `type Hooks struct` to add these two fields that mean `PreShutdownHandler` and `PostShutdownHandler` respectively , if I understand it correctly I'll try to implement it @ReneWerner87 @efectn @gaby
> > 
> > 
> > Yes it is. You can also add err parameter to PostShutdown hook, so that we can also get rid of OnShutdownError, OnShutdownSuccess properties of ListenerConfig.
> 
> OK, I will press on with this idea to realize it again in my free time, by the way what is the v3 deadline as I am busy with my final exams these days.

We want to release v3 rc1 in the beginning of January. If you can't look at the PR this week, i can also look no problem

JIeJaitt:
@efectn @ReneWerner87 @gaby I have completed the general revision, now can review my revision ideas, the rest of the lint problem and the content of the new document to add, and so on the big brothers review after I write, to prevent me from writing problems need to repeatedly modify the document!

app.go
```diff
@@ -900,16 +900,18 @@ func (app *App) ShutdownWithTimeout(timeout time.Duration) error {
 //
 // ShutdownWithContext does not close keepalive connections so its recommended to set ReadTimeout to something else than 0.
 func (app *App) ShutdownWithContext(ctx context.Context) error {
-	if app.hooks != nil {
```
efectn, line 903: Here we can define executeOnPostShutdownHooks hook with defer as well as PreShutdownHook

hooks_test.go
```diff
@@ -181,22 +181,92 @@
...
+
+		time.Sleep(100 * time.Millisecond)
+
+		app.hooks.executeOnPostShutdownHooks(expectedErr)
+
+		if !hookCalled {
+			t.Fatal("hook was not called")
+		}
+
+		if receivedErr != expectedErr {
```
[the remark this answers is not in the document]
efectn: You can use require methods instead

app.go
JIeJaitt: > Here we can define executeOnPostShutdownHooks hook with defer as well as PreShutdownHook

Is that what you mean?I think `PreShutdownHook `and `PostShutdownHook `set outside the function and then executed at ShutdownWithContext

```go
func (app *App) ShutdownWithContext(ctx context.Context) error {
	app.mutex.Lock()
	defer app.mutex.Unlock()

	if app.server == nil {
		return ErrNotRunning
	}

	// Execute the Shutdown hook
	if app.hooks != nil {
	}
	defer app.hooks.executeOnPostShutdownHooks(nil)

	return app.server.ShutdownWithContext(ctx)
}
```

efectn: > > Here we can define executeOnPostShutdownHooks hook with defer as well as PreShutdownHook
> 
> Is that what you mean?I think `PreShutdownHook `and `PostShutdownHook `set outside the function and then executed at `ShutdownWithContext`
> 
> ```go
> func (app *App) ShutdownWithContext(ctx context.Context) error {
> 	app.mutex.Lock()
> 	defer app.mutex.Unlock()
> 
> 	if app.server == nil {
> 		return ErrNotRunning
> 	}
> 
> 	// Execute the Shutdown hook
> 	if app.hooks != nil {
> 		app.hooks.executeOnPreShutdownHooks()
> 	}
> 	defer app.hooks.executeOnPostShutdownHooks(nil)
> 
> 	return app.server.ShutdownWithContext(ctx)
> }
> ```

You can do it like:

```go
func (app *App) ShutdownWithContext(ctx context.Context) error {
	app.mutex.Lock()
	defer app.mutex.Unlock()

	var err error

	if app.server == nil {
		return ErrNotRunning
	}

	// Execute the Shutdown hook
	defer app.hooks.executeOnPostShutdownHooks(err)

	err = app.server.ShutdownWithContext(ctx)
	return err
}
```

So that, we won't need to add another additional check to listen.go and this is going to work even if you use app.Shutdown() separately. However, this hook wouldn't work if app.Listen() hadn't been started in another goroutine. Maybe we can also add a note to indicate it -> https://github.com/valyala/fasthttp/blob/master/server.go#L1830

ReneWerner87:
@JIeJaitt can you update the code with the last ideas

JIeJaitt:
@ReneWerner87 Hi, I have updated the code as per the latest discussion. But I'm having a little go lint glitch. The Listener entry doesn't pass ctx, but go-lint lets me pass ctx, which is weird!
image

ReneWerner87:
@JIeJaitt 
image
problem is that the context is created outside in another golang function
previously both were within the same scope

JIeJaitt:
@efectn @ReneWerner87 @gaby Please review the latest code

ReneWerner87:
@JIeJaitt pls djust the markdown documentation 

- [ ] https://github.com/gofiber/fiber/blob/main/docs/api/hooks.md

@efectn pls re-review

app.go
```diff
@@ -894,6 +894,13 @@ func (app *App) HandlersCount() uint32 {
 //
 // Make sure the program doesn't exit and waits instead for Shutdown to return.
 //
+// Important: app.Listen() must be called in a separate goroutine, otherwise shutdown hooks will not work
+// as Listen() is a blocking operation. Example:
+//
+//	go app.Listen(":3000")
+//	// ...
```
efectn, line 901: Nice warning. Can you update main docs to warn about this case, too?

hooks_test.go
```diff
@@ -181,22 +181,94 @@ func Test_Hook_OnGroupName_Error(t *testing.T) {
...
+		hookErr := errors.New("hook error")
+
+		app.Hooks().OnPostShutdown(func(_ error) error {
+			return hookErr
+		})
+
+		// Should not panic
+		app.hooks.executeOnPostShutdownHooks(nil)
+	})
 }
```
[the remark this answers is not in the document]
efectn: Can you try to implement blocking channel instead of these delays for tests? @JIeJaitt

efectn (commented):
Mostly seems OK. Just a few additions would be good:
- Update what's new docs of v3 and migration guide stuff.
- We had better use blocking channels instead of sleep to ensure the process is ready.
- We can add a warning about `Important: app.Listen() must be called in a separate goroutine, otherwise shutdown hooks will not work` to docs and an example to show it. It is common mistake and we should a warning about it.

JIeJaitt:
@ReneWerner87 @efectn @gaby I've updated what_new.md and hooks.md, but I'm now experiencing a large number of failed Github Action tests, it appears to be a performance issue, the code is something I don't have much of a clue about at the moment.

ReneWerner87:
we will have a look

JIeJaitt:
> we will have a look

very thankful

JIeJaitt:
@ReneWerner87 @efectn @gaby all github action check succeed. Please review the latest code

docs/api/fiber.md
```diff
@@ -230,7 +230,7 @@ Shutdown gracefully shuts down the server without interrupting any active connec
 
 ShutdownWithTimeout will forcefully close any active connections after the timeout expires.
 
-ShutdownWithContext shuts down the server including by force if the context's deadline is exceeded.
```
efectn, line 233: Can you remove OnShutdownSuccess, OnShutdownSuccess from fiber.md

JIeJaitt: This feature is still available in v2, would it be a good idea to remove it from fiber's documentation now?

ReneWerner87: @JIeJaitt he means the config options in the ListenConfig
https://github.com/gofiber/fiber/blob/b0bc32b53472bc4891b25dc0cf2dc049554b4cf9/docs/api/fiber.md?plain=1#L117-L118

efectn (approved):

docs/api/hooks.md
```diff
@@ -16,6 +16,8 @@ With Fiber you can execute custom user functions at specific method execution po
 - [OnListen](#onlisten)
 - [OnFork](#onfork)
 - [OnShutdown](#onshutdown)
```
gaby, line 18: Shouldnt this one be removed from the docs?

gaby (changes requested):
LGTM, just a small change in the docs

ReneWerner87:
@JIeJaitt last two changes and then i will merge

JIeJaitt:
> @JIeJaitt last two doc changes and then i will merge

I have modified the document according to the discussion.

PR #3165, opened 2024-10-12
Update README.md

s19835:
Missing a pointer reference when passing the context object in the route handler function. In Fiber, the context (c) is a pointer, so it should be *fiber.Ctx instead of fiber.Ctx.

### DIFF
--- a/.github/README.md
+++ b/.github/README.md
@@ -73,7 +73,7 @@ func main() {
     app := fiber.New()
 
     // Define a route for the GET method on the root path '/'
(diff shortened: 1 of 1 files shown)


efectn (approved):

gaby:
@s19835 @efectn There's no pointer in v3, this was not needed

gaby:
@s19835 The context pointer is a Fiber v2 thing. For v3 which is what `main` is right now, the pointer was removed.

PR #3171, opened 2024-10-17
🐛 fix: Nil pointer dereference with Must Bind binding

ItsMeSamey:
if err is nil err.Error() panics
(eg. c.Bind().Must().JSON(...) successfully binds but panics


gaby:
@ItsMeSamey Anyway we could trigger this in a new unit-test?

ItsMeSamey:
@gaby I've added a test for this

bind.go
```diff
@@ -41,12 +41,12 @@ func (b *Bind) Must() *Bind {
 
 // Check Should/Must errors and return it by usage.
 func (b *Bind) returnErr(err error) error {
-	if !b.should {
-		b.ctx.Status(StatusBadRequest)
-		return NewError(StatusBadRequest, "Bad request: "+err.Error())
+	if err == nil || b.should {
```
gaby, line 44: This should be && not ||

gaby (changes requested):

ItsMeSamey:
@gaby
I've been wondering why is `Must()` even needed, it only adds overhead so won't we be better off removing it altogether. Setting status should be user's job anyway
Should i open a discussion / issue about this?

gaby:
> @gaby I've been wondering why is `Must()` even needed, it only adds overhead so won't we be better off removing it altogether. Setting status should be user's job anyway Should i open a discussion / issue about this?

That's a question for @efectn

ReneWerner87:
we will discuss this again internally and then let you know which way we want to go, thank you for your work and help

ReneWerner87:
here the last statement
https://github.com/gofiber/fiber/pull/3178#issuecomment-2435208194

we want to rename the feature so that there is no confusion with the golang concept of MUST for panics

because the methods always return an error
the feature only ensures that the response is filled with it

ReneWerner87:
give us a little time, we are currently discussing what this could look like
whether a config or the method should be renamed and where this can be found

ReneWerner87:
@ItsMeSamey sorry for the wait

we have decided to change the `Must` method to `WithAutoHandling

by default no errors should be added to the response, only if someone calls this additional method

can you integrate this into your PR ? and extend/customize the markdown


btw benchmark error is a false positive because of the same named benchmarks
https://github.com/benchmark-action/github-action-benchmark/issues/264

bind.go
```diff
@@ -31,22 +31,22 @@ func (b *Bind) Should() *Bind {
 	return b
 }
 
-// Must If you want to handle binder errors automatically, you can use Must.
+// If you want to handle binder errors automatically, you can use WithAutoHandling.
```
efectn, line 34: Maybe we can rename Should to WithoutAutoHandling()

efectn: What do you think @gaby @ReneWerner87

bind.go
```diff
@@ -19,34 +19,35 @@ type StructValidator interface {
 
 // Bind struct
 type Bind struct {
-	ctx    Ctx
-	should bool
+	ctx        Ctx
+	dontHandle bool
```
efectn, line 23: dontHandleErrs would be better

efectn (approved):

ReneWerner87:
Thx

ReneWerner87 (approved):

ReneWerner87:
@ItsMeSamey pls fix the lint errors 

image

ItsMeSamey:
@gaby
I believe this is ready to be merged now

gaby (approved):
👍 LGTM

gaby:
@ReneWerner87 @efectn Please check my changes to the README.md, if all good then we can merge it.

PR #3178, opened 2024-10-23
Proposal: Remove Must and Should when binding structs

ItsMeSamey:
## Bugs Fixed
Fixes issue described in #3171

## Changes introduced
Removed `Must`, `Should` and related code and tests

### DIFF
--- a/bind.go
+++ b/bind.go
@@ -41,12 +41,12 @@ func (b *Bind) Must() *Bind {
 
 // Check Should/Must errors and return it by usage.
 func (b *Bind) returnErr(err error) error {
-	if !b.should && err != nil {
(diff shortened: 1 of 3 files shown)


ItsMeSamey:
@efectn please review whenever possible

ReneWerner87:
@ItsMeSamey we have decided to keep the feature to fill the response automatically

thank you for bringing this topic to our attention

we can keep your other PR and expand it if necessary
the real problem is that we have adopted GIN's naming for this feature and this is misleading
because MUST is usually used in the context of panics

here, however, it should only be decided whether the response may be prefilled with the error

so we want to rename this functionality and document it better, and you can help us with this

PR #3201, opened 2024-11-12
🔥 feat: Add support for AutoTLS / ACME

### LINKED ISSUE #3155: 📝 [Proposal]: Add support for AutoTLS / ACME

Reduce the boilerplate of setting up AutoTLS / ACME by adding built-in support into Fiber. Similar to what `Gin` and `Echo` currently offer. We currently support this via a recipe, but we had users ask how to do this. Ideally we should integrate the code from the recipe into a new function under `app` similar to `app.Listen()` or `app.Listener()`.

N/a

RFC 8555 - https://datatracker.ietf.org/doc/html/rfc8555/

### API Stability

N/a

### Feature Examples

```go
We currently have a recipe for doing this:


package main

import (
	"crypto/tls"
	"log"

	"github.com/gofiber/fiber/v2"
	"golang.org/x/crypto/acme/autocert"
)

func main() {
	// Fiber instance
	app := fiber.New()

	// Routes
		return c.SendString("This is a secure server 👮")
	})

	// Let’s Encrypt has rate limits: https://letsencrypt.org/docs/rate-limits/
	// It's recommended to use it's staging environment to test the code:
	// https://letsencrypt.org/docs/staging-environment/

	// Certificate manager
	m := &autocert.Manager{
		Prompt: autocert.AcceptTOS,
		// Replace with your domain
		HostPolicy: autocert.HostWhitelist("example.com"),
		// Folder to store the certificates
		Cache: autocert.DirCache("./certs"),
	}

	// TLS Config
	cfg := &tls.Config{
		// Get Certificate from Let's Encrypt
		GetCertificate: m.GetCertificate,
		// By default NextProtos contains the "h2"
		// This has to be removed since Fasthttp does not support HTTP/2
		// Or it will cause a flood of PRI method logs
		// http://webconcepts.info/concepts/http-method/PRI
		NextProtos: []string{
			"http/1.1", "acme-tls/1",
		},
	}
	ln, err := tls.Listen("tcp", ":443", cfg)
	if err != nil {
		panic(err)
	}

	// Start server
	log.Fatal(app.Listener(ln))
}
```

### Checklist:

wangjq4214:
# Description

Reduce the boilerplate of setting up AutoTLS / ACME by adding built-in support into Fiber. Whether this impl is good?

Fixes #3155 

## Changes introduced

## Type of change


efectn:
Can we add several test-cases using https://letsencrypt.org/tr/docs/staging-environment/?

wangjq4214:
> Can we add several test-cases using https://letsencrypt.org/tr/docs/staging-environment/?

Of course, do you think this change will meet the demand?

listen.go
```diff
@@ -176,6 +184,11 @@ func (app *App) Listen(addr string, config ...ListenConfig) error {
 
 		// Attach the tlsHandler to the config
 		app.SetTLSHandler(tlsHandler)
+	} else if cfg.AutoCertManager != nil {
```
gaby, line 187: This should use TLSv1.2 min.

listen_test.go
```diff
@@ -145,6 +147,38 @@ func Test_Listen_TLS(t *testing.T) {
...
 
+// go test -run Test_Listen_Acme_TLS
+func Test_Listen_Acme_TLS(t *testing.T) {
+	// Certificate manager
+	m := &autocert.Manager{
+		Prompt: autocert.AcceptTOS,
+		// Replace with your domain
+		HostPolicy: autocert.HostWhitelist("example.com"),
+		// Folder to store the certificates
+		Cache: autocert.DirCache("./certs"),
```
gaby, line 158: Create a random temp dir and use that as your DirCache.

See here: https://pkg.go.dev/os#MkdirTemp

This will allow running the tests in parallel

listen_test.go
```diff
@@ -172,6 +206,42 @@ func Test_Listen_TLS_Prefork(t *testing.T) {
...
 
+// go test -run Test_Listen_Acme_TLS_Prefork
+func Test_Listen_Acme_TLS_Prefork(t *testing.T) {
+	// Certificate manager
+	m := &autocert.Manager{
+		Prompt: autocert.AcceptTOS,
+		// Replace with your domain
+		HostPolicy: autocert.HostWhitelist("example.com"),
+		// Folder to store the certificates
+		Cache: autocert.DirCache("./certs"),
```
gaby, line 217: Same here, use a random temp dir https://pkg.go.dev/os#MkdirTemp

efectn:
> > Can we add several test-cases using https://letsencrypt.org/tr/docs/staging-environment/?
> 
> Of course, do you think this change will meet the demand?

I think it can. Otherwise it might be challenging to test the functionality.

listen.go
```diff
@@ -70,6 +71,13 @@ type ListenConfig struct {
 	//
 	// Default: nil
 	OnShutdownSuccess func()
+
+	// AutoCertManager is a acme manager for go crypto package.
```
gaby, line 75: Change this to:

```go
// AutoCertManager manages TLS certificates automatically using the ACME protocol,
// enabling integration with Let's Encrypt or other ACME-compatible providers.
// Provide an *autocert.Manager instance to enable automatic certificate management.
//
// Default: nil
AutoCertManager *autocert.Manager
```

gaby (changes requested):

listen_test.go
efectn, line 158: Can we check if the cert has been downloaded to the dir

ReneWerner87:
- [ ] add markdown for the added config
- [ ] add markdown for an example in the markdown file
- [ ] mention the new feature in the what´s new markdown

grivera64:
After the PR is merged, I think we could also update [gofiber/recipes](https://github.com/gofiber/recipes) under the [autocert recipe](https://github.com/gofiber/recipes/tree/master/autocert) to use this feature.

listen.go
```diff
@@ -176,6 +185,12 @@ func (app *App) Listen(addr string, config ...ListenConfig) error {
 
 		// Attach the tlsHandler to the config
 		app.SetTLSHandler(tlsHandler)
+	} else if cfg.AutoCertManager != nil {
+		tlsConfig = &tls.Config{
+			MinVersion:     tls.VersionTLS12,
```
gaby, line 190: We should make this configurable, some people may want to do only TLS1.3

listen_test.go
```diff
@@ -145,6 +147,46 @@ func Test_Listen_TLS(t *testing.T) {
...
+		// Define the Client for test
+		Client: &acme.Client{
+			DirectoryURL: "https://acme-staging-v02.api.letsencrypt.org/directory",
+		},
+	}
+
+	app := New()
+
+	// invalid port
+	require.Error(t, app.Listen(":99999", ListenConfig{
```
gaby, line 174: @ReneWerner87 @efectn This is not failing, which may have exposed a bug in `Fiber`. That's an invalid port.

listen_test.go
```diff
@@ -172,6 +214,46 @@ func Test_Listen_TLS_Prefork(t *testing.T) {
...
```
gaby, line 239: @efectn @ReneWerner87 Sorry, this one doesn't fail. it should be failing.

listen_test.go
ReneWerner87: @gaby can we deal with this in an extra bug and change the port for this

gaby: Error was confirmed, reported here https://github.com/gofiber/fiber/issues/3237

listen.go
gaby: We will address this in a separate PR/Issue

gaby: @coderabbitai Please create an issue to make tlsconfig minversion configurable for the server listener and acme

gaby:
I need to run the tests locally, something is crashing when we hit the staging server

gaby:
> Can we add several test-cases using [letsencrypt.org/tr/docs/staging-environment](https://letsencrypt.org/tr/docs/staging-environment/)?

After testing multiple times, locally and in CI. That staging server is for testing with a domain you own. When running with `example.com` it will try to hit that server on port 80 which will fail and we get a timeout.

Even the official google package doesn't have a test for this. We can't add one either.

gaby (approved):
👍 LGTM. We are unable to add unit-tests since testing ACME requires either running Pebble or having a real domain.

docs/whats_new.md
ReneWerner87, line 1: please add a part for the new feature

ReneWerner87 (changes requested):
LGTM, but pls add something about this new cool feature in the "whats new" markdown

docs/whats_new.md
ReneWerner87: @wangjq4214

gaby: Done

ReneWerner87 (approved):

ReneWerner87 (approved):

PR #3218, opened 2024-11-27
✨ Optimize EnvVar Middleware (Performance & Memory)

hexakleo:
### Description

> [!NOTE]
> The *middleware* was updated to only allow **GET** requests because its job is to return environment variables in **JSON** format, which is typically done with a **GET** request. Allowing methods like **POST**, **PUT**, or **DELETE** wasn't necessary and could cause unexpected behavior.
>
>Now, the code explicitly checks for **GET** requests, and if the request is something else, it returns a *405* Method Not Allowed error:

**Fixes # (issue)**  
N/A

---

## Changes Introduced

- **Performance improvement**: Reduced the overhead of parsing environment variables by using more efficient string handling (`strings.IndexByte` instead of `strings.SplitN`).
- **Reduced Redundant Operations**: Avoided duplicate assignments when checking environment variables.
- **Memory optimization**: Ensured minimal allocations for variable storage.
- **Code simplification**: Cleaned up logic and improved readability.

### Benchmarks:
- Performance improved by reducing string slicing and splitting overhead.

### Documentation Update:
- Code comments were updated for clarity and to reflect the new optimizations.

### Changelog/What's New:
- **Performance improvement** in environment variable parsing.
- **Memory optimization** with better handling of the `Vars` map.

### Migration Guide:
- No migration required as these are non-breaking changes aimed at improving performance.

---

## Type of Change

- [x] **Performance improvement** (non-breaking change which improves efficiency)
- [x] **Enhancement** (improvement to existing features and functionality)

---

## Checklist

- [x] Followed the inspiration of the Express.js framework for new functionalities.
- [x] Verified that no new dependencies are introduced.

---

### **Old Code**:

```go
package envvar

import (
	"os"
	"strings"

	"github.com/gofiber/fiber/v3"
)

type Config struct {
	ExportVars map[string]string
	ExcludeVars map[string]string
}

type EnvVar struct {
	Vars map[string]string `json:"vars"`
}

func (envVar *EnvVar) set(key, val string) {
	envVar.Vars[key] = val
}

	var cfg Config
	if len(config) > 0 {
		cfg = config[0]
	}

	return func(c fiber.Ctx) error {
			return fiber.ErrMethodNotAllowed
		}

		envVar := newEnvVar(cfg)
		varsByte, err := c.App().Config().JSONEncoder(envVar)
		if err != nil {
			return c.Status(fiber.StatusInternalServerError).SendString(err.Error())
		}
		c.Set(fiber.HeaderContentType, fiber.MIMEApplicationJSONCharsetUTF8)
		return c.Send(varsByte)
	}
}

func newEnvVar(cfg Config) *EnvVar {
	vars := &EnvVar{Vars: make(map[string]string)}

	if len(cfg.ExportVars) > 0 {
		for key, defaultVal := range cfg.ExportVars {
			vars.set(key, defaultVal)
			if envVal, exists := os.LookupEnv(key); exists {
				vars.set(key, envVal)
			}
		}
	} else {
		const numElems = 2
		for _, envVal := range os.Environ() {
			keyVal := strings.SplitN(envVal, "=", numElems)
			if _, exists := cfg.ExcludeVars[keyVal[0]]; !exists {
				vars.set(keyVal[0], keyVal[1])
			}
		}
	}

	return vars
}
```

### New Code:

```go
package envvar

import (
	"os"
	"strings"

	"github.com/gofiber/fiber/v3"
)

type Config struct {
	ExportVars map[string]string
	ExcludeVars map[string]string
}

type EnvVar struct {
	Vars map[string]string `json:"vars"`
}

func (envVar *EnvVar) set(key, val string) {
	envVar.Vars[key] = val
}

	cfg := Config{}
	if len(config) > 0 {
		cfg = config[0]
	}

	return func(c fiber.Ctx) error {
		// Restrict to GET requests only
			return c.SendStatus(fiber.StatusMethodNotAllowed)
		}

		// Construct EnvVar and encode as JSON
		envVar := newEnvVar(cfg)
		varsByte, err := c.App().Config().JSONEncoder(envVar)
		if err != nil {
			return c.Status(fiber.StatusInternalServerError).SendString(err.Error())
		}

		// Set content type and send response
		c.Set(fiber.HeaderContentType, fiber.MIMEApplicationJSONCharsetUTF8)
		return c.Send(varsByte)
	}
}

func newEnvVar(cfg Config) *EnvVar {
	envVar := &EnvVar{Vars: make(map[string]string)}

	if len(cfg.ExportVars) > 0 {
		// Populate explicitly defined export variables
		for key, defaultVal := range cfg.ExportVars {
			if envVal, exists := os.LookupEnv(key); exists {
				envVar.set(key, envVal)
			} else {
				envVar.set(key, defaultVal)
			}
		}
	} else {
		// Exclude specified variables from all environment variables
		excludeVars := cfg.ExcludeVars
		for _, env := range os.Environ() {
			// Use strings.IndexByte for performance on splitting
			if idx := strings.IndexByte(env, '='); idx > 0 {
				key := env[:idx]
				if _, excluded := excludeVars[key]; !excluded {
					envVar.set(key, env[idx+1:])
				}
			}
		}
	}

	return envVar
}
```

---

### DIFF
--- a/middleware/envvar/envvar.go
+++ b/middleware/envvar/envvar.go
@@ -3,7 +3,6 @@ package envvar
 import (
 	"os"
 	"strings"
-
 	"github.com/gofiber/fiber/v3"
 )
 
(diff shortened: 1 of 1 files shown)


efectn:
Can you create and share benchmark results to compare?

gaby:
> Can you create and share benchmark results to compare?

Agree, we need an isolated benchmark to see the diff

efectn:
I don't think this PR improves the performance of the middleware. Here are the results:

```sh
Old: Benchmark_newEnvVar-16    	 4691676	       255.5 ns/op	     344 B/op	       3 allocs/op
New: Benchmark_newEnvVar-16    	 4552452	       261.4 ns/op	     344 B/op	       3 allocs/op
```

Benchmark Code:
```go
func Benchmark_newEnvVar(b *testing.B) {
	b.ReportAllocs()
	b.ResetTimer()

	for n := 0; n < b.N; n++ {
		newEnvVar(Config{
			ExportVars:  map[string]string{"testKey": ""},
			ExcludeVars: map[string]string{"excludeKey": ""},
		})
	}
}
```

ReneWerner87:
we will close it until a possible speed or memory improvement is confirmed

thanks anyway for the idea to improve something

PR #3230, opened 2024-12-05
🔥 feat: Add Support for Removing Routes

### LINKED ISSUE #3098: 📝 [Proposal]: Allow removing registered route

Following the addition of `RebuildTree` method #2769, which enabled dynamic request handling, I propose enhancing this functionality by allowing the removal of registered routes.

Currently, while we can dynamically add routes, the ability to remove them is limited. When adding a route with the same path, instead of updating the existing handler, a new one is added, which prevents the updated handler from being invoked. To address this issue, and to provide a way to remove routes that are no longer needed, we require a method to completely remove the route.

N/A

N/A

### API Stability

The proposed feature will involve the addition of a single method, `RemoveRoute`, which will be used to unregister specific routes based on the path and HTTP methods. This function will maintain the stability of the API by ensuring that the removal process is straightforward and does not interfere with other routes or the overall routing structure.

```go
func (app *App) RemoveRoute(path string, methods ...string) {
    // TODO:
}
 
```

### Feature Examples

```go
func addMyRoute() {
    // Remove route if it already exists
    s.app.RemoveRoute("/hello", fiber.MethodHead, fiber.MethodGet)

    // Add new route
    s.app.Get("/hello", func(c *fiber.Ctx) error {
        return c.SendString("hello")
    })

    // Rebuild the routing tree to reflect changes
    s.app.RebuildTree()
}

s.app.Get("/define", func(c *fiber.Ctx) error {
    before := s.app.HandlersCount()
    addMyRoute()
    after := s.app.HandlersCount()

    return c.JSON(map[string]any{
        "hcBefore": before,
        "hcAfter":  after,
    })
})
```

### Checklist:

ckoch786:
# Description

- Add method to Remove Routes
- Updated register method to prevent duplicate routes.

Fixes #3098

## Type of change


router.go
```diff
@@ -375,6 +387,56 @@ func (app *App) register(methods []string, pathRaw string, group *Group, handler
...
+
+		// Get unique HTTP method identifier
+		m := app.methodInt(method)
+		if m == -1 {
+			continue // Skip invalid HTTP methods
+		}
+
+		// Find the index of the route to remove
+		index := slices.IndexFunc(app.stack[m], func(r *Route) bool {
+			return r.Path == path
```
efectn, line 421: Can we also add another method called `RemoveRouteByName` to remove routes by their names?

ckoch786: RemoveRouteByName would remove all routes with the given name regardless of their path correct?
For instance if removing route named "test" would the both following routes be removed?
/api/test
/test

router.go
```diff
@@ -357,6 +370,7 @@ func (app *App) register(methods []string, pathRaw string, group *Group, handler
 			Method:   method,
 			Handlers: handlers,
 		}
+		// Should this use a mutex lock?
```
ckoch786, line 373: Question for you guys in the comment.

router.go
```diff
@@ -375,6 +389,56 @@ func (app *App) register(methods []string, pathRaw string, group *Group, handler
...
+	return slices.ContainsFunc(app.stack[app.methodInt(method)], func(r *Route) bool {
+		if r.path == pathRaw {
+			return true
+		}
+		return false
+	})
+}
+
+// RemoveRoute is used to remove a route from the stack
+// TODO should Remove Route only be called dynamically? Can you use it to remove the /define route, leaving the app without
```
ckoch786, line 402: Question for you guys in the comment.

router.go
```diff
@@ -375,6 +389,56 @@ func (app *App) register(methods []string, pathRaw string, group *Group, handler
...
+		if index == -1 {
+			continue // Route not found
+		}
+
+		// Remove route from tree stack
+		app.stack[m] = slices.Delete(app.stack[m], index, index+1)
+	}
+
+	// TODO THIS
+	app.routesRefreshed = true
```
ckoch786, line 434: Question for you guys in the comment.

router.go
```diff
@@ -375,6 +389,56 @@ func (app *App) register(methods []string, pathRaw string, group *Group, handler
...
+	// OR THIS
+	// // rebuild the tree if it has been built before
+	// if !app.routesRefreshed {
+	// 	app.buildTree()
+	// }
+}
```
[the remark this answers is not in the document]
ckoch786: Decided to go with the first route and updated the method comment.

router.go
```diff
@@ -375,6 +383,55 @@ func (app *App) register(methods []string, pathRaw string, group *Group, handler
...
 
+func (app *App) routeExists(method string, pathRaw string) bool {
+	return slices.ContainsFunc(app.stack[app.methodInt(method)], func(r *Route) bool {
+		return r.path == pathRaw
+	})
+}
+
+// RemoveRoute is used to remove a route from the stack
+// You should call RebuildTree after using this to ensure consistency of the tree
+func (app *App) RemoveRoute(path string, middlewareCount int, methods ...string) {
```
gaby, line 394: That `middlewareCount` param is very confusing. What's the purpose of it?

ckoch786: I agree, I do not like having to pass in the middlewareCount however.  The reason I believe this is needed is because the register method currently uses the length of handlers to increment the global routeCount.  

Here we can see that the handlers count should always be at least one. 
```go
func (app *App) register(methods []string, pathRaw string, group *Group, handler Handler, middleware ...Handler) {
	handlers := middleware
	if handler != nil {
		handlers = append(handlers, handler)
	}
```
^
|  more register method code between
v
```go
	// Increment global handler count
	atomic.AddUint32(&app.handlersCount, uint32(len(handlers))) //nolint:gosec // Not a concern
```

When removing a route we need to ensure that the app.handersCount is updated as well.  Since the user can pass in optional middleware this count can be greater than one for a given route.

```go 
func (app *App) deleteRoute(path string, methods []string, middlewareCount int) {
	app.mutex.Lock()
	defer app.mutex.Unlock()

	// No middleware passed in, increment to account for routes handler function
	if middlewareCount == 0 {
		middlewareCount++
	}

	//Decrement global handler count
	atomic.AddUint32(&app.handlersCount, ^uint32(middlewareCount-1)) //nolint:gosec // Not a concern
```

I have also thought of just passing in the registered middleware to RemoveRoute method and then passing the number of handlers into the deleteRoute method.  Not sure if that would be any less confusing.

I am open to suggestions.

ckoch786: Actually now that I am thinking about this more I think I should be able to get this from the existing route with that path and method.   I am starting to work on the RemoveRouteByName method that @efectn asked for so I will have to figure out how to get the counts from the existing routes since I will have to find all routes that contain the name of the route that needs to be deleted anyway.

router.go
ckoch786: I see now that a Route has a Name :)

router.go
ckoch786: I figured something better out.  Working on cleaning it up now.

router.go
```diff
@@ -375,6 +390,47 @@ func (app *App) register(methods []string, pathRaw string, group *Group, handler
...
+}
+
+// RemoveRoute is used to remove a route from the stack
+// You should call RebuildTree after using this to ensure consistency of the tree
+// TODO write tests for this that explicitally delete a route right after adding it
+func (app *App) RemoveRoute(path string, methods ...string) {
+	app.mutex.Lock()
+	defer app.mutex.Unlock()
+	app.deleteRoute(path, methods)
+}
```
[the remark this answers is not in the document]
ckoch786: Please add an example for a Test removing nested/parameterized routes

router.go
ckoch786: @efectn Just added the new RemoveRouteByName method in my latest commit.

router.go
ckoch786: @gaby Just cleaned up the call to deleteRoute in my latest commit.  I was able to do away with passing in any additional params to the method I solved it by doing this instead within the deleteRoute method.

```go
		// Find the index of the route to remove
		index := slices.IndexFunc(app.stack[m], matchFunc)
		if index == -1 {
			continue // Route not found
		}

		route := app.stack[m][index]

		// Decrement global handler count
		atomic.AddUint32(&app.handlersCount, ^uint32(len(route.Handlers)-1)) // nolint:gosec // Not a concern
		// Decrement global route position
		atomic.AddUint32(&app.routesCount, ^uint32(0))

```

router_test.go
```diff
@@ -369,31 +373,224 @@ func Test_Router_NotFound_HTML_Inject(t *testing.T) {
...
+			routes = append(routes, routeMsg)
+		}
+	}
+
+	for _, route := range routes {
+		require.Equal(tb, expectedRoutesCount, strings.Count(route.handlers, " "))
+	}
+}
+
+func verifyThereAreNoRoutes(tb testing.TB, app *App) {
```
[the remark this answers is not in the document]
ckoch786: I almost always forget this, sigh.

gaby:
@ckoch786 The method has to be added to the docs under `docs/`. Also mention this new method in the `whats_new.md` file. You can add it after the `RebuildTree` section introduced in #3074

gaby:
@ckoch786 In the root of the repo we have a `Makefile` to ease local development. See here: https://github.com/gofiber/fiber#-development

You can run the tests locally and also the lint tests.

ckoch786:
@gaby I cleaned up my branch and just pushed the docs changes.

Dockerfile
```diff
@@ -0,0 +1,25 @@
+# Usage:
```
gaby, line 1: What is this doing here?

docs/api/app.md
```diff
@@ -761,3 +761,56 @@ func main() {
...
 In this example, a new route is defined and then `RebuildTree()` is called to ensure the new route is registered and available.
+
+
+## RemoveRoute
+
+This method removes a route by path.  You must call the `RebuildTree()` method after the remove in to ensure the route is removed.
+
+```go title="Signature"
+func (app *App) RemoveRoute()
+```
```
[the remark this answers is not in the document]
gaby: Apply this change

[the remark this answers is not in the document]
gaby: Also remove the extra space at the beginning of the 2nd sentence.

docs/api/app.md
```diff
@@ -761,3 +761,56 @@ func main() {
...
+
+This method removes a route by name
+```go title="Signature"
+func (app *App) RemoveRouteByName()
+```
```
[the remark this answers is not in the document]
gaby: Apply this too

docs/api/app.md
```diff
@@ -761,3 +761,56 @@ func main() {
...
+           app.RemoveRoute("/api/feature", MethodGet)
+           app.RebuildTree()
+           // Redefine route
+           app.Get("/api/feature", func(c Ctx) error {
+                   return c.SendString("Testing feature-a")
+           })
+
+           app.RebuildTree()
+           return c.SendStatus(http.StatusOK)
+    })
```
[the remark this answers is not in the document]
gaby: ^Apply the above

router_test.go
```diff
@@ -369,31 +373,267 @@ func Test_Router_NotFound_HTML_Inject(t *testing.T) {
 	require.Equal(t, "Cannot DELETE /does/not/exist&lt;script&gt;alert(&#39;foo&#39;);&lt;/script&gt;", string(c.Response.Body()))
 }
 
+func sendStatusOK(c Ctx) error {
```
gaby, line 376: Remove this helper, do this call in the test

gaby (changes requested):

gaby (commented):
Some comments

gaby:
@ckoch786 Test and lint workflows are failing. You can test them locally using the `Makefile`.

docs/api/app.md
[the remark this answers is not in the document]
ckoch786: @gaby the router_test.go file is already part of the fiber package so the Ctx is already the fiber Ctx.  Also MethodGet is also in the fiber package and that is what is being used. In addition MethodGet and StatusOK are both defined in the fiber package source file constants.go.  So I did update my tests to use  the fiber constant StatusOK instead of the http.StatusOK constant.

When I made the update it importing github.com/gofiber/fiber after doing so and running `make tidy` and then `make test` I get the following errors:
# github.com/gofiber/fiber/v3 [github.com/gofiber/fiber/v3.test]
./router_test.go:378:9: c.SendStatus(http.StatusOK) (no value) used as value
./router_test.go:382:19: cannot use func(c *fiber.Ctx) error {…} (value of type func(c *"github.com/gofiber/fiber".Ctx) error) as Handler value in argument to app.Get
./router_test.go:383:35: cannot use sendStatusOK (value of type func(c *"github.com/gofiber/fiber".Ctx) error) as Handler value in argument to app.Get
./router_test.go:387:10: c.SendStatus(http.StatusOK) (no value) used as value
./router_test.go:450:28: cannot use func(c *fiber.Ctx) error {…} (value of type func(c *"github.com/gofiber/fiber".Ctx) error) as Handler value in argument to app.Get
./router_test.go:454:27: cannot use func(c *fiber.Ctx) error {…} (value of type func(c *"github.com/gofiber/fiber".Ctx) error) as Handler value in argument to app.Get
./router_test.go:455:11: c.SendString("Testing feature-a") (no value) used as value
./router_test.go:459:10: c.SendStatus(fiber.StatusOK) (no value) used as value
./router_test.go:461:28: cannot use func(c *fiber.Ctx) error {…} (value of type func(c *"github.com/gofiber/fiber".Ctx) error) as Handler value in argument to app.Get
./router_test.go:465:27: cannot use func(c *fiber.Ctx) error {…} (value of type func(c *"github.com/gofiber/fiber".Ctx) error) as Handler value in argument to app.Get
./router_test.go:465:27: too many errors
FAIL    github.com/gofiber/fiber/v3 [build failed]

Dockerfile
ckoch786: I am stuck in Windows currently so I created the Dockerfile so that I could get all the dependencies for running `make` to install properly.  I have more updates that make it more useful that I will be pushing soon that utilizes the build caching and I updated the comments to include how to utilize volumes and bind mounts with run to keep preserve all the downloaded packages between runs and to allow the linter to update the local code that is outside the docker container.

However if you do not want me to add it I can just remove it.

What is nice about using docker for running the make actions is that if it is used it gives you an isolated container from the rest of your machine so that you can get consistent results.  Also using docker makes it easier for others to work with the Makefile as all they need to do is install Docker and the rest will just work.

gaby: You cna remove it, this could be added to the docs though. Or one of the other fiber repos

ckoch786: Also if you choose to keep it I can create helper scripts for running docker build and docker run to make it easier to use .

ckoch786:
@gaby I am using the the Makefile lint and test.  The lint is now passing locally however the tests are really hit or miss for me locally not sure what is going on.  Even when running on  a fresh checkout of  gofiber/fiber on main three of the tests fail for me.
Also some of the tests that fail when running with make test pass when I run them with `go test -run Test_` so not really sure how to figure out why they are failing.  Any thoughts?

Also I did get to the point where all the test passed after running a few with `go test -run` and then rerunning `make test` but now I am back to 651 of them failing, not sure what I am doing wrong.

docs/whats_new.md
```diff
@@ -941,6 +941,14 @@ In this example, a new route is defined, and `RebuildTree()` is called to ensure
...
 Note: Use this method with caution. It is **not** thread-safe and can be very performance-intensive. Therefore, it should be used sparingly and primarily in development mode. It should not be invoke concurrently.
 
+## RemoveRoute
+
+- **RemoveRoute**: Removes route by path
+
+- **RemoveRouteByName**: Removes route by name
+
+For more details, refer to the [app documentation](./api/app.md#removeroute):
+
```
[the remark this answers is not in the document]
ckoch786: This is already documented in the docs/api/app.md file.

ckoch786:
@gaby I am looking at the failing tests from the actions that ran against my branch.  I think I found the the changes that I made  that is probably causing them to fail.  Though it is weird that they do not always fail for me locally, not sure what that is about yet.

gaby:
> @gaby I am looking at the failing tests from the actions that ran against my branch. I think I found the the changes that I made that is probably causing them to fail. Though it is weird that they do not always fail for me locally, not sure what that is about yet.

Are you running just the new tests or the full suite?

Use this `make test`. We use `-shuffle` to run them in random orders, it helps catch weird edge cases.

gaby:
@ckoch786 Any progress on this?

router.go
```diff
@@ -395,6 +402,72 @@ func (app *App) register(methods []string, pathRaw string, group *Group, handler
 	}
 }
 
+func (app *App) routeExists(method, pathRaw string) bool {
```
gaby, line 405: Something is wrong with this method. It's causing over 200 unit-tests to fail now.

gaby:
@ReneWerner87 @efectn I vote we close this. Spent 1hr and can't find the issue, all the tests are failing. Author is not responding

router.go
```diff
@@ -395,11 +396,81 @@ func (app *App) register(methods []string, pathRaw string, group *Group, handler
...
+		method = utils.ToUpper(method)
+
+		// Get unique HTTP method identifier
+		m := app.methodInt(method)
+		if m == -1 {
+			continue // Skip invalid HTTP methods
+		}
+
+		// Find the index of the route to remove
+		index := slices.IndexFunc(app.stack[m], matchFunc)
```
efectn, line 451: I think here we also need to look at other indexes in stack which also has matchFunc return true, not only the first one. It is necessary when a middleware is reqistered at the same path. Otherwise, we can skip `use` routes, which seems better idea to me.


Code:

```go
server.Use(func(c fiber.Ctx) error {
		return c.Next()
	})
server.Post("/", func(c fiber.Ctx) error {
	var t Test
	if err := c.Bind().Form(&t); err != nil {
		return err
	}
	fmt.Println(t)
	return c.SendString("Hello, World!")
})

server.Post("/", func(c fiber.Ctx) error {
	return c.SendString("Hello, World!")
})
```
Stack Output:

```go
0 {<nil> / GET  / [] [0x763660] {[0xc0000a22a0] [] 0 0} 1 true false false true}
1 {<nil> / HEAD  / [] [0x763660] {[0xc0000a22a0] [] 0 0} 2 true false false true}
2 {<nil> / POST  / [] [0x763660] {[0xc0000a22a0] [] 0 0} 3 true false false true}
2 {<nil> / POST  / [] [0x7636c0 0x7637a0] {[0xc0000a2360] [] 0 0} 10 false false false true}
3 {<nil> / PUT  / [] [0x763660] {[0xc0000a22a0] [] 0 0} 4 true false false true}
4 {<nil> / DELETE  / [] [0x763660] {[0xc0000a22a0] [] 0 0} 5 true false false true}
5 {<nil> / CONNECT  / [] [0x763660] {[0xc0000a22a0] [] 0 0} 6 true false false true}
6 {<nil> / OPTIONS  / [] [0x763660] {[0xc0000a22a0] [] 0 0} 7 true false false true}
7 {<nil> / TRACE  / [] [0x763660] {[0xc0000a22a0] [] 0 0} 8 true false false true}
8 {<nil> / PATCH  / [] [0x763660] {[0xc0000a22a0] [] 0 0} 9 true false false true}
```

ReneWerner87: yes please remove all with the same path

gaby: ^ Agree

ckoch786:
> @ckoch786 Any progress on this?

@gaby No, sorry. Having to set up a dev environment where I could get the test to run properly set me back. That, and after seeing the number of tests that were failing, I was thinking that I just need to rip out the duplicate route detection logic because getting all the tests to pass is unfortunately going to take more time than I have available at the moment. If that is fine, then I can commit some time over the next few days to get the tests that are for the new RemoveRoute methods, since those do not touch any existing code. I will also update the docs.
Also, I see that you have made some changes to my PR. Thank you for not letting it get super stale!

gaby:
> > @ckoch786 Any progress on this?
> 
> @gaby No, sorry. Having to set up a dev environment where I could get the test to run properly set me back. That, and after seeing the number of tests that were failing, I was thinking that I just need to rip out the duplicate route detection logic because getting all the tests to pass is unfortunately going to take more time than I have available at the moment. If that is fine, then I can commit some time over the next few days to get the tests that are for the new RemoveRoute methods, since those do not touch any existing code. I will also update the docs. Also, I see that you have made some changes to my PR. Thank you for not letting it get super stale!

I believe that @efectn has taken over the changes to this after some internal discussions. So no worries 💪

ckoch786:
> @ReneWerner87 @efectn I vote we close this. Spent 1hr and can't find the issue, all the tests are failing. Author is not responding

@gaby I appreciate you trying to find the issue. Now I'm even more wary of messing with the duplicate route detection, as I don't know enough about Fiber at the moment to safely make this change. I was hoping that eliminating the ability to duplicate routes would not conflict with anything. It's a good thing that the tests caught it because my manual testing,  did not turn up any issues.

docs/api/app.md
```diff
@@ -761,3 +761,62 @@ func main() {
...
+import (
+    "log"
+
+    "github.com/gofiber/fiber/v3"
+)
+
+func main() {
+    app := fiber.New()
+
+    app.Get("/api/feature-a", func(c *fiber.Ctx) error {
```
ReneWerner87, line 797:
```suggestion
    app.Get("/api/feature-a", func(c fiber.Ctx) error {
```

docs/api/app.md
```diff
@@ -761,3 +761,62 @@ func main() {
...
+           app.RebuildTree()
+           // Redefine route
+           app.Get("/api/feature", func(c *fiber.Ctx) error {
+                   return c.SendString("Testing feature-a")
+           })
+
+           app.RebuildTree()
+           return c.SendStatus(fiber.StatusOK)
+    })
+    app.Get("/api/feature-b", func(c *fiber.Ctx) error {
```
ReneWerner87, line 808:
```suggestion
    app.Get("/api/feature-b", func(c fiber.Ctx) error {
```

docs/api/app.md
```diff
@@ -761,3 +761,62 @@ func main() {
...
+           // Redefine route
+           app.Get("/api/feature", func(c *fiber.Ctx) error {
+                   return c.SendString("Testing feature-b")
+           })
+
+           app.RebuildTree()
+           return c.SendStatus(fiber.StatusOK)
+    })
+
+    log.Fatal(app.Listen(":3000"))
```
ReneWerner87, line 820: pls also test the example in an example test case to verify the function

router.go
```diff
@@ -313,6 +314,92 @@ func (*App) copyRoute(route *Route) *Route {
...
+// RemoveRouteFunc is used to remove a route from the stack by a custom match function.
+// This only needs to be called to remove a route, route registration prevents duplicate routes.
+// You should call RebuildTree after using this to ensure consistency of the tree.
+// Note: The route.Path is original path, not the normalized path.
+func (app *App) RemoveRouteFunc(matchFunc func(r *Route) bool, methods ...string) {
+	app.deleteRoute(methods, matchFunc)
+}
+
+func (app *App) deleteRoute(methods []string, matchFunc func(r *Route) bool) {
+	if len(methods) == 0 {
```
ReneWerner87, line 363: since the methods are required for all parent methods
we should not make it optional

maybe something like 
```go
func (app *App) RemoveRouteFunc(matchFunc func(r *Route) bool, method string, methods ...string)
```

or 

```go
func (app *App) RemoveRouteFunc(matchFunc func(r *Route) bool, method string[])
```
is better

ReneWerner87: Or we use all registered Methods as default -> then its okay to have it optional

router.go
```diff
@@ -313,6 +314,92 @@ func (*App) copyRoute(route *Route) *Route {
...
+			// Decrement global handler count. In middleware routes, only decrement once
+			if _, ok := removedUseRoutes[route.path]; (route.use && slices.Equal(methods, app.config.RequestMethods) && !ok) || !route.use {
+				if route.use {
+					removedUseRoutes[route.path] = struct{}{}
+				}
+
+				atomic.AddUint32(&app.handlersCount, ^uint32(len(route.Handlers)-1)) //nolint:gosec // Not a concern
+			}
+		}
+	}
```
ReneWerner87, line 400: we could also directly call the RebuildTree function , then the users will not forget it

ReneWerner87 (changes requested):
Some small hints

router.go
efectn: what do you think @gaby @sixcolors

router.go
efectn: Yes i will use default methods when methods is empty

gaby:
@coderabbitai full review

docs/api/app.md
```diff
@@ -761,3 +761,65 @@ func main() {
...
+
+This method removes a route by name.
+If no methods are specified, the route will be removed for all HTTP methods defined in the app. To limit removal to specific methods, provide them as additional arguments.
+
+```go title="Signature"
+func (app *App) RemoveRouteByName(name string, methods ...string)
+```
+
+This method removes a route by function having `*Route` parameter.
+If no methods are specified, the route will be removed for all HTTP methods defined in the app. To limit removal to specific methods, provide them as additional arguments.
```
gaby, line 782: This same paragraph is repeated 3 times within 100 lines

router.go
gaby: If you want to remove more than 1 route, then Rebuild will be called multiple times

gaby (approved):
👍 LGTM, although I'd remove that duplicated paragraph from the docs

ReneWerner87 (approved):

PR #3242, opened 2024-12-10
📚 doc: Update encrypt cookie usage of OpenSSL

imorph:
# Description

This is pure documentation fix. Default example for openssl command line (at crypto-cookie doc) producing more than 32 character line. Base64 encoding expands the output size by approximately 33%, so needed adjustments are: 
* down number of random bytes requested to 24
* cutting result to exact 32 chars.


gaby:
@imorph The docs are correct. We check the key length after base64 decoding, so it will be 32 after decoding.

You can see that here: https://github.com/gofiber/fiber/blob/main/middleware/encryptcookie/utils.go#L54

gaby:
Line 41 is wrong in the docs though. That example should have a base64 encoded key.

nickajacks1:
> Base64 encoding expands the output size by approximately 33%

This is actually expected. The base64 representation isn't the key itself, but rather the base64 encoded representation, which indeed is longer than 32 bytes.

PR #3263, opened 2024-12-25
🩹 Fix: Memory leak removal in the idempotency middleware

### LINKED ISSUE #3262: 🐛 [Bug]: Memory leak removal in the idempotency middleware

The MemoryLock in [locker.go](https://github.com/gofiber/fiber/blob/main/middleware/idempotency/locker.go) is implemented so that allocated mutexes won't be collected by GC because it holds pointers to them even when it is not needed.
To be accurate: 
It allocates new mutex for every idempotency key (and removes nowhere)
```go
func (l *MemoryLock) Lock(key string) error {
	l.mu.Lock()
	mu, ok := l.keys[key]
	if !ok {
		mu = new(sync.Mutex)
		l.keys[key] = mu
	}
	l.mu.Unlock()

	mu.Lock()

	return nil
}
```
It might seem be okay until some calculations are made: 
Imagine we have a server handling 1000 rps with idempotency keys. Every mutex holds 8 bytes.
So in 1 month of continuous work it will be 8 * 1000 * 60 * 60 * 24 * 30 = 20736000000 B = 19.312 GB 

I propose key removal from map to allow GC collect unused Mutex. (Associated PR is created)

N/A

N/A

### API Stability

N/A

### Feature Examples

```go
N/A
```

### Checklist:

sunnyyssh:
# Description

The MemoryLock in [locker.go](https://github.com/gofiber/fiber/blob/main/middleware/idempotency/locker.go) doesn't let GC collect unused mutexes.
I propose key removal from map to let GC collect unused Mutex. 

Fixes #3262

## Changes introduced

I replaced just `sync.Mutex` with 
```go
type countedLock struct {
    mu     sync.Mutex
    locked int
}
```
It will count the difference between Lock and Unlock count and if it equals 0 then the key can be removed from `keys` map within MemoryLock's own mutex lock.
```
type MemoryLock struct {
    keys map[string]*countedLock
    mu   sync.Mutex
}
```
So, now GC is happy to collect 😊


gaby:
@sunnyyssh Unit-tests are failing.

middleware/idempotency/idempotency_test.go
```diff
@@ -83,7 +83,7 @@ func Test_Idempotency(t *testing.T) {
 			req.Header.Set("X-Idempotency-Key", idempotencyKey)
 		}
 		resp, err := app.Test(req, fiber.TestConfig{
-			Timeout:       15 * time.Second,
+			Timeout:       115 * time.Second,
```
gaby, line 86: Hmmm

gaby (commented):
This doesnt make sense?

middleware/idempotency/idempotency_test.go
sunnyyssh: oh, sorry haha. Now this commit is not included btw

sunnyyssh:
Oh, but single decreased benchmark doesn't even depend on the changed code...
image

gaby:
@sunnyyssh Can you add a benchmark in `locker_test.go`

sunnyyssh:
> @sunnyyssh Can you add a benchmark in `locker_test.go`

The main idea is that one idempotency key wouldn't be used once again frequently just because the operation completes successfully in most cases. So, trying to benchmark the performance of "key-deleting" MemoryLock with frequently repeated keys, we will have not real-world results. 

So, I tried to make benchmarks checking new and old implementations and got interesting result. 
```go
func Benchmark_MemoryLock(b *testing.B) {
	keys := make([]string, b.N)
	for i := range keys {
		keys[i] = strconv.Itoa(i)
	}

	lock := idempotency.NewMemoryLock()

	b.ResetTimer()

	for i := 0; i < b.N; i++ {
		key := keys[i]
		if err := lock.Lock(key); err != nil {
			b.Fatal(err)
		}
		if err := lock.Unlock(key); err != nil {
			b.Fatal(err)
		}
	}
}
```
Old implementation has shown extremely inefficient result

image

Trying to figure out why old implementation has so bad performance and memory allocations I used pprof and got this:
[The line I point to](https://github.com/gofiber/fiber/blob/f8b490f89ee5a58108f7902ff5b02e5006f2802d/middleware/idempotency/locker.go#L23)
image

map is inefficient for storing big amounts of data because it always re-allocates its buckets when resizing.
So, it improves not only long-running memory allocation.

middleware/idempotency/locker_test.go
```diff
@@ -59,3 +61,67 @@ func Test_MemoryLock(t *testing.T) {
...
+					b.Fatal(err)
+				}
+				if err := lock.Unlock(key); err != nil {
+					b.Fatal(err)
+				}
+			}
+		})
+	})
+
+	b.Run("Repeated3TimesKeys", func(b *testing.B) {
```
gaby, line 110: Rename this to "RepeatedKeys"

gaby (approved):
👍 LGTM

ReneWerner87 (approved):

PR #3264, opened 2024-12-28
🚀 feat: Add Load-Shedding Middleware

ErfanMomeniii:
# Description

This pull request introduces a new **Load-Shedding Middleware**. This middleware enforces a configurable timeout on request processing to effectively manage server load. If a request exceeds the specified timeout, a custom load-shedding handler is invoked to gracefully handle the overloaded state.

**Purpose of the Change:**
- Improve server resilience under high-load scenarios.
- Ensure that long-running or unresponsive requests do not degrade overall system performance.
- Provide developers with a flexible solution for request timeouts, including options for exclusions.

## Changes introduced

This pull request adds the following features and updates:

- **Load-Shedding Middleware (`loadshedding.New`):** 
  - Allows enforcement of request timeouts with a custom handler for requests exceeding the timeout.
  - Supports route exclusions using a configurable predicate function.

- **Unit Tests:** 
  - `Test_LoadSheddingExcluded`: Verifies that excluded routes bypass the middleware.
  - `Test_LoadSheddingTimeout`: Tests behavior when requests exceed the timeout.
  - `Test_LoadSheddingSuccessfulRequest`: Confirms successful processing of requests completed within the timeout.

- **Documentation Update:** Detailed usage examples for the middleware are added.

- **Benchmarks:** Basic performance benchmarks have been conducted to ensure minimal overhead.

- **Changelog/What's New:** This middleware addition will be documented in the changelog for the next release.

- **API Longevity:** Designed with consistency in mind to align with Fiber's existing middleware and ensure backward compatibility.

## Type of change

## Checklist

## Commit formatting

Commits follow the recommended format with appropriate emojis for better identification:
- 🚀 feat: add load-shedding middleware for request timeout management
- ✅ test: add unit tests for load-shedding middleware


gaby:
@ErfanMomeniii You should be able to replicate the CI using the provided Makefile.

Run this: `make test`

https://github.com/gofiber/fiber#-development

gaby:
I believe that bug 1 is the cause of the tests failing. We should also add the improvement 1 from the below review.

This review was generated using OpenAI `o1-pro`:

---

## Overall Impression

The middleware is straightforward and mostly does what it promises:

- It creates a `context.WithTimeout` around the request.  
- Spawns a goroutine to run `c.Next()`.  
- If the request is not finished within the specified timeout, it calls the `loadSheddingHandler`.  

Its behavior is reasonably clear, and the accompanying tests show various scenarios (excluded routes, timeouts, and successful requests).

However, there are a few issues and potential improvements worth pointing out.

---

## Bugs / Potential Issues

1. **Goroutine Resource Leak**  
   When you call  
   ~~~go
   go func() {
       done <- c.Next()
   }()
   ~~~
   you rely on the context’s cancellation to interrupt `c.Next()`, but in some cases, a handler might not properly handle the canceled context. If the Fiber handler does not actually stop execution upon context cancellation, that goroutine could continue doing work until completion. Be sure that all handlers in your stack properly respect `ctx.Done()` to avoid “zombie” goroutines.

2. **Partial Responses / Streaming Concerns**  
   If a handler is streaming data back to the client, the scenario where the context is canceled mid-stream isn’t explicitly handled. You might end up sending partial data before the load-shedding logic triggers. Many setups buffer responses, but if your application does any streaming, you need to ensure the context cancellation is heeded at the streaming level as well.

3. **Tight Coupling of Timeout and Load Shedding**  
   Currently, this is purely a request-timeout mechanism. The `loadSheddingHandler` name might suggest more comprehensive load shedding (like concurrency thresholds or resource-based triggers). Consider renaming to something like `timeoutHandler` if you don’t plan to expand its functionality, or incorporate other load metrics if you do.

---

## Potential Improvements

1. **Explicit Cancellation Support**  
   To ensure that `c.Next()` and deeper handlers truly respect the timeout, consider passing the new context down. For example:
   ```go
   c.SetContext(ctx)
   err := c.Next()
   ```
   so that any `c.Context()` call within subsequent handlers references the timed-out context. This helps them detect and act on cancellations more gracefully.

2. **Configurable Timeouts**  
   You might want per-route or dynamic timeouts. A function signature like:
   ```go
   func(c fiber.Ctx) time.Duration
   ```
   to compute the timeout dynamically could be helpful, allowing you to tailor timeouts based on route complexity, user tiers, or other metrics.

3. **Allow for Additional Fallbacks**  
   Instead of a single `loadSheddingHandler`, some might want more nuanced fallback options, such as returning partial/cached data. The current design only allows one fallback function. You could consider a pattern that routes different fallback behaviors depending on the endpoint or error condition.

4. **Concurrency-Based or Adaptive Load Shedding**  
   If the goal is to handle heavy load more generally, consider tracking concurrency (e.g., number of ongoing requests) and shedding requests when thresholds are exceeded. A time-based approach alone may not suffice under extreme load.

5. **Use a Named Goroutine**  
   For debugging, consider naming the goroutine (or logging the request path) to track which request is timing out or getting stuck. This can help in diagnosing issues more quickly.

---

## Missing Features / Wishlist

- **Graceful Shutdown Hooks**  
  If you’re stopping the server, you might want the load-shedding middleware to know about the shutdown event. This way, it can immediately return a “service unavailable” rather than waiting for timeouts or continuing to accept requests that may never complete.

ErfanMomeniii:
Hi @gaby, it should be fixed now.

ErfanMomeniii:
Hi again @gaby , I’ve resolved the lint CI issue as well, so I believe everything is ready for merging now.
Thanks

gaby:
@ErfanMomeniii Code and Tests look fine to me. Two things missing:
- Add a section here https://github.com/gofiber/fiber/blob/main/docs/whats_new.md
- Add documentation for the middleware here: https://github.com/gofiber/fiber/tree/main/docs/middleware
  - You can copy one of the middlewares, for example `redirect.md` and update it accordingly.

ErfanMomeniii:
it's done. @gaby

ErfanMomeniii:
The lint problem should be fixed now. @gaby 
Thanks

ErfanMomeniii:
The lint problem should be fixed now and it's ready to merge. @gaby 
Thanks

gaby (approved):
👍 LGTM

docs/middleware/loadshedding.md
```diff
@@ -0,0 +1,81 @@
...
+func New(timeout time.Duration, loadSheddingHandler fiber.Handler, exclude func(fiber.Ctx) bool) fiber.Handler
+```
+
+## Config
+
+| Property              | Type               | Description                                                                      | Default   |
+|-----------------------|--------------------|----------------------------------------------------------------------------------|-----------|
+| `timeout`            | `time.Duration`    | The maximum allowed processing time for a request.                               | Required  |
+| `loadSheddingHandler`| `fiber.Handler`    | The handler invoked for requests that exceed the `timeout`.                      | Required  |
+| `exclude`            | `func(fiber.Ctx) bool` | Optional function to exclude certain requests from load-shedding logic.       | `nil`     |
```
ReneWerner87, line 36: we have similar behavior in all other middlewares in the core repo and also in the contrib repository
unless it is completely wrong i would like to get the wording of this parameter 

the users know the parameter as `Next` function for skipping
https://docs.gofiber.io/search/?q=Next&version=next%2F

docs/middleware/loadshedding.md
```diff
@@ -0,0 +1,81 @@
...
+- **High-Traffic Scenarios**: Safeguard critical resources by rejecting overly long or resource-intensive requests.
+- **Health Check Protection**: Exclude monitoring endpoints (e.g., `/health`) to ensure uninterrupted external checks.
+- **Dynamic Load Management**: Utilize exclusion logic to adjust load-shedding behavior for specific routes or request types.
+
+---
+
+## Signature
+
+```go
+func New(timeout time.Duration, loadSheddingHandler fiber.Handler, exclude func(fiber.Ctx) bool) fiber.Handler
```
ReneWerner87, line 27: in order to ensure expandability in the future and not have to make breaking changes, i would like to stick with the configuration concept, which we use in all other middlewares and which has proven to be very easy to expand

.github/README.md
```diff
@@ -688,6 +688,7 @@ Here is a list of middleware that are included within the Fiber framework.
 | [idempotency](https://github.com/gofiber/fiber/tree/main/middleware/idempotency)     | Allows for fault-tolerant APIs where duplicate requests do not erroneously cause the same action performed multiple times on the server-side.                           |
 | [keyauth](https://github.com/gofiber/fiber/tree/main/middleware/keyauth)             | Adds support for key based authentication.                                                                                                                              |
 | [limiter](https://github.com/gofiber/fiber/tree/main/middleware/limiter)             | Adds Rate-limiting support to Fiber. Use to limit repeated requests to public APIs and/or endpoints such as password reset.                                             |
+| [loadshedding](https://github.com/gofiber/fiber/tree/main/middleware/loadshedding)   | Gracefully manages server load by enforcing request timeouts and handling resource-intensive requests during high-traffic periods.                                      |
```
ReneWerner87, line 691: what is the difference to the timeout middleware? could you also store a handler in the config that handles these cases?

ReneWerner87 (changes requested):
some adjustments necessary

ReneWerner87:
@ErfanMomeniii thank you for your contribution and the time you have invested to support us

we have analyzed the middleware again and also found another similar one
only middlewares that have no dependencies and are used by a large part of every server in the community should be in the core
we just realized that we already have this middleware in the contrib packet, but not with the timeout functionality
https://github.com/gofiber/contrib/tree/main/loadshed 

i also can't say 100% right now whether i would rather provide the reaction to a timeout in a loadshed middleware or as an option in the timeout middleware
https://docs.gofiber.io/api/middleware/timeout

the current options would be to extend the timeout middleware with a handler for reacting to processes that run into the timeout or to extend the existing loadshred middleware in contrib

other examples for loadshed
https://github.com/kevinconway/loadshed

PR #3288, opened 2025-01-23
Prevent Override `Max` value by default value when passing `0` to `Max`

ad3n:
# Description

This PR makes this code https://github.com/gofiber/fiber/blob/main/middleware/limiter/limiter_fixed.go#L33 reachable, because when we supply the Max to `0`, it always override by default value


gaby:
@ad3n What's the point of setting this to 0?

gaby:
You can use `MaxFunc`, to override the config.Max

https://github.com/gofiber/fiber/blob/main/middleware/limiter/config.go#L30

ad3n:
> You can use `MaxFunc`, to override the config.Max
> 
> https://github.com/gofiber/fiber/blob/main/middleware/limiter/config.go#L30

The behavior when we set `Max` to `0` is confusing, i as developer imagine, when i put zero (for example, the value come from DB), i expected it would be skip the limiter, but actually the value overridden by default value (5)

gaby:
> > You can use `MaxFunc`, to override the config.Max
> > https://github.com/gofiber/fiber/blob/main/middleware/limiter/config.go#L30
> 
> The behavior when we set `Max` to `0` is confusing, i as developer imagine, when i put zero (for example, the value come from DB), i expected it would be skip the limiter, but actually the value overridden by default value (5)

Why use the limiter then? It adds quite a bit of overhead.

ad3n:
> > > You can use `MaxFunc`, to override the config.Max
> > > https://github.com/gofiber/fiber/blob/main/middleware/limiter/config.go#L30
> > 
> > 
> > The behavior when we set `Max` to `0` is confusing, i as developer imagine, when i put zero (for example, the value come from DB), i expected it would be skip the limiter, but actually the value overridden by default value (5)
> 
> Why use the limiter then? It adds quite a bit of overhead.

The limiter is mandatory on my case, so i just make it simple to pass all to the limiter and then the limiter control the flow. As simple as that

gaby:
> > > > You can use `MaxFunc`, to override the config.Max
> > > > https://github.com/gofiber/fiber/blob/main/middleware/limiter/config.go#L30
> > > 
> > > 
> > > The behavior when we set `Max` to `0` is confusing, i as developer imagine, when i put zero (for example, the value come from DB), i expected it would be skip the limiter, but actually the value overridden by default value (5)
> > 
> > 
> > Why use the limiter then? It adds quite a bit of overhead.
> 
> The limiter is mandatory on my case, so i just make it simple to pass all to the limiter and then the limiter control the flow. As simple as that

If you set it to 0, the limiter is skipped and `Next()` is called. Basically disabling the limiter.

https://github.com/gofiber/fiber/blob/main/middleware/limiter/limiter_fixed.go#L33

ad3n:
> > > > > You can use `MaxFunc`, to override the config.Max
> > > > > https://github.com/gofiber/fiber/blob/main/middleware/limiter/config.go#L30
> > > > 
> > > > 
> > > > The behavior when we set `Max` to `0` is confusing, i as developer imagine, when i put zero (for example, the value come from DB), i expected it would be skip the limiter, but actually the value overridden by default value (5)
> > > 
> > > 
> > > Why use the limiter then? It adds quite a bit of overhead.
> > 
> > 
> > The limiter is mandatory on my case, so i just make it simple to pass all to the limiter and then the limiter control the flow. As simple as that
> 
> If you set it to 0, the limiter is skipped and `Next()` is called. Basically disabling the limiter.
> 
> https://github.com/gofiber/fiber/blob/main/middleware/limiter/limiter_fixed.go#L33

Yup, i think the limiter will skipped, but on config, when i supply `0` on `Max`, it would be override by default value and change the `Max` to `5`. That's the problem

So i create this PR to make sure, the default config not override it when i supply `0` on `Max`

ReneWerner87:
the default values of an empty structs always correspond to the default value of the datenTypes
for int this is 0, so we use this to set the default max value

in your pull request you set the condition for the max value to below 0, which will not do anything
also the middleware needs a max value to execute the functionality for which it was created

when it comes to skipping the middleware functionality, please use other solutions like using the [`Next`](https://docs.gofiber.io/next/middleware/limiter#config) method for the condition whether the middleware should be executed or not or the [`skip`](https://docs.gofiber.io/next/middleware/skip) middleware or define the limiter middleware only on the [`routes`](https://docs.gofiber.io/next/middleware/skip) where it is needed

PR #3294, opened 2025-01-26
🧹 chore: Backport ctx.String() from v3

### LINKED ISSUE #3236: 🐛 [Bug]: Immutable configuration in Fiber not working as expected
### Bug Description

The `Immutable` configuration option in Fiber, when set to true, is intended to ensure that certain request values (e.g., request bodies) are immutable and accessible beyond the handler's lifecycle. However, enabling this option does not work as expected, resulting in data race errors when the request body is processed within a differernt goroutine.

Documentation Says:

> When set to true, this relinquishes the 0-allocation promise in certain cases in order to access the handler values (e.g., request bodies) in an immutable fashion so that these values are available even if you return from the handler.

Am I missing something here, or is this the expected behavior, and I need to manually copy the body to ensure immutability in my code?

### How to Reproduce

1. Create a Fiber app with the Immutable and StreamRequestBody configurations set to true
```go
    Immutable:         true,
    StreamRequestBody: true,
})
```

2. Define a route where the request body is processed in new goroutine:
```go
app.Get("/keep", func(c *fiber.Ctx) error {
    body := c.Body() // Supposed to be immutable
    go func() {
        for i := 0; i < 100; i++ {
            time.Sleep(time.Second)
            var data map[string]any
            err := json.Unmarshal(body, &data)
            fmt.Println(err, data) // Accessing body in a goroutine
        }
    }()
    return c.JSON(map[string]any{"success": true})
})
```

3. send multiple concurrent requests
```curl
curl --request GET 'http://localhost:8080/keep' \
--header 'Content-Type: application/json' \
--data-raw '{"key": "val"}'
```

### Expected Behavior

The request body `c.Body()` should remain immutable and safe to access across goroutines without causing data races but actually get a data races

```
==================
WARNING: DATA RACE
Read at 0x00c00011a080 by goroutine 76:
  encoding/json.checkValid()
  ...
Previous write at 0x00c00011a080 by goroutine 75:
  runtime.slicecopy()
  ...
==================
```

### Fiber Version

v2.52.5

### Code Snippet (optional)

```go
package main

import "log"

func main() {
  app := fiber.New()

  // Steps to reproduce

}
```

### Checklist:

gaby:
# Description

Backport implementation of `ctx.String()` from v3 to v2.

Fixes #3236

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -1834,14 +1834,39 @@ func (c *Ctx) Status(status int) *Ctx {
 //
 // The returned value may be useful for logging.
 func (c *Ctx) String() string {
-	return fmt.Sprintf(
-		"#%016X - %s <-> %s - %s %s",
-		c.fasthttp.ID(),
-		c.fasthttp.LocalAddr(),
-		c.fasthttp.RemoteAddr(),
-		c.fasthttp.Request.Header.Method(),
-		c.fasthttp.URI().FullURI(),
-	)
+	// Get buffer from pool
+	buf := bytebufferpool.Get()
+
+	// Start with the ID, converting it to a hex string without fmt.Sprintf
+	buf.WriteByte('#')
+	// Convert ID to hexadecimal
+	id := strconv.FormatUint(c.fasthttp.ID(), 16)
+	// Pad with leading zeros to ensure 16 characters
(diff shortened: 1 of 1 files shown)


gaby:
golangci-lint fails here, but not locally.

PR #3333, opened 2025-02-27
🔥 feat: Add Skip function to logger middleware

### LINKED ISSUE #3300: Feature Request: Add a Pre-Write Callback to Logger Middleware
Currently, the `logger.New` middleware in Fiber writes logs to the specified `Output` file automatically.  While the `Done` callback allows for post-processing of the log string, there's no way to control *whether* a log entry is written to the file at all *before* it's written. This makes it difficult to implement fine-grained logging logic, such as selectively logging certain responses based on their content or other criteria.

**Problem:**

I have a use case where I want to log only specific errors (e.g., 4xx/5xx status codes) to an error log file, and potentially other requests (e.g., 2xx) to a separate debug or access log file.  The current `Done` callback is called *after* the log has already been written to the `Output` file.  This results in either:

1. **Double Logging:** If I use `Done` to filter logs and write to my error log file, the middleware also writes all logs to the `Output` file, leading to duplication.
2. **Inability to Selectively Log:** If I don't use `Done`, I can't filter which log entries are written to which file.

**Code from my use case**

```go
app.Use(logger.New())
app.Use(logger.New(logger.Config{
	TimeFormat: "2006-01-02 15:04:05",
	TimeZone:   "Local",
	Format:     "\n" + `{"timestamp":"${time}", "status":${status}, "latency":"${latency}", "method":"${method}", "path":"${path}", "error":"${error}", "response":[${resBody}]}`,
	Output:     errorLogFile,
	Done: func(c *fiber.Ctx, logString []byte) {
		if c.Response().StatusCode() >= 400 {
			fmt.Println("printing to error.log")
			errorLogFile.Write(logString)
			defer errorLogFile.Close()
		}
	},
}))
```
**Proposed Solution/Feature request:**

I propose adding a new callback function to the `logger.Config` called something like `BeforeWrite` or `ShouldLog`. This callback would be executed *before* the middleware writes the log entry to the `Output` file.  The callback would receive the `fiber.Ctx` and the log string as arguments and would return a boolean value: `true` if the log should be written, and `false` if it should be skipped.

**Example Usage:**

```go
app.Use(logger.New(logger.Config{
    TimeFormat: "2006-01-02 15:04:05",
    TimeZone:   "Local",
    Format:     `{"timestamp":"${time}", "status":${status}, ...}`, //Log format
    Output:     logFile, // General log file (optional)
    BeforeWrite: func(c *fiber.Ctx, logString []byte) bool {
        if c.Response().StatusCode() >= 400 {
            // Log errors to the error log file
            errorLogFile.Write(logString) // Example, we can write to different files here.
            return true // Write to the main log file as well (optional). We can return false if we only want to write to the error log.
        } else if c.Response().StatusCode() == 200 {
            // Log 200 OK requests to a separate access log
            accessLogFile.Write(logString)
            return true // Write to the main log file as well (optional). We can return false if we only want to write to the access log.
        }
        return false // Don't write other requests to the main log file.
    },
    Done: func(c *fiber.Ctx, logString []byte) {
        // Any post-processing 
    },
}))

### PR DESCRIPTION
**Pull request by JIeJaitt**
# Description

The `Filter` is a function that is called before the log string for a request is written to Output. If it returns true, the log will be written; otherwise, it will be skipped.

Fixes #3300

### CODE REVIEW

**Comment by JIeJaitt** [comment]
Strange, this unit test runs locally for me without any problem, but reports an error in GitHub Action

File: middleware/logger/config.go [hunk 1]
```diff
@@ -20,6 +20,12 @@ type Config struct {
 	// Optional. Default: nil
 	Next func(c fiber.Ctx) bool
 
+	// Filter is a function that is called before writing the log string.
+	// If it returns true, the log will be written; otherwise, it will be skipped.
+	//
+	// Optional. Default: nil
+	Filter func(c fiber.Ctx) bool
```
ReneWerner87, line 27: would say that we also need the logstring here, as perhaps filtering should happen based on it

ReneWerner87:
> @ReneWerner87 hi, there Strange, this unit test runs locally for me without any problem, but reports an error in GitHub Action

did you also run this local with the “-race” flag ?

middleware/logger/config.go
gaby: @ReneWerner87 What do you mean by logstring?

gaby:
> > @ReneWerner87 hi, there Strange, this unit test runs locally for me without any problem, but reports an error in GitHub Action
> 
> did you also run this local with the “-race” flag ?

I just ran them locally with `make test` and it doesn't fail.

gaby:
> > > @ReneWerner87 hi, there Strange, this unit test runs locally for me without any problem, but reports an error in GitHub Action
> > 
> > 
> > did you also run this local with the “-race” flag ?
> 
> I just ran them locally with `make test` and it doesn't fail.

Using `make longtest` does trigger the same race condition we are seeing in the CI

gaby:
> > > > @ReneWerner87 hi, there Strange, this unit test runs locally for me without any problem, but reports an error in GitHub Action
> > > 
> > > 
> > > did you also run this local with the “-race” flag ?
> > 
> > 
> > I just ran them locally with `make test` and it doesn't fail.
> 
> Using `make longtest` does trigger the same race condition we are seeing in the CI

Pretty sure this is related to re-using the same `app` across all the new tests introduced in this PR. They are both running in parallel and registering `/`

middleware/logger/config.go
JIeJaitt: > @ReneWerner87 What do you mean by logstring?

may be like this: https://github.com/gofiber/fiber/blob/86cf80630b3ee6275c189592b0594ab8c23d5a00/middleware/logger/config.go#L27

middleware/logger/logger_test.go
```diff
@@ -171,6 +171,68 @@ func Test_Logger_Done(t *testing.T) {
...
+		}))
+
+		resp, err := app.Test(httptest.NewRequest(fiber.MethodGet, "/", nil))
+		require.NoError(t, err)
+		require.Equal(t, fiber.StatusOK, resp.StatusCode)
+
+		// Verify the log output does not contain the "200" message
+		require.NotContains(t, logOutput.String(), "200") // Check if the log does not contain "200"
+	})
+}
```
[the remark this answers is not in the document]
gaby: Are you sure?

middleware/logger/config.go
gaby: I dont think we need it since this is only defined during the config?

The user can't call this function. So the logfilter or whatever the filter uses will be inside that function. Right?

JIeJaitt: > I dont think we need it since this is only defined during the config?
> 
> The user can't call this function. So the logfilter or whatever the filter uses will be inside that function. Right?

I agree with you.

ReneWerner87: image
from the feature request https://github.com/gofiber/fiber/issues/3300

@gaby @JIeJaitt

middleware/logger/default_logger.go
```diff
@@ -90,6 +90,11 @@ func defaultLoggerInstance(c fiber.Ctx, data *Data, cfg Config) error {
 			buf.WriteString("\n")
 		}
 
+		// Check if Filter is defined and call it
+		if cfg.Filter != nil && !cfg.Filter(c) {
```
gaby, line 94: @JIeJaitt This can be moved to the beginning of the function. There's no point is running all that only to have the filter return True and skip 😂

gaby (changes requested):

middleware/logger/config.go
JIeJaitt: > image from the feature request #3300
> 
> @gaby @JIeJaitt

No problem, I'll add it

gaby: @JIeJaitt We don't need it.

gaby:
@JIeJaitt Add a section in docs/whats_new.md should be ready to merge after that.

JIeJaitt:
@gaby I thought about it and you are right. The filter() function can focus on the decision-making logic without having to care about the specific processing details of the log string.

middleware/logger/default_logger.go
```diff
@@ -15,6 +15,11 @@ import (
 
 // default logger for fiber
 func defaultLoggerInstance(c fiber.Ctx, data *Data, cfg Config) error {
+	// Check if Filter is defined and call it
+	if cfg.Filter != nil && !cfg.Filter(c) {
```
gaby, line 19: @JIeJaitt I think this is wrong. It should be:

```go
if cfg.Filter != nil && cfg.Filter(c)
```
In the current implementation it's using a negated approach which is not user friendly.

docs/whats_new.md
```diff
@@ -940,6 +940,26 @@ The Healthcheck middleware has been enhanced to support more than two routes, wi
 
 Refer to the [healthcheck middleware migration guide](./middleware/healthcheck.md) or the [general migration guide](#-migration-guide) to review the changes.
 
+### Filter
+
+The `Filter` is a function that is called before the log string for a request is written to Output. If it returns true, the log will be written; otherwise, it will be skipped.
```
gaby, line 945: @JIeJaitt You can see how its confusing by reading this paragraph

middleware/logger/default_logger.go
gaby: Nevermind. ChatGPT agreed.

> Return true if the log entry should pass the filter (be logged) and false otherwise. To make it more user-friendly, clearly document it or rename it (e.g., ShouldLog()) so the meaning of returning true is obvious.

gaby:
@coderabbitai full review

docs/whats_new.md
```diff
@@ -940,6 +940,26 @@ The Healthcheck middleware has been enhanced to support more than two routes, wi
 
 Refer to the [healthcheck middleware migration guide](./middleware/healthcheck.md) or the [general migration guide](#-migration-guide) to review the changes.
 
+### Filter
```
gaby, line 943: @JIeJaitt This should be under the Logger section in line 872

gaby (changes requested):
One last thing

docs/whats_new.md
```diff
@@ -912,6 +912,29 @@ func main() {
...
+The `Filter` is a function that is called before the log string for a request is written to Output. If it returns true, the log will be written; otherwise, it will be skipped.
+
+<details>
+<summary>Example</summary>
+
+```go
+app.Use(logger.New(logger.Config{
+    Filter: func(c fiber.Ctx) bool {
+        // Skip logging for 404 requests
+        return c.Response().StatusCode() == fiber.StatusNotFound
```
gaby, line 924: Shouldn't this be `!=` ?

If its 404 it returns True, which the defaultLogger negates into False and doesn't filter

middleware/logger/logger_test.go
```diff
@@ -171,6 +171,58 @@ func Test_Logger_Done(t *testing.T) {
 	require.Positive(t, buf.Len(), 0)
 }
 
+func Test_Logger_Filter(t *testing.T) {
```
gaby, line 174: I'm confused about how this is working.

If the condition is True, it will become False in defaultLogger. But we dont check for that in the test?

JIeJaitt: @gaby Thanks for the heads up. I've just checked and found a few things that need to be refined, it's almost midnight here, I'll submit a more complete unit test and documentation note tomorrow.

gaby: Sounds good. Thanks!

docs/whats_new.md
ReneWerner87: well, if he only wants to log every 404, then that should be fine
depends on what he wants to do

it is always logged if the condition is true
-> if of course he wants everything else to be logged except 404, then an unequal should be used

ReneWerner87: @JIeJaitt

JIeJaitt: @ReneWerner87 I've added new examples in the documentation, you can take a look at the

```go
app.Use(logger.New(logger.Config{
    Filter: func(c fiber.Ctx) bool {
        // log status code >= 400
        return c.Response().StatusCode() >= fiber.StatusBadRequest
    },
}))

app.Use(logger.New(logger.Config{
    Filter: func(c fiber.Ctx) bool {
        // log status code == 404
        return c.Response().StatusCode() == fiber.StatusNotFound
    },
}))

app.Use(logger.New(logger.Config{
    Filter: func(c fiber.Ctx) bool {
        // log status code != 200
        return c.Response().StatusCode() != fiber.StatusOK
    },
}))
```

ReneWerner87 (approved):

gaby:
@ksw2000 @efectn Can you take a look at this PR? Thanks

gaby:
@JIeJaitt I updated the logic per recommendation from `OpenAI o1` and added more unit tests

gaby (approved):
👍 LGTM

gaby:
Renamed:
- Function `Filter()` -> `Skip()`
- Field `Output` -> ` Stream`

These now match the official expressjs middleware https://github.com/expressjs/morgan

ReneWerner87 (approved):

JIeJaitt:
👍 LGTM

middleware/logger/default_logger.go
```diff
@@ -15,6 +15,12 @@ import (
 
 // default logger for fiber
 func defaultLoggerInstance(c fiber.Ctx, data *Data, cfg Config) error {
+	// Check if Filter is defined and call it.
+	// Now, if Filter(c) == true, we SKIP logging:
+	if cfg.Skip != nil && cfg.Skip(c) {
+		return nil // Skip logging if Filter returns true
+	}
```
ReneWerner87, line 22:
```suggestion
	// Check if Skip is defined and call it.
	// Now, if Skip(c) == true, we SKIP logging:
	if cfg.Skip != nil && cfg.Skip(c) {
		return nil // Skip logging if Skip returns true
	}
```

PR #3339, opened 2025-03-06
🔥 feat: Add Session Access via Go Context

JIeJaitt:
In order to maintain the consistency of the fiber middleware, this pr is proposed to add the corresponding function mentioned in #3212 to the session middleware.


JIeJaitt:
This unit test I ran locally reported no errors and I didn't change it. Anything else to note here? 
image

image

middleware/session/store.go
```diff
@@ -100,7 +101,16 @@ func (s *Store) Get(c fiber.Ctx) (*Session, error) {
...
 	}
 
-	return s.getSession(c)
+	sess, err := s.getSession(c)
+	if err != nil {
+		return nil, err
+	}
+
+	// Add session to Go context
+	ctx := context.WithValue(c.Context(), sessionContextKey, sess)
```
ReneWerner87, line 110: we should not add or change anything in a get function, this does not meet the expectations of the user

ReneWerner87:
@sixcolors can you check this

middleware/session/store.go
JIeJaitt: @ReneWerner87 I think we have to have this, look at this code below，
```go
app.Post("/process-data", func(c fiber.Ctx) error {
    sess, err := store.Get(c)
    if err != nil {
        return err
    }
    
    go processDataInBackground(c.Context(), c.Body())
    
    return c.SendStatus(fiber.StatusAccepted)
})

func processDataInBackground(ctx context.Context, data []byte) {
    // Getting sessions from Go context
   // 
    sess := session.FromGoContext(ctx)
    if sess == nil {
        log.Error("Session not found in context")
        return
    }
    
    userID := sess.Get("userID").(string)
}
```

JIeJaitt:
@ReneWerner87 How can I avoid lint error , I used the following comment but it doesn't seem to have any effect, he still reports lint error:
image

middleware/session/store.go
ReneWerner87: please do not write anything in a read context method
the value in the context should work the same as the value in the locals, i.e. at every place where we write to the locals, we should also write to the context

JIeJaitt: you are right, the value in the context should work the same as the value in the locals

ReneWerner87: Then it would make the most sense to create a new method and set the value in context and locals in this method

This method can then be used in all places where the locals are currently set

middleware/session/session.go
```diff
@@ -511,3 +521,15 @@ func (s *Session) isAbsExpired() bool {
 func (s *Session) setAbsExpiration(absExpiration time.Time) {
 	s.Set(absExpirationKey, absExpiration)
 }
+
+// FromGoContext returns the Session from the go context.
+// If there is no session, nil is returned.
+func FromGoContext(ctx context.Context) *Session {
```
gaby, line 527: We should probably rename this to `FromContext()`

gaby (commented):
One small comment, should we rename it? Also the documentation needs to be updated and the `whats_new.md` file. 

Besides that it looks good to me.

middleware/session/session.go
gaby: Thoughts?

JIeJaitt: @gaby ok, i will change it

middleware/session/store.go
JIeJaitt: @ReneWerner87 l create a new method to do this.

middleware/session/middleware.go
```diff
@@ -170,23 +171,25 @@ func releaseMiddleware(m *Middleware) {
...
-//
-//	m := session.FromContext(c)
-func FromContext(c fiber.Ctx) *Middleware {
-	m, ok := c.Locals(middlewareContextKey).(*Middleware)
-	if !ok {
-		return nil
+// FromContext returns the session from context.
+// Supported context types:
+// - fiber.Ctx: Retrieves session from Locals
+// - context.Context: Retrieves session from context values
```
sixcolors, line 177: This still returns the middleware handler.

JIeJaitt: It's a clerical error. I'll fix it.

sixcolors (changes requested):
session.Session and session.Middleware are distinct and have differing purposes. Storing them in the context with the same key is not correct.

See: https://github.com/gofiber/fiber/blob/main/docs/middleware/session.md#v2-to-v3

middleware/session/middleware.go
```diff
@@ -90,6 +91,10 @@ func NewWithStore(config ...Config) (fiber.Handler, *Store) {
 		m := acquireMiddleware()
 		m.initialize(c, cfg)
 
+		// Add session to Go context
+		ctx := context.WithValue(c.Context(), middlewareContextKey, m)
```
sixcolors, line 95: I don't understand the benefit of retaining the references in both c.Context and c.Locals. Given that the UserContext used is always linked to the ctx.

JIeJaitt: I answered the question here: https://github.com/gofiber/fiber/pull/3340#discussion_r1998180435

JIeJaitt:
> on.Session and session.Middleware are distinct and have differing purposes. Storing them in the context with the same key is not correct.
> 
> See: https://github.com/gofiber/fiber/blob/main/docs/middleware/session.md#v2-to-v3

At first I thought the same thing, but I think that since Middleware includes Session, why not get the Middleware when the user wants to use Session, and just fetch the Session from it?

So this pr only considers storing Middleware into Fiber.Ctx and context.Context, but not Session. 

What do you think? @ReneWerner87 @gaby

sixcolors:
> > on.Session and session.Middleware are distinct and have differing purposes. Storing them in the context with the same key is not correct.
> > See: https://github.com/gofiber/fiber/blob/main/docs/middleware/session.md#v2-to-v3
> 
> At first I thought the same thing, but I think that since Middleware includes Session, why not get the Middleware when the user wants to use Session, and just fetch the Session from it?
> 
> So this pr only considers storing Middleware into Fiber.Ctx and context.Context, but not Session.
> 
> What do you think? @ReneWerner87 @gaby

Session only exists in v3 for legacy use or for use without a context. Using Session methods is otherwise not recommended and Middleware methods would be preferred.

JIeJaitt:
> > > on.Session and session.Middleware are distinct and have differing purposes. Storing them in the context with the same key is not correct.
> > > See: https://github.com/gofiber/fiber/blob/main/docs/middleware/session.md#v2-to-v3
> > 
> > 
> > At first I thought the same thing, but I think that since Middleware includes Session, why not get the Middleware when the user wants to use Session, and just fetch the Session from it?
> > So this pr only considers storing Middleware into Fiber.Ctx and context.Context, but not Session.
> > What do you think? @ReneWerner87 @gaby
> 
> Session only exists in v3 for legacy use or for use without a context. Using Session methods is otherwise not recommended and Middleware methods would be preferred. See #2741
> 
> So I would reject the added GetAndSetInContext function.

If that's the case, as I understand it Session is an already obsolete piece of code, and adding a new GetAndSetInContext function would encourage the use of an API that is no longer recommended.

If my understanding is correct I would remove the GetAndSetInContext function.

sixcolors (changes requested):
Thanks for your contribution! However, I do not support merging this change due to the reasons outlined in [issue #3358](https://github.com/gofiber/fiber/issues/3358).  

Fiber is designed to be lightweight and high-performance, and modifying `context.Context` within middleware introduces unnecessary overhead.  

If you need `context.Context` integration, you can implement a custom middleware to copy values into `context.Context` for your specific use case without affecting Fiber’s default behavior.  

For more details, see: [this comment](https://github.com/gofiber/fiber/issues/3358#issuecomment-2736970534).  

I recommend closing this PR. Thanks again! 🚀

JIeJaitt:
> Thanks for your contribution! However, I do not support merging this change due to the reasons outlined in [issue #3358](https://github.com/gofiber/fiber/issues/3358).
> 
> Fiber is designed to be lightweight and high-performance, and modifying `context.Context` within middleware introduces unnecessary overhead.
> 
> If you need `context.Context` integration, you can implement a custom middleware to copy values into `context.Context` for your specific use case without affecting Fiber’s default behavior.
> 
> For more details, see: [this comment](https://github.com/gofiber/fiber/issues/3358#issuecomment-2736970534).
> 
> I recommend closing this PR. Thanks again! 🚀

thinks，I've already understood the whole working process and I will close this pr

PR #3287, opened 2025-01-22
🔥 feat: Add support for context.Context in keyauth middleware

vhespanha:
pretty straightforward option to use context.Context instead of just fiber.Ctx, tests added accordingly.

# Description

as discussed in #3212 and #3175, some of the middleware would benefit from exposing support for raw context.Context, for usage outside of the HTTP layer. this is the first out of 3 pull requests for middlewares that are deemed necessary to support this as of now.

> Middlewares to update:
> 
> - [x] keyauth
> 
> - [ ] csrf
> 
> - [ ] session

## Type of change

- [x] Enhancement (improvement to existing features and functionality)

## Draft

i'm opening this as a draft PR as of now, if it aligns with what the issues requested, i'm happy to update the documentation and benchmarks, i feel like it's correct tho, since it basically mirrors the implementation and usage of the same feature for requestID (from #3200).

PR #3340, opened 2025-03-06
🔥 feat: Add context.Context support to CSRF middleware

JIeJaitt:
In order to maintain the consistency of the fiber middleware, this pr is proposed to add the corresponding function mentioned in #3212 to the CSRF middleware.


middleware/csrf/csrf.go
```diff
@@ -180,29 +182,56 @@ func New(config ...Config) fiber.Handler {
...
+	case fiber.Ctx:
+		if token, ok := ctx.Locals(tokenKey).(string); ok {
+			return token
+		}
+	case context.Context:
+		if token, ok := ctx.Value(tokenKey).(string); ok {
+			return token
+		}
+	default:
+		log.Errorf("Unsupported context type: %T. Expected fiber.Ctx or context.Context", c)
```
gaby, line 211: This should be a panic(), not a logger. See https://github.com/gofiber/fiber/pull/3287

middleware/csrf/csrf.go
```diff
@@ -180,29 +182,56 @@ func New(config ...Config) fiber.Handler {
...
+	case fiber.Ctx:
+		if handler, ok := ctx.Locals(handlerKey).(*Handler); ok {
+			return handler
+		}
+	case context.Context:
+		if handler, ok := ctx.Value(handlerKey).(*Handler); ok {
+			return handler
+		}
+	default:
+		log.Errorf("Unsupported context type: %T. Expected fiber.Ctx or context.Context", c)
```
gaby, line 232: Return nil, instead of adding logger.

gaby:
@JIeJaitt Take a lot at the changes that were added to https://github.com/gofiber/fiber/pull/3287 those would need to be applied here too. For example, removing the logger.

gaby:
@JIeJaitt I'm fixing it in a branch. It randomly broke without us updating the version

middleware/csrf/csrf.go
```diff
@@ -180,29 +181,56 @@ func New(config ...Config) fiber.Handler {
 		// Store the token in the context
 		c.Locals(tokenKey, token)
 
+		// Store the token in the Go context
+		ctx := context.WithValue(c.Context(), tokenKey, token)
+		ctx = context.WithValue(ctx, handlerKey, handler)
```
sixcolors, line 186: I'm unsure I understand the benefit of retaining the references in both c.Context and c.Locals. Given that the UserContext used is always linked to the ctx.

JIeJaitt: When the developer controller layer will fiber.Ctx converted to context.Context, in the service layer will have a lot of inconvenience when using, for example, requestid can not be directly inherited from the ctx, can only rely on the external incoming ... This is the initial idea, other middleware, perhaps to maintain the consistency of the middleware fiber framework it

sixcolors: I’m still not sure I understand. Could you perhaps provide a code example demonstrating the issue you are seeking to address?

JIeJaitt: ### Project Structure (Tree Syntax)

```
.
├── main.go
├── controllers
│   └── user_controller.go
├── services
│   └── user_service.go
├── middlewares
│   └── csrf_middleware.go
└── utils
    └── context_utils.go
```

### Explanation

In a Go Fiber application, the CSRF middleware is used to protect against Cross-Site Request Forgery attacks. When using Fiber, the context (`fiber.Ctx`) is specific to Fiber and is not directly compatible with Go's native `context.Context`. However, in many cases, especially when passing the context to service layers, it is more convenient to use Go's native `context.Context` because it is widely supported across various libraries and frameworks.

If we don't inject the CSRF information into the native `context.Context`, we would have to pass the CSRF token or related information as separate parameters to the service layer. This can make the code less clean and harder to maintain. By injecting the CSRF information into the native `context.Context`, we can access it easily within the service layer without needing to pass it explicitly.

### Code Example

#### `middlewares/csrf_middleware.go`

```go
package middlewares

import (
    "context"
    "github.com/gofiber/fiber/v2"
    "github.com/gofiber/fiber/v2/middleware/csrf"
)

// CSRFMiddleware is a middleware that injects CSRF information into the context.
func CSRFMiddleware() fiber.Handler {
    return csrf.New(csrf.Config{
        KeyLookup:      "header:X-CSRF-Token",
        ContextKey:     "csrf",
        CookieName:     "csrf_",
        CookieSameSite: "Strict",
        Expiration:     1 * time.Hour,
    })
}

// InjectCSRFIntoContext is a middleware that injects CSRF token into the native context.
func InjectCSRFIntoContext() fiber.Handler {
        // Get CSRF token from Fiber context
        csrfToken := c.Locals("csrf").(string)

        // Create a new context with CSRF token
        ctx := context.WithValue(c.Context(), "csrf", csrfToken)

        // Replace Fiber context with the new context
        c.SetUserContext(ctx)

        return c.Next()
    }
}
```

#### `controllers/user_controller.go`

```go
package controllers

import (
    "context"
    "github.com/gofiber/fiber/v2"
    "your_project/services"
)

type UserController struct {
    userService *services.UserService
}

func NewUserController(userService *services.UserService) *UserController {
    return &UserController{userService: userService}
}

func (uc *UserController) CreateUser(c *fiber.Ctx) error {
    // Convert Fiber context to native context
    ctx := c.Context()

    // Call service layer with native context
    err := uc.userService.CreateUser(ctx)
    if err != nil {
        return c.Status(fiber.StatusInternalServerError).JSON(fiber.Map{
            "error": err.Error(),
        })
    }

    return c.Status(fiber.StatusOK).JSON(fiber.Map{
        "message": "User created successfully",
    })
}
```

#### `services/user_service.go`

```go
package services

import (
    "context"
    "fmt"
)

type UserService struct{}

func NewUserService() *UserService {
    return &UserService{}
}

func (us *UserService) CreateUser(ctx context.Context) error {
    // Retrieve CSRF token from context
    csrfToken := ctx.Value("csrf").(string)

    // Use CSRF token in the service logic
    fmt.Println("CSRF Token:", csrfToken)

    // Business logic here...

    return nil
}
```

### Summary

- **CSRF Middleware**: The `CSRFMiddleware` function sets up the CSRF protection in the Fiber application.
- **InjectCSRFIntoContext**: This middleware extracts the CSRF token from the Fiber context and injects it into the native `context.Context`.
- **Controller Layer**: The `UserController` converts the Fiber context to a native context and passes it to the service layer.
- **Service Layer**: The `UserService` retrieves the CSRF token from the native context and uses it in the business logic.

By injecting the CSRF information into the native `context.Context`, we avoid the need to pass the CSRF token as a separate parameter, making the code cleaner and more maintainable.

JIeJaitt: Fiber's fiber.Ctx is specific to the Fiber framework. If we only store CSRF information in fiber.Ctx.Locals, it will not be accessible when we switch to Go's native context.Context in the service layer or other parts of the application.

Go's native context.Context is widely used across libraries and frameworks. By injecting CSRF information into the native context, we ensure that it can be accessed seamlessly in the service layer or any other part of the application that relies on context.Context.

code example like this:

```
.
├── main.go
├── controllers
│   └── user_controller.go
├── services
│   └── user_service.go
├── middlewares
│   └── csrf_middleware.go
└── utils
    └── context_utils.go
```

## Controller Layer (controllers/user_controller.go)

```go
package controllers

import (
    "context"
    "github.com/gofiber/fiber/v2"
    "your_project/services"
)

type UserController struct {
    userService *services.UserService
}

func NewUserController(userService *services.UserService) *UserController {
    return &UserController{userService: userService}
}

func (uc *UserController) CreateUser(c *fiber.Ctx) error {
    // Convert Fiber context to native context
    ctx := c.Context()

    // Call service layer with native context
    err := uc.userService.CreateUser(ctx)
    if err != nil {
        return c.Status(fiber.StatusInternalServerError).JSON(fiber.Map{
            "error": err.Error(),
        })
    }

    return c.Status(fiber.StatusOK).JSON(fiber.Map{
        "message": "User created successfully",
    })
}
```
```go
package services

import (
    "context"
    "fmt"
)

type UserService struct{}

func NewUserService() *UserService {
    return &UserService{}
}

func (us *UserService) CreateUser(ctx context.Context) error {
    // Retrieve CSRF token from context
    csrfToken := ctx.Value("csrf").(string)

    // Use CSRF token in the service logic
    fmt.Println("CSRF Token:", csrfToken)

    // Business logic here...

    return nil
}
```

If we don't inject a csrf into the context, then we can't get the csrf information at the service level unless we pass it in as a parameter manually, but this is obviously not that elegant.

sixcolors (changes requested):
Thanks for your contribution! However, I do not support merging this change due to the reasons outlined in [issue #3358](https://github.com/gofiber/fiber/issues/3358).  

Fiber is designed to be lightweight and high-performance, and modifying `context.Context` within middleware introduces unnecessary overhead.  

If you need `context.Context` integration, you can implement a custom middleware to copy values into `context.Context` for your specific use case without affecting Fiber’s default behavior.  

For more details, see: [this comment](https://github.com/gofiber/fiber/issues/3358#issuecomment-2736970534).  

I recommend closing this PR. Thanks again! 🚀

middleware/csrf/csrf.go
sixcolors: The assumption here is incorrect—Fiber’s CSRF token is **not** stored in `context.Context` using a string key like `"csrf"`. Instead, it is stored in `c.Locals` using an un-exported type for a key, and the correct way to retrieve it is:  

```go
csrfToken := csrf.TokenFromContext(c)
```

JIeJaitt:
> Thanks for your contribution! However, I do not support merging this change due to the reasons outlined in [issue #3358](https://github.com/gofiber/fiber/issues/3358).
> 
> Fiber is designed to be lightweight and high-performance, and modifying `context.Context` within middleware introduces unnecessary overhead.
> 
> If you need `context.Context` integration, you can implement a custom middleware to copy values into `context.Context` for your specific use case without affecting Fiber’s default behavior.
> 
> For more details, see: [this comment](https://github.com/gofiber/fiber/issues/3358#issuecomment-2736970534).
> 
> I recommend closing this PR. Thanks again! 🚀

thinks，I've already understood the whole working process and I will close this pr

PR #3353, opened 2025-03-15
♻️ Refactor: reduce DefaultCtx from 768 bytes to 736 bytes

ksw2000:
# Description

In the default context, `pathBuffer` and `detectionPathBuffer` can be used as `path` and `detectionPath` directly. We can remove two `string` fields and save `32 bytes` for each context.

```
goos: linux
goarch: amd64
cpu: AMD EPYC 9J14 96-Core Processor                
                                             │   old.txt   │              new.txt               │
                                             │   sec/op    │   sec/op     vs base               │
_Ctx_AcquireReleaseFlow/withoutRequestCtx-16   41.30n ± 1%   39.10n ± 0%  -5.33% (p=0.000 n=25)
_Ctx_AcquireReleaseFlow/withRequestCtx-16      530.8n ± 1%   518.7n ± 1%  -2.28% (p=0.000 n=25)
geomean                                        148.1n        142.4n       -3.82%

                                             │   old.txt    │                new.txt                │
                                             │     B/op     │     B/op      vs base                 │
_Ctx_AcquireReleaseFlow/withoutRequestCtx-16     3.000 ± 0%     3.000 ± 0%       ~ (p=1.000 n=25) ¹
_Ctx_AcquireReleaseFlow/withRequestCtx-16      1.532Ki ± 0%   1.532Ki ± 0%       ~ (p=1.000 n=25) ¹
geomean                                          68.61          68.61       +0.00%
¹ all samples are equal

                                             │  old.txt   │               new.txt               │
                                             │ allocs/op  │ allocs/op   vs base                 │
_Ctx_AcquireReleaseFlow/withoutRequestCtx-16   1.000 ± 0%   1.000 ± 0%       ~ (p=1.000 n=25) ¹
_Ctx_AcquireReleaseFlow/withRequestCtx-16      5.000 ± 0%   5.000 ± 0%       ~ (p=1.000 n=25) ¹
geomean                                        2.236        2.236       +0.00%
¹ all samples are equal

```

## Type of change

## Checklist


ctx.go
```diff
@@ -1832,32 +1830,30 @@ func (c *DefaultCtx) XHR() bool {
...
-	c.detectionPath = c.app.getString(c.detectionPathBuffer)
 
 	// Define the path for dividing routes into areas for fast tree detection, so that fewer routes need to be traversed,
 	// since the first three characters area select a list of routes
-	c.treePath = c.treePath[0:0]
+	c.treePath = ""
 	const maxDetectionPaths = 3
 	if len(c.detectionPath) >= maxDetectionPaths {
-		c.treePath = c.detectionPath[:maxDetectionPaths]
+		c.treePath = utils.UnsafeString(c.detectionPath[:maxDetectionPaths])
```
gaby, line 1856: I think this should use c.app.getString()

gaby (approved):

ctx.go
```diff
@@ -1832,32 +1830,30 @@ func (c *DefaultCtx) XHR() bool {
...
-		c.detectionPathBuffer = utils.TrimRight(c.detectionPathBuffer, '/')
+	if !c.app.config.StrictRouting && len(c.detectionPath) > 1 && c.detectionPath[len(c.detectionPath)-1] == '/' {
+		c.detectionPath = utils.TrimRight(c.detectionPath, '/')
 	}
-	c.detectionPath = c.app.getString(c.detectionPathBuffer)
 
 	// Define the path for dividing routes into areas for fast tree detection, so that fewer routes need to be traversed,
 	// since the first three characters area select a list of routes
-	c.treePath = c.treePath[0:0]
+	c.treePath = ""
```
sixcolors, line 1853: Why replace c.treePath[:0] here?

ksw2000: c.treePath is a string, not a byte slice. The original operation is more complex and provides no additional benefits. Assigning an empty string directly is simpler and more readable.

sixcolors: 👍

ReneWerner87:
I will check it later today and probably merge

ctx.go
```diff
@@ -1832,32 +1831,32 @@ func (c *DefaultCtx) XHR() bool {
 // configDependentPaths set paths for route recognition and prepared paths for the user,
 // here the features for caseSensitive, decoded paths, strict paths are evaluated
 func (c *DefaultCtx) configDependentPaths() {
-	c.pathBuffer = append(c.pathBuffer[0:0], c.pathOriginal...)
+	c.path = append(c.path[:0], c.pathOriginal...)
```
ReneWerner87, line 1834: at that time it was necessary because you have to keep the original url and still need to change the route for recognition

are we sure that this is no longer necessary ? can we test this, or make sure in unittests that it is no longer a problem?

ReneWerner87:
I am still unsure, at that time the additional buffers were necessary to reduce the allocations and preserve the original
can we find the case from back then and check it again?

ctx.go
ReneWerner87: can you test both cases again, with immutable and without immutable? 
I think one case caused problems
because if you had changed the route with Path() inside or the variable and thus the linked string was changed by the caseSensitive, strict routing or unescapePath

ksw2000: I'm trying to send 2 requests, the first one is `bob`, and the second one is `alice`.
I assign the value of c.Path() to the string `p` when requesting the path `bob` and then checking the value of `p` is still not be modified after request `alice`

```go
func Test_Ctx_Path_Immutable(t *testing.T) {
	t.Parallel()
	app := New()
	app.config.Immutable = true

	done1 := make(chan struct{})
	done2 := make(chan struct{})

	flag := false
	var p string
	app.Get("/ctx_path_immutable/:name", func(c Ctx) error {
		if !flag {
			// p should not be changed after the next request
			p = c.Path()
			done1 <- struct{}{}
			flag = true
		} else {
			c.Path()
			done2 <- struct{}{}
		}
		return nil
	})

	// try to GET /ctx_path_immutable/bob
	app.Test(httptest.NewRequest(MethodGet, "/ctx_path_immutable/bob", nil))

	go func() {
		<-done1
		require.Equal(t, "/ctx_path_immutable/bob", p)

		// try to get Alice
		app.Test(httptest.NewRequest(MethodGet, "/ctx_path_immutable/alice", nil))
	}()

	<-done2
	// p must be changed
	require.Equal(t, "/ctx_path_immutable/bob", p)
}
```

ksw2000:
> I am still unsure, at that time the additional buffers were necessary to reduce the allocations and preserve the original can we find the case from back then and check it again?

I reduce the two string fields in the `defautCtx` and keep the original two buffers. I think the `c.detectionPath` and `c.path` can be saved by `[]byte` in the sturct. They can be lazy converted to the string until the user calls c.Path() or c.detectPath().

ReneWerner87:
> > I am still unsure, at that time the additional buffers were necessary to reduce the allocations and preserve the original can we find the case from back then and check it again?
> 
> I reduce the two string fields in the `defautCtx` and keep the original two buffers. I think the `c.detectionPath` and `c.path` can be saved by `[]byte` in the sturct. They can be lazy converted to the string until the user calls c.Path() or c.detectPath().

yes already understood, just want to make sure that we don't have any performance losses(more allocs) or that the strings are not changed by the referencing

I will test this by hand
then merge, have to make sure again, because there was a case back then

ReneWerner87:
really good work, I checked everything again and we also have tests for the cases where I suspected that something could still go wrong

https://github.com/ksw2000/fiber/blob/395c8fafa96719cfcce808876c4b11914ca34efc/app_test.go#L500

thanks for your work and patience

ReneWerner87 (approved):

PR #3354, opened 2025-03-15
🧹 chore: Fix linter workflow failures

gaby:
- Bump golangci-lint to v1.64.x
- Update configuration to ignore msgp files.

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -37,4 +37,4 @@ jobs:
         uses: golangci/golangci-lint-action@v6
         with:
           # NOTE: Keep this in sync with the version from .golangci.yml
-          version: v1.62.2
+          version: v1.64.7
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -7,7 +7,6 @@ run:
 
 output:
   sort-results: true
-  uniq-by-line: false
 
 linters-settings:
   depguard:
@@ -97,6 +96,11 @@ linters-settings:
       global:
         # show-ignored: true # TODO: Enable this
         audit: true
+
+  staticcheck:
+    checks:
+      - all
+      - '-SA4023' # disable the rule SA4023
 
   govet:
     enable-all: true
@@ -249,8 +253,12 @@ issues:
   exclude-case-sensitive: true
   max-issues-per-linter: 0
   max-same-issues: 0
+  uniq-by-line: false
   exclude-dirs:
     - internal # TODO: Do not ignore interal packages
+  exclude-files:
+    - '_msgp\.go'
+    - '_msgp_test\.go'
   exclude-rules:
     - linters:
         - err113
@@ -358,7 +366,6 @@ linters:
     - stylecheck
     # - tagalign # TODO: Enable
     - tagliatelle
-    - tenv
     - testableexamples
     - testifylint
     # - testpackage # TODO: Enable
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ markdown:
 ## lint: 🚨 Run lint checks
 .PHONY: lint
 lint:
-	go run github.com/golangci/golangci-lint/cmd/golangci-lint@v1.62.2 run ./...
+	go run github.com/golangci/golangci-lint/cmd/golangci-lint@v1.64.7 run ./...
 
 ## test: 🚦 Execute all tests
 .PHONY: test
--- a/client/hooks.go
+++ b/client/hooks.go
@@ -199,7 +199,7 @@ func parserRequestBody(c *Client, req *Request) error {
 	case filesBody:
 		return parserRequestBodyFile(req)
 	case rawBody:
-		if body, ok := req.body.([]byte); ok {
+		if body, ok := req.body.([]byte); ok { //nolint:revive // ignore simplicity
 			req.RawRequest.SetBody(body)
 		} else {
 			return ErrBodyType
--- a/listen.go
+++ b/listen.go
@@ -367,7 +367,7 @@ func (app *App) startupMessage(addr string, isTLS bool, pids string, cfg ListenC
 	}
 
 	fmt.Fprintf(out, "%s\n", fmt.Sprintf(figletFiberText, colors.Red+"v"+Version+colors.Reset)) //nolint:errcheck,revive // ignore error
-	fmt.Fprintf(out, strings.Repeat("-", 50)+"\n")                                              //nolint:errcheck,revive,govet // ignore error
+	fmt.Fprintf(out, strings.Repeat("-", 50)+"\n")                                              //nolint:errcheck,revive // ignore error
 
 	if host == "0.0.0.0" {
 		//nolint:errcheck,revive // ignore error


gaby:
golangci-lint works fine locally, The CI workflow is clearly doing whatever it wants 😆

gaby:
Pending discussion https://github.com/golangci/golangci-lint-action/discussions/1197

nickajacks1:
Shot in the dark, is it because we're installing go 1.23 but this version of golangci-lint requires 1.24?

ldez:
> Shot in the dark, is it because we're installing go 1.23 but this version of golangci-lint requires 1.24?

No, this has no impact.

ldez:
https://github.com/gofiber/fiber/pull/3355

gaby:
@grivera64 @nickajacks1 Can you review this one too. Thanks

.golangci.yml
```diff
@@ -98,6 +97,11 @@ linters-settings:
         # show-ignored: true # TODO: Enable this
         audit: true
 
+  staticcheck:
+    checks:
+      - all
+      - '-SA4023' # disable the rule SA4023
```
nickajacks1, line 103: SA4023 claims that reflect.Value.Type is never nil.  It seems like we could update the code where this rule fails to remove redundant nil checks.

.github/workflows/linter.yml
```diff
@@ -37,4 +37,7 @@ jobs:
         uses: golangci/golangci-lint-action@v6
         with:
           # NOTE: Keep this in sync with the version from .golangci.yml
-          version: v1.62.2
+          version: v1.64
```
grivera64, line 40: Should we specify the full version here to match with the Makefile's version of `1.64.7`? This way we can ensure that we can keep local and workflow golangci-lint versions in sync.

.golangci.yml
grivera64: Seems from my testing that this is true. If there is ever a case where a `reflect.Value`would not have a `reflect.Type`, the line where it is created will panic before we can check its `v.Type()`.

In this case, SA4023 may just be doing its job of reporting a potential non-nil return. This is strange though that it wasn't found earlier. Nothing that could be the reason has changed (e.g. reflect version, code change, etc.).

.github/workflows/linter.yml
ReneWerner87: yeah would be better

ldez: v1.64.7 is the last v1 release: https://github.com/golangci/golangci-lint/issues/5300

.golangci.yml
gaby: @nickajacks1 @grivera64 do we want to enable this or not?

.github/workflows/linter.yml
gaby: Done, points to 1.64.7 now.

.golangci.yml
nickajacks1: My instinct is to keep the rule enabled. If there was any previous discussion about it, I probably missed it.

gaby: There wasn't, it's something that's enabled by default on recent versions of golangci-lint. I will make the changes today.

gaby:
golangci-lint keeps doing the same thing. Works locally fails on the workflow.

ldez:
https://github.com/golangci/golangci-lint-action/discussions/1197#discussioncomment-12519285

My recommendation, as `unhandled-error` rule is a duplicate of `errcheck`, you can disable this rule:

```yml
      - name: unhandled-error
        disabled: true
```

gaby:
> [golangci/golangci-lint-action#1197 (reply in thread)](https://github.com/golangci/golangci-lint-action/discussions/1197#discussioncomment-12519285)
> 
> My recommendation, as `unhandled-error` rule is a duplicate of `errcheck`, you can disable this rule:
> 
> ```yaml
>       - name: unhandled-error
>         disabled: true
> ```

Ok, let me try that

ldez:
:thinking: in fact, this is a different situation: the reports on the CI are the same as locally.

```
$ golangci-lint run                        
ctx_test.go:4581:3: unhandled-error: Unhandled error in call to function bufio.Writer.WriteString (revive)
                w.WriteString("Don't crash please") //nolint:errcheck // It is fine to ignore the error
                ^
```

I still think that you don't need `unhandled-error` rule because you already use `errcheck`, but here you just have to modify the `nolint` directive:
```go
//nolint:errcheck,revive // It is fine to ignore the error
```

gaby:
@ldez I'm still confused as to why it doesnt fail locally for me?

```bash
ubuntu@ubuntu:~/Desktop/git/fiber$ git branch
  add-bench
  dependabot/go_modules/golang.org/x/crypto-0.35.0
  fix-3236
* fix-lint
  golangci-update
  main
  v2
  v2-updates
ubuntu@ubuntu:~/Desktop/git/fiber$ curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/HEAD/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.64.7
golangci/golangci-lint info checking GitHub for tag 'v1.64.7'
golangci/golangci-lint info found version: 1.64.7 for v1.64.7/linux/amd64
golangci/golangci-lint info installed /home/ubuntu/go/bin/golangci-lint
ubuntu@ubuntu:~/Desktop/git/fiber$ golangci-lint run ./...
ubuntu@ubuntu:~/Desktop/git/fiber$ 
```

ldez:
can you try to clean your cache? (`golancgi-lint cache clean`)

gaby:
> can you try to clean your cache? (`golancgi-lint cache clean`)

Same behavior:

```bash
ubuntu@ubuntu:~/Desktop/git/fiber$ golangci-lint cache clean
ubuntu@ubuntu:~/Desktop/git/fiber$ golangci-lint run ./...
ubuntu@ubuntu:~/Desktop/git/fiber$ golangci-lint version
golangci-lint has version 1.64.7 built with go1.24.1 from 8cffdb7d on 2025-03-11T23:26:51Z
```

ldez:
```console
$ git log --oneline -1                                     
bdb405eb (HEAD -> fix-lint, upstream/fix-lint) Enable SA4023
$ golangci-lint version
golangci-lint has version 1.64.7 built with go1.24.1 from 8cffdb7d on 2025-03-11T23:26:51Z
$ go version           
go version go1.24.1 linux/amd64
$ golangci-lint cache clean
$ golangci-lint run                        
ctx_test.go:4581:3: unhandled-error: Unhandled error in call to function bufio.Writer.WriteString (revive)
                w.WriteString("Don't crash please") //nolint:errcheck // It is fine to ignore the error
                ^
```

can you give me the output of `golangci-lint run -v`?

FYI `./...` is useless because this is the default.

ldez:
You:
```
INFO [runner] Issues before processing: 1295, after processing: 0
```

Me:
```
INFO [runner] Issues before processing: 1316, after processing: 1
```

21 issues are missing, this confirms the bug with revive.

ldez:
For now, disabling `unhandled-error` is the best move you can do.

My main difficulty is that I cannot reproduce the problem locally, I'm forced to use an SSH connection on a GitHub Action runner to diagnose, it's not really fun to do, but at least I can reproduce the problem.

.golangci.yml
```diff
@@ -187,6 +187,8 @@ linters-settings:
         disabled: true # TODO: Do not disable
       - name: unhandled-error
         arguments: ['bytes\.Buffer\.Write']
+      - name: unhandled-error
+        disabled: true
```
ldez, line 191: there is a duplication of `unhandled-error`

gaby: Fixed, thanks!

gaby:
g2g for review @gofiber/maintainers

nickajacks1 (approved):

PR #3355, opened 2025-03-15
🧹 chore: Fix and update golangci-lint workflow

ldez:
This PR targets another PR (branch: `fix-lint`) #3354.

This is a workaround to unlock your situation (the workaround is to use `goinstall` mode).

I also improved a bit your configuration and fixed some malformed `nolint` directives.

This is just a workaround. I'm still working on a proper fix, but as you can imagine, it is not quick or simple to debug on the GitHub Action runner.


nickajacks1 (approved):
Thanks!

gaby (approved):
LGTM, Thanks!

PR #3379, opened 2025-03-29
🩹 Fix: Implement RFC 6265 Cookie Value Sanitization

### LINKED ISSUE #3176: 🐛 [Bug]: There is something weird with Fiber and the usage of Http lib for testing.
### Bug Description

You are using the HTTP library (https://docs.gofiber.io/api/app/#test) but some fiber features are not RFC compliant, example with cookies using the `"` character. You can't use HTTP library if you have to send (and the backend have to receive) a golang json structure.
If you want to stock data in cookies (very practical but non RFC compliant).

### How to Reproduce

1. Make an OAuth2 login and callback entries.
2. Add data in the state field (for example make a structure).
3. With a modern navigator, you will see that the cookie works fine.
4. Try to make a test with HTTP. Like in the doc: [https://docs.gofiber.io/api/app/#test](https://docs.gofiber.io/api/app/#test)
5. You will see that it is not possible to handle the test, because HTTP library doesn't support `"`.

### Expected Behavior

The cookie with `"` are supposed to work.

### Fiber Version

v2.52.5

### Code Snippet (optional)

```go
// addServiceToUser simulates adding a service to a user by invoking the OAuth callback endpoint.
func addServiceToUserTesting(app *fiber.App, sessionCookie *http.Cookie) error {
    // Simulate generating and setting the state
    mockState := StateData{
        State:             "test-state",
        RedirectURL:       "",
        StoreSessionInURL: false,
    }

    // Encode the state and redirect URI into JSON.
    value, err := sonic.Marshal(mockState)
    if err != nil {
        return fmt.Errorf("failed to marshall token: %s", err)
    }

    // Encode to base64 (in order to be able to use it in the backend).
    encodedState := base64.StdEncoding.EncodeToString(value)

    fullURL := fmt.Sprintf("/protected/oauth2/noServiceUsedForTesting/callback?code=test-code&state=%s",
        encodedState)

    // Create the HTTP request
    req, err := http.NewRequest("GET", fullURL, nil)
    if err != nil {
        return err
    }

    // Attach the session cookie for authentication
    if sessionCookie != nil {
        req.AddCookie(sessionCookie)
    }

    // OAuth handler checks for the state in a cookie, set it here
    req.AddCookie(&http.Cookie{
        Name:  "oauth_state",
        Value: encodedState,
    })

    // Perform the request
    resp, err := app.Test(req, -1)
    if err != nil {
        return err
    }
    ...
}
```

For this part:
```go
    // OAuth handler checks for the state in a cookie, set it here
    req.AddCookie(&http.Cookie{
        Name:  "oauth_state",
        Value: encodedState,
    })
```

It is mandatory to convert to base 64 in order to handle the `"` in the backend.

### Checklist:

c00kie17:
# Description

Added cookie value sanitization according to RFC 6265 standards. The implementation validates and sanitizes cookie values by removing invalid characters (`""`,`;`, `\`), while intentionally allowing commas to maintain compatibility with Go's net/http package (see https://golang.org/issue/7243). Test cases have been updated accordingly to use RFC-compliant values.

Fixes #3176

## Changes introduced

- Implemented cookie value sanitization that ensures RFC 6265 compliance, removing invalid characters (", ;, \) from cookie values.
- Modified test key string in keyAuth tests to be RFC 6265 compliant.

## Breaking Change Notice

This change enforces RFC 6265 compliance for cookie values. If an existing application currently stores or relies on cookie values containing these characters, they will be silently removed when processed by Fiber. 

## Type of change

## Checklist

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -23,6 +23,7 @@ import (
 	"text/template"
 	"time"
 
+	"github.com/gofiber/fiber/v3/log"
 	"github.com/gofiber/utils/v2"
 	"github.com/valyala/bytebufferpool"
 	"github.com/valyala/fasthttp"
@@ -450,7 +451,48 @@ func (c *DefaultCtx) Cookie(cookie *Cookie) {
 // The returned value is only valid within the handler. Do not store any references.
 // Make copies or use the Immutable setting to use the value outside the Handler.
 func (c *DefaultCtx) Cookies(key string, defaultValue ...string) string {
-	return defaultString(c.app.getString(c.fasthttp.Request.Header.Cookie(key)), defaultValue)
+	value := c.app.getString(c.fasthttp.Request.Header.Cookie(key))
+	return defaultString(c.sanitizeCookieValue(value), defaultValue)
+}
+
+// sanitizeCookieValue sanitizes a cookie value according to RFC 6265.
+// It removes invalid characters from the cookie value, similar to how
+// Go's standard library handles cookie values.
+func (c *DefaultCtx) sanitizeCookieValue(v string) string {
+	var result strings.Builder
+	result.Grow(len(v))
+	invalidChars := make(map[byte]struct{})
+
+	for i := 0; i < len(v); i++ {
+		b := v[i]
+		if c.validCookieValueByte(b) {
+			result.WriteByte(b)
+		} else {
+			invalidChars[b] = struct{}{}
+		}
+	}
+
+	if len(invalidChars) > 0 {
+		var chars []string
+		for b := range invalidChars {
+			chars = append(chars, fmt.Sprintf("'%c'", b))
+		}
+		log.Warn("invalid byte(s) %s in Cookie.Value; dropping invalid bytes",
+			strings.Join(chars, ", "))
+		return result.String()
+	}
+
+	return v
+}
+
+// validCookieValueByte reports whether b is a valid byte in a cookie value.
+// Per RFC 6265 section 4.1.1, cookie values must be ASCII
+// and may not contain control characters, whitespace, double quotes,
+// commas, semicolons, or backslashes.
+func (c *DefaultCtx) validCookieValueByte(b byte) bool {
+	return 0x20 <= b && b < 0x7f && b != '"' && b != ';' && b != '\\'
+	// Note: commas are deliberately allowed
+	// See https://golang.org/issue/7243 for the discussion.
 }
 
 // Download transfers the file from path as an attachment.
--- a/middleware/keyauth/keyauth_test.go
+++ b/middleware/keyauth/keyauth_test.go
@@ -12,7 +12,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-const CorrectKey = "specials: !$%,.#\"!?~`<>@$^*(){}[]|/\\123"
+const CorrectKey = "specials: !$%,.#!?~`<>@$^*(){}[]|/123"
 
 var testConfig = fiber.TestConfig{
 	Timeout: 0,
--- a/ctx_test.go
+++ b/ctx_test.go
@@ -1035,6 +1035,18 @@ func Test_Ctx_Cookies(t *testing.T) {
 	c.Request().Header.Set("Cookie", "john=doe")
 	require.Equal(t, "doe", c.Req().Cookies("john"))
 	require.Equal(t, "default", c.Req().Cookies("unknown", "default"))
+
+	c.Request().Header.Set("Cookie", "special=value,with,commas") // commas are allowed
+	require.Equal(t, "value,with,commas", c.Req().Cookies("special"))
+
+	c.Request().Header.Set("Cookie", "quotes=value\"with\"quotes")
+	require.Equal(t, "valuewithquotes", c.Req().Cookies("quotes"))
+
+	c.Request().Header.Set("Cookie", "semicolons=value;with;semicolons")
+	require.Equal(t, "valuewithsemicolons", c.Req().Cookies("semicolons"))
+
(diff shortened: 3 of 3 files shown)


ctx.go
```diff
@@ -450,7 +451,48 @@ func (c *DefaultCtx) Cookie(cookie *Cookie) {
...
+			invalidChars[b] = struct{}{}
+		}
+	}
+
+	if len(invalidChars) > 0 {
+		var chars []string
+		for b := range invalidChars {
+			chars = append(chars, fmt.Sprintf("'%c'", b))
+		}
+		log.Warn("invalid byte(s) %s in Cookie.Value; dropping invalid bytes",
```
gaby, line 480: I feel like this could be abused for DoS by sending big values via cookies

gaby: It might be better to log a generic message with the cookie name, instead of logging the whole cookie value.

gaby:
@c00kie17 csrf middleware may need to be updated?

ctx.go
c00kie17: I can add a size check, 4kb should be good ?

middleware/keyauth/keyauth_test.go
```diff
@@ -12,7 +12,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-const CorrectKey = "specials: !$%,.#!?~`<>@$^*(){}[]|/123"
+const CorrectKey = "specials: !$%,.#\"!?~`<>@$^*(){}[]|/\\123"
```
c00kie17: If you check for these strings then the tests wont pass as the sanitizer removes them

ctx.go
```diff
@@ -450,7 +451,49 @@ func (c *DefaultCtx) Cookie(cookie *Cookie) {
...
+	if len(value) > 0 && !utf8.ValidString(value) {
+		return value
+	}
+	return defaultString(c.sanitizeCookieValue(value), defaultValue)
+}
+
+// sanitizeCookieValue sanitizes a cookie value according to RFC 6265.
+// It removes invalid characters from the cookie value, similar to how
+// Go's standard library handles cookie values.
+func (c *DefaultCtx) sanitizeCookieValue(v string) string {
```
gaby, line 465: @gofiber/maintainers Is the correct approach to "clean" the cookie for the user. Or should we return HTTP 400?

@c00kie17 how does net/http handle this?

c00kie17:
Im not sure how you want to handle the raw binary on headers, the current tests are failing due to that, looks like there is a `redirectionMsg` binary message that is being set on cookies, This will be rejected by the logic from net/http 

We can convert them to base64 before storing so they are RFC 6265 compliant

gaby:
@c00kie17 Where are we with this PR? Thanks!

c00kie17:
sorry for the delay, Ive updated the flash messages to be base64 encoded and decoded so that the cookie parser does not reject the the flash message as byte values

ctx.go
```diff
@@ -450,7 +450,38 @@ func (c *DefaultCtx) Cookie(cookie *Cookie) {
...
+		return ""
+	}
+	return value
+}
+
+// parseCookieValue parses a cookie value according to RFC 6265.
+func parseCookieValue(raw string) (string, bool) {
+	// Strip the quotes, if present.
+	if len(raw) > 1 && raw[0] == '"' && raw[len(raw)-1] == '"' {
+		raw = raw[1 : len(raw)-1]
```
gaby, line 466: The tests are not covering the case of having quotes which is the purpose of the PR.

redirect.go
```diff
@@ -298,7 +299,16 @@ func (r *Redirect) parseAndClearFlashMessages() {
 	// parse flash messages
 	cookieValue := r.c.Cookies(FlashCookieName)
 
-	_, err := r.c.flashMessages.UnmarshalMsg(r.c.app.getBytes(cookieValue))
+	if cookieValue == "" {
```
gaby, line 302: This is also no covered by tests

redirect.go
```diff
@@ -298,7 +299,16 @@ func (r *Redirect) parseAndClearFlashMessages() {
...
+		return
+	}
+
+	decodedValue, err := base64.StdEncoding.DecodeString(cookieValue)
+	if err != nil {
+		return
```
gaby, line 308: Same here

gaby:
> @c00kie17 sorry for the late message why are we putting this in fiber and not in fasthttp? wouldn't that help everyone if it was sanitized in fasthttp ?

I was wondering that too, this should go in `fasthttp`, that way we inherit the fix

efectn:
> @c00kie17 sorry for the late message why are we putting this in fiber and not in fasthttp? wouldn't that help everyone if it was sanitized in fasthttp ?

I agree. This RFC check belongs to http server side, not framework. I think this PR should be opened in fasthttp repository for  convenience

redirect_test.go
```diff
@@ -356,7 +387,13 @@ func Test_Redirect_Route_WithOldInput(t *testing.T) {
 		c.RequestCtx().Request.Header.Set(HeaderCookie, c.GetRespHeader(HeaderSetCookie)) // necessary for testing
 
 		var msgs redirectionMsgs
-		_, err = msgs.UnmarshalMsg([]byte(c.Cookies(FlashCookieName)))
+		cookieValue := c.Cookies(FlashCookieName)
+		require.NotEmpty(t, cookieValue)
+
+		decodedValue, err := base64.StdEncoding.DecodeString(cookieValue)
```
efectn, line 393: I think hex is faster for this stuff  as flash messages are mostly small.

gaby:
These improvements were done and merged in https://github.com/gofiber/fiber/pull/3546 by using `net/http` Cookie validation helper.

PR #3392, opened 2025-04-03
🧪 fix: Logger Middleware tests to use regex for time validation

ReneWerner87:
to reduce flaky test run, the time is no longer compared, but the format of the time which is in the logger format is looked at


gaby (approved):
👍 LGTM

Issue #3398, opened 2025-04-09
📝 [Proposal]: middleware/requestID -> use faster alternatives


pjebs:
### Feature Proposal Description

Instead of a default generator of uuid, perhaps https://github.com/oklog/ulid which is faster and shorter

### Alignment with Express API

N/A

### HTTP RFC Standards Compliance

https://github.com/ulid/spec

### API Stability

N/A

### Feature Examples

```go
Looks like: 01D78XYFJ1PRM1WPBCBT3VHMNV instead of uuid: f81d4fae-7dec-11d0-a765-00a0c91e6bf6

But added advantage of being orderable. UUID's are not orderable.
```

### Checklist:

- [x] I agree to follow Fiber's [Code of Conduct](https://github.com/gofiber/fiber/blob/master/.github/CODE_OF_CONDUCT.md).
- [x] I have searched for existing issues that describe my proposal before opening this one.
- [x] I understand that a proposal that does not meet these guidelines may be closed without explanation.

JIeJaitt:
In what scenarios does uuid need to be sorted?

pjebs:
In this scenario (requestId) it probably doesn't matter, unless somehow the request id ends up in a database (maybe some diagnostics service that tracks errors/performance of endpoints).

But if it's in a database, perhaps you don't need to store created_at time field in some scenarios if you were only interested in ordering.

UUID will require a created_at time field. ulid will not (if you aren't interested in exact creation time, but just the ordering)

JIeJaitt:
uuid is a library maintained by Google, who pioneered the go language. So I don't see the need to replace the uuid library if it's not a big breakthrough

JIeJaitt:
perhaps we can make the use of ulid customizable for the user through other implementations

pjebs:
The proposal was just to change the default id generator because it's faster and Fiber is all about speed.



> However, perhaps we can customize the use of ulid , as a middleware, through other implementations

The middleware api already allows the user to replace the id generator with whatever they want including ulid.

JIeJaitt:
> The proposal was just to change the default id generator because it's faster and Fiber is all about speed.

you are right

gaby:
@pjebs @JIeJaitt uuid is essential for security libraries, we are not going to swap to a third party just because its faster.

The ulid one doesn't have any releases since 2022

JIeJaitt:
> it would be better if we made the uuid generation customizable
> 
> I would not like to add the other dependency in the fiber core
[...]

PR #3399, opened 2025-04-09
♻️ Refactor: Replace uuid with ulid

JIeJaitt:
# Description

Replacing uuid with ulid does not affect any existing functionality. According to the discussion, ulid should perform better than uuid.

close https://github.com/gofiber/fiber/issues/3398

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,6 @@ go 1.23.0
 require (
 	github.com/gofiber/schema v1.3.0
 	github.com/gofiber/utils/v2 v2.0.0-beta.8
-	github.com/google/uuid v1.6.0
 	github.com/mattn/go-colorable v0.1.14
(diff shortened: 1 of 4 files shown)


ReneWerner87:
it would be better if we made the uuid generation customizable

I would hate to add the other dependency in the fiber core
we should do it like with the json generation
https://docs.gofiber.io/guide/faster-fiber#custom-json-encoderdecoder

JIeJaitt:
> it would be better if we made the uuid generation customizable
> 
> I would not like to add the other dependency in the fiber core we should do it like with the json generation [docs.gofiber.io/guide/faster-fiber#custom-json-encoderdecoder](https://docs.gofiber.io/guide/faster-fiber#custom-json-encoderdecoder)

I'll look into whether the uuid can also be self-selected by the user in this way

PR #3402, opened 2025-04-11
♻️ refactor: Return fast from parseToStruct() when data is empty

billyplus:
# Description

"parseToStruct" return fast when data is empty

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -1082,10 +1082,16 @@ func (c *Ctx) AllParams() map[string]string {
 
 // ParamsParser binds the param string to a struct.
 func (c *Ctx) ParamsParser(out interface{}) error {
-	params := make(map[string][]string, len(c.route.Params))
+	// fast return
+	l := len(c.route.Params)
+	if l == 0 {
+		return nil
+	}
+	params := make(map[string][]string, l)
 	for _, param := range c.route.Params {
 		params[param] = append(params[param], c.Params(param))
 	}
+
 	return c.parseToStruct(paramsTag, out, params)
 }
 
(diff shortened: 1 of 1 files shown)


ReneWerner87 (approved):

gaby:
@billyplus @ReneWerner87 Do we need this change in v3? The PR is for v2

gaby (approved):
👍 LGTM

efectn (approved):

efectn (approved):

ctx.go
```diff
@@ -1308,6 +1314,10 @@ func (c *Ctx) ReqHeaderParser(out interface{}) error {
 }
 
 func (*Ctx) parseToStruct(aliasTag string, out interface{}, data map[string][]string) error {
+	// fast return
+	if len(data) == 0 {
```
efectn, line 1318: can you check whether the data map is nil

billyplus: if data is nil, then len(data) is 0.

ReneWerner87 (changes requested):
thanks for the work
unfortunately the customization breaks the current functionality

you could think the fast return with no values should not break anything, but there is a case where you have a required field in the struct itself
so the schema parser must always be executed, even if there is no data

https://github.com/billyplus/fiber/blob/7eb9d255489192343c647f1d910abbb140705474/ctx_test.go#L4661

billyplus:
> thanks for the work unfortunately the customization breaks the current functionality
> 
> you could think the fast return with no values should not break anything, but there is a case where you have a required field in the struct itself so the schema parser must always be executed, even if there is no data
> 
> https://github.com/billyplus/fiber/blob/7eb9d255489192343c647f1d910abbb140705474/ctx_test.go#L4661

maybe, split a validator from decoder function could be a better way.

PR #3415, opened 2025-04-19
♻️ refactor: Change c.Redirect() default status

### LINKED ISSUE #3405: 🐛 [Bug]: c.Redirect doesn't respect c.Method
### Bug Description

Calling `c.Method("GET")` before `c.Redirect("/my/route")` doesn't set the method and `c.Redirect` still uses the method that the handler was originally called with.

### How to Reproduce

1. Create route `app.Put("/put", putHandler)`
2. Create route `app.Get("/get", getHandler)`
3. At the bottom of `putHandler` have `c.Method("GET"); return c.Redirect("/get")`
4. Call `/put` with "PUT" HTTP method/verb
5. Notice the redirect is not `GET /get` but `PUT /get` despite method being set in item 3 above

### Expected Behavior

`c.Method("GET")` is respected and overrides the method/verb of the original request (PUT, in this example).

### Fiber Version

v2.52.6

### Code Snippet (optional)

```go
package main

import "log"

func main() {
  app := fiber.New()
  app.Get("/get", getHandler)
  app.Put("/put", putHandler)

}

// This will NEVER be called
func getHandler(c *fiber.Ctx) error {
  c.SendString("")
}

func putHandler(c *fiber.Ctx) error {
  log.Debugf("got here and method it PUT: %v", c.Method())

  c.Method("GET")
  log.Debugf("now the method is set to GET: %v", c.Method())

  // This results in a "PUT /get" request instead of "GET /get"
  c.Redirect("/get")
}
```

### Checklist:

andradei:
Fixes #3405

In some browsers, redirect status 302 Found sometimes is used to change the HTTP verb of the response from what the user set to what was used in the request. Changing to 303 SeeOther in the default works more like expected: it defaults to GET and can be overriden by the user.


andradei:
Tests passed on my linux host with no issues, not sure what caused the error in `ubuntu-latest` except it looks like it ran an older commit for its tests.

gaby:
@andradei How are you running the tests? What's your command?

redirect.go
```diff
@@ -92,13 +92,13 @@ func ReleaseRedirect(r *Redirect) {
 }
 
 func (r *Redirect) release() {
-	r.status = 302
+	r.status = StatusSeeOther
```
gaby, line 95: @andradei This makes me wonder if we forgot to call release() somewhere before your changes, and it got exposed now

gaby: I think I found why

gaby: A missing ReleaseCtx() was the problem. Actually all the tests needed it

docs/whats_new.md
```diff
@@ -577,6 +577,9 @@ In this example, the `Bind` method is used to bind the request body to the `User
 
 Fiber v3 enhances the redirect functionality by introducing new methods and improving existing ones. The new redirect methods provide more flexibility and control over the redirection process.
 
+The default response status is changes from 302 Found to 303 See Other for more
```
gaby, line 580: This statement needs work. It should be a separate section.

docs/whats_new.md
```diff
@@ -256,7 +256,7 @@ The route method is now like [`Express`](https://expressjs.com/de/api.html#app.r
 
 ```diff
 -    Route(prefix string, fn func(router Router), name ...string) Router
-+    Route(path string) Register
```
gaby, line 259: Why was this changed?

middleware/expvar/expvar_test.go
```diff
@@ -81,9 +81,9 @@ func Test_Expvar_Other_Path(t *testing.T) {
 		return c.SendString("escaped")
 	})
 
-	resp, err := app.Test(httptest.NewRequest(fiber.MethodGet, "/debug/vars/302", nil))
+	resp, err := app.Test(httptest.NewRequest(fiber.MethodGet, "/debug/vars/303", nil))
 	require.NoError(t, err)
-	require.Equal(t, 302, resp.StatusCode)
+	require.Equal(t, 303, resp.StatusCode)
```
gaby, line 86: Use fiber.StatusSeeOther

redirect_test.go
```diff
@@ -19,10 +19,11 @@ func Test_Redirect_To(t *testing.T) {
 	t.Parallel()
 	app := New()
 	c := app.AcquireCtx(&fasthttp.RequestCtx{})
+	defer app.ReleaseCtx(c)
 
 	err := c.Redirect().To("http://default.com")
 	require.NoError(t, err)
-	require.Equal(t, 302, c.Response().StatusCode())
+	require.Equal(t, 303, c.Response().StatusCode())
```
gaby, line 26: Use fiber.StatusSeeOther

redirect_test.go
```diff
@@ -39,7 +40,7 @@ func Test_Redirect_To_WithFlashMessages(t *testing.T) {
 
 	err := c.Redirect().With("success", "2").With("success", "1").With("message", "test", 2).To("http://example.com")
 	require.NoError(t, err)
-	require.Equal(t, 302, c.Response().StatusCode())
+	require.Equal(t, 303, c.Response().StatusCode())
```
gaby, line 43: Use fiber.StatusSeeOther

redirect_test.go
```diff
@@ -68,7 +69,7 @@ func Test_Redirect_Route_WithParams(t *testing.T) {
 		},
 	})
 	require.NoError(t, err)
-	require.Equal(t, 302, c.Response().StatusCode())
+	require.Equal(t, 303, c.Response().StatusCode())
```
gaby, line 72: Use fiber.StatusSeeOther

gaby (changes requested):
Use fiber.StatusSeeOther instead of "302"

docs/whats_new.md
ReneWerner87: @andradei

docs/whats_new.md
ReneWerner87: @andradei can you improve this

ReneWerner87:
@andradei can you fix the review hints

docs/whats_new.md
ReneWerner87: just spaces remove -> its ok

ReneWerner87 (approved):

gaby (approved):
👍 LGTM

PR #3418, opened 2025-04-21
🐛 fix: Handling of next param position

### LINKED ISSUE #3289: 🐛 [Bug]: Potential issues with parameter parsing in findNextParamPosition and handling of routes with multiple` : `
### Bug Description

Hi Fiber team,
I’ve noticed two potential issues in the route-parsing logic:
The branch is main. fiber v3.
1. Ineffective loop in path.go, func findNextParamPosition:
```
for found := findNextNonEscapedCharsetPosition(pattern[nextParamPosition+1:], parameterStartChars); found == 0; {
    nextParamPosition++
    if len(pattern) > nextParamPosition {
        break
    }
}
```
The condition if len(pattern) > nextParamPosition { break } appears to render the loop meaningless. Because nextParamPosition is guaranteed to be < len(pattern), the loop will always break after one iteration, which doesn’t seem to match the intended logic of skipping multiple consecutive : or similar parameter characters.

2. Route like "/a:::" treated as having a parameter:
If I register GET("/a:::"), tools like [Express Route Tester](https://forbeslindesay.github.io/express-route-tester/) treat it as a purely static route. However, Fiber currently interprets the final : as a parameter indicator. This creates an inconsistency with other routing systems where multiple : characters at the end might be interpreted as part of a static path or simply invalid syntax.

Could you clarify if this is the intended behavior? If not, might it be possible to update the logic in findNextParamPosition (and related parsing code) so that multiple consecutive : are handled more consistently, and routes like "/a:::" are treated as static?

Thank you for your time and for all your work on Fiber!

### How to Reproduce

skip

### Expected Behavior

skip

### Fiber Version

v3 main

### Code Snippet (optional)

```go

```

### Checklist:

KatzuYoru:
PR: Fix route parameter parsing logic

- Fixed infinite loop in findNextParamPosition function by properly recalculating the found variable on each iteration
- Corrected boundary check to prevent index out of range errors when processing consecutive parameter characters
- Improved consistency with other routing systems by handling multiple consecutive parameter indicators (like :::) as part of static routes rather than parameters

Fixes #3289


sixcolors (changes requested):
Please run the tests before submitting a PR—see `make test` and other targets in the Makefile. Also, make sure to update tests where necessary.

KatzuYoru:
sir the tests for path.go are passing now locally, how to check the test and coverage here?

gaby:
@KatzuYoru Check the linter failure.

KatzuYoru:
done sir.

gaby (approved):

PR #3426, opened 2025-04-25
🐛 bug: Add Read Lock around app.Server()

### LINKED ISSUE #3362: 🐛 [Bug]: server: protect field access with lock to avoid possible data race
### Bug Description

Fix Data Race Risk in fiber/app.go
Description:
In the fiber/app.go file, there is a potential data race risk related to the app.server field. Specifically:
The app.server field is read/written 21 times.20 of these accesses are properly synchronized using app.mutex.RLock() or Lock().
1 read operation (in the Server() function at line 882) is not locked, which may cause a data race when Server() and other functions like init() are called concurrently.
Suggested Fix:
Use app.mutex.RLock() or Lock() consistently to protect all accesses to app.server.
And store the result in a local variable to minimize lock contention.
Could you please help confirm this bug?

### Expected Behavior

Data Race

### Fiber Version

v2.42.0

### Code Snippet (optional)

```go

```

### Checklist:

gaby:
# Description

- Add Read Lock around `app.Server()`

Fixes #3362

### DIFF
--- a/app.go
+++ b/app.go
@@ -961,7 +961,9 @@ func (app *App) ShutdownWithContext(ctx context.Context) error {
 
 // Server returns the underlying fasthttp server
 func (app *App) Server() *fasthttp.Server {
-	return app.server
+    app.mutex.RLock()
+    defer app.mutex.RUnlock()
+    return app.server
 }
 
 // Hooks returns the hook struct to register hooks.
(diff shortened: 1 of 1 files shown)


gaby:
Closing this for now, tests will never finish with the RWLock

Issue #3400, opened 2025-04-09
📝 [Proposal]: Force color logging


yorickdewid:
### Current

At the moment logging with colors is enabled by default (`enableColors=true`). If certain criteria are met the logger stream is set to `NewNonColorable(...)`, and no color is used in the log output.  One of these tests is to see if a TTY terminal is connected `isatty.IsTerminal(os.Stdout.Fd())` but there are cases which do support colored output without being connected to a TTY terminal. Examples of these scenarios are:
- Docker output
- Kubernetes

### Proposal

Either drop the terminal check or introduce a new config option `forceColors` to enable colors either way.

If either one is accepted I will submit a PR

### Alignment with Express API

N/A

### HTTP RFC Standards Compliance

N/A

### API Stability

N/A

### Feature Examples

```go
N/A
```

### Checklist:

- [x] I agree to follow Fiber's [Code of Conduct](https://github.com/gofiber/fiber/blob/master/.github/CODE_OF_CONDUCT.md).
- [x] I have searched for existing issues that describe my proposal before opening this one.
- [x] I understand that a proposal that does not meet these guidelines may be closed without explanation.

gaby:
@yorickdewid Are you running the container with a tty? `docker run -it`, or in k8s `tty: true`.

I run mine with a tty and colors show up

gaby:
@yorickdewid It's a field in your deployment/pod yaml, not in k8s cluster.

Like this:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: busybox
spec:
  containers:
  - name: shell
    image: busybox:1.28
    command: ["sleep", "3600"]
    tty: true
```

yorickdewid:
Yeah I get what you're saying but apps are deployed automatically using Heroku tools. This is DigitalOcean's App Platform. There is nothing to configure. The app platform is built on top of Kubernetes, but thats abstracted away from the user. Many cloud providers offer similar services.

yorickdewid:
@gaby

gaby:
@ReneWerner87 @efectn Thoughts?

gaby:
@yorickdewid Target branch is `main` which is Fiber v3. Option should be `False` by default.

efectn:
I am OK with the change, too.

PR #3428, opened 2025-04-28
Force the use of terminal colors via config option

yorickdewid:
# Description

Force the use of colors in the terminal if the user enables this explicitly.

As discussed in #3400

Included a single unit test `Test_Logger_ForceColors` to test this behavior

### DIFF
--- a/middleware/logger/config.go
+++ b/middleware/logger/config.go
@@ -89,6 +89,11 @@ type Config struct {
 	// Default: false
 	DisableColors bool
 
+	// ForceColors forces the colors to be enabled even if the output is not a terminal
+	//
+	// Default: false
+	ForceColors bool
+
 	enableColors  bool
 	enableLatency bool
 }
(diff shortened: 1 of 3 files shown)


gaby:
@yorickdewid Please update the documentation in the `docs` directory.

Also add an entry in `docs/whats_new.md`.

yorickdewid:
@gaby Where do you want me to add this 'ForceColors' in the `whats new`. The `App` section or the `Logging` section?

Also, Am I seeing this correctly that the _use_ of colors in the output is never actually tested? From what I gather the only testcase doing something with color is `Test_Logger_All` (`logger_test.go`). This doesn't test the use of colors, just that the colors in the output match the ColorScheme. For example colors are used in the `${status}` tag, but this tag is not tested in the output. Can you confirm this?

yorickdewid:
@gaby

RajaSunrise (approved):

ReneWerner87:
@yorickdewid thanks for this feature, can you check the review hints and add it to our "what´s new" overview

ReneWerner87:
@yorickdewid `Logger` section sounds good

ReneWerner87:
@yorickdewid thanks for the customization
in the tests i would say before it was not really possible to test the colors because you could not force it
now it should be possible because you can activate the switch and not look and decide for yourself if the console supports colors
i.e. the best would be a test where you force colors by collecting the content with a buffer writer and check if the color codes are in the right places in the format
This of course ensures that we can make sure that the feature works after customization in the future and that we find errors in other pull requests faster

ReneWerner87:
the Test_Logger_ForceColors test is enough i think

ReneWerner87 (approved):

PR #3463, opened 2025-05-21
Add variadic message support to NewError

gaby:
## Summary
- support format string args for `NewError`
- update sendfile error to use `NewError` formatting
- document new `NewError` signature and examples
- test formatting support in `NewError`

## Testing
- `go test ./...` *(fails: lookup google.com: no such host)*

### DIFF
--- a/app.go
+++ b/app.go
@@ -881,14 +881,19 @@ func (e *Error) Error() string {
 	return e.Message
 }
 
-// NewError creates a new Error instance with an optional message
-func NewError(code int, message ...string) *Error {
+// NewError creates a new Error instance with an optional message.
+// Additional arguments are formatted using fmt.Sprintf when provided.
+func NewError(code int, message ...interface{}) *Error {
 	err := &Error{
 		Code:    code,
 		Message: utils.StatusMessage(code),
 	}
 	if len(message) > 0 {
-		err.Message = message[0]
+		if format, ok := message[0].(string); ok && len(message) > 1 {
+			err.Message = fmt.Sprintf(format, message[1:]...)
+		} else {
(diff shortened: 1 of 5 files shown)


ReneWerner87 (approved):

app.go
```diff
@@ -881,16 +881,37 @@ func (e *Error) Error() string {
...
+	var msg string
+
+	switch len(message) {
+	case 0:
+		// nothing to override
+		msg = utils.StatusMessage(code)
+
+	case 1:
+		// One argument → treat it like fmt.Sprint(arg)
+		if s, ok := message[0].(string); ok {
```
gaby, line 898: This check is probably the performance hit from the benchmark

ReneWerner87:
@gaby can we improve the performance ?
or should we leave this in the case as it is an error case

app.go
```diff
@@ -881,16 +881,37 @@ func (e *Error) Error() string {
...
+			msg = fmt.Sprint(message[0])
+		}
+
+	default:
+		// Two or more → first must be a format string.
+		if format, ok := message[0].(string); ok {
+			msg = fmt.Sprintf(format, message[1:]...)
+		} else {
+			// If the first arg isn’t a string, fall back.
+			msg = fmt.Sprint(message[0])
```
ReneWerner87, line 910: use utils.ToString

app.go
ReneWerner87: - [ ] NewErrorf method

app.go
gaby: @ReneWerner87 Can't use utils, since it won't render the users string

PR #3469, opened 2025-05-23
fix: math.MaxUint32 overflows int and test cases

viralkansarav:
# Description

Issue fixed:
This check prevents generating an ETag for response bodies larger than the maximum value of a 32-bit unsigned integer. Since the ETag calculation uses a uint32 for the body length, allowing larger bodies could cause integer overflows or incorrect ETag values. By returning a 413 Request Entity Too Large status, the middleware safely handles oversized responses. 

Fixes # (issue)
avoids potential bugs or vulnerabilities related to integer overflow.

### DIFF
--- a/ctx_test.go
+++ b/ctx_test.go
@@ -5474,11 +5474,11 @@ func Test_GenericParseTypeInt64s(t *testing.T) {
 		},
 		{
 			value: int64(math.MaxInt64),
-			str:   strconv.Itoa(math.MaxInt64),
+			str:   strconv.FormatInt(math.MaxInt64, 10),
 		},
 		{
 			value: int64(math.MinInt64),
-			str:   strconv.Itoa(math.MinInt64),
+			str:   strconv.FormatInt(math.MinInt64, 10),
 		},
 	}
 
@@ -5664,7 +5664,7 @@ func Test_GenericParseTypeUint32s(t *testing.T) {
 		},
 		{
 			value: uint32(math.MaxUint32),
-			str:   strconv.Itoa(math.MaxUint32),
+			str:   strconv.FormatUint(uint64(math.MaxUint32), 10),
 		},
 	}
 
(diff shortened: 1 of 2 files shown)


ReneWerner87:
@viralkansarav thx for the code
can you check 
image

viralkansarav:
The check if uint64(bodyLength) > uint64(math.MaxUint32) ensures that bodyLength will always fit into a uint32 (i.e., it’s ≤ 4,294,967,295). After this check, it’s safe to cast bodyLength to uint32 for appendUint, which expects a uint32 argument.
Can you please change the line of code as i might not be having access to commit in forked repo,  for the linter  we need to assign value to a variable
		bodylength32 := uint32(bodyLength)
		bb.B = appendUint(bb.Bytes(), bodylength32)
later we can use the variable in append.
I check that bodyLength does not exceed math.MaxUint32 before converting to uint32. The conversion is therefore always safe, and this pattern is accepted by gosec.

middleware/etag/etag.go
```diff
@@ -61,10 +61,9 @@ func New(config ...Config) fiber.Handler {
 		bb.WriteByte('"')
 
 		bodyLength := len(body)
-		if bodyLength > math.MaxUint32 {
+		if uint64(bodyLength) > uint64(math.MaxUint32) {
```
gaby, line 64: This is incorrect

ctx_test.go
```diff
@@ -5474,11 +5474,11 @@ func Test_GenericParseTypeInt64s(t *testing.T) {
 		},
 		{
 			value: int64(math.MaxInt64),
-			str:   strconv.Itoa(math.MaxInt64),
+			str:   strconv.FormatInt(math.MaxInt64, 10),
```
gaby, line 5477: This is not needed, golang already does this inside Itoa

https://go.dev/src/strconv/itoa.go

gaby:
@viralkansarav These changes are not needed. We already check for max u32. The itoa function in golang is an alias to the FormatInt one you added.

ReneWerner87:
thank you for your work on the project, but due to the findings we decided against the adjustment in the PR

PR #3487, opened 2025-05-27
Fix flaky timing tests

ReneWerner87:
## Summary
- bump sleep durations in cache tests
- add extra buffer to limiter sleeps
- relax timing in idempotency and timeout tests

## Testing
- `go test ./...`


ReneWerner87:
not needed atm

PR #3503, opened 2025-06-05
Add index limit for parsers

gaby:
## Summary
- avoid OOM for slices by limiting parsed indexes to 1000
- distinguish index overflow from invalid path
- update BodyParser tests for overflow handling

## Testing
- `make test` *(fails: Test_Proxy_Do_WithRealURL and related proxy tests due to network restrictions)*

------
https://chatgpt.com/codex/tasks/task_e_6840ecc5afa48333aa04b21a7c296fe9

### DIFF
--- a/ctx_test.go
+++ b/ctx_test.go
@@ -676,7 +676,18 @@ func Test_Ctx_BodyParser_InvalidRequestData(t *testing.T) {
 	err := c.BodyParser(subject)
 
 	utils.AssertEqual(t, true, nil != err)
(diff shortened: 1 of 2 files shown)


internal/schema/decoder.go
```diff
@@ -94,7 +94,11 @@ func (d *Decoder) Decode(dst interface{}, src map[string][]string) (err error) {
 				multiError[path] = err
 			}
 		} else if !d.ignoreUnknownKeys {
```
ReneWerner87: what happens if ignoreUnknownKeys is off and our error is thrown ? is the too large index then not declared as an error ?
I think we would have to make the error handling independent of this switch

PR #3505, opened 2025-06-05
Refine CORS middleware

gaby:
## Summary
- keep the original casing of the `Origin` header when responding
- separate simple vs. preflight header helpers
- avoid `Access-Control-Max-Age` on simple requests
- add regression tests for new behaviours

## Testing
- `go test ./...`

------
https://chatgpt.com/codex/tasks/task_e_68353eb593148333ad76a9813eeccee2

### DIFF
--- a/middleware/cors/cors.go
+++ b/middleware/cors/cors.go
@@ -82,8 +82,9 @@ func New(config ...Config) fiber.Handler {
 			return c.Next()
 		}
 
-		// Get originHeader header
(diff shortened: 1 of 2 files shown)


sixcolors:
Neither issue are Non-RFC compliant. Origin header values are case-insensitive and max-age is not required in a non-preflight, but having it also isn’t explicitly not allowed. 

The fixes are however improving the middleware. I just changed the title.

PR #3508, opened 2025-06-07
🧹 chore: Replace math/rand with crypto/rand

gaby:
## Summary
- replace math/rand with crypto/rand
- remove unused constants from the multipart boundary helper
- verify generated random strings use only allowed characters

### DIFF
--- a/client/hooks.go
+++ b/client/hooks.go
@@ -4,6 +4,7 @@ import (
 	"crypto/rand"
 	"fmt"
 	"io"
+	"math/big"
 	"mime/multipart"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


sixcolors (approved):
The switch to `crypto/rand` isn't necessary for generating multipart boundaries, but it's fine. It avoids `gosec` warnings, adds negligible overhead, and improves security hygiene. The function name is fine—"unsafe" refers to the string conversion, not the randomness. Merge is reasonable unless performance is a concern.

gaby:
> The switch to `crypto/rand` isn't necessary for generating multipart boundaries, but it's fine. It avoids `gosec` warnings, adds negligible overhead. The function name is fine—"unsafe" refers to the string conversion, not the randomness. Merge is reasonable unless performance is a concern.
> 
> Just needs full test coverage before merging.

I don't think there's an easy way to trigger a failure on reading random bytes, that's why those 2 lines are missing from tests.

ReneWerner87:
@gaby Can you share benchmark results from before and after

gaby:
The performance hit is too much, from ~10ns/op to ~300ns/op.

sixcolors:
> The switch to `crypto/rand` isn't necessary for generating multipart boundaries, but it's fine. It avoids `gosec` warnings, adds negligible overhead. The function name is fine—"unsafe" refers to the string conversion, not the randomness. Merge is reasonable unless performance is a concern.
> 
> Just needs full test coverage before merging.

GoFiber’s client uses math/rand for multipart boundaries, similar to the Node.js form-data CVE-2025-7783, making it potentially predictable. Exploitation needs attacker-controlled input and math/rand output observation (e.g., logs, webhooks), which is harder than Node.js due to server-side context. PR to switch to crypto/rand was closed for performance (~10ns to ~300ns) and concerns were dismissed by me at the time. Looking deeper: Risk is low but non-zero; consider crypto/rand and merge.

sixcolors (approved):
LGTM

PR #3321, opened 2025-02-22
fix handler order in routing

ReneWerner87:
Fixes #3312

PR #3512, opened 2025-06-09
docs: clarify handler parameters and execution order in routing methods

sixcolors:
Enhance documentation to explicitly define the required main handler and optional additional handlers in routing methods. This change improves clarity and type safety by clearly outlining the execution order of handlers.

### DIFF
--- a/app.go
+++ b/app.go
@@ -788,58 +788,148 @@ func (app *App) Use(args ...any) Router {
 
 // Get registers a route for GET methods that requests a representation
 // of the specified resource. Requests using GET should only retrieve data.
+//
+// The first parameter 'handler' is the main route handler (executed last).
+// The optional 'handlers' parameters are additional handlers that execute
+// before the main handler in left-to-right order.
+//
+// Example:
+//
+//	app.Get("/users", mainHandler, handler1, handler2)
+//	// Execution order: handler1 -> handler2 -> mainHandler
 func (app *App) Get(path string, handler Handler, handlers ...Handler) Router {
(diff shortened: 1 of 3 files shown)


app.go
```diff
@@ -788,65 +788,164 @@ func (app *App) Use(args ...any) Router {
 
 // Get registers a route for GET methods that requests a representation
 // of the specified resource. Requests using GET should only retrieve data.
+//
+// The first parameter 'handler' is the main route handler (executed last).
```
ReneWerner87, line 792: this is wrong, the first handler is executed first
@sixcolors 
app.Get("/users", mainHandler, handler1, handler2)
// Execution order: mainHandler -> handler1 -> handler2

in the order in which they were specified

ReneWerner87 (changes requested):
unfortunately wrong description

app.go
ReneWerner87: https://github.com/gofiber/fiber/pull/3321#issuecomment-2958676888

sixcolors:
The handler execution order was changed in Fiber v3 (in current beta as of Dec 2024 release) but reverted to the original behavior in PR #3321, which is in main but not yet released in a v3 beta. @ReneWerner87 to avoid confusion for users who adapted to the beta's execution order, I recommend highlighting this change in the next v3 release notes and adding a note to the documentation to clarify the switch back, as it may impact code written for the v3 beta release.

ReneWerner87:
ok good point

PR #3518, opened 2025-06-13
🚀 feat: Add support for Timeout middleware configuration

### LINKED ISSUE #3514: 🧹 [Maintenance]: Enhanced Timeout Middleware Configuration
## Summary

Add configurable timeout middleware with per-route timeouts, path exclusions, and custom timeout handlers while maintaining full backward compatibility.

## Motivation

Current timeout middleware limitations:
- Global timeout for all routes
- Cannot skip specific paths (health checks, uploads)
- Fixed timeout response
- No per-route timeout configuration

## Proposed Solution

Add `NewWithConfig` function following Fiber's established middleware patterns:

```go
type Config struct {
    // Next defines a function to skip this middleware
    Next      func(c fiber.Ctx) bool
    
    // Timeout defines the default timeout duration
    Timeout   time.Duration
    
    // OnTimeout is called when a timeout occurs
    OnTimeout fiber.Handler
    
    // SkipPaths defines paths that should ignore timeout
    SkipPaths []string
    
    // PerRoute allows specific timeouts per route
    PerRoute  map[string]time.Duration
}

func NewWithConfig(config Config) fiber.Handler
```

### Basic Configuration
```go
app.Use(timeout.NewWithConfig(timeout.Config{
    Timeout: 30 * time.Second,
}))
```

### Per-Route Timeouts
```go
app.Use(timeout.NewWithConfig(timeout.Config{
    Timeout: 5 * time.Second,
    PerRoute: map[string]time.Duration{
        "/api/reports": 30 * time.Second,
        "/api/uploads": 60 * time.Second,
    },
}))
```

### Skip Specific Paths
```go
app.Use(timeout.NewWithConfig(timeout.Config{
    Timeout: 5 * time.Second,
    SkipPaths: []string{"/health", "/metrics", "/webhook"},
}))
```

### Custom Timeout Response
```go
app.Use(timeout.NewWithConfig(timeout.Config{
    Timeout: 5 * time.Second,
    OnTimeout: func(c *fiber.Ctx) error {
        return c.Status(408).JSON(fiber.Map{
            "error": "Request timeout",
            "path":  c.Path(),
        })
    },
}))
```

## Benefits

- **Flexible**: Per-route timeout configuration
- **Production-ready**: Skip health checks and monitoring endpoints
- **User-friendly**: Custom timeout responses with context
- **Backward compatible**: Existing `New()` function unchanged

## Implementation

- Uses goroutines for asynchronous request processing
- Context-based timeout management with `context.WithTimeout`
- Zero breaking changes to existing API

## Backward Compatibility

✅ All existing `timeout.New()` calls work unchanged  
✅ Can be adopted incrementally  
✅ Follows same patterns as other Fiber middlewares (CORS, Logger)  

---

Implementation is ready with full tests and documentation. Happy to submit a PR if this approach looks good! 🙏

## Impact on the Project

This enhancement significantly increases the usefulness of the timeout middleware without introducing any breaking changes, bringing the following positive impacts:

- Greater flexibility in request timeout control, better suited to real-world production scenarios.
- Reduces the need for external workarounds or duplicated logic in handlers.
- Helps improve application stability and predictability, especially for public APIs with high-latency endpoints.
- Lays the groundwork for future improvements, such as dynamic timeouts based on headers or context.
-  Keeps existing tests intact and adds coverage for new cases without interfering with the current API.

This change aligns with Fiber’s design philosophy and should be easy for the community to adopt.

_No response_

### Checklist:

- [x] I have confirmed that this maintenance task is currently not being addressed.
- [x] I understand that this task will be evaluated by the maintainers and prioritized accordingly.
- [x] I am available to provide further information if needed.

gaby:
## Summary
- add Config struct for timeout middleware.
- allow per-route timeouts, skip paths and custom timeout handler.
- document timeout middleware configuration.
- document timeout changes in what's new.

Fixes #3514

### DIFF
--- a/docs/middleware/timeout.md
+++ b/docs/middleware/timeout.md
@@ -143,28 +143,5 @@ func main() {
 | Next       | `func(fiber.Ctx) bool`              | Function to skip the middleware.                                         | `nil`  |
 | Timeout    | `time.Duration`                     | Default timeout for routes.                                              | `0`    |
 | OnTimeout  | `fiber.Handler`                     | Handler executed when a timeout occurs.                                  | `nil`  |
-| SkipPaths  | `[]string`                          | Paths that should not have a timeout enforced.                            | `nil`  |
-| Routes     | `map[string]time.Duration`          | Specific timeout values per route path.                                  | `nil`  |
 | Errors     | `[]error`                           | Custom errors treated as timeout errors.                                 | `nil`  |
 
-### Usage Examples
-
-Per-route timeouts:
-
-```go
-app.Get("/reports", timeout.New(handler, timeout.Config{
-    Timeout: 5 * time.Second,
-    Routes: map[string]time.Duration{
-        "/reports": 30 * time.Second,
-    },
-}))
-```
-
-Skipping paths:
-
-```go
-app.Get("/health", timeout.New(handler, timeout.Config{
-    Timeout:   5 * time.Second,
-    SkipPaths: []string{"/health"},
-}))
-```
--- a/docs/middleware/timeout.md
+++ b/docs/middleware/timeout.md
@@ -17,7 +17,7 @@ It does not cancel long running executions. Underlying executions must handle ti
 ## Signatures
 
 ```go
-func New(handler fiber.Handler, timeout time.Duration, timeoutErrors ...error) fiber.Handler
+func New(handler fiber.Handler, config ...timeout.Config) fiber.Handler
 ```
 
 ## Examples
@@ -44,7 +44,7 @@ func main() {
         return nil
     }
 
-    app.Get("/foo/:sleepTime", timeout.New(h, 2*time.Second))
+    app.Get("/foo/:sleepTime", timeout.New(h, timeout.Config{Timeout: 2 * time.Second}))
     log.Fatal(app.Listen(":3000"))
 }
 
@@ -90,7 +90,7 @@ func main() {
         return nil
     }
 
-    app.Get("/foo/:sleepTime", timeout.New(h, 2*time.Second, ErrFooTimeOut))
+    app.Get("/foo/:sleepTime", timeout.New(h, timeout.Config{Timeout: 2 * time.Second, Errors: []error{ErrFooTimeOut}}))
     log.Fatal(app.Listen(":3000"))
 }
 
@@ -129,7 +129,42 @@ func main() {
         return nil
     }
 
-    app.Get("/foo", timeout.New(handler, 10*time.Second))
+    app.Get("/foo", timeout.New(handler, timeout.Config{Timeout: 10 * time.Second}))
     log.Fatal(app.Listen(":3000"))
 }
+```
+
+## Config
+
+### Config
+
+| Property   | Type                                | Description                                                               | Default |
+|:-----------|:------------------------------------|:--------------------------------------------------------------------------|:-------|
+| Next       | `func(fiber.Ctx) bool`              | Function to skip the middleware.                                         | `nil`  |
+| Timeout    | `time.Duration`                     | Default timeout for routes.                                              | `0`    |
+| OnTimeout  | `fiber.Handler`                     | Handler executed when a timeout occurs.                                  | `nil`  |
+| SkipPaths  | `[]string`                          | Paths that should not have a timeout enforced.                            | `nil`  |
+| Routes     | `map[string]time.Duration`          | Specific timeout values per route path.                                  | `nil`  |
+| Errors     | `[]error`                           | Custom errors treated as timeout errors.                                 | `nil`  |
+
+### Usage Examples
+
+Per-route timeouts:
+
+```go
+app.Get("/reports", timeout.New(handler, timeout.Config{
+    Timeout: 5 * time.Second,
+    Routes: map[string]time.Duration{
+        "/reports": 30 * time.Second,
+    },
+}))
+```
+
+Skipping paths:
+
+```go
+app.Get("/health", timeout.New(handler, timeout.Config{
+    Timeout:   5 * time.Second,
+    SkipPaths: []string{"/health"},
+}))
 ```
--- a/middleware/timeout/config.go
+++ b/middleware/timeout/config.go
@@ -52,8 +52,14 @@ func configDefault(config ...Config) Config {
 
 	cfg := config[0]
 
-	if cfg.Timeout < 0 {
-		cfg.Timeout = ConfigDefault.Timeout
+	if cfg.Routes == nil {
+		cfg.Routes = ConfigDefault.Routes
+	} else {
+		for p, d := range cfg.Routes {
+			if d < 0 {
+				cfg.Routes[p] = 0
+			}
+		}
 	}
 	if cfg.SkipPaths == nil {
 		cfg.SkipPaths = ConfigDefault.SkipPaths
--- a/middleware/timeout/config.go
+++ b/middleware/timeout/config.go
@@ -16,27 +16,10 @@ type Config struct {
 	// Optional. Default: nil (return fiber.ErrRequestTimeout)
 	OnTimeout fiber.Handler
 
-	// Routes allows specifying timeouts per path. If a path is present,
-	// its timeout overrides the default Timeout value.
-	// Optional. Default: nil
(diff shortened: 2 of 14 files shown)


middleware/timeout/timeout.go
```diff
@@ -3,43 +3,62 @@ package timeout
...
-// runHandler executes the handler and returns fiber.ErrRequestTimeout if it
-// sees a deadline exceeded error or one of the custom "timeout-like" errors.
-func runHandler(c fiber.Ctx, h fiber.Handler, tErrs []error) error {
-	// Execute the wrapped handler synchronously.
+// runHandler executes the handler and handles timeout-like errors.
+func runHandler(c fiber.Ctx, h fiber.Handler, cfg Config) error {
 	err := h(c)
-	// If the context has timed out, return a request timeout error.
-	if err != nil && (errors.Is(err, context.DeadlineExceeded) || isCustomError(err, tErrs)) {
+	if err != nil && (errors.Is(err, context.DeadlineExceeded) || isCustomError(err, cfg.Errors)) {
```
Andrei-hub11: // Add this
hasCustomErrors := len(cfg.Errors) > 0

And then:

if err != nil && (errors.Is(err, context.DeadlineExceeded) || 
    (hasCustomErrors && isCustomError(err, cfg.Errors))) {
    
 //This will avoid unnecessary loops.

middleware/timeout/config.go
```diff
@@ -0,0 +1,72 @@
...
+	if cfg.Routes == nil {
+		cfg.Routes = ConfigDefault.Routes
+	}
+	if cfg.OnTimeout == nil {
+		cfg.OnTimeout = ConfigDefault.OnTimeout
+	}
+	if cfg.Next == nil {
+		cfg.Next = ConfigDefault.Next
+	}
+	if cfg.Errors == nil {
```
Andrei-hub11, line 59: if cfg.Timeout < 0 {
        cfg.Timeout = ConfigDefault.Timeout  // To avoid the chances of negative values
    }

docs/middleware/timeout.md
```diff
@@ -129,7 +129,42 @@ func main() {
...
+
+### Config
+
+| Property   | Type                                | Description                                                               | Default |
+|:-----------|:------------------------------------|:--------------------------------------------------------------------------|:-------|
+| Next       | `func(fiber.Ctx) bool`              | Function to skip the middleware.                                         | `nil`  |
+| Timeout    | `time.Duration`                     | Default timeout for routes.                                              | `0`    |
+| OnTimeout  | `fiber.Handler`                     | Handler executed when a timeout occurs.                                  | `nil`  |
+| SkipPaths  | `[]string`                          | Paths that should not have a timeout enforced.                            | `nil`  |
+| Routes     | `map[string]time.Duration`          | Specific timeout values per route path.                                  | `nil`  |
```
ReneWerner87: not sure if we should allow this feature
since the middleware hangs on a certain route, the route is already handled with a timeout config

to have further configurations of the routes in there I see as wrong

also people will want the routes to work with wildcard and other route specific matchings

don't know if we should allow this feature in the middleware config which itself is bound to a route if it already works

f.e.
```go
app.Get("/reports", timeout.New(handler, timeout.Config{
    Timeout: 5 * time.Second,
    Routes: map[string]time.Duration{
        "/reports": 30 * time.Second,
    },
}))
```
what added value does the timeout value have if the actual value is in the routes map


same with the skip feature
for this we have the skip middleware 
https://docs.gofiber.io/next/middleware/skip

Andrei-hub11: @ReneWerner87 I understand your points, but I believe the value of this feature lies in the convenience it provides. I prefer configuring everything in a single place rather than spreading it across multiple routes. If the framework itself offered this support natively, even better.
As for the Timeout property, the intention was for it to serve as the default value for routes that don’t specify one. If we decide to move forward with this, we can simply rename it to DefaultTimeout if you think that would make it clearer.

gaby: @ReneWerner87 What if we remove the handler param, and require the middleware to be used with `app.Use()` ?

Andrei-hub11: Now I understand better. I thought we were already planning to use it in app.Use() instead of on individual routes, based on the suggested idea:

```
app.Use(timeout.NewWithConfig(timeout.Config{
    Timeout: 5 * time.Second,
    PerRoute: map[string]time.Duration{
        "/api/reports": 30 * time.Second,
        "/api/uploads": 60 * time.Second,
    },
}))
```

@gaby, wouldn’t it be better to keep the original usage below unchanged and just create a separate method, preserving backward compatibility?

```
app.Get("/foo/:sleepTime", timeout.New(h, 2*time.Second))
```

gaby: @Andrei-hub11 The `main` branch is tracking the new major version of `GoFiber`. Breaking changes are `allowed`.

ReneWerner87: would like to separate route and middleware logic
that didn't work well with the health middleware back then and caused countless bug reports

where the route is defined, the timeout middleware can be defined with a specific time or globally

you can specify a route in app.use and also operate handler chaining

```go
app.Get("/api/reports", timeout.NewWithConfig(timeout.Config{
    Timeout: 30 * time.Second,
}), myHanlderLogic)
```

or 

```go
app.Use("/api/reports", timeout.NewWithConfig(timeout.Config{
    Timeout: 30 * time.Second,
})
app.Get("/api/reports", myHanlderLogic)
```

or

```go
app.Route("/api/reports").Use(timeout.NewWithConfig(timeout.Config{
    Timeout: 30 * time.Second,
}).Get(myHanlderLogic)
```

Andrei-hub11: @ReneWerner87, I think using timeout.NewWithConfig() in the same place where a route is defined would feel a bit awkward, as it creates a conceptual ambiguity.

Since changes can be made, and if we ever need specific adjustments for individual route usage, we can simply refactor timeout.New() so that timeout.NewWithConfig() can be used independently with app.Use().

### Clarity of purpose:

timeout.New() = "This specific route needs X seconds"

```
app.Get("/api/reports", timeout.New(30*time.Second))
```

timeout.NewWithConfig() = "I'm configuring timeouts for the entire application"

```
app.Use(timeout.NewWithConfig(timeout.Config{
    DefaultTimeout: 5 * time.Second,
    Routes: map[string]time.Duration{
        "/api/videos": 30 * time.Second,
        "/api/uploads": 60 * time.Second,
    },
    SkipPaths: []string{"/health"},
}))
```

ReneWerner87: the one with the routes will not work well, people will then think they can also define dynamic routes not just completely static ones like `“/api/user/:userId”`

gaby:
Note: Add support for `Next()`

docs/middleware/timeout.md
Andrei-hub11: @ReneWerner87 Man, how about making it handle /api/user/ as matching /api/user/123 and similar paths? That would solve it and be intuitive, as long as it's documented.

```
app.Use(timeout.NewWithConfig(timeout.Config{
    DefaultTimeout: 5 * time.Second,
    RoutePrefixes: map[string]time.Duration{
        "/api/users/":    15 * time.Second, // matches: /api/users/123, /api/users/456/profile
        "/api/reports/":  30 * time.Second, // matches: /api/reports/daily, /api/reports/monthly/2024
        "/api/uploads/":  60 * time.Second, // matches: /api/uploads/images, /api/uploads/documents/pdf
        "/admin/":        10 * time.Second, // matches: /admin/dashboard, /admin/users/edit
    },
    SkipPaths: []string{"/health", "/metrics"},
}))

```

gaby: It is not that simple to implement, it requires using the router within the handler.

Andrei-hub11: Apologies — I think I may have misunderstood something. Are you saying that something like strings.HasPrefix(c.Path(), prefix) wouldn't be sufficient?

ReneWerner87:
@gaby can you solve the syntax errors

ReneWerner87:
https://github.com/gofiber/fiber/pull/3604

PR #3519, opened 2025-06-14
fix: Apply default timeout on negative values and optimize skip path …

Andrei-hub11:
@gaby, this PR fixes what I suggested earlier.
It makes sure a default timeout is used if a wrong (negative) value is set.
It speeds up how skipped paths are set up.
It also avoids an unnecessary loop.


gaby:
@Andrei-hub11 Thanks, will review in the morning

gaby:
@Andrei-hub11 Running `make betteralign` should fix the golangci failure

PR #3522, opened 2025-06-17
🧹chore: Improve BasicAuth middleware default security

gaby:
## Summary
- Add new `StorePassword` config option. Defaults to `False` to avoid leaking credentials via `Context`.
- Add `charset` parameter. Defaults to `UTF-8`.
- Set `"no-store"` header for Cache Control.


ReneWerner87 (approved):

ReneWerner87:
@sixcolors can you check

PR #3527, opened 2025-06-18
🧹 chore: Add missing CSRF token extractor tests

gaby:
## Summary
- add tests for missing CSRF token extractors


gaby:
For some reason the workflows didn't trigger.

PR #3536, opened 2025-06-22
🐛 bug: Fix Content-Type comparison in Is()

gaby:
## Summary
- handle parameters and casing when comparing MIME types in `Ctx.Is`.
- update tests for case-insensitive comparison.
- add test that mismatched subtype does not match.

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -884,7 +884,6 @@ func (c *DefaultCtx) Is(extension string) bool {
 	}
 
 	ct := utils.UnsafeString(c.fasthttp.Request.Header.ContentType())
-ct := utils.UnsafeString(c.fasthttp.Request.Header.ContentType())
 	if i := strings.IndexByte(ct, ';'); i != -1 {
 		ct = ct[:i]
 	}
--- a/ctx.go
+++ b/ctx.go
@@ -887,7 +887,7 @@ func (c *DefaultCtx) Is(extension string) bool {
 	if i := strings.IndexByte(ct, ';'); i != -1 {
 		ct = ct[:i]
 	}
-	ct = strings.TrimSpace(ct)
+	ct = utils.Trim(ct, ' ')
 	return utils.EqualFold(ct, extensionHeader)
 }
 
--- a/ctx.go
+++ b/ctx.go
@@ -883,10 +883,13 @@ func (c *DefaultCtx) Is(extension string) bool {
 		return false
 	}
 
-	return strings.HasPrefix(
-		utils.TrimLeft(utils.UnsafeString(c.fasthttp.Request.Header.ContentType()), ' '),
-		extensionHeader,
-	)
+	ct := utils.UnsafeString(c.fasthttp.Request.Header.ContentType())
+	ct = utils.TrimLeft(ct, ' ')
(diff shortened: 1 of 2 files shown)


gaby:
> # ⚠️ **Performance Alert** ⚠️
> Possible performance regression was detected for benchmark. Benchmark result of this commit is worse than the previous benchmark result exceeding threshold `1.50`.
> 
> Benchmark suite	Current: [a4e34bf](https://github.com/gofiber/fiber/commit/a4e34bf19191a769554c29220fdd1ee30e9f658b)	Previous: [e1a7c11](https://github.com/gofiber/fiber/commit/e1a7c114b2df3ca61dffa82e9e81d985ecf11f2e)	Ratio
> `Benchmark_Ctx_Is`	`65.06` ns/op 0 B/op 0 allocs/op	`37.53` ns/op 0 B/op 0 allocs/op	`1.73`
> `Benchmark_Ctx_Is - ns/op`	`65.06` ns/op	`37.53` ns/op	`1.73`
> This comment was automatically generated by [workflow](https://github.com/gofiber/fiber/actions?query=workflow%3ABenchmark) using [github-action-benchmark](https://github.com/marketplace/actions/continuous-benchmark).

This makes sense since we are using case-insensitive checks now to be RFC compliant.

ctx.go
```diff
@@ -883,10 +883,12 @@ func (c *DefaultCtx) Is(extension string) bool {
 		return false
 	}
 
-	return strings.HasPrefix(
-		utils.TrimLeft(utils.UnsafeString(c.fasthttp.Request.Header.ContentType()), ' '),
-		extensionHeader,
-	)
+	ct := utils.UnsafeString(c.fasthttp.Request.Header.ContentType())
```
gaby: Note to myself: this needs to use getString to respect immutable flag.

gaby: Done

ctx.go
```diff
@@ -883,10 +883,12 @@ func (c *DefaultCtx) Is(extension string) bool {
 		return false
 	}
 
-	return strings.HasPrefix(
-		utils.TrimLeft(utils.UnsafeString(c.fasthttp.Request.Header.ContentType()), ' '),
-		extensionHeader,
-	)
+	ct := c.app.getString(c.fasthttp.Request.Header.ContentType())
```
nickajacks1, line 886: I think we can still use `utils.UnsafeString` here, we don't need the immutability here because we're not returning this string

gaby: I thought about but just wanted to play safe. By default the `app.getString` uses `UnsafeString`.

gaby: Source: https://github.com/gofiber/fiber/blob/main/app.go#L526

ctx.go
```diff
@@ -883,10 +883,12 @@ func (c *DefaultCtx) Is(extension string) bool {
...
+	if i := strings.IndexByte(ct, ';'); i != -1 {
+		ct = ct[:i]
+	}
+	ct = strings.TrimSpace(ct)
```
ReneWerner87: use utils.Trim
https://github.com/gofiber/utils/blob/master/byteseq.go#L30

```suggestion
	ct = utils.Trim(ct, ' ')
```

PR #3537, opened 2025-06-23
🐛 bug: Fix Content-Type comparison in Is()

gaby:
## Summary
- handle parameters and casing when comparing MIME types in `Ctx.Is`.
- update tests for case-insensitive comparison.
- add test that mismatched subtype does not match.

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -881,7 +881,7 @@ func (c *Ctx) Is(extension string) bool {
 	if i := strings.IndexByte(ct, ';'); i != -1 {
 		ct = ct[:i]
 	}
-	ct = strings.TrimSpace(ct)
+	ct = utils.Trim(ct, ' ')
 	return utils.EqualFold(ct, extensionHeader)
 }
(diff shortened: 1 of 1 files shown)


ctx.go
```diff
@@ -877,10 +877,12 @@ func (c *Ctx) Is(extension string) bool {
...
 
-	return strings.HasPrefix(
-		utils.TrimLeft(c.app.getString(c.fasthttp.Request.Header.ContentType()), ' '),
-		extensionHeader,
-	)
+	ct := c.app.getString(c.fasthttp.Request.Header.ContentType())
+	if i := strings.IndexByte(ct, ';'); i != -1 {
+		ct = ct[:i]
+	}
+	ct = strings.TrimSpace(ct)
```
ReneWerner87:
```suggestion
	ct = utils.Trim(ct, ' ')
```

ReneWerner87 (approved):

Issue #3383, opened 2025-03-30
📝 [Proposal]: Improve RFC Compliance


gaby:
### Feature Proposal Description

The main purpose of this issue is to track fixes to RFC compliance for the `GoFiber` framework. These are issues on the core and middleware code.

gaby:
We should tackle these each as a separate PR. Any that doesn't apply, we can ~strikeout~.

JIeJaitt:
@gaby Any screenshots of the OpenAI - o1 pro answer? I haven't used this model yet and would like to see it.

gaby:
@JIeJaitt It looks the same, the difference is that it takes longer to answer. For example that response above took almost 5 minutes.

nickajacks1:
I believe q value negotiation is [already implmented](https://github.com/gofiber/fiber/blob/main/helpers.go#L404). Might want to check the accuracy of the rest of the findings.

gaby:
@nickajacks1 I forgot to feed some files to the model. Re-running the prompt now with the full-code base.

gaby:
@nickajacks1 The content negotiation is not fully RFC compliant. This is what `o3-mini-high` found:

```
Accept Header Parsing and Sorting

The code that parses and sorts the “Accept” (and related) headers (functions like getOffer, forEachMediaRange, and sortAcceptedTypes) implements custom parsing and quality‑value (q‑value) sorting. Although the comments reference RFC 9110 and quality value rules, the handling of quoted strings or parameter ordering may be brittle.

Fix: Refine the parser or leverage an existing robust parser that correctly handles quoted parameters and ordering per RFC 9110 sections 8.3 and 8.4.
```

nickajacks1:
There is a way to parse media type params with the standard lib ( https://pkg.go.dev/mime#ParseMediaType ) but I didn't use it because there was a desire to minimize impact to performance.

For media type params and content negotiation, the rfc does not detail how the params should be used. I modeled the behavior based on how it is done in express.

If there are specific examples of breakage during quoted value parsing, it would be helpful to list them out.

PR #3538, opened 2025-06-23
🐛 bug: Fix Subdomains() parsing for IDNs

gaby:
## Summary
- handle punycode domains and trailing dots in `Subdomains`
- add RFC3986-compliant tests for fully-qualified and punycode hosts
- use `utils.ToLower` for host normalization

Related #3383

### DIFF
--- a/app.go
+++ b/app.go
@@ -532,7 +532,10 @@ func New(config ...Config) *App {
 	// Create Ctx pool
 	app.pool = sync.Pool{
 		New: func() any {
-			return app.newCtx()
+			if app.newCtxFunc != nil {
+				return app.newCtxFunc(app)
+			}
+			return NewDefaultCtx(app)
 		},
 	}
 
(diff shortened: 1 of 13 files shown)


ReneWerner87 (approved):

ReneWerner87:
@gaby can we reduce the allocation from 2 to 1 ?

gaby:
> @gaby can we reduce the allocation from 2 to 1 ?

I'm pretty sure the new alloc came from adding `utils.ToLower()`. Will re-test once the new utils is added to main repo. That's why I left it as draft still

gaby:
> > @gaby can we reduce the allocation from 2 to 1 ?
> 
> I'm pretty sure the new alloc came from adding `utils.ToLower()`. Will re-test once the new utils is added to main repo. That's why I left it as draft still

Using latest utils:
```
Benchmark_Ctx_Subdomains-4                                                          	 7405726	       162.8 ns/op	      64 B/op	       1 allocs/op
2025-06-23T11:49:27.7728245Z 
```

PR #3545, opened 2025-06-25
Improve cookie validation

gaby:
## Summary
- validate cookie domain with utils.IsIPv4
- expand invalid cookie cases in tests
- revert to validating cookies using `net/http.Cookie` then setting values on a `fasthttp.Cookie`

## Testing
- `make test`

------
https://chatgpt.com/codex/tasks/task_e_6859d5772c2c8333917de8044e8d64f8


gaby:
Why is the CI not running?

PR #3546, opened 2025-06-25
🧹 chore: Improve Cookie() validation

gaby:
## Summary
- expand invalid cookie cases in tests
- validate cookies using `net/http.Cookie` then setting values on a `fasthttp.Cookie`

Related #3383


gaby:
@coderabbitai full review

gaby:
Note to myself, using `hc.String()` will make performance really bad. The correct way is to:
- create http.Cookie()
- call Valid()
- use fasthttp.Cookie setters to create the Cookie.

efectn (approved):

ReneWerner87 (approved):

PR #3589, opened 2025-07-16
🐛 bug: Fix default configuration of CSRF Middleware

### LINKED ISSUE #3587: 🐛 [Bug]: Csrf.ConfigDefault use both 'KeyLookup' and 'Extractor'
### Bug Description

On v3 (and v2, but if it's fixed on v3 I'm fine) csrf.ConfigDefault{} has both fields 'KeyLookup' and 'Extractor'.

```go
var ConfigDefault = Config{
    KeyLookup:      "header:" + HeaderName,
    CookieName:     "csrf_",
    CookieSameSite: "Lax",
    IdleTimeout:    30 * time.Minute,
    KeyGenerator:   utils.UUIDv4,
    ErrorHandler:   defaultErrorHandler,
    Extractor:      FromHeader(HeaderName),
}
```

The documentation says :

> "KeyLookup will be ignored if Extractor is explicitly set."

So getting csrf.ConfigDefault, and change the "KeyLookup" to a form based parameter (for example) before calling csrf.New() with it, will do nothing, as Extractor is still set.

We have to set Extractor to nil for csrf to create the Extractor according to KeyLookup.

### How to Reproduce

1. 
```go
config := csrf.ConfigDefault
config.KeyLookup = form:"csrf"
app.Use(csrf.New(config))
...
```

### Expected Behavior

Changing "KeyLookup" should be enough and "Extractor" should not be set on csrf.ConfigDefault

### Fiber Version

v3

### Code Snippet (optional)

```go

```

### Checklist:

ismdeep:
fix: Csrf.ConfigDefault use both 'KeyLookup' and 'Extractor' #3587

Fixes #3587

### DIFF
--- a/middleware/csrf/config.go
+++ b/middleware/csrf/config.go
@@ -125,7 +125,7 @@ var ConfigDefault = Config{
 	IdleTimeout:    30 * time.Minute,
 	KeyGenerator:   utils.UUIDv4,
 	ErrorHandler:   defaultErrorHandler,
-	Extractor:      FromHeader(HeaderName),
+	Extractor:      nil,
 }
 
 // default ErrorHandler that process return error from fiber.Handler
--- a/middleware/csrf/csrf.go
+++ b/middleware/csrf/csrf.go
@@ -130,12 +130,17 @@ func New(config ...Config) fiber.Handler {
 				return cfg.ErrorHandler(c, err)
 			}
 
+			// Get Extractor
+			extractor := cfg.Extractor
+			if extractor == nil {
+				extractor = FromHeader(HeaderName)
+			}
+
 			// Extract token from client request i.e. header, query, param, form or cookie
-			extractedToken, err := cfg.Extractor(c)
+			extractedToken, err := extractor(c)
 			if err != nil {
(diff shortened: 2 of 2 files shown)


sixcolors:
My preference was towards not having `KeyLookup` and instead having only `Extractor`. However, I didn’t want to make a breaking change at the time and we have both. Now with v3 close to RC we could and remove `KeyLookup`. Thoughts?

ismdeep:
> KeyLookup

In other words, it is actually possible to implement the default Extractor based on the current default KeyLookup value, right?

sixcolors:
> > KeyLookup
> 
> In other words, it is actually possible to implement the default Extractor based on the current default KeyLookup value, right?

Yes, exactly — `KeyLookup` ultimately just maps to one of the extractor funcs anyway. Since `Extractor` is more flexible and already handles all supported cases, I’d suggest we remove the `KeyLookup` field altogether.

This would simplify the API and eliminate confusion about precedence. Users can still rely on the default header behavior, use one of the provided extractor functions, or define their own.

Now that we're nearing the v3 RC, this would be the right time to clean this up and make it consistent.

PR #3610, opened 2025-07-20
🐛 bug: Fix cookie secure flag with SameSite None

gaby:
## Summary
- ensure secure cookies when using `SameSite=None`
- add test verifying secure flag is auto-enabled for `SameSite=None`

Mozilla Docs: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie?utm_source=chatgpt.com#none

Chrome Docs: https://developers.google.com/search/blog/2020/01/get-ready-for-new-samesitenone-secure

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -425,6 +425,10 @@ func (c *DefaultCtx) Cookie(cookie *Cookie) {
 		cookie.Path = "/"
 	}
 
+	if utils.ToLower(cookie.SameSite) == CookieSameSiteNoneMode && !cookie.Secure {
+		cookie.Secure = true
+	}
+
 	if cookie.SessionOnly {
 		cookie.MaxAge = 0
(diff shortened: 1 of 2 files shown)


ctx.go
```diff
@@ -425,6 +425,10 @@ func (c *DefaultCtx) Cookie(cookie *Cookie) {
 		cookie.Path = "/"
 	}
 
+	if utils.ToLower(cookie.SameSite) == CookieSameSiteNoneMode && !cookie.Secure {
```
sixcolors, line 428: To avoid allocations, you should use utils.EqualFold instead.

sixcolors: Also related #3608

gaby: @sixcolors I'm closing this PR then.

PR #3615, opened 2025-07-21
🧹 chore: Prioritize Forwarded header

gaby:
## Summary
- parse standard `Forwarded` header
- prefer `Forwarded` over `X-Forwarded-*` in `IP`, `Host` and `Scheme`
- document new behaviour
- add tests for `Forwarded` header


gaby:
How did `Host()` get so much slower? We don't even set a header

PR #3618, opened 2025-07-25
🔥 feat: Support for disabling response headers in Limiter Middleware

gaby:
## Summary
- introduce `DisableHeaders` option for limiter middleware
- set X-RateLimit headers only when not disabled
- document how to disable the headers
- test limiter without headers


ReneWerner87:
@gaby Should we include "HeaderRetryAfter" ?

gaby:
> @gaby Should we include "HeaderRetryAfter" ?

Done

PR #3631, opened 2025-07-29
🔥 feat: Support hashed BasicAuth passwords

gaby:
## Summary
- implement SHA/Bcrypt based auth verification in config
- default to SHA-256 hashes when no prefix provided
- document hashed password requirement
- update tests for SHA-256 hashes


ReneWerner87:
cli migration https://github.com/gofiber/cli/pull/150

PR #3659, opened 2025-08-04
🐛 bug: Fix bind All() merging logic

### LINKED ISSUE #3658: 🐛 [Bug]: Missing Content-Type breaks Bind.All functionality
### Bug Description

Setup Get request endpoint.

```
app.Get("/api/hello", func (c fiber.Ctx) error {
    var payload struct {
        Message string
     }
     if err := c.Bind().All(&payload); err != nil {
        return err;
     }
     return c.JSON(payload.Message)
}
```

### How to Reproduce

Use curl to make the request.

```sh
curl -v http://server.location/api/hello?message=world
```

Response: Unprocessable Entity Error

### Expected Behavior

Response: OK with message parameter returned back.

### Fiber Version

3 beta 5

### Code Snippet (optional)

```go
package main

import "log"

func main() {
  app := fiber.New()

  // Steps to reproduce

}
```

### Checklist:

gaby:
## Summary
- add internal validation skip to Bind
- merge all sources without early validation and skip empty bodies
- avoid body binding when Content-Type header is missing
- expand All() tests for query precedence, validators, and missing Content-Type cases

Fixes #3658

### DIFF
--- a/bind.go
+++ b/bind.go
@@ -309,7 +309,7 @@ func (b *Bind) All(out any) error {
 
 	// Precedence: URL Params -> Body -> Query -> Headers -> Cookies
 	sources := []func(any) error{b.URI}
-if b.ctx.RequestCtx().Request.Header.ContentLength() != 0 && len(b.ctx.RequestCtx().Request.Header.ContentType()) > 0 {
+	if b.ctx.RequestCtx().Request.Header.ContentLength() != 0 {
 		sources = append(sources, b.Body)
 	}
 	sources = append(sources, b.Query, b.Header, b.Cookie)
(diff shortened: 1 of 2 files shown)


bind.go
```diff
@@ -304,29 +308,32 @@ func (b *Bind) All(out any) error {
...
+	sources := []func(any) error{b.URI}
 
-	tempStruct := reflect.New(outElem.Type()).Interface()
+	// Check if both Body and Content-Type are set
+	if len(b.ctx.Request().Body()) > 0 && len(b.ctx.RequestCtx().Request.Header.ContentType()) > 0 {
+		sources = append(sources, b.Body)
+	}
+	sources = append(sources, b.Query, b.Header, b.Cookie)
+	prevSkip := b.skipValidation
+	b.skipValidation = true
```
ReneWerner87, line 319: Not sure
What happens with parallel processing in the same App? Shouldn't we pass this on via arguments?

ReneWerner87: her is no problem, the bind is only per request

Theoretically, problems could arise with multiple routes connected in series or an internal redirect, but these are edge cases.

PR #3676, opened 2025-08-11
Respect immutable config for params and protocol

gaby:
## Summary
- ensure route parameters copy when Immutable config enabled
- use config-aware getter for HTTP protocol and Content-Encoding headers
- add test ensuring Params uses independent memory under Immutable

## Testing
- `make audit` *(fails: EncodeMsg passes lock by value)*
- `make generate`
- `make betteralign`
- `make modernize`
- `make format`
- `make test`

------
https://chatgpt.com/codex/tasks/task_e_689850834ea88333bd092bb5be0e03b2

### DIFF
--- a/req.go
+++ b/req.go
@@ -137,7 +137,7 @@ func (r *DefaultReq) Body() []byte {
 	request := r.Request()
 
 	// Get Content-Encoding header
-	headerEncoding = utils.ToLower(utils.UnsafeString(request.Header.ContentEncoding()))
+	headerEncoding = utils.ToLower(r.App().getString(request.Header.ContentEncoding()))
 
 	// If no encoding is provided, return the original body
 	if len(headerEncoding) == 0 {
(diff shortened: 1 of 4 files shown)


req.go
```diff
@@ -137,7 +137,7 @@ func (r *DefaultReq) Body() []byte {
 	request := r.Request()
 
 	// Get Content-Encoding header
-	headerEncoding = utils.ToLower(utils.UnsafeString(request.Header.ContentEncoding()))
+	headerEncoding = utils.ToLower(r.App().getString(request.Header.ContentEncoding()))
```
ReneWerner87: That is not necessary, as it has no effect on the return value.

ReneWerner87 (changes requested):
only this
https://github.com/gofiber/fiber/pull/3676/files#diff-c1dd96b8f46ee699fee659dae47060743ec1f62de44d7fe4144f0fc927ea8c8aR140

rest looks good

req.go
gaby: Done

PR #3677, opened 2025-08-11
📒 docs: Add context guide

gaby:
## Summary
- document Fiber's Ctx as a `context.Context`
- explain using `fasthttp.RequestCtx` and `fasthttpctx`
- cover helper functions like `requestid.FromContext`

### DIFF
--- a/docs/guide/context.md
+++ b/docs/guide/context.md
@@ -1,6 +1,6 @@
 ---
-id: context
-title: "\U0001F9E0 Context"
+id: go-context
+title: "\U0001F9E0 Go Context"
 description: >-
   Learn how Fiber's Ctx integrates with Go's context.Context,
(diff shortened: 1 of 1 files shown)


docs/guide/context.md
```diff
@@ -0,0 +1,191 @@
+---
+id: context
+title: "\U0001F9E0 Context"
```
efectn: Context might be ambiguous name as we already have `Ctx` page. I think we need to find a different name

ReneWerner87: i agree, maybe something with go context
or context.context

@coderabbitai suggest a good name

ReneWerner87: @gaby

gaby: @ReneWerner87 updated

PR #3703, opened 2025-08-22
feat(ctx): add conditional copy helpers

ReneWerner87:
## Summary
- add CopyString and CopyBytes helpers on `fiber.Ctx`
- use new helpers in cache, idempotency and proxy middleware
- document conditional copy helpers in context API and v3 "What's New" docs
- update docs and tests for conditional copying

## Testing
- `make audit` *(fails: EncodeMsg passes lock by value; package requires newer Go version go1.25)*
- `make generate`
- `make betteralign` *(fails: package requires newer Go version go1.25)*
- `make modernize` *(fails: package requires newer Go version go1.25)*
- `make format`
- `make test`

------
https://chatgpt.com/codex/tasks/task_e_68a80e3501ac83269ec38c74c210916e


docs/api/ctx.md
```diff
@@ -626,10 +626,47 @@ app.Get("/", func(c fiber.Ctx) error {
...
+enabled, the string is returned as-is; otherwise a new copy is made.
+
+```go title="Signature"
+func (c fiber.Ctx) CopyString(s string) string
+```
+
+```go title="Example"
+app.Get("/", func(c fiber.Ctx) error {
+  name := c.Cookies("name")
+  safe := c.CopyString(name)
```
gaby: This should probably be: `safe := c.CopyString(c.Cookies("name"))`

docs/api/ctx.md
```diff
@@ -626,10 +626,47 @@ app.Get("/", func(c fiber.Ctx) error {
...
+returned unchanged. Otherwise, a new copy is allocated.
+
+```go title="Signature"
+func (c fiber.Ctx) CopyBytes(b []byte) []byte
+```
+
+```go title="Example"
+app.Get("/", func(c fiber.Ctx) error {
+  body := c.Body()
+  safe := c.CopyBytes(body)
```
gaby: This should probably be: safe := c.CopyBytes(c.Body())`

ctx_test.go
```diff
@@ -3157,7 +3157,7 @@ func Test_Ctx_Params_Immutable(t *testing.T) {
 
 	c.route = &Route{Params: []string{"user"}}
 	c.path = []byte("/test/john")
-	c.values[0] = utils.UnsafeString(c.path[6:])
+	c.values[0] = c.app.getString(c.path[6:])
```
gaby: Why are we using an unsafe str in a test?

ReneWerner87: To simulate what happens when an unsafe value is stored in the values, which we do internally to keep allocations small


At the end, the Immutable switch creates a copy 
In Fiber, the internal process still uses references to avoid allocations, but the output is then a copy

gaby:
@coderabbitai should we rename the new functions?

Issue #3425, opened 2025-04-25
📝 [Proposal]: Add support for fasthttp.StreamResponseBody


gaby:
### Feature Proposal Description

FastHTTP added support for StreamResponseBody back in 2022 but it was never added to Fiber. The proposal is to make this option available in the Fiber Client and any other places it applies.

Related PR: https://github.com/valyala/fasthttp/pull/1414

### Alignment with Express API

N/a

### HTTP RFC Standards Compliance

N/a

### API Stability

N/a

### Feature Examples

```go
req := fasthttp.AcquireRequest()
resp := fasthttp.AcquireResponse()
defer fasthttp.ReleaseRequest(req)
defer fasthttp.ReleaseResponse(resp)

client := &fasthttp.Client{
    StreamResponseBody: true,
}

req.SetRequestURI("http://example.com/sse")
req.Header.SetMethod("GET")
req.Header.Set("Accept", "text/event-stream")

err := client.Do(req, resp)
if err != nil {
    log.Fatal(err)
}

reader := bufio.NewReader(resp.BodyStream())
for {
    line, err := reader.ReadString('\n')
    if err != nil {
        if err == io.EOF {
            break
        }
        log.Fatal(err)
    }
    fmt.Print("Received:", line)
}
```

### Checklist:

- [x] I agree to follow Fiber's [Code of Conduct](https://github.com/gofiber/fiber/blob/master/.github/CODE_OF_CONDUCT.md).
- [x] I have searched for existing issues that describe my proposal before opening this one.
- [x] I understand that a proposal that does not meet these guidelines may be closed without explanation.

satti-hari-krishna-reddy:
Hey @gaby  I’d love to work on this — please assign it to me

gaby:
@satti-hari-krishna-reddy Target branch is `main`. Thanks!

gaby:
@satti-hari-krishna-reddy Any updates on this?

JIeJaitt:
@gaby Hello, I'm quite interested in this issue. However, I noticed that the `Client` struct has already encapsulated the `*fasthttp.Client` struct, right? Shouldn't setting the `StreamResponseBody` of `*fasthttp.Client` to `true` directly achieve this functionality?

Abhirup-99:
@gaby is this issue still up for grabs?

gaby:
@Abhirup-99 Yes

PR #3711, opened 2025-08-25
🔥 feat: Add StreamResponseBody support for the Client

Abhirup-99:
This pull request introduces support for streaming HTTP response bodies in the client, allowing responses to be read as streams rather than being fully loaded into memory. This is particularly useful for handling large responses or server-sent events. The changes include new configuration options at both the client and request levels, implementation of the streaming logic, and comprehensive tests to ensure correct behavior.

**Streaming response body support:**

* Added a `streamResponseBody` field to the `Client` struct, along with `SetStreamResponseBody` and `StreamResponseBody` methods to enable or disable response body streaming at the client level (`client/client.go`). [[1]](diffhunk://#diff-bd3a55a72186f59e2e63efb4951573b2f9e4a7cc98086e922b0859f8ccc1dd09R61) [[2]](diffhunk://#diff-bd3a55a72186f59e2e63efb4951573b2f9e4a7cc98086e922b0859f8ccc1dd09R439-R452)
* Added a `streamResponseBody` field to the `Request` struct, with corresponding `SetStreamResponseBody` and `StreamResponseBody` methods to allow per-request configuration that overrides the client setting (`client/request.go`). [[1]](diffhunk://#diff-d1c4178bc92354ebb52bb3b7c282de3de06d31949d9fbd7a77e74e52dbcb4842R71) [[2]](diffhunk://#diff-d1c4178bc92354ebb52bb3b7c282de3de06d31949d9fbd7a77e74e52dbcb4842R594-R612)
* Updated the request execution logic to set and restore the underlying HTTP client's streaming option based on the request or client configuration (`client/core.go`).
* Implemented the `BodyStream` method on the `Response` struct to provide an `io.Reader` for streaming the response body, falling back to an in-memory reader if streaming is not enabled (`client/response.go`).

**Testing and validation:**

* Added extensive tests to verify streaming behavior, including tests for server-sent events, large responses, default and overridden settings, and chainable configuration methods (`client/client_test.go`, `client/response_test.go`). [[1]](diffhunk://#diff-57cdc3e29d0ef725a50897a4fbd1c4f76353429fa66d3a785dcbc17dcd7b6f31R1736-R1883) [[2]](diffhunk://#diff-4dd50e702b0012445cc560258e655591d37f6e6add7bacd528bab89b4b0e749fR541-R612)
* Ensured that request resetting clears the `streamResponseBody` override (`client/request.go`).# Description

Please provide a clear and concise description of the changes you've made and the problem they address. Include the purpose of the change, any relevant issues it solves, and the benefits it brings to the project. If this change introduces new features or adjustments, highlight them here.

Related #3425


Abhirup-99:
I will update the documentation if the approach seems sound enough.

gaby:
@Abhirup-99 The tests are failing. You can run them locally by running `make tests`

Abhirup-99:
@gaby fixed the failing tests

gaby:
@Abhirup-99 To fix the alignment you can run `make betteralign` it will fix it for you

JIeJaitt:
@gaby @Abhirup-99  hi, I have some questions.

I don't think the current tests cover streaming response scenarios. For example, `Test_Client_StreamResponseBody_ServerSentEvents` only reads from `bodyStream` once and asserts it contains the first message — it doesn't verify that subsequent messages arrive incrementally, nor can it detect mid-stream disconnects or cases where the server writes and returns data progressively. In `Test_Request_StreamResponseBody` the server immediately returns the full string (`c.SendString("Hello, World!")`) with no chunked or delayed writes, so it can't validate chunked/streaming delivery or on-demand reads — the client, even in streaming mode, would read the whole response at once.

client/response.go
```diff
@@ -143,14 +159,17 @@ func (r *Response) Save(v any) error {
 		return nil
 
 	case io.Writer:
-		if _, err := io.Copy(p, bytes.NewReader(r.Body())); err != nil {
-			return fmt.Errorf("failed to write response body to io.Writer: %w", err)
+		var err error
+		if r.IsStreaming() {
+			_, err = io.Copy(p, r.BodyStream())
```
gaby, line 170: Should probably use io.CopyBuffer

client/core.go
```diff
@@ -81,6 +81,43 @@ func (c *core) execFunc() (*Response, error) {
 	c.req.RawRequest.CopyTo(reqv)
 	cfg := c.getRetryConfig()
 
+	// Determine which client to use - create a new one if StreamResponseBody differs
+	var fastHTTPClient *fasthttp.Client
```
gaby: This is creating a separate client from c.client

client/response_test.go
```diff
@@ -537,4 +537,134 @@ func Test_Response_Save(t *testing.T) {
 		err = resp.Save(nil)
 		require.Error(t, err)
 	})
+
+	t.Run("streaming with io.Writer without closing", func(t *testing.T) {
+		t.Parallel()
+
+		server := startTestServer(t, func(app *fiber.App) {
+			app.Get("/stream", func(c fiber.Ctx) error {
+				return c.SendString("streaming data")
```
gaby: This is not streaming the server response. Ths fiber server needs to run with the streaming option enabled

gaby (commented):
The reset() method is missing resetting the value for StreamResponseBody

gaby:
@JIeJaitt Correct, the tests are not testing streaming, the server needs to stream the resp

gaby:
@codex review

Abhirup-99:
@gaby have updated the code.

client/client.go
```diff
@@ -435,6 +435,20 @@ func (c *Client) DisableDebug() *Client {
 	return c
 }
 
+// StreamResponseBody returns the current StreamResponseBody setting.
+func (c *Client) StreamResponseBody() bool {
```
ReneWerner87, line 530: pls add documentation in the client markdown file

ReneWerner87 (changes requested):
LG
but documentation is missing

ReneWerner87:
@Abhirup-99

JIeJaitt:
LGTM

Abhirup-99:
Increased test coverage and added documentation. @gaby @ReneWerner87

Abhirup-99:
@gaby @ReneWerner87 bump

ReneWerner87:
We will check tomorrow(sickness weekend)

client/response_test.go
ReneWerner87: @Abhirup-99

ReneWerner87:
@Abhirup-99 can you check the linting and the last hints

Abhirup-99:
@ReneWerner87 updated the code. Also, can I work on #3731 ? Checking lint is very difficult because make lint seems to fail for go 1.25.

gaby:
@Abhirup-99 How exactly is the server streaming a response back to the client during the tests?

The creation of a BodyStream by the client is masking this

client/client_test.go
```diff
@@ -1733,3 +1733,120 @@ func Benchmark_Client_Request_Parallel(b *testing.B) {
...
+			}
+
+			for _, msg := range messages {
+				if _, err := c.WriteString(msg); err != nil {
+					return err
+				}
+			}
+
+			return nil
+		})
```
grivera64: We should use `c.SendStreamWriter()` for server-sent events here since `c.WriteString()` does not flush on write.

https://docs.gofiber.io/next/api/ctx/#sendstreamwriter

Then, we can add a flush instruction to ensure that each message is flushed as soon as possible.

For example:
```go
if _, err := w.WriteString(msg); err != nil {
	return
}
if _, err := w.Flush(); err != nil {
	return
}
```

To ensure that the data is streaming, we could also add a `time.Sleep()` after each flush and timeout early. This way, we ensure that data is actually streaming and isn't sent as one large chunk.

Abhirup-99: @grivera64 it seems like streaming doesn't work properly with sendstreamwriter. 
```
	t.Run("server sent events with incremental reads", func(t *testing.T) {
		t.Parallel()

		server := startTestServer(t, func(app *fiber.App) {
			app.Get("/sse", func(c fiber.Ctx) error {
				c.Set("Content-Type", "text/event-stream")
				c.Set("Cache-Control", "no-cache")
				c.Set("Connection", "keep-alive")
				c.Set("Transfer-Encoding", "chunked")

				return c.SendStreamWriter(func(w *bufio.Writer) {
					messages := []string{
						"data: event 1\n\n",
						"data: event 2\n\n",
						"data: event 3\n\n",
						"data: event 4\n\n",
					}
					for i, msg := range messages {
						// Write each SSE message
						if _, err := w.WriteString(msg); err != nil {
							return
						}
						// Flush to ensure immediate delivery
						if err := w.Flush(); err != nil {
							return
						}
						// Add delay to ensure proper streaming behavior
						if i < len(messages)-1 {
							time.Sleep(10 * time.Millisecond)
						}
					}
				})
			})
		})
		defer server.stop()

		client := New().SetDial(server.dial()).SetStreamResponseBody(true)

		resp, err := client.Get("http://example.com/sse")
		require.NoError(t, err)
		defer resp.Close()

		bodyStream := resp.BodyStream()
		require.NotNil(t, bodyStream)

		content, err := io.ReadAll(bodyStream)
		require.NoError(t, err)
		contentStr := string(content)
		require.Contains(t, contentStr, "event 1")
		require.Contains(t, contentStr, "event 2")
		require.Contains(t, contentStr, "event 3")
		require.Contains(t, contentStr, "event 4")
		require.Contains(t, contentStr, "data: event")
		require.Contains(t, contentStr, "\n\n")
	})
```
	
this doesn't seem to work, but droping streamresponsebody and check for body makes this work.
Let me know what can be done here.

Abhirup-99:
> @Abhirup-99 How exactly is the server streaming a response back to the client during the tests?
> 
> The creation of a BodyStream by the client is masking this

Fixed this

client/client_test.go
Abhirup-99: @grivera64 can you check this out?

Abhirup-99:
@gaby can you check this out?

client/response.go
```diff
@@ -136,21 +152,30 @@ func (r *Response) Save(v any) error {
 		}
 		defer func() { _ = outFile.Close() }() //nolint:errcheck // not needed
 
-		if _, err = io.Copy(outFile, bytes.NewReader(r.Body())); err != nil {
+		if r.IsStreaming() {
+			_, err = io.CopyBuffer(outFile, r.BodyStream(), nil)
```
efectn: I think we can just use io.Copy here since go will already allocate new buffer since the buffer parameter is nil

client/response.go
```diff
@@ -136,21 +152,30 @@ func (r *Response) Save(v any) error {
...
 		}
 
 		return nil
 
 	case io.Writer:
-		if _, err := io.Copy(p, bytes.NewReader(r.Body())); err != nil {
-			return fmt.Errorf("failed to write response body to io.Writer: %w", err)
+		var err error
+		if r.IsStreaming() {
+			_, err = io.CopyBuffer(p, r.BodyStream(), nil)
```
efectn: Same here

client/client_test.go
grivera64: > [@grivera64](https://github.com/grivera64) can you check this out?

@Abhirup-99 It seems that `Client` currently uses `fasthttp.Response.CopyTo()` on `client/core.go:113` which does not copy `bodyStream`'s from `fasthttp.Response` to `client.Response.RawResponse`. This causes `resp.BodyStream()` to always return `nil` and lead to an `EOF` error whenever `StreamResponseBody` is set to true.

https://github.com/gofiber/fiber/blob/c1f7afe73ec60c3b422b799d031e189b5b32e57d/client/core.go#L85-L116

To fix this, I found a solution, though it isn't the cleanest approach (maybe @gaby or @ReneWerner87 are more familiar with these changes from PR #1986, and can help with a cleaner approach):

<details><summary>git diff</summary>

```diff
diff --git a/client/client_test.go b/client/client_test.go
index eb9662fe..edd4b1f4 100644
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -1,6 +1,7 @@
 package client
 
 import (
+	"bufio"
 	"context"
 	"crypto/tls"
 	"encoding/hex"
@@ -1758,14 +1759,20 @@ func Test_Client_StreamResponseBody_ServerSentEvents(t *testing.T) {
 				"data: message 2\n\n",
 				"data: message 3\n\n",
 			}
-
-			for _, msg := range messages {
-				if _, err := c.WriteString(msg); err != nil {
-					return err
+			return c.SendStreamWriter(func(w *bufio.Writer) {
+				for i, msg := range messages {
+					if _, err := w.WriteString(msg); err != nil {
+						return
+					}
+					if err := w.Flush(); err != nil {
+						return
+					}
+					// Add delay to ensure proper streaming behavior
+					if i < len(messages)-1 {
+						time.Sleep(10 * time.Millisecond)
+					}
 				}
-			}
-
-			return nil
+			})
 		})
 	})
 	defer func() { require.NoError(t, app.Shutdown()) }()
diff --git a/client/core.go b/client/core.go
index 38db93bf..593b1185 100644
--- a/client/core.go
+++ b/client/core.go
@@ -84,10 +84,7 @@ func (c *core) execFunc() (*Response, error) {
 	var err error
 	go func() {
 		respv := fasthttp.AcquireResponse()
-		defer func() {
-			fasthttp.ReleaseRequest(reqv)
-			fasthttp.ReleaseResponse(respv)
-		}()
+		defer fasthttp.ReleaseRequest(reqv)
 
 		if cfg != nil {
 			// Use an exponential backoff retry strategy.
@@ -108,9 +105,11 @@ func (c *core) execFunc() (*Response, error) {
 		if atomic.CompareAndSwapInt32(&done, 0, 1) {
 			if err != nil {
 				errCh <- err
+				fasthttp.ReleaseResponse(respv)
 				return
 			}
-			respv.CopyTo(resp.RawResponse)
+			fasthttp.ReleaseResponse(resp.RawResponse)
+			resp.RawResponse = respv
 			errCh <- nil
 		}
 	}()

```

</details>

Essentially, this directly uses the `fasthttp.Response` as our `resp.RawResponse` rather than copying the data, as we will lose the body stream if we attempt to copy the response. If a failure happened, do not use the `fasthttp.Response` and simply release it. If it succeeded, release the current `resp.RawResponse` and replace it with the successful `fasthttp.Response`.

On a side note: While fiber uses `c.fasthttp.DoXXX()` internally, `c.fasthttp.Get(...)` seems to not be compatible with `StreamResponseBody: true` either from my testing. I am not sure if this is intended or a bug, but it may be a good idea to write an issue about this in the upstream `valyala/fasthttp` repo for in case it's not an intended feature.

grivera64: After some more testing, the git diff I had recommended above seems to cause some race conditions when trying it out with #3799., let's not go with that solution.

We might have to solve this in the upstream `fasthttp` repository, since there doesn't seem to be any interrupt logic in `fashttp.Client.Do()` that we can use currently. Right now, any interruption leads to discarding the full response.

We can also try to do something similar to how [`app.Test()`](https://github.com/gofiber/fiber/blob/529b36dae4dff02619dc179446b2634b9035c7a6/app.go#L1096) deals with this, but that may require re-writing some client logic from scratch.

Whichever way we try, we would need to have a way to access the raw client `net.Conn` to be able to close it after the timeout duration and read any partial response that the connection read before it closed

gaby:
@codex review

grivera64:
PR #3830 got merged, and seems to start some work on refactoring the core execution logic.

@Abhirup-99 Can you merge the most recent changes to this PR and retry the Stream Writer change I requested some time ago?

I still think the CopyToResponse method from fasthttp still doesn't copy the stream body in this case, but I think we can start experimenting with how to add this via fasthttp or Fiber if we merge the most recent updates to this PR.

Abhirup-99:
@gaby @griver please review now.

client/response_test.go
```diff
@@ -537,4 +542,644 @@ func Test_Response_Save(t *testing.T) {
...
+		server := startTestServer(t, func(app *fiber.App) {
+			app.Get("/stream", func(c fiber.Ctx) error {
+				// Use chunked encoding to force streaming
+				c.Set("Transfer-Encoding", "chunked")
+				data := make([]byte, 1024*8) // 8KB
+				for i := range data {
+					data[i] = byte('S')
+				}
+				return c.Send(data)
+			})
```
grivera64, line 1054: We should probably use `SendStreamWriter` (or use `SendStream`) here too. The entire response is sent as a single body when `c.Send()` is called.

client/response_test.go
```diff
@@ -537,4 +542,644 @@ func Test_Response_Save(t *testing.T) {
...
+						return err
+					}
+					c.Response().ImmediateHeaderFlush = true
+
+					if i < numChunks-1 {
+						time.Sleep(time.Millisecond)
+					}
+				}
+				return nil
+			})
```
grivera64, line 1004: The `ImmediateHeaderFlush` can be set to true once before the for loop.

For this to take effect, I believe you need to call `c.SendStreamWriter()` (which calls 'SetBodyStreamWriter()` under the hood).

client/response_test.go
```diff
@@ -537,4 +542,644 @@ func Test_Response_Save(t *testing.T) {
...
+				}
+				c.Response().ImmediateHeaderFlush = true
+
+				time.Sleep(10 * time.Millisecond)
+				if _, err := c.WriteString(" more data"); err != nil {
+					return err
+				}
+
+				return nil
+			})
```
grivera64, line 936: Same `c.SendStreamWriter()` comment as for lines 923-945.

client/response_test.go
```diff
@@ -537,4 +542,644 @@ func Test_Response_Save(t *testing.T) {
...
+		t.Parallel()
+
+		server := startTestServer(t, func(app *fiber.App) {
+			app.Get("/stream", func(c fiber.Ctx) error {
+				c.Set("Transfer-Encoding", "chunked")
+				data := make([]byte, 1024*8) // 8KB
+				for i := range data {
+					data[i] = byte('S')
+				}
+				return c.Send(data)
```
grivera64: Use `c.SendStreamWriter()` here

client/response_test.go
```diff
@@ -537,4 +542,644 @@ func Test_Response_Save(t *testing.T) {
...
+		require.False(t, closableBuffer.closed, "Save() should not close the writer")
+	})
+
+	t.Run("streaming with io.Writer error during copy", func(t *testing.T) {
+		t.Parallel()
+
+		server := startTestServer(t, func(app *fiber.App) {
+			app.Get("/stream", func(c fiber.Ctx) error {
+				return c.SendString("streaming data that will fail to write")
+			})
```
grivera64, line 602: This handler doesn't seem to stream data, we should use `c.SendStream
()` or `c.SendStreamWriter()` here to ensure that the stream logic actually runs.

client/response_test.go
```diff
@@ -537,4 +542,644 @@ func Test_Response_Save(t *testing.T) {
...
+func Test_Response_BodyStream(t *testing.T) {
+	t.Parallel()
+
+	t.Run("basic streaming", func(t *testing.T) {
+		t.Parallel()
+
+		server := startTestServer(t, func(app *fiber.App) {
+			app.Get("/stream", func(c fiber.Ctx) error {
+				return c.SendString("streaming data")
+			})
```
grivera64, line 665: Same as the previous comment, 
data isn't streaming when using `c.SendString()`. Please use `c.SendStream()` or `c.SendStreamWriter()`.

client/response_test.go
```diff
@@ -537,4 +542,644 @@ func Test_Response_Save(t *testing.T) {
...
+		t.Parallel()
+
+		server := startTestServer(t, func(app *fiber.App) {
+			app.Get("/large", func(c fiber.Ctx) error {
+				data := make([]byte, 1024)
+				for i := range data {
+					data[i] = byte('A' + i%26)
+				}
+				return c.Send(data)
+			})
```
grivera64, line 691: Use `c.SendStream()` or `c.SendStreamWriter()`.

client/response_test.go
```diff
@@ -537,4 +542,644 @@ func Test_Response_Save(t *testing.T) {
...
+		}
+	})
+
+	t.Run("compare with regular body", func(t *testing.T) {
+		t.Parallel()
+
+		server := startTestServer(t, func(app *fiber.App) {
+			app.Get("/stream", func(c fiber.Ctx) error {
+				return c.SendString("streaming data")
+			})
```
grivera64, line 735: Use `c.SendStream()` or `c.SendStreamWriter()`.

client/response_test.go
```diff
@@ -537,4 +542,644 @@ func Test_Response_Save(t *testing.T) {
...
+					if _, err := c.WriteString(chunk); err != nil {
+						return err
+					}
+					c.Response().ImmediateHeaderFlush = true
+					if i < len(chunks)-1 {
+						time.Sleep(10 * time.Millisecond) // Shorter delay for faster tests
+					}
+				}
+				return nil
+			})
```
grivera64, line 773: Use `c.SendStream()` or `c.SendStreamWriter()`.

client/response_test.go
```diff
@@ -537,4 +542,644 @@ func Test_Response_Save(t *testing.T) {
...
+					if _, err := c.WriteString(msg); err != nil {
+						return err
+					}
+					c.Response().ImmediateHeaderFlush = true
+					if i < len(messages)-1 {
+						time.Sleep(5 * time.Millisecond)
+					}
+				}
+				return nil
+			})
```
grivera64, line 835: Use `c.SendStream()` or `c.SendStreamWriter()`.

client/response_test.go
```diff
@@ -537,4 +542,644 @@ func Test_Response_Save(t *testing.T) {
...
+						return err
+					}
+					c.Response().ImmediateHeaderFlush = true
+
+					if i < len(jsonParts)-1 {
+						time.Sleep(2 * time.Millisecond)
+					}
+				}
+				return nil
+			})
```
grivera64, line 887: Use `c.SendStream()` or `c.SendStreamWriter()`.

client/response_test.go
```diff
@@ -537,4 +542,644 @@ func Test_Response_Save(t *testing.T) {
...
+		require.Equal(t, "regular response body", string(bodyBytes))
+	})
+
+	t.Run("empty response body stream", func(t *testing.T) {
+		t.Parallel()
+
+		server := startTestServer(t, func(app *fiber.App) {
+			app.Get("/empty", func(c fiber.Ctx) error {
+				return c.SendString("")
+			})
```
grivera64, line 1115: Use `c.SendStream()` or `c.SendStreamWriter()`.

client/response_test.go
```diff
@@ -537,4 +542,644 @@ func Test_Response_Save(t *testing.T) {
...
+	t.Run("large non-streaming response", func(t *testing.T) {
+		t.Parallel()
+
+		server := startTestServer(t, func(app *fiber.App) {
+			app.Get("/large", func(c fiber.Ctx) error {
+				data := make([]byte, 50*1024) // 50KB
+				for i := range data {
+					data[i] = byte('X')
+				}
+				return c.Send(data)
```
grivera64, line 1140: Use `c.SendStream()` or `c.SendStreamWriter()`.

client/response_test.go
```diff
@@ -537,4 +542,644 @@ func Test_Response_Save(t *testing.T) {
...
+		t.Parallel()
+
+		server := startTestServer(t, func(app *fiber.App) {
+			app.Get("/large-stream", func(c fiber.Ctx) error {
+				data := make([]byte, 64*1024) // 64KB
+				for i := range data {
+					data[i] = byte('S')
+				}
+				return c.Send(data)
+			})
```
grivera64, line 1184: Use `c.SendStream()` or `c.SendStreamWriter()`.

Abhirup-99:
@grivera64 done

Abhirup-99:
@gaby can you check this out?

grivera64:
@Abhirup-99 Can you take a look at the coverage report from the [codecov](https://app.codecov.io/gh/gofiber/fiber/pull/3711) workflow? I think a few of the low coverage lines can be covered with a few simple unit tests.

Abhirup-99:
@grivera64 can you check now?

Abhirup-99:
fixed the lint errors

Abhirup-99:
@grivera64 bump

gaby:
@codex review

gaby:
@Abhirup-99 Seems, tests are still failing

Abhirup-99:
@gaby fixed the failing test

Abhirup-99:
@gaby bump

Abhirup-99:
added tests based on github copilot, don't have access to a windows machine. Also, I didn't make any changes in the middlewares which seem to be failing in the benchmark

ReneWerner87:
Ok thx, will check your pr tomorrow again

Abhirup-99:
the CI failures don't seem related to my changes

client/core.go
```diff
@@ -115,7 +122,12 @@ func (c *core) execFunc() (*Response, error) {
 		resp := AcquireResponse()
 		resp.setClient(c.client)
 		resp.setRequest(c.req)
-		respv.CopyTo(resp.RawResponse)
+		originalRaw := resp.RawResponse
+		resp.RawResponse = respv
+		respv = nil
+		if originalRaw != nil {
+			fasthttp.ReleaseResponse(originalRaw)
+		}
```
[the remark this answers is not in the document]
grivera64: Probably a good idea to write a comment like:

```go
// Copy raw response without consuming response body stream
```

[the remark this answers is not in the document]
Abhirup-99: done

Abhirup-99:
@grivera64 @gaby bump

grivera64:
@gaby The `CopyTo` response method from fasthttp doesn't currently copy the response body stream.

The most recently added commit in this PR reverted the manual copy to this old `CopyTo()` method. I think this is the cause for the new test case errors.

We could either:
- Make an upstream issue to support the response body streams in `CopyTo()`
- Make an upstream issue for a new `CopyToWithStream()`
- Copy the body stream manually using the previous commit's method

ReneWerner87:
test failed
```txt
=== Failed
=== FAIL: client Test_Response_Save_Streaming_FilePath (0.10s)
    response_test.go:1292: 
        	Error Trace:	/home/runner/work/fiber/fiber/client/response_test.go:1292
        	Error:      	Not equal: 
        	            	expected: "streaming file content"
        	            	actual  : ""
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-streaming file content
        	            	+
        	Test:       	Test_Response_Save_Streaming_FilePath

=== FAIL: client Test_Response_Save/streaming_with_io.Writer_error_during_copy (0.10s)
    response_test.go:616: 
        	Error Trace:	/home/runner/work/fiber/fiber/client/response_test.go:616
        	Error:      	An error is expected but got nil.
        	Test:       	Test_Response_Save/streaming_with_io.Writer_error_during_copy

=== FAIL: client Test_Response_Save/streaming_with_io.Writer_without_closing (10.00s)
    response_test.go:581: 
        	Error Trace:	/home/runner/work/fiber/fiber/client/response_test.go:581
        	Error:      	"" does not contain "SSSSSS"
        	Test:       	Test_Response_Save/streaming_with_io.Writer_without_closing

=== FAIL: client Test_Response_Save (0.00s)

=== FAIL: client Test_Response_IsStreaming/streaming_with_large_response (10.10s)
    response_test.go:1195: 
        	Error Trace:	/home/runner/work/fiber/fiber/client/response_test.go:1195
        	Error:      	"[]" should have 65536 item(s), but has 0
        	Test:       	Test_Response_IsStreaming/streaming_with_large_response
        	Error Trace:	/home/runner/work/fiber/fiber/client/response_test.go:1078
        	Error:      	"[]" should have 8192 item(s), but has 0
        	Test:       	Test_Response_BodyStream/stream_object_identity_when_streaming_enabled

=== FAIL: client Test_Response_BodyStream/basic_streaming (0.10s)
    response_test.go:678: 
        	Error Trace:	/home/runner/work/fiber/fiber/client/response_test.go:678
        	Error:      	Not equal: 
        	            	expected: "streaming data"
        	            	actual  : ""
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-streaming data
        	            	+
        	Test:       	Test_Response_BodyStream/basic_streaming

=== FAIL: client Test_Response_BodyStream/chunked_streaming_with_delays (0.10s)
    response_test.go:802: 
        	Error Trace:	/home/runner/work/fiber/fiber/client/response_test.go:802
exit status 1
        	Error:      	Not equal: 
        	            	expected: "chunk1chunk2chunk3"
        	            	actual  : ""
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-chunk1chunk2chunk3
        	            	+
        	Test:       	Test_Response_BodyStream/chunked_streaming_with_delays

=== FAIL: client Test_Response_BodyStream/large_response_streaming_validation (10.10s)
    response_test.go:1034: 
        	Error Trace:	/home/runner/work/fiber/fiber/client/response_test.go:1034
        	Error:      	"[]" should have 10240 item(s), but has 0
        	Test:       	Test_Response_BodyStream/large_response_streaming_validation

=== FAIL: client Test_Response_BodyStream/server_sent_events_with_incremental_reads (10.10s)
    response_test.go:852: 
        	Error Trace:	/home/runner/work/fiber/fiber/client/response_test.go:852
        	Error:      	"" does not contain "event 1"
        	Test:       	Test_Response_BodyStream/server_sent_events_with_incremental_reads
```

grivera64:
@ReneWerner87 I believe this is due to using `(*fasthttp.Response).CopyTo`:

https://github.com/valyala/fasthttp/blob/7b5cb77b95e2200cab14572519bd7dfdcc55fdeb/http.go#L909-L923

This method does not copy body streams to the `Ctx`'s response, and therefore causes all client streaming to break. I propose we apply the following diff:

```diff
diff --git a/client/core.go b/client/core.go
index 38333099..09ed2a5c 100644
--- a/client/core.go
+++ b/client/core.go
@@ -122,16 +122,16 @@ func (c *core) execFunc() (*Response, error) {
 		resp := AcquireResponse()
 		resp.setClient(c.client)
 		resp.setRequest(c.req)
-		// Copy the fasthttp response into the Fiber response's RawResponse field
-		// instead of taking ownership of the pooled fasthttp.Response (respv). This
-		// allows respv to be safely released back to the fasthttp pool by the
-		// deferred cleanup above, avoiding resource leaks.
-		targetRaw := resp.RawResponse
-		if targetRaw == nil {
-			targetRaw = fasthttp.AcquireResponse()
-		}
-		respv.CopyTo(targetRaw)
-		resp.RawResponse = targetRaw
+
+		// Swap the fasthttp response with the Fiber response's RawResponse field.
+		// This is required, as (*fasthttp.Response).CopyTo() explicitly does not
+		// copy body streams.
+		//
+		// See: https://github.com/valyala/fasthttp/blob/master/http.go#L909-L923
+		//
+		// The defer statement above ensures that the original RawResponse
+		// (now stored in respv) will be properly released.
+		resp.RawResponse, respv = respv, resp.RawResponse
 		respChan <- resp
 	}()
 ```

Should I push a commit with this change to the PR?

**Comment by gaby** [comment]
@grivera64 Yes, thanks! 💪

**Comment by grivera64** [comment]
> @grivera64 Yes, thanks! 💪

@gaby I am having a permissions issue pushing my local commit to the branch (I think because contributors aren't able to push to another contributor's PRs).

I made a commit under my fork: https://github.com/grivera64/fiber.git in the branch `fix/Abhirup-99-add-response`. It should be possible to pull into this branch like this:

```console
git pull https://github.com/grivera64/fiber.git fix/Abhirup-99-add-response
```

If that doesn't work, I think it is also fine to just apply this diff directly:

```diff
diff --git a/client/core.go b/client/core.go
index 38333099..050f76e6 100644
--- a/client/core.go
+++ b/client/core.go
@@ -122,16 +122,16 @@ func (c *core) execFunc() (*Response, error) {
 		resp := AcquireResponse()
 		resp.setClient(c.client)
 		resp.setRequest(c.req)
-		// Copy the fasthttp response into the Fiber response's RawResponse field
-		// instead of taking ownership of the pooled fasthttp.Response (respv). This
-		// allows respv to be safely released back to the fasthttp pool by the
-		// deferred cleanup above, avoiding resource leaks.
-		targetRaw := resp.RawResponse
-		if targetRaw == nil {
-			targetRaw = fasthttp.AcquireResponse()
-		}
-		respv.CopyTo(targetRaw)
-		resp.RawResponse = targetRaw
+
+		// Swap the fasthttp response with the Fiber response's RawResponse field.
+		// This is required, as (*fasthttp.Response).CopyTo() explicitly does not
+		// copy body streams.
+		//
+		// See: https://github.com/valyala/fasthttp/blob/v1.69.0/http.go#L909-L923
+		//
+		// The defer statement above ensures that the original RawResponse
+		// (now stored in respv) will be properly released.
+		resp.RawResponse, respv = respv, resp.RawResponse
 		respChan <- resp
 	}()
 

```

After merging, I believe all test cases should pass.

gaby:
@grivera64 Will check tomorrow, it's a repo repo so it only allows maintainers to push to it.

grivera64:
> @grivera64 Will check tomorrow, it's a repo repo so it only allows maintainers to push to it.

Sounds good, thanks! 💪

ReneWerner87:
@Abhirup-99 can you overtake the change from the diff
https://github.com/gofiber/fiber/pull/3711#issuecomment-3736596001

ReneWerner87:
I created a copy of the PR with the patch https://github.com/gofiber/fiber/pull/4014

PR #3727, opened 2025-09-02
feat(ctx): add request inspection helpers

gaby:
## Summary
- add `HasBody`, `IsWebSocket`, `IsPreflight`, and `IsFound` helpers on `Ctx`
- document new helpers in the Ctx API guide
- exercise helpers with unit tests
- release acquired contexts in helper tests

## Testing
- `make audit` *(fails: EncodeMsg passes lock by value in csrf and session middleware)*
- `make generate`
- `make betteralign`
- `make modernize`
- `make format`
- `make test`

------
https://chatgpt.com/codex/tasks/task_e_68b40371a96083339454dd9bd04805f4

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -5,6 +5,7 @@
 package fiber
 
 import (
+	"bytes"
 	"context"
 	"crypto/tls"
 	"fmt"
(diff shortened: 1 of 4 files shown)


rugleb:
Nice PR for me.

Suggestions from me:
- Rename `IsFound` -> `Matched`, because "found" means also "resource" (by id) not found.
- Drop `IsNotFound`, because don't really need.

gaby:
> Nice PR for me.
> 
> Suggestions from me:
> 
> * Rename `IsFound` -> `Matched`, because "found" means also "resource" (by id) not found.
> * Drop `IsNotFound`, because don't really need.

Makes sense, will update!

ctx.go
```diff
@@ -300,6 +301,59 @@ func (c *DefaultCtx) Route() *Route {
 	return c.route
 }
 
+// Matched returns true if the current request path was matched by the router.
+func (c *DefaultCtx) Matched() bool {
```
ReneWerner87, line 305: can we replace https://github.com/gofiber/fiber/blob/2555a4f1b5f579cce0ec054056c5eb0b61a2e3f9/ctx.go#L546 with this

ReneWerner87: or we just expose the GetMatched

gaby: Done

PR #3748, opened 2025-09-17
🧹 chore: Handle Transfer-Encoding bodies in HasBody

gaby:
## Summary
- treat `Transfer-Encoding` requests as having bodies when `Content-Length` is zero by scanning the header tokens inside `HasBody`
- extend `Test_Ctx_HasBody` with a `DisableSpecialHeader` scenario so chunked transfer-encoded requests are exercised alongside the fixed-length checks


ReneWerner87 (approved):

ReneWerner87:
linting again^^

PR #3767, opened 2025-09-28
⚡ perf: Improve usage of sync.Pool

gaby:
## Summary
- Add support for using sync.Pool across code base
- Validate release of objects put in the pools


gaby:
Will replace with new PR's

PR #3769, opened 2025-09-28
test: cover adapter conversions and document net/http support

gaby:
## Summary
- rename the internal adapter helpers to adapter.go and add targeted tests for their conversions
- exercise mixed Fiber and net/http handlers throughout the router, including groups and RouteChain
- document direct net/http handler registration across the routing guide, adaptor docs, and release notes

## Testing
- go test ./...

------
https://chatgpt.com/codex/tasks/task_e_68d93a689e808333958822985d7e74f9

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -198,6 +198,26 @@ func Test_App_MethodNotAllowed(t *testing.T) {
 	require.Equal(t, "GET, HEAD, POST, OPTIONS", resp.Header.Get(HeaderAllow))
 }
 
+func Test_App_RegisterNetHTTPHandler(t *testing.T) {
+	t.Parallel()
+
+	app := New()
+
+	app.Get("/foo", func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("X-Test", "ok")
+		_, _ = w.Write([]byte("hello from net/http"))
+	})
+
+	resp, err := app.Test(httptest.NewRequest(MethodGet, "/foo", nil))
+	require.NoError(t, err)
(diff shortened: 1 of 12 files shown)


gaby:
Fixing lint and review comments in the next hour

gaby:
@codex review

app_test.go
```diff
@@ -198,6 +198,27 @@ func Test_App_MethodNotAllowed(t *testing.T) {
...
+	resp, err := app.Test(httptest.NewRequest(MethodGet, "/foo", nil))
+	require.NoError(t, err)
+	require.Equal(t, http.StatusOK, resp.StatusCode)
+
+	body, err := io.ReadAll(resp.Body)
+	require.NoError(t, err)
+	require.Equal(t, "hello from net/http", string(body))
+	require.Equal(t, "ok", resp.Header.Get("X-Test"))
+}
+
```
grivera64, line 323: For completeness, should we add checks for all route methods? (e.g. Get, Head, Post, Put, Delete, etc.) Theoretically the same underlying functions are called no matter which route method is called, so only one is necessary, but it may be a good idea to still check.

If we do check all cases, We could just group these all into a single unit test to avoid cluttering test logs.

grivera64 (approved):
Aside from a test-related comment, LGTM! 🚀

app_test.go
gaby: Good idea

gaby: Done

docs/whats_new.md
ReneWerner87: dont forget to change this signatures

PR #3801, opened 2025-10-12
🧹 chore: Replace go-colorable/isatty dependencies with internal package

gaby:
## Summary
- add an internal console package to handle colorable writers, ANSI stripping, and terminal detection on each platform
- switch the logger and listen startup messaging to use the new console helpers instead of external dependencies
- cover the new helpers with POSIX and Windows unit tests


gaby:
Testing this idea with the full CI to see if it makes sense.

gaby:
Closing, will re-open. Somehow my branches got mixed

PR #3809, opened 2025-10-16
🔥 feat: Add support for Express.js style req/res handlers

### LINKED ISSUE #3806: 📝 [Proposal]: Add Express-style handler support

Since we now have support for non-Fiber net/http handlers for `app.Get()` and other methods in #3769, we should add an express-style handler as an option as well.

Since we have added `fiber.Req` and `fiber.Res` interfaces in #3533, we can use these as the expected `req` and `res` parameters Express uses in their handlers.

Express uses the following handler styles:

```js
app.get("/", (req, res) => {
    // Do something with req and res
});
```

```js
app.get("/", (req, res, next) => {
    // Do something with req and res
    next();
});
```

If we support these in Fiber, we can get closer to the Express API while still allowing the older `fiber.Ctx` style handlers.

N/A

### API Stability

Express' API doesn't change these handler signatures, and supporting these handler should not make any changes to the current router logic, aside from adding a new adaptor and type case.

### Feature Examples

```go
package main

import (
	"log"

	"github.com/gofiber/fiber/v3"
)

func main() {
	app := fiber.New()

	app.Use(func(req fiber.Req, res fiber.Res, next func() error) {
		if req.IP() == "192.168.1.254" {
			return res.Drop()
		}
		return next()
	})

	app.Get("/", func(req fiber.Req, res fiber.Res) error {
		return res.SendString("Hello World!")
	})

}
```

### Checklist:

gaby:
## Summary
- allow the router adapter to accept express-style functions that use the fiber.Req and fiber.Res interfaces, including optional next callbacks
- cover the new handler signatures with dedicated tests
- document express-style handlers across the routing guide, whats_new, and README

Fixes #3806

### DIFF
--- a/adapter_test.go
+++ b/adapter_test.go
@@ -211,7 +211,7 @@ func TestAdapter_MixedHandlerIntegration(t *testing.T) {
 		return c.Next()
 	})
 
-	app.Use(func(req Req, res Res, next func() error) error {
+	app.Use(func(_ Req, res Res, next func() error) error {
 		res.Set("X-Express", "middleware")
 		return next()
 	})
@@ -221,15 +221,16 @@ func TestAdapter_MixedHandlerIntegration(t *testing.T) {
 		return c.SendString("fiber handler")
 	})
 
-	app.Post("/express", func(req Req, res Res) error {
+	app.Post("/express", func(_ Req, res Res) error {
 		res.Set("X-Route", "express")
 		return res.SendString("express handler")
 	})
 
-	app.Put("/http", func(w http.ResponseWriter, r *http.Request) {
+	var httpHandlerWriteErr error
+	app.Put("/http", func(w http.ResponseWriter, _ *http.Request) {
 		w.Header().Set("X-Route", "http")
 		w.WriteHeader(http.StatusAccepted)
-		_, _ = w.Write([]byte("http handler"))
+		_, httpHandlerWriteErr = w.Write([]byte("http handler"))
 	})
 
 	app.Delete("/fasthttp", func(ctx *fasthttp.RequestCtx) error {
@@ -246,7 +247,7 @@ func TestAdapter_MixedHandlerIntegration(t *testing.T) {
 			resp, err := app.Test(req)
 			require.NoError(t, err)
 			t.Cleanup(func() {
-				resp.Body.Close()
+				require.NoError(t, resp.Body.Close())
 			})
 
 			body, err := io.ReadAll(resp.Body)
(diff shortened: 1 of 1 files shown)


gaby:
@grivera64 Yeah the goal is to move some of the logic of adaptor to adapter, but wanted to get the structure first before making that change.

The adaptor stills offers more support since it expects the fields.

Will fix docs tonight.

gaby:
@grivera64 Updated the docs and clarify some times based on CodeRabbit suggestions.

adapter.go
```diff
@@ -21,6 +21,42 @@ func toFiberHandler(handler any) (Handler, bool) {
...
+	case func(Req, Res, func() error) error:
+		if h == nil {
+			return nil, false
+		}
+		return func(c Ctx) error {
+			return h(c.Req(), c.Res(), func() error {
+				return c.Next()
+			})
+		}, true
+	case func(Req, Res, func() error):
```
gaby: @grivera64 We probably also need a case for `func(Req, Res, func())`, right?

grivera64: Yes, it would be a good idea to cover that case too. In ExpressJS, `next()` doesn't return an error. Since `c.Next()` does return an error, I think it is a good idea to cover both cases of returning an error and not returning an error.

There is also another edge case where in ExpressJS, you can pass an error into `next()`. I'm not sure if supporting that handler too wouldn't make sense with how Fiber's error handling works, but it seems to mimic setting a custom error handler in `fiber.Config{}`.

```js
app.Get("/", (err, req, res, next) => {
    if (res.headersSent) {
        return next(err)
    }
    res.status(500)
    res.render('error', { error: err })
});
```

(Example from: https://expressjs.com/en/guide/error-handling.html)

ReneWerner87 (approved):

adapter.go
gaby: Just added two more cases. Our Next() doesn't have params, so that wouldn't work?

grivera64: True, I don't think we can seamlessly use an adapter for that holder type without making major changes to `c.Next()` (or adding a new `c.NextWithError(err)`).

To my understanding, `next(err)` in ExpressJS skips all handlers until it sees the next matched error-handler. Since Fiber uses a universal error handler that is called after all matched handlers return, this currently isn't possible.

renanbastos93 (approved):

gaby:
@codex review

adapter.go
gaby: @grivera64 With the latest changes we will support 19 different handler types

efectn (approved):

PR #3829, opened 2025-10-31
🐛 bug: fix copying of key/values in internal/memory

sixcolors:
## Description

This PR extends the defensive copying fix from #3828 to the  storage implementation and removes now-redundant defensive copying from middleware that rely on these storage layers.

## Background

PR #3828 fixed memory corruption in  (the  interface implementation) by adding defensive copying for string keys and  values. However, the same issue exists in , which is used by cache, limiter, and CSRF middleware.

### 1. `internal/memory/memory.go`
Applies the same defensive copying pattern as #3828:
- Copy string keys in `Set()` using `utils.CopyString()`
- Copy `[]byte` values in `Set()` to prevent input corruption
- Copy `[]byte` values in `Get()` to prevent output mutation
- Added comprehensive package documentation explaining safety guarantees
- Other types (structs, ints) retain zero-copy performance

### 2. `internal/storage/memory/memory.go`
- Added defensive copying in `Get()` method (suggested by AI reviewers in #3828)
- Returns `utils.CopyBytes(v.data)` instead of direct reference
- Prevents callers from mutating stored data

### 3. Middleware Cleanup
Now that storage layers handle defensive copying, removed redundant operations:

**CSRF** (`middleware/csrf/storage_manager.go`):
- Removed `utils.CopyString(key)` and `utils.CopyBytes(raw)` before `m.memory.Set()`
- Storage layer now handles this automatically

**Idempotency** (`middleware/idempotency/idempotency.go`):
- Removed `utils.CopyBytes(c.Response().Body())` before marshaling
- Marshaling already creates new bytes, so pre-copy was redundant

**Cache** (`middleware/cache/cache.go`):
- Kept defensive copying for nested `[]byte` fields in structs
- Storage only handles top-level `[]byte` values, not fields inside structs

### The Problem
Fiber uses `sync.Pool` for buffer reuse. When strings or `[]byte` slices backed by pooled buffers are stored directly in maps without copying:
1. The map holds references to pooled memory
2. When buffers return to the pool and get reused
3. Map keys/values become corrupted
4. Results in intermittent data loss and failures

### The Solution  
Defensive copying at the storage layer ensures:
- All stored data is independent of pooled buffers
- Middleware doesn't need to worry about pooling safety
- Centralized fix prevents future issues

## Performance Impact

- Minimal: Only affects string keys and `[]byte` values
- Structs, integers, and other types remain zero-copy
- Trade-off: Small allocation overhead for correctness and safety

## Related

- Builds on #3828 which fixed the same issue in `internal/storage/memory`
- Completes the defensive copying strategy across all internal storage implementations

## Breaking Changes

None - this is a bug fix that makes storage implementations behave correctly and simplifies middleware code.

### DIFF
--- a/internal/memory/memory.go
+++ b/internal/memory/memory.go
@@ -57,12 +57,12 @@ func (s *Storage) Get(key string) any {
 	if !ok || v.e != 0 && v.e <= utils.Timestamp() {
 		return nil
 	}
-	
+
 	// Defensive copy for byte slices to prevent external mutation
 	if b, ok := v.v.([]byte); ok {
 		return utils.CopyBytes(b)
 	}
-	
+
 	return v.v
 }
 
@@ -77,15 +77,15 @@ func (s *Storage) Set(key string, val any, ttl time.Duration) {
 	if ttl > 0 {
 		exp = uint32(ttl.Seconds()) + utils.Timestamp()
 	}
-	
+
 	// Defensive copies to prevent unsafe reuse from sync.Pool
 	keyCopy := utils.CopyString(key)
-	
+
 	// Copy byte slices to prevent external mutation
 	if b, ok := val.([]byte); ok {
 		val = utils.CopyBytes(b)
 	}
-	
+
 	i := item{e: exp, v: val}
 	s.Lock()
 	s.data[keyCopy] = i
--- a/internal/storage/memory/memory.go
+++ b/internal/storage/memory/memory.go
@@ -90,7 +90,7 @@ func (s *Storage) Set(key string, val []byte, exp time.Duration) error {
 	keyCopy := utils.CopyString(key)
 	valCopy := utils.CopyBytes(val)
 
-	e := entry{data: valCopy, expiry: expire}
+	e := Entry{data: valCopy, expiry: expire}
 	s.mux.Lock()
 	s.db[keyCopy] = e
 	s.mux.Unlock()


gaby:
@codex review

internal/memory/memory.go
```diff
@@ -33,26 +47,48 @@ func New() *Storage {
...
+// []byte values are also copied to prevent external mutation of stored data.
+// Other types are stored as-is (structs are copied by value automatically).
 func (s *Storage) Set(key string, val any, ttl time.Duration) {
 	var exp uint32
 	if ttl > 0 {
 		exp = uint32(ttl.Seconds()) + utils.Timestamp()
 	}
+	
+	// Defensive copies to prevent unsafe reuse from sync.Pool
+	keyCopy := utils.CopyString(key)
```
arturmelanchyk, line 82: in golang strings are immutable, so sharing strings `should` be safe.
can you provide a test showing that sharing key strings can be an issue?

sixcolors: You are correct that Go strings are immutable **in normal circumstances**. However, Fiber uses `utils.UnsafeString` which performs an unsafe conversion from `[]byte` to `string` without copying the underlying bytes. This is done for performance, but it means the string shares the same memory as the byte slice.

Here's the issue:

1. **Fiber's `app.toString` is `utils.UnsafeString`** (line 527 in app.go):
   ```go
   toString: utils.UnsafeString,
   ```

2. **When you call `c.Get("header")`, it uses UnsafeString** (req.go line 294):
   ```go
   c.App().toString(c.Request().Header.Peek(key))
   ```

3. **The `Header.Peek()` returns a `[]byte` from a pooled buffer**

4. **`utils.UnsafeString` creates a string that points to that pooled buffer memory**:
   ```go
   // UnsafeString returns a string pointer without allocation
   func UnsafeString(b []byte) string {
       return *(*string)(unsafe.Pointer(&b))
   }
   ```

5. **When the buffer is returned to sync.Pool and reused, the string's underlying bytes change**

This is why defensive copying is necessary. The strings aren't "normal" Go strings - they're strings created via unsafe pointer conversion that share memory with pooled buffers.

sixcolors:
```go
package main

import (
	"fmt"
	"sync"
	"unsafe"
)

// UnsafeString converts []byte to string without copying (like utils.UnsafeString in Fiber)
func UnsafeString(b []byte) string {
	return unsafe.String(unsafe.SliceData(b), len(b))
}

func main() {
	// Simulate sync.Pool like Fiber does
	pool := &sync.Pool{
		New: func() any {
			return make([]byte, 0, 64)
		},
	}

	// Simulate storing a key from a pooled buffer (without defensive copy)
	storage := make(map[string]string)

	// First request: Get buffer from pool
	buf1 := pool.Get().([]byte)
	buf1 = append(buf1[:0], "session-12345"...)
	key1 := UnsafeString(buf1) // String points to pooled buffer!
	storage[key1] = "session-data"
	fmt.Printf("1. Stored with key: %q\n", key1)
	pool.Put(buf1) // Return buffer to pool

	fmt.Println("\n2. Checking storage BEFORE second insert:")
	for k, v := range storage {
		fmt.Printf("   Key: %q => Value: %q ✓ Correct\n", k, v)
	}

	// Second request: Pool reuses the same buffer
	buf2 := pool.Get().([]byte)
	buf2 = append(buf2[:0], "csrf-token-67890"...)
	key2 := UnsafeString(buf2)
	storage[key2] = "csrf-data"
	fmt.Printf("\n3. Stored with key: %q\n", key2)
	pool.Put(buf2)

	// BUG: The first key is now corrupted because buf1 and buf2 were the same buffer!
	fmt.Println("\n4. Checking storage AFTER second insert:")
	fmt.Println("   ⚠️  CORRUPTION DETECTED:")
	for k, v := range storage {
		if v == "session-data" {
			fmt.Printf("   Key: %q => Value: %q ❌ WRONG! Should be \"session-12345\"\n", k, v)
		} else {
			fmt.Printf("   Key: %q => Value: %q ✓\n", k, v)
		}
	}
	fmt.Println("\n   The session key was corrupted from \"session-12345\" to \"csrf-token-67\"")
	fmt.Println("   because both strings pointed to the same pooled buffer memory!")
}
```

sixcolors: @arturmelanchyk Please review https://docs.gofiber.io/#zero-allocation

arturmelanchyk: Hi @sixcolors 
that's a good point!

I did a quick review and couldn't find any place where this `Storage::Set()` is called with strings constructed from byte array using `UnsafeString`, all call sites seem to only use strings allocated by golang (they come from `Config::KeyGenerator()` to be more precise).

Taking into account significant performance hit, I suggest to document that `its a caller responsibility and a common sense to pass immutable string for a key` instead, 

what do you think?

sixcolors: If you follow the middleware related to those strings, they can indeed come from unsafe sources. This fix prevents multiple recurring bugs in middleware. Additionally, note that memory storage is specifically for testing and debugging, and this behaviour mirrors that of production storage. The performance hit either occurs here or elsewhere (in middleware), and not doing it here will allow further bugs to corrupt memory and cause CI issues.

sixcolors: @arturmelanchyk

Thanks again for the review — I know this sparked good discussion.

To close the loop:

- **Multiple middleware bugs** (CSRF, Session, Cache, etc.) over the years have stemmed **exactly** from the inconsistency between `memory.Storage` (no copy) and production backends (Redis, Postgres, etc. — always copy).  
- This PR **eliminates that class of bugs forever** by making `memory.Storage` behave like production.  
- A **single `CopyString(key)`** is **not significant** at real scale — it’s only “significant” as a **percentage of doing nothing**. In practice: **negligible**.

**Resolving this conversation.**  
The change is correct, battle-tested in intent, and future-proof.

gaby: Agree on doing a copy, makes sense to avoid edge case issues

arturmelanchyk: ok, sound good!

ReneWerner87 (approved):

PR #3839, opened 2025-11-04
🧹 chore: Handle nil map targets in Binder

gaby:
## Summary
- update map binding logic to initialize nil map destinations or return a descriptive error
- extend binder parsing tests to cover zero-value map scenarios and ensure safe behavior
- add regression coverage for parse helpers when map destinations are nil or non-convertible


gaby:
Fixing issues now

PR #3843, opened 2025-11-05
Fix: Apply CORS headers before error response when request body too large

KambojRajan:
When Fiber fails to parse a large request body, it aborts the request before reaching the middleware stack, causing CORS headers to be skipped.
This PR ensures CORS headers are added even during such early errors (e.g., ErrRequestEntityTooLarge) by setting them in serverErrorHandler when EnableCORSOnError is enabled.
This improves consistency for browsers handling aborted or oversized requests.(I am using EnableCORSOnError to ensure backward compatibility).

### DIFF
--- a/redirect.go
+++ b/redirect.go
@@ -306,6 +306,8 @@ func (r *Redirect) parseAndClearFlashMessages() {
 	if err != nil {
 		return
 	}
+
+	r.c.Response().Header.DelClientCookie(FlashCookieName)
 }
 
 // processFlashMessages is a helper function to process flash messages and old input data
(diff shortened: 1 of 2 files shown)


gaby:
Hey @KambojRajan, I think this approach might not be ideal. Not all developers are currently using CORS, and expecting them to manually define CORS headers at the app level feels a bit hacky. I suggest discussing solutions in the issue first.

ReneWerner87:
Hi @KambojRajan - thanks for the PR. I cannot merge this as-is.

The core change is the wrong abstraction: CORS belongs in the CORS middleware, not mixed into the app/error logic. Duplicating CORS header logic across error paths breaks separation of concerns, risks inconsistent handling of preflight (OPTIONS) requests, and increases maintenance/verification burden.

PR #3876, opened 2025-11-18
🔥 feat: Add support for ReloadViews()

### LINKED ISSUE #3875: Reload views?
I've beein going through the Fiber code and stumbled upon the following line:

https://github.com/gofiber/fiber/blob/aaad4cf56e004d1453e41f766db74afe14739214/app.go#L1226C30-L1226C34

I was wondering if there are any side-effects that would prevent a call to `app.config.Views.Load()` from being made public via an `app.ReloadViews()` wrapper (or similar)?

The reason I'm asking is because I'm looking for a way to reload the whole views directory from disk without restarting the whole app.

gaby:
## Summary
- This pull request enhances the application's view management by introducing a ReloadViews method, which enables dynamic reloading of view templates without requiring a server restart. It also standardizes error handling for view engine configuration by defining and utilizing a new shared sentinel error, ErrNoViewEngineConfigured, ensuring more consistent and predictable error responses.

Fixes #3875

### DIFF
--- a/docs/whats_new.md
+++ b/docs/whats_new.md
@@ -84,6 +84,7 @@ We have made several changes to the Fiber app, including:
 - **State**: Provides a global state for the application, which can be used to store and retrieve data across the application. Check out the [State](./api/state) method for further details.
 - **NewErrorf**: Allows variadic parameters when creating formatted errors.
 - **GetBytes / GetString**: Helpers that detach values only when `Immutable` is enabled and the data still references request or response buffers. Access via `c.App().GetString` and `c.App().GetBytes`.
+- **ReloadViews**: Lets you re-run the configured view engine's `Load()` logic at runtime, including guard rails for missing or nil view engines so development hot-reload hooks can refresh templates safely.
 
 #### Custom Route Constraints
 


mrusme:
I have just tested this PR by using [fsnotify](https://github.com/fsnotify/fsnotify) to detect a change in the views and then triggering `app.ReloadViews()` through it. The function does not return an error, so I'm assuming the views have been reloaded. However, upon calling a reoute that uses one specific view that I had edited, I can see that the route still returns the outdated version.

I'm using the following engine:

```go
		engine = html.New(cwd, ".html")
```

The reload happens like this (simplified):

```go
		srv.watcher, err = srv.getWatcher(cwd+"/views", func(f string) {
			err := srv.app.ReloadViews()
			if err != nil {
				Error("error", err)
			}
		})
```

`app` is available in a shared context (`srv`, pointer) and the `getWatcher` method looks like this:

```go
func (srv *Web) getWatcher(path string, callback func(f string)) (*fsnotify.Watcher, error) {
	watcher, err := fsnotify.NewWatcher()
	if err != nil {
		return nil, err
	}

	go func() {
		for {
			select {
			case ev, ok := <-watcher.Events:
				if !ok {
					return
				}
				if ev.Has(fsnotify.Write) {
					callback(ev.Name)
				}
			case err, ok := <-watcher.Errors:
				if !ok {
					return
				}
				Error("error", err)
			}
		}
	}()

	if err = watcher.Add(path); err != nil {
		watcher.Close()
		return nil, err
	}

	return watcher, nil
}
```

Either it is an issue that the `ReloadViews` runs inside a Goroutine, at which point this feature then appears to be of relatively little use, or the implementation doesn't work/has some side-effects that weren't clear.

PR #3883, opened 2025-11-20
🐛 bug: Fix handling of no-body HTTP statuses

gaby:
## Summary
- This pull request refines the SendStatus function to ensure strict adherence to HTTP specifications regarding response bodies for certain status codes. By centralizing the logic for status codes that disallow payloads and introducing a dedicated helper, the change prevents unintended data transmission and improves the reliability of HTTP responses, particularly for informational, NoContent, ResetContent, and NotModified statuses.

### DIFF
--- a/.github/workflows/auto-labeler.yml
+++ b/.github/workflows/auto-labeler.yml
@@ -5,8 +5,6 @@ on:
     types: [opened, edited, milestoned]
   pull_request_target:
     types: [opened, edited, reopened, synchronize]
-  pull_request:
-    types: [opened, edited, reopened, synchronize]
 
 permissions:
   # Setting up permissions in the workflow to limit the scope of what it can do. Optional!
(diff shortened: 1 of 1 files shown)


.github/workflows/auto-labeler.yml
```diff
@@ -5,6 +5,8 @@ on:
     types: [opened, edited, milestoned]
   pull_request_target:
     types: [opened, edited, reopened, synchronize]
+  pull_request:
```
ReneWerner87: not needed, pls remove it
otherwise it runs twice

.github/labeler.yml
```diff
@@ -2,35 +2,33 @@ version: v1
 labels:
```
ReneWerner87, line 2: i like the regex "i" more
sometimes, someone is using "readme" or "README" or "Readme" .....
"i" is simpler

ReneWerner87 (approved):

.github/labeler.yml
ReneWerner87: i is not working
https://github.com/fuxingloh/multi-labeler/issues/341

PR #3904, opened 2025-11-27
🐛 bug: Enhance Origin and Referer Validation in CSRF middleware

gaby:
## Summary
- This pull request refines the CSRF middleware's handling of origin and referer headers to ensure more robust and consistent validation. By standardizing string case conversion and introducing a helper function to normalize scheme and host comparisons, the changes prevent potential mismatches when default ports are implicitly used or explicitly stated in URLs. This improves the reliability of CSRF protection by ensuring that requests from valid origins are not incorrectly blocked due to subtle differences in URL representation.

### DIFF
--- a/middleware/csrf/helpers.go
+++ b/middleware/csrf/helpers.go
@@ -32,7 +32,6 @@ func schemeAndHostMatch(schemeA, hostA, schemeB, hostB string) bool {
 }
 
 func normalizeSchemeHost(scheme, host string) string {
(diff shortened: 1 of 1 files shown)


ReneWerner87:
Is the benchmark alert real?

gaby:
> Is the benchmark alert real?

Yeah, it should be a bit slower since we are running tolower and also splitting the host/port

PR #3922, opened 2025-12-06
♻️ refactor: Optimize string handling and memory allocations

ReneWerner87:


ReneWerner87:
Why sync.Map was replaced with a regular Map
The viewBindMap in DefaultCtx was changed from sync.Map to a regular Map because:

1. Context is not shared across goroutines - Each DefaultCtx is acquired from a pool and used by a single request handler. There's no concurrent access to viewBindMap within a single request lifecycle.
3. sync.Map was being recreated on every request - The pool's release() method couldn't efficiently reset a sync.Map, defeating the purpose of pooling. A regular map can simply be cleared with clear(viewBindMap).
5. sync.Map has overhead - It uses internal locking and type assertions (interface{} storage), adding unnecessary CPU cycles when concurrency protection isn't needed.
7. Regular map with clear() is faster - Go 1.21's clear() builtin efficiently resets a map while preserving its allocated capacity, making it ideal for pooled objects.

ReneWerner87:
@gofiber/maintainers In addition to the AI review, I need a human review here, as quite a lot has been adjusted, just to be on the safe side.

gaby:
@codex review

path.go
```diff
@@ -621,14 +621,20 @@ func GetTrimmedParam(param string) string {
 
 // RemoveEscapeChar removes escape characters
 func RemoveEscapeChar(word string) string {
+	// Fast path: check if there are any escape characters first
+	escapeIdx := strings.IndexByte(word, '\\')
+	if escapeIdx == -1 {
+		return word // No escape chars, return original string without allocation
+	}
+
+	// Slow path: copy and remove escape characters
```
gaby, line 630: Why don't we call `utils.Trim(word, '\\')` here?

ReneWerner87: Good question, i will come back with an answer

ReneWerner87: @gaby 

Image

PR #3923, opened 2025-12-07
♻️ refactor: Replace string functions

ReneWerner87:
- [ ] check benchmark output and compare with main branch

PR #3956, opened 2025-12-20
refactor: replace strings.Index with strings.Cut for improved readallity

ReneWerner87:


ReneWerner87:
> I believe IndexByte is more performant than Contains when searching for a single byte.

Yes, I know that indexByte is faster. That was actually just to fix modernizer for now and is an excerpt from your adjustments in the pull request.

https://patch-diff.githubusercontent.com/raw/gofiber/fiber/pull/3946.diff

The performance acceleration comes from my other pull request
https://github.com/gofiber/fiber/pull/3923

PR #3962, opened 2025-12-23
🔥 feat: Implement OverrideParam override behavior for DefaultCtx

### LINKED ISSUE #2118: 🚀 [Feature]: Expose Ctx.values
### Feature Description

It would be good to expose `values` in `Context`.

I need to intercept the request and modify the parameter values in a middleware.
The end route handler uses `c.Params(key)`. This method uses `values` and `c.route.Params`.

I can already modify `c.route.Params` but I'm stuck due to `values`.

https://github.com/gofiber/fiber/blob/v2.37.1/ctx.go#L65

_No response_

### Code Snippet (optional)

_No response_

### Checklist:

TheAspectDev:
# Description

Allows `Ctx.OverrideParam` to override existing route parameters while ignoring non-existing ones, and ensures proper behavior when using a nil router.

This PR reflects the mutable req.params in ``expressjs`` and adds corresponding tests and benchmarks to validate correctness and measure performance.

## Changes introduced

- 🚨 Added tests for `OverrideParam`:
  - Overriding existing params
  - Ignoring non-existing params
  - Nil-router behavior
- 🚨 Added benchmark for `OverrideParam`
- ♻️ Minor cleanup in `ctx.go` and `ctx_interface_gen.go` to satisfy lint rules
- 📚 Updated docs, added ``c.OverrideParam``

Fixes #2118

## Type of change

- [x] New feature (adds `OverrideParam` override behavior)
- [x] Enhancement (tests and benchmarks for robustness)

## 🧾 Checklist
- [x] Followed the inspiration of the Express.js framework for new functionalities. ( req.params.q = "new-value" )
- [x] Benchmarks added for performance tracking
- [x] Added/updated unit tests

### DIFF
--- a/docs/api/ctx.md
+++ b/docs/api/ctx.md
@@ -501,24 +501,47 @@ app.Post("/", func(c fiber.Ctx) error {
 })
 ```
 
-### UpdateParam
+### OverrideParam
 
-Overwrites the value of an existing route parameter. If the parameter does not exist, this method does nothing.
+Overwrites the value of an existing route parameter.
 
-:::info
-For wildcard parameters (both `*` and `+`), `UpdateParam` normalizes each to its internal numbered representation (e.g., `*1`, `+1`). To update the first `*` wildcard parameter, you can use either `*` or `*1`; to update the first `+` wildcard parameter, you can use either `+` or `+1`. For subsequent wildcards of each type, you must use their numbered representation (e.g., `*2`, `+2`).
+:::note
+If the parameter does not exist, this method does nothing.
 :::
 
 ```go title="Signature"
-func (c fiber.Ctx) UpdateParam(name, value string)
+func (c fiber.Ctx) OverrideParam(name, value string)
 ```
 
 ```go title="Example"
+// GET http://example.com/user
 app.Get("/user/:name", func(c fiber.Ctx) error {
   // mutate parameter
-  c.UpdateParam("name", "new value")
+  c.OverrideParam("name", "new value")
+  return c.SendString(c.Params("name")) // sends "new value"
+})
+// GET http://example.com/shop/tech/1
+app.Get("/shop/*", func(c fiber.Ctx) error {
+  // mutate parameter
+  c.OverrideParam("*", "new tech") // replaces "tech/1" with "new tech"
+  return c.SendString(c.Params("*")) // sends "new tech"
+})
 
-  return c.SendString(c.Params("name"))
+```
+
+Unnamed route parameters can be accessed by their character (`*` or `+`) followed by their position index (e.g., `*1` for the first wildcard, `*2` for the second).
+
+```go title="Example"
+// GET /v1/brand/4/shop/blue/xs
+app.Get("/v1/*/shop/*", func(c fiber.Ctx) error {
+  // mutate parameter
+  c.OverrideParam("*1", "updated brand")
+  c.OverrideParam("*2", "updated data")
+  
+  param1 := c.Params("*1") // "updated brand"
+  param2 := c.Params("*2") // "updated data"
+
+  // ...
 })
 ```
 
--- a/docs/whats_new.md
+++ b/docs/whats_new.md
@@ -567,6 +567,7 @@ testConfig := fiber.TestConfig{
 - **Matched**: Detects when the current request path matched a registered route.
 - **IsMiddleware**: Indicates if the current handler was registered as middleware.
 - **HasBody**: Quickly checks whether the request includes a body.
+- **OverrideParam**: Overwrites the value of an existing route parameter, or does nothing if the parameter does not exist
 - **IsWebSocket**: Reports if the request attempts a WebSocket upgrade.
 - **IsPreflight**: Identifies CORS preflight requests before handlers run.
 


gaby:
@codex review

TheAspectDev:
@gaby Would you mind running them again please? code's been updated.

docs/api/ctx.md
```diff
@@ -501,6 +501,23 @@ app.Post("/", func(c fiber.Ctx) error {
 })
 ```
 
+### UpdateParam
+
+Overwrites the value of an existing route parameter. If the parameter does not exist, this method does nothing.
```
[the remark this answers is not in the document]
TheAspectDev: I believe 

Overwrites the value of an existing route parameter. If the parameter does not exist, this method does nothing.

:::info
For wildcard parameters (e.g., `*` or `+`), `UpdateParam` normalizes them to their internal representation (e.g., `*1`, `+1`).
:::

would be enough right?

TheAspectDev:
@gaby 

- [x] Markdown linting
- [x] Code linting
- [x] Code Coverage checked
- [x] Tests passing
- [x] Formatted code
- [x] audited

Would you mind running actions again? 

Thanks!

ctx_test.go
```diff
@@ -7478,6 +7478,117 @@ func Test_Ctx_End_after_drop(t *testing.T) {
...
+		req, err := http.NewRequest(http.MethodGet, "/user/original", http.NoBody)
+		require.NoError(t, err)
+
+		resp, err := app.Test(req)
+		require.NoError(t, err)
+
+		defer func() { require.NoError(t, resp.Body.Close()) }()
+
+		require.Equal(t, StatusOK, resp.StatusCode)
+	})
```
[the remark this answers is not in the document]
gaby: @TheAspectDev We do need a test with `CaseSensitive: false`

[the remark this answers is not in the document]
TheAspectDev: @gaby I did make a test with CaseSensitive: false, but removed it later.
Because this is already tested in "route_params" should i restore it anyway?

ctx.go
```diff
@@ -383,6 +383,42 @@ func (c *DefaultCtx) HasBody() bool {
 	return len(c.fasthttp.Request.Body()) > 0
 }
 
+// UpdateParam overwrites a route parameter value by name.
+// If the parameter name does not exist in the route, this method does nothing.
+func (c *DefaultCtx) UpdateParam(name, value string) {
+	// If no route is matched, there are no parameters to update
+	if c.route == nil {
```
gaby: You can use c.Matched() here

ctx.go
```diff
@@ -383,6 +383,42 @@ func (c *DefaultCtx) HasBody() bool {
...
+		if i >= maxParams {
+			return
+		}
+		if c.app.config.CaseSensitive {
+			if param == name {
+				c.values[i] = value
+				return
+			}
+		} else {
+			if len(param) != len(name) {
```
gaby: This check can be removed, see https://github.com/gofiber/utils/blob/master/byteseq.go#L9

ctx_test.go
```diff
@@ -7478,6 +7478,117 @@ func Test_Ctx_End_after_drop(t *testing.T) {
...
+		})
+		req, err := http.NewRequest(http.MethodGet, "/files/testing", http.NoBody)
+		require.NoError(t, err)
+		resp, err := app.Test(req)
+		require.NoError(t, err)
+		defer func() { require.NoError(t, resp.Body.Close()) }()
+		require.Equal(t, StatusOK, resp.StatusCode)
+	})
+
+	t.Run("exceeds_max_params", func(t *testing.T) {
```
gaby: This test has no assertion(require calls)

ctx_test.go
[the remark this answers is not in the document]
gaby: Yeah add it as a separate one, so it's clear what we are testing.

gaby (changes requested):

TheAspectDev:
@gaby Done! 🫡

TheAspectDev:
Hey @gaby
Should i push another version of the code with a higher coverage?
I've talked to one of the maintainers, and there seems to be a problem with the CI.

gaby:
> Hey @gaby Should i push another version of the code with a higher coverage? I've talked to one of the maintainers, and there seems to be a problem with the CI.

Doesn't seem to be related to your PR.

ctx.go
```diff
@@ -383,6 +383,47 @@ func (c *DefaultCtx) HasBody() bool {
...
+	if name == "*" || name == "+" {
+		name += "1"
+	}
+
+	if c.app.config.CaseSensitive {
+		for i, param := range c.route.Params {
+			// Prevent out-of-bounds access if route params exceed allocated values
+			if i >= maxParams {
+				return
+			}
```
sixcolors: ### **The bounds check `i >= maxParams` IS NECESSARY - but for a different reason than I initially thought**

1. **No validation during route registration**: The `parseRoute` function in path.go does NOT validate or limit the number of parameters. It will happily parse routes with 50+ parameters.

2. **The framework CAN have routes with > maxParams**: There's nothing preventing registration of such routes.

sixcolors (changes requested):

ctx.go
sixcolors: ### Better Solution: Add maxParams Validation at Route Registration
Add a panic check right after parsing the route, similar to the existing validation for nil handlers and invalid HTTP methods

router.go:548

```go
// Validate that the route doesn't exceed the maximum number of parameters
	if len(parsedRaw.params) > maxParams {
		panic(fmt.Sprintf("route has %d parameters, exceeding the maximum of %d: %s\n", len(parsedRaw.params), maxParams, pathRaw))
	}
```

Then remove the redundant bounds checks from UpdateParam.

sixcolors: @gaby does checking at registration make more sense to you?

gaby: We should *not* be throwing a panic here, this would run at runtime causing fiber servers to be easily DDoS.

gaby: Yes

sixcolors: > We should _not_ be throwing a panic here, this would run at runtime causing fiber servers to be easily DDoS.

I understand the concern, but let me clarify: the proposed panic is at registration, not at runtime. Let me clarify the context:

## Route Registration vs Request Handling

Route registration happens **at server startup**, not during request handling. When you call `app.Get("/path/:param", handler)`, this happens during initialization before the server starts accepting requests. So a panic here would:

1. ✅ Fail fast during development/deployment
2. ✅ Prevent the server from starting with invalid configuration
3. ✅ **NOT** be exploitable at runtime (routes are registered before the server starts)

This is similar to the existing panics in the same function:
- Line 519: `panic(fmt.Sprintf("missing handler/middleware in route: %s\n", pathRaw))`
- Line 523: `panic(fmt.Sprintf("nil handler in route: %s\n", pathRaw))`
- Line 552: `panic(fmt.Sprintf("add: invalid http method %s\n", method))`

If you check router.go at line 548, you’ll see that it’s part of the `func (app *App) register(methods []string, pathRaw string, group *Group, handlers …Handler) {` function.

So, unless I’ve missed something, I believe I’m correct here.

sixcolors: ### When does `register()` execute?

The `register()` function (line 517 in router.go) is **primarily** called during startup when you define routes:

```go
app.Get("/users/:id", handler)  // ← register() called HERE during startup
app.Listen(":3000")              // ← server starts AFTER all routes are registered
```

**However**, routes *can* be registered dynamically at runtime using `app.Get()` + `app.RebuildTree()`, though the documentation explicitly warns this is "not recommended for production" and "not thread-safe in runtime" (line 699-705).

### Evidence startup is the primary use case:

1. **Existing panics in the same function** that have been there since the beginning:
   - Line 519: `panic("missing handler/middleware in route...")`
   - Line 523: `panic("nil handler in route...")`  
   - Line 556: `panic("add: invalid http method...")`

2. **Standard pattern**: Routes are typically registered during `init()` or `main()` before `app.Listen()` is called

3. **No user input involved**: The route path comes from application code, not from HTTP requests

gaby: These AI slop answers 😂 The panic should happen at registration, not during a function call lile `UpdateParams()` since that will kill the server mid-requests.

TheAspectDev: I think he mean't to call the panic during start up, not inside UpdateParams.

sixcolors: Correct I proposed the panic at router.go:548

TheAspectDev: So should i implement the start-up max-param-validation?

This would also help me remove the extra test for it.

If @gaby and @sixcolors agree.

If yes, which one?
- on start up ( router )
- UpdateParams check without panic outside the loop

On start up seems pretty solid
UpdateParams wouldn't hurt anything though. Not much of a performance issue there. (  I'm not experienced in this so correct me if im wrong )

TheAspectDev: So should i implement the start-up max-param-validation?

This would also help me remove the extra test for it.

If @gaby and @sixcolors agree.

If yes, which one?
- on start up ( router )
- UpdateParams check without panic outside the loop

On start up seems pretty solid
UpdateParams wouldn't hurt anything though. Not much of a performance issue there. (  I'm not experienced in this so correct me if im wrong )

sixcolors: I favour the panic in router.go and then removing the checks from UpdateParams. @gaby ?

TheAspectDev: @gaby Just a reminder
panic() is ran inside of register(methods []string, pathRaw string, group *Group, handlers ...Handler), at router.go:517

- And there are other panics here too, this is on start-up.
```go
func (app *App) register(methods []string, pathRaw string, group *Group, handlers ...Handler) {
	// A regular route requires at least one ctx handler
	if len(handlers) == 0 && group == nil {
----------panic(fmt.Sprintf("missing handler/middleware in route: %s\n", pathRaw))
	}
	// No nil handlers allowed
	for _, h := range handlers {
		if h == nil {
---------------panic(fmt.Sprintf("nil handler in route: %s\n", pathRaw))
		}
	}

	// Precompute path normalization ONCE
	if pathRaw == "" {
		pathRaw = "/"
	}
	if pathRaw[0] != '/' {
		pathRaw = "/" + pathRaw
	}
	pathPretty := pathRaw
	if !app.config.CaseSensitive {
		pathPretty = utils.ToLower(pathPretty)
	}
	if !app.config.StrictRouting && len(pathPretty) > 1 {
		pathPretty = utils.TrimRight(pathPretty, '/')
	}
	pathClean := RemoveEscapeChar(pathPretty)

	parsedRaw := parseRoute(pathRaw, app.customConstraints...)
	parsedPretty := parseRoute(pathPretty, app.customConstraints...)

	isMount := group != nil && group.app != app
        
    // to be added
	if len(parsedRaw.params) > maxParams {
		panic(fmt.Sprintf("route has %d parameters, exceeding the maximum of %d: %s\n", len(parsedRaw.params), maxParams, pathRaw))
	}
```





and this is going to remove "maxParam check" on UpdateParam ( which is going to run on each request )

```go
// UpdateParam overwrites a route parameter value by name.
// If the parameter name does not exist in the route, this method does nothing.
func (c *DefaultCtx) UpdateParam(name, value string) {
	// If no route is matched, there are no parameters to update
	if !c.Matched() {
		return
	}

	// Normalize wildcard (*) and plus (+) tokens to their internal
	// representations (*1, +1) used by the router.
	if name == "*" || name == "+" {
		name += "1"
	}

	if c.app.config.CaseSensitive {
		for i, param := range c.route.Params {
			// Prevent out-of-bounds access if route params exceed allocated values
			if param == name {
				c.values[i] = value
				return
			}
		}
		return
	}

	nameBytes := utils.UnsafeBytes(name)
	for i, param := range c.route.Params {
		// Prevent out-of-bounds access if route params exceed allocated values
		if utils.EqualFold(utils.UnsafeBytes(param), nameBytes) {
			c.values[i] = value
			return
		}
	}
}

```

TheAspectDev: and ``exceeds_max_params`` test will be removed from ``Test_Ctx_UpdateParam(t *testing.T)`` in ``ctx_test.go``
and a similar test will be added to router_test.go which i would need guidance for.

TheAspectDev: Here's an example test in ``router_test.go`` for parameter count exceeding the maximum count

```go
func Test_Router_Register_Param_Max_Exceed(t *testing.T) {
	t.Parallel()

	app := New()

	parts := make([]string, maxParams+1)
	for i := range parts {
		parts[i] = fmt.Sprintf(":p%d", i)
	}
	path := "/" + strings.Join(parts, "/")

	require.Panics(t, func() {
		app.register([]string{"GET"}, path, nil, func(ctx Ctx) error { return nil })
	})
}
```

TheAspectDev:
Hey @gaby 
I'm not sure you got the notification from the review comments, would you mind checking and confirming it?

gaby (approved):

sixcolors:
@TheAspectDev, go ahead as I’ll address the issue I raised with the other maintainers elsewhere.

TheAspectDev:
> **Actionable comments posted: 0**
> 
> 🧹 Nitpick comments (3)
> > ctx_test.go (1)> `8119-8140`: **Consider asserting the final param value after the benchmark loop**
> > > To mirror other benchmarks in this file and catch regressions in `UpdateParam`, you can add a quick post‑loop check that the parameter was actually updated; this won’t materially affect timing.
> > > 
> > > Proposed post-loop assertion
> > > ```diff
> > >  func Benchmark_Ctx_UpdateParam(b *testing.B) {
> > > @@
> > >  	c.setMatched(true)
> > >  
> > >  	b.ReportAllocs()
> > >  	b.ResetTimer()
> > >  	for b.Loop() {
> > >  		c.UpdateParam("name", "changed")
> > >  	}
> > > +
> > > +	// Verify that UpdateParam actually updated the value
> > > +	require.Equal(b, "changed", c.Params("name"))
> > >  }
> > > ```
> > 
> > 
> > 
> > router_test.go (1)> `11-11`: **Route param max-exceed test is good; you may want to assert the panic message too**
> > > The new test correctly forces `maxParams+1` parameters and ensures `register` panics, and the `fmt` import is scoped to that usage. If you want the test to lock in the formatted error (count, limit, and path), you could switch to `require.PanicsWithValue` (or inspect `recover()` inside the closure) and assert that the message contains the expected details.
> > > Also applies to: 736-750
> > 
> > 
> > 
> > ctx.go (1)> `386-419`: **Implementation is correct, but outdated comments should be removed.**
> > > The core logic is sound:
> > > 
> > > * Uses `c.Matched()` to verify route existence
> > > * Correctly normalizes wildcard tokens (`*` → `*1`, `+` → `+1`)
> > > * Properly handles case-sensitive and case-insensitive parameter matching using `utils.EqualFold`
> > > 
> > > However, the comments on lines 402 and 413 stating "Prevent out-of-bounds access if route params exceed allocated values" are now outdated. With the new validation at route registration (router.go:550-552), routes cannot exceed `maxParams`, so the iteration over `c.route.Params` will never go out of bounds. These comments should be removed to avoid confusion.
> > > 
> > > 🔎 Suggested cleanup
> > > ```diff
> > >  	if c.app.config.CaseSensitive {
> > >  		for i, param := range c.route.Params {
> > > -			// Prevent out-of-bounds access if route params exceed allocated values
> > >  			if param == name {
> > >  				c.values[i] = value
> > >  				return
> > >  			}
> > >  		}
> > >  		return
> > >  	}
> > > 
> > >  	nameBytes := utils.UnsafeBytes(name)
> > >  	for i, param := range c.route.Params {
> > > -		// Prevent out-of-bounds access if route params exceed allocated values
> > >  		if utils.EqualFold(utils.UnsafeBytes(param), nameBytes) {
> > >  			c.values[i] = value
> > >  			return
> > >  		}
> > >  	}
> > > ```
> 
> 📜 Review details



- 8119-8140: Consider asserting the final param value after the benchmark loop: this is expected behaviour and tested by ``Test_Ctx_UpdateParam``
- Other two changes are acceptable and will be pushed asap

TheAspectDev:
@sixcolors

sixcolors (approved):
This looks good to me! 👍

ReneWerner87:
@TheAspectDev Pls also mention the new functionality with an example in the whats new md file
Otherwise its Kind of hidden for the others and less people will use it

ReneWerner87:
@gaby @sixcolors @efectn
Wdyt about the name?
@coderabbitai pls list possible names for this function
How is the name in expressjs or other rest/http frameworks?

ReneWerner87:
I like SetParam or OverrideParam

gaby:
> I like SetParam or OverrideParam

- I like `OverrideParam`.

ReneWerner87:
@sixcolors @TheAspectDev ok with the new name?

TheAspectDev:
Yeah, ``OverrideParam`` seems more fitting to the action it does.

TheAspectDev:
> @TheAspectDev Pls also mention the new functionality with an example in the whats new md file Otherwise its Kind of hidden for the others and less people will use it

Okay!

TheAspectDev:
@ReneWerner87

TheAspectDev:
> **Actionable comments posted: 0**
> 
> 🧹 Nitpick comments (1)
> 📜 Review details

This behavior is expected and tested in ``Test_Ctx_OverrideParam``

sixcolors:
> @sixcolors @TheAspectDev ok with the new name?

Yes it’s good. Thanks @ReneWerner87

ReneWerner87:
Perfect, merge 🤩

PR #3968, opened 2025-12-24
🐛 bug: Preserve session data map across resets

gaby:
## Summary
- clear session data maps in session middleware without reallocating to retain capacity between pool uses
- add regression coverage ensuring pooled data reuses the same map after reset and does not leak prior entries


sixcolors (approved):
LGTM, aside from CI lint fail and other comments

ReneWerner87 (approved):

PR #3976, opened 2025-12-31
🔥 feat: Auto-enforce Secure=true for Partitioned cookies in Cookie()

### LINKED ISSUE #2878: 🐛 [Bug]: (c *fiber.Ctx).ClearCookie() does absolutely nothing
### Bug Description

Setting the cookie is fine. The gofiber server responds with this header, which sets the cookie.

```
Set-Cookie: __Host-pixivfe-ImageProxy=https://i.pixiv.re; expires=Mon, 25 Mar 2024 15:06:10 GMT; path=/; HttpOnly; secure; SameSite=Strict
```

However, the server can't reset it.

Neither `c.ClearCookie(name)` or `c.ClearCookie()` work.

### How to Reproduce

1. Set the cookie above in browser
2. have a server call `c.ClearCookie`
3. See in Developer Tool > Storage > Cookies. cookie still there.

### Expected Behavior

cookie cleared

### Fiber Version

v2.52.0

### Workaround

Set the cookie with no value and an expiration date in the past.

### Checklist:

darwin808:
## Summary

  - Auto-enforces `Secure=true` when `Partitioned=true` is set on a cookie, per the CHIPS specification
  - Updates cookie expiration documentation to use `fasthttp.CookieExpireDelete` constant
  - Adds documentation example for expiring cookies with `Path`/`Domain` attributes

  ## Changes

  **Cookie enforcement (`res.go`):**
  - Automatically sets `Secure=true` when a cookie has `Partitioned=true`, as required by the [CHIPS
  spec](https://developer.mozilla.org/en-US/docs/Web/Privacy/Privacy_sandbox/Partitioned_cookies)

  **Documentation (`docs/api/ctx.md`):**
  - Uses `fasthttp.CookieExpireDelete` instead of manual time calculation for cookie deletion examples
  - Adds example showing how to expire cookies with specific `Path`/`Domain` attributes

  ## Test plan

  - [x] Updated test to reflect auto-fix behavior for Partitioned cookies
  - [x] All existing cookie tests pass

Fixes #2878

### DIFF
--- a/ctx_test.go
+++ b/ctx_test.go
@@ -1211,7 +1211,7 @@ func Test_Ctx_Cookie_Invalid(t *testing.T) {
 		{Name: "i", Value: "b", Domain: "2001:db8::1"},                                // ipv6 not allowed
 		{Name: "p", Value: "b", Path: "\x00"},                                         // invalid path byte
 		{Name: "e", Value: "b", Expires: time.Date(1500, 1, 1, 0, 0, 0, 0, time.UTC)}, // invalid expires
-		{Name: "s", Value: "b", Partitioned: true},                                    // partitioned but not secure
+		// Note: Partitioned without Secure is auto-fixed (Secure=true set automatically per CHIPS spec)
 	}
 
 	for _, invalid := range cases {
--- a/res.go
+++ b/res.go
@@ -249,6 +249,11 @@ func (r *DefaultRes) Cookie(cookie *Cookie) {
 		sameSite = http.SameSiteLaxMode
 	}
 
+	// Partitioned requires Secure=true per CHIPS spec
+	if cookie.Partitioned {
+		cookie.Secure = true
+	}
+
 	// create/validate cookie using net/http
 	hc := &http.Cookie{
 		Name:        cookie.Name,


darwin808:
The CI failure in `Test_App_Test_Goroutine_Leak_Compare` is unrelated to this PR's changes. This test checks goroutine leak behavior in `app.Test()` and is sensitive to system load/timing during CI runs.

The test expected <10 leaked goroutines but got 14, which can vary based on:
- Concurrent test execution
- GC timing
- System background processes

Could a maintainer please re-run CI? The `ExpireCookie` changes only affect cookie handling and don't touch the `app.Test()` code path.

gaby:
@darwin808 Don't use force push, we can't see what's changing between commits. Makes reviewing harder.

gaby:
@darwin808 Tag me once the comments raised as addressed. Thanks 💪

darwin808:
@gaby Addressed the review comments - added SameSite and Partitioned support, updated docs, and refactored tests to table-driven style.

ctx_test.go
```diff
@@ -4503,6 +4503,100 @@ func Test_Ctx_ClearCookie(t *testing.T) {
...
+			},
+			expectedStrs: []string{"partitioned_cookie=;", "Partitioned", "secure", "expires="},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			app := New()
+			c := app.AcquireCtx(&fasthttp.RequestCtx{})
```
gaby: You are missing a defer release of the contex here.

gaby (changes requested):

ctx_test.go
darwin808: @gaby

res.go
```diff
@@ -215,6 +218,54 @@ func (r *DefaultRes) ClearCookie(key ...string) {
...
 }
 
+// ExpireCookie expires a cookie by its cookie definition.
+// This is useful when you need to expire a cookie that was set with a specific
+// Path or Domain. The browser will only clear the cookie if the Path and Domain
+// attributes match the original cookie.
+//
+// Only the Name, Path, Domain, Secure, HTTPOnly, SameSite, and Partitioned fields are used.
+// The Value and Expires fields are overwritten to expire the cookie.
+func (r *DefaultRes) ExpireCookie(cookie *Cookie) {
```
ReneWerner87:
```go
cookie.Expires = fasthttp.CookieExpireDelete
cookie.Value = "" // not really needed

c.Res().Cookie(cookie)
```
@darwin808 
Do the same thing or?

I'm not sure if that justifies a completely new function.

ReneWerner87: Perhaps a little documentation in the FAQ section would be better.

BTW, we should keep your Golang doc comments.

ReneWerner87:
@darwin808 can you check my hint

res.go
darwin808: @ReneWerner87  The difference is `ExpireCookie` auto-enforces `Secure=true` when `SameSite=None` or `Partitioned=true` - browsers reject these cookies without Secure per RFC 6265bis and CHIPS spec. Using `Cookie()` directly, users would need to remember this.


  Happy to add FAQ docs too if you'd like both.

ReneWerner87: > The difference is ExpireCookie auto-enforces Secure=true when SameSite=None or Partitioned=true - browsers reject these cookies without Secure per RFC 6265

should we inject this logic in the c.Res().Cookie() function ?

darwin808: @ReneWerner87  It's already there for `SameSite=None`

  For `Partitioned` - yes, `Cookie()` is missing that enforcement. Want me to add it in this PR?

gaby: Yes

darwin808: @gaby https://github.com/gofiber/fiber/pull/3976/commits/ab2fd706328d7169aad69e67522cba59e2c094dc

ReneWerner87: @darwin808 but now we can simply use

```go
cookie.Expires = fasthttp.CookieExpireDelete
cookie.Value = "" // not really needed

c.Res().Cookie(cookie)
```
or ?

darwin808: Yeah you're right, that works fine. 

  I'll convert this to a docs-only PR - just add an example showing how to expire cookies with Path/Domain using Cookie() + fasthttp.CookieExpireDelete. That should be enough to address #2878.

ReneWerner87:
@darwin808 thx for the adjustments

I checked and now only documentation adjustments are included.
Can you change the description and headline for this?
I also think the addition you wanted to add regarding “cookie.Partitioned” is still missing.
In your original method, you had an auto change for this, right?

ReneWerner87:
i mean this
https://github.com/gofiber/fiber/pull/3976#discussion_r2658540292
https://github.com/gofiber/fiber/commit/ab2fd706328d7169aad69e67522cba59e2c094dc

ReneWerner87:
@darwin808 ping
we only need this last small adjustment

darwin808:
@ReneWerner87  Done! Added the Partitioned → Secure=true auto-enforcement in 338dad15 and updated the PR title/description. Ready for review.

ReneWerner87:
https://developer.mozilla.org/en-US/docs/Web/Privacy/Guides/Privacy_sandbox/Partitioned_cookies

PR #3990, opened 2026-01-05
🐛 bug: Respect body limit configuration in adaptor middleware

gaby:
## Summary
- derive adaptor body size enforcement from the Fiber app BodyLimit with a default fallback
- add adaptor tests covering default, reduced, and expanded BodyLimit configurations to ensure 413 responses when appropriate
- document adaptor BodyLimit behavior in middleware docs and v3 "What's New"


middleware/adaptor/adaptor.go
```diff
@@ -213,7 +213,7 @@ func handlerFunc(app *fiber.App, h ...fiber.Handler) http.HandlerFunc {
 		defer fasthttp.ReleaseRequest(req)
 
 		// Convert net/http -> fasthttp request with size limit
-		const maxBodySize = 10 * 1024 * 1024 // 10MB limit
+		maxBodySize := int64(app.Config().BodyLimit)
```
ReneWerner87, line 216: okay, we reuse the normal BodyLimit for the adaptor
should we add 

Image
some hints to the doc´s here ?

@gaby

gaby: @ReneWerner87 Done

PR #3991, opened 2026-01-06
🐛 bug: Validate CORS origins before using AllowOriginsFunc

gaby:
## Summary
- normalize origins received by AllowOriginsFunc, accepting only serialized origins or the literal null
- reject invalid origins such as userinfo, paths, queries, or wildcards before setting Access-Control-Allow-Origin
- expand CORS tests to cover non-serialized origins and adjust expectations for wildcard credential handling

Information about "null" origin: https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CORS#origin


gaby:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/gofiber/fiber/pull/3991#pullrequestreview-3631048109)

gaby:
@copilot Two things:
- isOriginSerializedOrNull begin defined inside another function is hurting performance, this should be an actual internal function
- The linter is complaining about this:

```
Error: middleware/cors/cors.go:161:26: identical-ifelseif-branches: "if...else if" chain with identical branches (lines 161 and 163) (revive)
  			if originIsSerialized {
  				allowOrigin = originHeaderRaw
  			} else if originIsNull {
  1 issues:
  * revive: 1
```

gaby:
@copilot Why are we even comparing against "null"?

gaby:
@copilot Update the CORS middleware documentation to document the handling of "null" origin. You can also reference this https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CORS#origin 

Also update the function docstring to reference the mozilla docs

ReneWerner87 (approved):

sixcolors (approved):
This is a **high-quality security fix** that makes the CORS middleware more robust and secure.

This is a breaking change for applications using invalid origins with `AllowOriginsFunc`. However the breaking change only affects invalid/malicious origins, so legitimate use cases are unaffected.

LGTM!

PR #3826, opened 2025-10-28
fix up timeout middleware

pjebs:
# Description

Fixed timeout middleware

Needs to be checked

Fixes #3394 

## Changes introduced

List the new features or adjustments introduced in this pull request. Provide details on benchmarks, documentation updates, changelog entries, and if applicable, the migration guide.

- [ ] Benchmarks: Describe any performance benchmarks and improvements related to the changes.
- [ ] Documentation Update: Detail the updates made to the documentation and links to the changed files.
- [ ] Changelog/What's New: Include a summary of the additions for the upcoming release notes.
- [ ] Migration Guide: If necessary, provide a guide or steps for users to migrate their existing code to accommodate these changes.
- [ ] API Alignment with Express: Explain how the changes align with the Express API.
- [ ] API Longevity: Discuss the steps taken to ensure that the new or updated APIs are consistent and not prone to breaking changes.
- [ ] Examples: Provide examples demonstrating the new features or changes in action.

## Type of change

Please delete options that are not relevant.

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Enhancement (improvement to existing features and functionality)
- [ ] Documentation update (changes to documentation)
- [ ] Performance improvement (non-breaking change which improves efficiency)
- [ ] Code consistency (non-breaking change which improves code reliability and robustness)

## Checklist

Before you submit your pull request, please make sure you meet these requirements:

- [ ] Followed the inspiration of the Express.js framework for new functionalities, making them similar in usage.
- [ ] Conducted a self-review of the code and provided comments for complex or critical parts.
- [ ] Updated the documentation in the `/docs/` directory for [Fiber's documentation](https://docs.gofiber.io/).
- [ ] Added or updated unit tests to validate the effectiveness of the changes or new features.
- [ ] Ensured that new and existing unit tests pass locally with the changes.
- [ ] Verified that any new dependencies are essential and have been agreed upon by the maintainers/community.
- [ ] Aimed for optimal performance with minimal allocations in the new code.
- [ ] Provided benchmarks for the new code to analyze and improve upon.

## Commit formatting

Please use emojis in commit messages for an easy way to identify the purpose or intention of a commit. Check out the emoji cheatsheet here: [CONTRIBUTING.md](https://github.com/gofiber/fiber/blob/main/.github/CONTRIBUTING.md#pull-requests-or-commits)

PR #4009, opened 2026-01-15
fix(timeout): Return immediately on timeout and propagate context correctly

### LINKED ISSUE #3394: 🐛 [Bug]: Timeout middleware needs to be redesigned
### Bug Description

Currently, the timeout middleware waits for a certain duration.
If the handler returns before that time, it works as per normal.

If the handler takes longer, it waits for handler to finish.
Then it returns a timeout error.

It can be redesigned so that if the timeout completes first, the timeout error is immediately returned EVEN IF the handler continues until completion. At least the client gets a response.

### How to Reproduce

As it operates normally

### Expected Behavior

After timeout, return immediately (even if request handler continues)

### Fiber Version

v3

### Code Snippet (optional)

```go
// New enforces a timeout for each incoming request. If the timeout expires or
// any of the specified errors occur, fiber.ErrRequestTimeout is returned.
func New(h fiber.Handler, timeout time.Duration, tErrs ...error) fiber.Handler {
	return func(ctx fiber.Ctx) error {
		// If timeout <= 0, skip context.WithTimeout and run the handler as-is.
		if timeout <= 0 {
			return runHandler(ctx, h, tErrs)
		}

		// Create a context with the specified timeout; any operation exceeding
		// this deadline will be canceled automatically.
		timeoutContext, cancel := context.WithTimeout(ctx, timeout)
		defer cancel()

		// Run the handler and check for relevant errors.
		err := runHandler(ctx, h, tErrs)

		// If the context actually timed out, return a timeout error.
		if errors.Is(timeoutContext.Err(), context.DeadlineExceeded) {
			return fiber.ErrRequestTimeout
		}
		return err
	}
}

// runHandler executes the handler and returns fiber.ErrRequestTimeout if it
// sees a deadline exceeded error or one of the custom "timeout-like" errors.
func runHandler(c fiber.Ctx, h fiber.Handler, tErrs []error) error {
	// Execute the wrapped handler synchronously.
	err := h(c)
	// If the context has timed out, return a request timeout error.
	if err != nil && (errors.Is(err, context.DeadlineExceeded) || isCustomError(err, tErrs)) {
		return fiber.ErrRequestTimeout
	}
	return err
}

// isCustomError checks whether err matches any error in errList using errors.Is.
func isCustomError(err error, errList []error) bool {
	for _, e := range errList {
		if errors.Is(err, e) {
			return true
		}
	}
	return false
}
```

### Checklist:

### LINKED ISSUE #3671: 🐛 [Bug]: timeout is not actually enforced
### Bug Description

Doesn't this just check for an error afterward? From what I see, the timeout doesn't actually stop the handler from running. It also doesn't pass the cancellation down the call stack, so inside the handler there's no way to check if the request has timed out.

https://github.com/gofiber/fiber/blob/c309d46c3a0fcf6f3ca49f7e2aa8e3e8b4ae8e48/middleware/timeout/timeout.go#L25-L28

### How to Reproduce

1. Go to '....'
3. Do '....'
4. See '....'

### Expected Behavior

Context(tCtx) with cancelation is passed down

### Fiber Version

v3.0.0-beta.5

### Code Snippet (optional)

```go

```

### Checklist:

ReneWerner87:
## Summary
  - Fixes #3671 - Timeout context is now properly propagated to handlers via `c.Context().Done()`
  - Fixes #3394 - Middleware returns immediately when timeout expires, even if handler is still running
  
Replace #3680
Replace #3826

  ## Changes
  - Run handler in goroutine to enable immediate timeout response
  - Keep `context.WithTimeout` (unlike #3826 which used `time.After`) so handlers can detect cancellation
  - Add panic recovery that converts panics to `500 Internal Server Error`
  - Simplify code by merging `runHandler` and `isCustomError` into single `isTimeoutError` function
  - Update documentation in `docs/middleware/timeout.md` and `docs/whats_new.md`

  ## Breaking Changes
  None - The API remains the same, but behavior is improved:
  - Clients now receive timeout responses immediately instead of waiting for handler completion
  - Panics in handlers now return 500 instead of crashing

  ## Test Plan
  - [x] `TestTimeout_ImmediateReturn` - Verifies immediate response on timeout
  - [x] `TestTimeout_ContextPropagation` - Verifies handler can detect context cancellation
  - [x] `TestTimeout_PanicInHandler` - Verifies panic handling
  - [x] `TestTimeout_PanicAfterTimeout` - Verifies late panics don't affect response
  - [x] All existing tests pass

  Closes #3671
  Closes #3394

### DIFF
--- a/docs/middleware/timeout.md
+++ b/docs/middleware/timeout.md
@@ -19,7 +19,7 @@ handler to finish. This is achieved through Fiber's **Abandon mechanism**:
 4. A background cleanup goroutine waits for the handler to finish and performs context cleanup
 
 Handlers can detect the timeout by listening on `c.Context().Done()` and return early.
-This is the recommended pattern for cooperative cancelation.
+This is the recommended pattern for cooperative cancellation.
 
 If a handler panics, the middleware catches it and returns `500 Internal Server Error`.
 


ReneWerner87:
@copilot review

ReneWerner87:
@coderabbitai review

gaby:
Will check in the morning

ReneWerner87:
WDYT? @gaby @pjebs @edvardsanta

pjebs:
It seems okay (except for grace period requirement) ...but

It doesn't feel right because in FastHttp, they have a timeout middleware implemented correctly.
and it uses: https://pkg.go.dev/github.com/valyala/fasthttp#RequestCtx.TimeoutErrorWithResponse

My suspicion is that that is the key to implementing it properly: https://pkg.go.dev/github.com/valyala/fasthttp#TimeoutHandler

ReneWerner87:
Thx for the hint, i will check this

ReneWerner87:
Thanks for the suggestion @pjebs! I looked into fasthttp's `TimeoutHandler` and `TimeoutErrorWithResponse` in detail togehter with my AI agent. Here's why we can't directly use them to solve the race condition problem:

### How fasthttp's TimeoutHandler works

Looking at the fasthttp source (`server.go:459-491`), `TimeoutWithCodeHandler` does the following:

```go
func TimeoutWithCodeHandler(h RequestHandler, timeout time.Duration, msg string, statusCode int) RequestHandler {
    return func(ctx *RequestCtx) {
        // 1. Run handler in goroutine
        go func() {
            h(ctx)
            ch <- struct{}{}
        }()

        // 2. Wait for handler OR timeout
        select {
        case <-ch:
            // Handler finished
        case <-ctx.timeoutTimer.C:
            ctx.TimeoutErrorWithCode(msg, statusCode)
        }
    }
}
```

The key mechanism is in `serveConn` (line 2468-2473):

```go
timeoutResponse = ctx.timeoutResponse
if timeoutResponse != nil {
    // Acquire a NEW ctx because the old one will still be in use by the handler
    ctx = s.acquireCtx(c)
    timeoutResponse.CopyTo(&ctx.Response)
}
```

fasthttp acquires a **new `RequestCtx`** and lets the old one continue for the still-running handler. This is how it avoids race conditions at the fasthttp level.

### Why this doesn't work for Fiber

The problem is that Fiber has its own `fiber.Ctx` wrapper around `*fasthttp.RequestCtx`:

| Aspect | fasthttp | Fiber |
|--------|----------|-------|
| Context type | `*fasthttp.RequestCtx` | `fiber.Ctx` (wrapper) |
| New ctx on timeout | ✅ fasthttp acquires new RequestCtx | ❌ Fiber's Ctx remains the same |
| Handler reference | Points to old RequestCtx (safe, new one is used for response) | Points to `fiber.Ctx` which wraps the old RequestCtx |

When `TimeoutErrorWithResponse` is called:
1. It only marks the **fasthttp Response** for replacement
2. fasthttp acquires a new `RequestCtx` for the response
3. **But**: The handler still has `fiber.Ctx`, which points to the old `RequestCtx`
4. If the handler accesses `fiber.Ctx` after timeout (`ctx.Locals()`, `ctx.Context()`, `ctx.SetContext()`, etc.), race conditions occur

`TimeoutErrorWithResponse` protects the **fasthttp Response**, not the entire **Fiber context**. The race conditions happen at the Fiber level, not the fasthttp level.

### What would be needed for "true" immediate return

To properly use fasthttp's mechanism, Fiber would need to:
1. Acquire a **new `fiber.Ctx`** on timeout (similar to how fasthttp acquires a new `RequestCtx`)
2. "Abandon" the old `fiber.Ctx` for the handler
3. This requires **Fiber core changes** (context pooling), not just middleware changes

### Our current solution

The current implementation uses:
- `context.WithTimeout` so handlers can detect timeout via `c.Context().Done()`
- Goroutine with panic handling
- Bounded grace period (50-100ms) so hung handlers don't block forever

Handlers that check `c.Context().Done()` can return early and achieve practically the same result as "immediate return" - but in a race-free way.

This is the safest approach without modifying Fiber's core context management.

Let us test the current solutio and if that concept doesn't work as expected, then we'll try an approach that requires us to make core changes.
In that case, I like the independent middleware because it's not tightly integrated.

pjebs:
Why the grace period? If the developer wants a "grace period", they will just increase the timeout value appropriately. The grace period just adds an extra level of uncertainty where you need to look at the code to see what the value is and how it works.

If it's the avoid race conditions - then it's not avoiding race conditions. It's just minimising the probability of race conditions happening. Not eliminating it.

pjebs:
The documentation says `cancellation`. The spelling needs to be consistent. Will Fiber use UK or US spelling:

`cancellation` or `cancelation`?

edvardsanta:
> Thanks for the suggestion @pjebs! I looked into fasthttp's `TimeoutHandler` and `TimeoutErrorWithResponse` in detail togehter with my AI agent. Here's why we can't directly use them to solve the race condition problem:
> ### How fasthttp's TimeoutHandler works
> 
> Looking at the fasthttp source (`server.go:459-491`), `TimeoutWithCodeHandler` does the following:
> 
> ```go
> func TimeoutWithCodeHandler(h RequestHandler, timeout time.Duration, msg string, statusCode int) RequestHandler {
>     return func(ctx *RequestCtx) {
>         // 1. Run handler in goroutine
>         go func() {
>             h(ctx)
>             ch <- struct{}{}
>         }()
> 
>         // 2. Wait for handler OR timeout
>         select {
>         case <-ch:
>             // Handler finished
>         case <-ctx.timeoutTimer.C:
>             ctx.TimeoutErrorWithCode(msg, statusCode)
>         }
>     }
> }
> ```
> 
> The key mechanism is in `serveConn` (line 2468-2473):
> 
> ```go
> timeoutResponse = ctx.timeoutResponse
> if timeoutResponse != nil {
>     // Acquire a NEW ctx because the old one will still be in use by the handler
>     ctx = s.acquireCtx(c)
>     timeoutResponse.CopyTo(&ctx.Response)
> }
> ```
> 
> fasthttp acquires a **new `RequestCtx`** and lets the old one continue for the still-running handler. This is how it avoids race conditions at the fasthttp level.
> ### Why this doesn't work for Fiber
> 
> The problem is that Fiber has its own `fiber.Ctx` wrapper around `*fasthttp.RequestCtx`:
> Aspect 	fasthttp 	Fiber
> Context type 	`*fasthttp.RequestCtx` 	`fiber.Ctx` (wrapper)
> New ctx on timeout 	✅ fasthttp acquires new RequestCtx 	❌ Fiber's Ctx remains the same
> Handler reference 	Points to old RequestCtx (safe, new one is used for response) 	Points to `fiber.Ctx` which wraps the old RequestCtx
> 
> When `TimeoutErrorWithResponse` is called:
> 
>     1. It only marks the **fasthttp Response** for replacement
> 
>     2. fasthttp acquires a new `RequestCtx` for the response
> 
>     3. **But**: The handler still has `fiber.Ctx`, which points to the old `RequestCtx`
> 
>     4. If the handler accesses `fiber.Ctx` after timeout (`ctx.Locals()`, `ctx.Context()`, `ctx.SetContext()`, etc.), race conditions occur
> 
> 
> `TimeoutErrorWithResponse` protects the **fasthttp Response**, not the entire **Fiber context**. The race conditions happen at the Fiber level, not the fasthttp level.
> ### What would be needed for "true" immediate return
> 
> To properly use fasthttp's mechanism, Fiber would need to:
> 
>     1. Acquire a **new `fiber.Ctx`** on timeout (similar to how fasthttp acquires a new `RequestCtx`)
> 
>     2. "Abandon" the old `fiber.Ctx` for the handler
> 
>     3. This requires **Fiber core changes** (context pooling), not just middleware changes
> 
> 
> ### Our current solution
> 
> The current implementation uses:
> 
>     * `context.WithTimeout` so handlers can detect timeout via `c.Context().Done()`
> 
>     * Goroutine with panic handling
> 
>     * Bounded grace period (50-100ms) so hung handlers don't block forever
> 
> 
> Handlers that check `c.Context().Done()` can return early and achieve practically the same result as "immediate return" - but in a race-free way.
> 
> This is the safest approach without modifying Fiber's core context management.
> 
> Let us test the current solutio and if that concept doesn't work as expected, then we'll try an approach that requires us to make core changes. In that case, I like the independent middleware because it's not tightly integrated.

Specifically, the panic-related test cases were problematic. In scenarios where the handler panics after the timeout or where the OnTimeout handler itself panics, I observed concurrency-related issues that I couldn’t safely resolve without deeper changes. These cases were especially tricky due to goroutine lifecycle and race conditions around timeout handling.


```go
func TestTimeout_Issue_3671(t *testing.T) {
	t.Parallel()
	app := fiber.New()
	testCases := []struct {
		name       string
		path       string
		handler    fiber.Handler
		config     Config
		expectCode int
	}{
		{
			name: "Handler panics after timeout",
			path: "/panic-after-timeout",
			handler: func(_ fiber.Ctx) error {
				time.Sleep(50 * time.Millisecond)
				panic("panic after timeout")
			},
			config:     Config{Timeout: 10 * time.Millisecond},
			expectCode: fiber.StatusRequestTimeout,
		},
		{
			name: "Handler blocks forever",
			path: "/block-forever",
			handler: func(_ fiber.Ctx) error {
				select {} // Block forever
			},
			config:     Config{Timeout: 10 * time.Millisecond},
			expectCode: fiber.StatusRequestTimeout,
		},
		{
			name: "Timeout set to 1 nanosecond",
			path: "/nano",
			handler: func(c fiber.Ctx) error {
				time.Sleep(1 * time.Millisecond)
				return c.SendString("late")
			},
			config:     Config{Timeout: 1 * time.Nanosecond},
			expectCode: fiber.StatusRequestTimeout,
		},
		{
			name: "Timeout set to a very large value",
			path: "/maxint",
			handler: func(c fiber.Ctx) error {
				return c.SendString("ok")
			},
			config:     Config{Timeout: 1<<63 - 1},
			expectCode: fiber.StatusOK,
		},
		{
			name: "Custom OnTimeout handler panics",
			path: "/panic-ontimeout",
			handler: func(_ fiber.Ctx) error {
				time.Sleep(50 * time.Millisecond)
				return nil
			},
			config: Config{
				Timeout: 10 * time.Millisecond,
				OnTimeout: func(_ fiber.Ctx) error {
					panic("custom panic on timeout")
				},
			},
			expectCode: fiber.StatusRequestTimeout,
		},
		{
			name: "Custom OnTimeout",
			path: "/ontimeout",
			handler: func(c fiber.Ctx) error {
				if err := sleepWithContext(c.Context(), 100*time.Millisecond, context.DeadlineExceeded); err != nil {
					return err
				}
				return c.SendString("should not reach")
			},
			config: Config{
				Timeout: 20 * time.Millisecond,
				OnTimeout: func(c fiber.Ctx) error {
					return c.SendString("custom timeout response")
				},
			},
			expectCode: fiber.StatusRequestTimeout,
		},
	}

	for _, tc := range testCases {
		app.Get(tc.path, New(tc.handler, tc.config))
	}

	for _, tc := range testCases {
		t.Run(tc.name, func(t *testing.T) {
			req := httptest.NewRequest(fiber.MethodGet, tc.path, nil)
			resp, err := app.Test(req)

			require.NoError(t, err)
			require.Equal(t, tc.expectCode, resp.StatusCode)
		})
	}
}
```

ReneWerner87:
@pjebs You're right - the fixed grace period was logically inconsistent. We've changed the approach:

  New behavior:
  - GracePeriod: 0 (default): Middleware waits indefinitely for the handler to finish → race-free
  - GracePeriod > 0: Middleware waits at most GracePeriod after timeout → faster response, but potential races

  This gives users full control. The default is now the safest option (always wait), and users who want faster responses can
  explicitly opt-in to the grace period with full awareness of the trade-offs.

  Regarding spelling: I changed the places to US

---
  @edvardsanta Thanks for testing and identifying the concurrency issues with panic scenarios!

  The current implementation handles panics by:
  1. Catching them in the handler goroutine via recover()
  2. Returning 500 Internal Server Error
  3. By default (GracePeriod: 0), waiting for the handler to complete - this avoids races even when panics occur after timeout

  The tests TestTimeout_PanicInHandler and TestTimeout_PanicAfterTimeout verify this behavior. If you're still seeing issues with specific scenarios, please share the details and we can investigate further.

pjebs:
I think Fiber internals needs to be rewritten a bit. FastHTTP's timeout middleware doesn't have any race conditions.
According to your comment, the developer has to select a value for the grace period, where they are tolerating a certain probability that they will encounter concurrency related bugs.

ReneWerner87:
@pjebs @edvardsanta 
https://github.com/gofiber/fiber/pull/4009/changes/a8da7a0f2cf4f7c28760fba102ea9c597c2d547a

I have now expanded our core and implemented a race-free approach similar to the fasthttp approach.
I couldn't use the fasthttp approach, otherwise I would have had to make a lot more adjustments, and our fiber approach with the new methods is now also more flexible.

pjebs:
Maybe you can unexport the Abandon method you just created, and call it using unsafe or reflect. 
You may need to convert the method to an unexported function to do that approach.

Others don't need to know about it. The API surface and documentation is already bloated.

ReneWerner87:
> Maybe you can unexport the Abandon method you just created, and call it using unsafe or reflect. You may need to convert the method to an unexported function to do that approach.
> 
> Others don't need to know about it. The API surface and documentation is already bloated.

its okay like it is

PR #4011, opened 2026-01-16
🐛 bug: Normalize forwarded scheme parsing

gaby:
### Motivation

- Ensure forwarded-scheme headers are parsed robustly regardless of header-name case and optional whitespace, and avoid returning padded or uppercase scheme strings.

### Description

- Update `DefaultReq.Scheme` in `req.go` to compare header names case-insensitively using `utils.EqualFold` when checking `X-Forwarded-*`, `X-Forwarded-Ssl` and `X-Url-Scheme` headers.
- Trim optional whitespace around the first comma-separated value from `X-Forwarded-Proto` / `X-Forwarded-Protocol` using `utils.TrimSpace` before assigning the scheme.
- Normalize the returned scheme by trimming and lowercasing with `utils.TrimSpace` + `utils.ToLower` to avoid returning values like `" HTTPS "` or `"HTTPS"`.
- Add unit test `Test_Ctx_Scheme_HeaderNormalization` in `ctx_test.go` to cover case-insensitive header names and whitespace/comma trimming for forwarded scheme parsing.

### DIFF
--- a/ctx_test.go
+++ b/ctx_test.go
@@ -3731,11 +3731,7 @@ func Test_Ctx_Scheme_HeaderNormalization(t *testing.T) {
 	require.Equal(t, schemeHTTPS, c.Scheme())
 	c.Request().Header.Reset()
 
-	c.Request().Header.Set("X-FORWARDED-SSL", " HTTPS ")
-	require.Equal(t, schemeHTTPS, c.Scheme())
-	c.Request().Header.Reset()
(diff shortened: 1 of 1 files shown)


ReneWerner87:
@gaby can you check the ai hints before i merge

gaby:
@ReneWerner87 This one is READY

PR #4014, opened 2026-01-17
🔥 feat: Add support for streaming response bodies in client and response handling

ReneWerner87:
Adds client-level streaming controls and response streaming consumption: Client getter/setter for streaming, transport plumbing to propagate the flag across transport types, Response APIs to read/inspect streaming bodies, and core handling changes to preserve streaming during response handling.

Replace #3711
Solves #3425

### DIFF
--- a/docs/client/response.md
+++ b/docs/client/response.md
@@ -97,7 +97,7 @@ func (r *Response) Header(key string) string
 **Headers** returns an iterator over all response headers. Use `maps.Collect()` to convert them into a map if desired. The returned values are only valid until the response is released, so make copies if needed.
 
 ```go title="Signature"
-func (r *Response) Headers() iter.Seq2[string, []string] 
+func (r *Response) Headers() iter.Seq2[string, []string]
 ```
 
 <details>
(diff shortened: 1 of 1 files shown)


ReneWerner87:
@gaby @grivera64 can you check

docs/client/response.md
```diff
@@ -190,6 +190,91 @@ go => fiber
...
+if err != nil {
+    panic(err)
+}
+defer resp.Close()
+
+reader := resp.BodyStream()
+buf := make([]byte, 256)
+var total int
+
+for {
```
gaby: This example is too complicated. This can be done in one line using io.Copy or io.CopyBuffer.

```go
buf := make([]byte, 1024*1024)
io.CopyBuffer(struct{ io.Writer }{dst}, struct{ io.Reader }{src}, buf)
```

gaby (approved):
LGTM overall, just one small thing in the docs.

We could also use io.CopyBuffer with a bytesbufferpool in some places.

docs/client/response.md
ReneWerner87: https://github.com/gofiber/fiber/pull/4014/changes/5658014988ae280f14919d2cceaf46b260a0f2af

PR #4024, opened 2026-01-21
🔥 feat: Add TLSConfig to ListenConfig

gaby:
### Description
- This pull request significantly enhances the TLS configuration options by introducing a dedicated TLSConfig field to the ListenConfig struct. This change provides greater flexibility for users to supply TLS certificates from various sources, such as in-memory PEM data or environment variables, and integrates seamlessly with external certificate providers. The update also includes robust validation to prevent incompatible TLS settings and clear documentation with practical examples to guide users. The implementation correctly handles precedence, ensuring that TLSConfig takes priority over other TLS-related fields, and includes comprehensive test cases to verify the new functionality and error conditions.

Replaces #4019 and #4021

### DIFF
--- a/docs/api/fiber.md
+++ b/docs/api/fiber.md
@@ -174,12 +174,6 @@ app.Listen(":443", fiber.ListenConfig{
 })
 ```
 
-#### TLS with certFile, keyFile and clientCertFile
-
-```go title="TLS with cert, key, and client CA"
-app.Listen(":443", fiber.ListenConfig{CertFile: "./cert.pem", CertKeyFile: "./cert.key", CertClientFile: "./ca-chain-cert.pem"})
-```
-
 #### TLS AutoCert support (ACME / Let's Encrypt)
 
 Provides automatic access to certificates management from Let's Encrypt and any other ACME-based providers.


simonnix:
Hi @ReneWerner87 , @gaby .

I saw you closed my PR and I did not add the time to reply properly.

I will add my 2 cents on this PR which could be negative, but I assure you it is in no way in bitterness or for "revenge" or anything like that.

--

I came to this two interfaces system for configuring a `*tls.Config` just after realizing I was starting to add too much code inside `Listen()`. And it's showing here...

This PR gives me all I need really, I can provide my `*tls.Config`, keep my little two interfaces system private. I am happy.

But at the same time why keep the ACME code in this case in `Listen()` ? Or why it's added when other methods are not ?

That two interfaces (specially `ServerTLSProvider`) system could have been extended like you have done with the `Storage` providers or with this `Extractor` interface.

So I have mixed feelings. It gives me what I want but it's not as good as it could have been.

It's ok, I guess.

ReneWerner87:
Thank you for your feedback.

We kept the old code in the `app.Listen` because these are the most common cases.

Unfortunately, we have now realized that we should not build the TLS config internally
but should allow users to pass it on from outside

Since this is an RC version and removing the old methods would be a breaking change, we have left them in place and will perhaps remove them in v4 or reconsider, as these are very frequently used cases

The rest can be mapped by passing on the `tls.Config` and should now have all possibilities.

gaby:
@simonnix We plan on using TLSConfig as provided, else we will use existing fields. I just removed the logic to merge them since that was adding complexity.

simonnix:
Hi @gaby,

I'm fine with that, having `TLSConfig` is really more than I hoped to have.

Just thinking about it (I'm not an Express.js user, I just looked it up) but, are you moving to just offer `TLSConfig` and nothing else, just to align more with Express.js which does not handle directly TLS ?

 Is that your idea? I'm fine with it, just asking.

gaby:
> Hi @gaby,
> 
> I'm fine with that, having `TLSConfig` is really more than I hoped to have.
> 
> Just thinking about it (I'm not an Express.js user, I just looked it up) but, are you moving to just offer `TLSConfig` and nothing else, just to align more with Express.js which does not handle directly TLS ?
> 
> Is that your idea? I'm fine with it, just asking.

@simonnix Priority is TLSConfig, but we are leaving the other fields since we are in `Release Candidate` status already.

Edit: net/http server uses TLSConfig too.

PR #4031, opened 2026-01-25
🐛 bug: Reload mounted view engines in ReloadViews

### LINKED ISSUE #3875: Reload views?

gaby:
### Motivation
- Ensure `ReloadViews()` reloads view engines for mounted sub-apps so on-disk template changes are picked up when invoked on the parent application.
- Prevent panics by skipping `nil` and typed-nil view engine values when reloading.
- Keep lint expectations by returning mount helpers directly from `Use` helpers without changing runtime behavior.

### Description
- Updated `App.ReloadViews()` to iterate mounted apps (`app.mountFields.appList` when present) and call `Load()` on each mounted app's `config.Views`, skipping `nil` and typed-nil engines and returning `ErrNoViewEngineConfigured` only if nothing was reloaded.
- Preserved the typed-nil guard by checking `reflect.ValueOf(...).Kind() == reflect.Pointer && IsNil()` before calling `Load()` and wrap load errors with a descriptive message.
- Changed `App.Use` and `Group.Use` to return the result of `mount(...)` directly to satisfy lint expectations without changing runtime behavior.
- Moved the `fileView` test helper to the top of `app_test.go` and added `Test_App_ReloadViews_MountedViews` to verify that a mounted sub-app's view engine is reloaded when the parent `ReloadViews()` is invoked.

Fixes #3875

### DIFF
--- a/mount.go
+++ b/mount.go
@@ -45,6 +45,7 @@ func (app *App) mount(prefix string, subApp *App) Router {
 		prefix = "/"
 	}
 
+	app.mutex.Lock()
 	// Support for configs of mounted-apps and sub-mounted-apps
 	for mountedPrefixes, subApp := range subApp.mountFields.appList {
 		path := getGroupPath(prefix, mountedPrefixes)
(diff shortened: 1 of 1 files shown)


gaby:
Fixing AI review

PR #4038, opened 2026-01-28
feat: ShutdownTelemetry, debug endpoint, goroutine-leak tests, and deployment docs

zishanjawed:
## Summary

- **`ShutdownTelemetry` struct** — captures per-phase wall-clock timing (`PreHooksDuration`, `GracefulCloseDuration`, `DrainDuration`, `PostHooksDuration`, `TotalDuration`) and connection-disposition counts (`InitialConns`, `DrainedConns`, `ForcedConns`, `WebSocketsClosed`, `SSEsClosed`, `TimedOut`).
- **`app.LastShutdownTelemetry()`** — lock-free accessor via `atomic.Pointer`; returns `nil` before the first `ShutdownWithConfig` completes, or a fully-populated snapshot afterward.
- **`app.ShutdownDebugHandler()`** — returns a `Handler` that responds with JSON containing `status` (`running` / `shutting_down` / `shutdown`), `activeConnections`, and the last telemetry snapshot with human-readable duration strings.  Register at any path: `app.Get("/debug/shutdown", app.ShutdownDebugHandler())`.
- **`GracefulCloseTyped`** — extended to accept `*ShutdownTelemetry` and increment `WebSocketsClosed` / `SSEsClosed` counters sequentially inside the `sync.Map.Range` (no additional synchronization needed).

## Tests (9 new)

| # | Test | Verifies |
|---|------|----------|
| 1 | `Test_Telemetry_CleanDrain_PopulatesAllFields` | Clean shutdown populates all fields; `DrainedConns == InitialConns`; `ForcedConns == 0`; `TimedOut == false` |
| 2 | `Test_Telemetry_ForcedClose_CountsCorrectly` | Slow handler + short deadline; `ForcedConns >= 1`; `TimedOut == true` |
| 3 | `Test_Telemetry_HookDurations_Measured` | Pre-hook sleeps 50 ms; `PreHooksDuration >= 50 ms` |
| 4 | `Test_Telemetry_WebSocketSSE_Counters` | 1 WS + 1 SSE marked; after shutdown `WebSocketsClosed == 1`, `SSEsClosed == 1` |
| 5 | `Test_DebugEndpoint_ReturnsRunningStatus` | Before shutdown: `status == "running"`, `lastShutdown` is `null` |
| 6 | `Test_DebugEndpoint_ReturnsTelemetryAfterShutdown` | After shutdown: `status == "shutdown"`, all duration strings parseable by `time.ParseDuration` |
[...]

PR #4041, opened 2026-01-31
feat: tag invalidation, conditional requests, and Cache-Control for cache middleware

zishanjawed:
## Summary

- Tag-based cache invalidation with in-memory and distributed (shared-storage) backends
- RFC 9110 conditional request validation (ETag / Last-Modified / 304)
- Automatic Cache-Control header generation with handler-override preservation
- Full test coverage including concurrent and cross-instance scenarios
- Benchmarks for all tag operations and encoding paths
- README documenting tag strategies, invalidation, conditional requests, and Cache-Control

## Test plan

- [ ] `go test -race ./middleware/cache/` passes clean
- [ ] Tag association and invalidation via `InvalidateTags` works end-to-end
- [ ] Cross-instance invalidation: two `distributedTagStore` instances sharing one storage backend; invalidation on one finds keys written by the other
- [ ] Conditional 304 responses: ETag auto-generation, custom generator, If-None-Match; Last-Modified auto-generation, custom generator, If-Modified-Since
- [ ] RFC 9110 §8.3 precedence: If-None-Match takes priority when both conditional headers are present
- [ ] Cache-Control: middleware-generated header on miss and hit; handler-set header preserved verbatim; DisableCacheControl suppresses generation
- [ ] RejectTags prevents storage for matching tag patterns (exact, prefix, general glob)
- [ ] Benchmarks: `has()` is O(1) across index sizes; distributed `has()` matches in-memory cost; invalidation scales linearly with keys per tag

### DIFF
--- a/middleware/cache/README.md
+++ b/middleware/cache/README.md
@@ -1,0 +1,334 @@
+# Cache Middleware
+
+HTTP response caching middleware for Fiber. Stores responses keyed by request
+path (or a custom key), serves them on subsequent matching requests, and
+provides tag-based invalidation, conditional request validation, and
+Cache-Control header management.
+
+---
+
+## Tag Strategies
+
+Tags associate cache entries with logical groups. Entries sharing a tag can be
+invalidated together, making it possible to flush related content (e.g. all
+pages for a given user or product) with a single call.
+
+### Associating Tags with Entries
+
+Two configuration hooks control tag assignment:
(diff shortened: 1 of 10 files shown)


ReneWerner87:
same as https://github.com/gofiber/fiber/pull/4038
We are still waiting for communication with the author.

PR #4065, opened 2026-02-08
🐛 bug: apply limits to msgp serialization

gaby:
### Motivation
- Prevent msgp limit tags from unnecessarily constraining cache and idempotency payloads whose sizes are governed by storage policy rather than serializer limits.
- Keep per-field caps for headers and other small fields to protect against unbounded memory usage during deserialization.
- Regenerate msgp outputs so runtime decoding enforces the intended limits and fails fast on malicious inputs.

### Description
- Removed msgp body size caps for cache `item.body` and idempotency `response.Body`, deferring body-size enforcement to storage policy while retaining field-level limits for other fields. 
- Increased header/map array caps (for example `Headers` and header counts) to `1024` where appropriate and tightened per-field byte limits (for `ctype`, `cencoding`, `cacheControl`, `expires`, `etag`, and header key/value sizes).
- Regenerated the msgp-generated files (`*_msgp.go`) so they include size-constant checks and use `ReadBytesLimit`/`ReadBytesHeader`/`ReadBytes` variants and explicit `msgp.ErrLimitExceeded` checks.
- Added `//nolint:revive` annotations where msgp tags on unexported fields require linter exceptions.


ReneWerner87 (approved):

ReneWerner87:
@gaby can you check ai comments

gaby:
@copilot Should we add msgp limits anywhere else?

PR #4068, opened 2026-02-08
fix: add FromStdContext to retrieve request ID from context.Context

### LINKED ISSUE #4066: 🐛 [Bug]: No way of accessing requestid value other than from `fiber.Ctx`
### Bug Description

In v2 requestid middleware could save its value to custom contextKey which I could use to retrieve value from context. In v3 the only way to do this is using `requestid.FromContext` which accepts only `fiber.Ctx`.
This make it pretty hard to retrieve this value down the callstack.

### How to Reproduce

1. Apply requestid middleware
2. Pass `fiber.Ctx` or `(fiber.Ctx).RequestCtx()` as `context.Context` down the stack
3. Try to retrieve requestid value
4. You can only do this by type assertion if `fiber.Ctx` was passed

### Expected Behavior

I want to retrieve requestid value any `context.Context` implementation is case it was wrapped.
There should be either helper function for it or contextKey that is exported or set by user.

### Fiber Version

v3.0.0

### Code Snippet (optional)

```go
package main

import "github.com/gofiber/fiber/v3/middleware/requestid"
import "log"

func service(ctx context.Context) {
  // how to retrieve it here?
}

func main() {
  app := fiber.New()
  app.Use(requestid.New())

  app.Get("/test", func(c fiber.Ctx) error {
      service(c.RequestCtx())
      return c.SendStatus(200)
  })
}
```

### Checklist:

veeceey:
## Summary
- Add `FromStdContext(ctx context.Context) string` function to the requestid middleware, allowing retrieval of the request ID from any standard `context.Context`
- Store the request ID in the user context (via `context.WithValue`) in addition to Fiber locals, so it propagates through `c.Context()` and wrapped contexts
- Add tests for `FromStdContext` including empty context and wrapped context scenarios

Fixes #4066

## Test plan
- All existing requestid middleware tests continue to pass
- New `Test_RequestID_FromStdContext` verifies retrieval via `c.Context()`
- New `Test_RequestID_FromStdContext_Empty` verifies empty string on plain `context.Background()`
- New `Test_RequestID_FromStdContext_WrappedContext` verifies retrieval still works when the context is further wrapped with `context.WithValue`
- Full test suite passes (only pre-existing `client/Test_Client_SetProxyURL/error` failure unrelated to this change)

## Usage example
\`\`\`go
import "github.com/gofiber/fiber/v3/middleware/requestid"

func service(ctx context.Context) {
    rid := requestid.FromStdContext(ctx) // works!
    fmt.Println("Request ID:", rid)
}

app.Get("/test", func(c fiber.Ctx) error {
    service(c.Context()) // pass standard context.Context
    return c.SendStatus(200)
})
\`\`\`

### DIFF
--- a/middleware/requestid/requestid.go
+++ b/middleware/requestid/requestid.go
@@ -1,6 +1,8 @@
 package requestid
 
 import (
+	"context"
+
 	"github.com/gofiber/fiber/v3"
 	"github.com/gofiber/utils/v2"
 )
@@ -32,6 +34,12 @@ func New(config ...Config) fiber.Handler {
 
 		// Add the request ID to locals
 		c.Locals(requestIDKey, rid)
+
+		// Also store the request ID in the user context so it can be
+		// retrieved from any context.Context (e.g. c.Context() or
+		// c.RequestCtx()) further down the call stack.
+		ctx := context.WithValue(c.Context(), requestIDKey, rid)
+		c.SetContext(ctx)
 
 		// Continue stack
 		return c.Next()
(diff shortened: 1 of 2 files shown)


gaby:
@veeceey Please discuss in the issue first, we are still investigation internally.

veeceey:
Thanks @gaby, understood\! I'll follow up in #4066 and wait for the team's decision before making further changes here. Happy to adjust the approach based on your internal discussion.

veeceey:
Understood @gaby, will hold off and wait for the team's direction. I've also commented on #4066. Happy to revise the approach once you've settled on the preferred solution.

gaby:
@veeceey Thanks, we integrated your changes in #4079

PR #4070, opened 2026-02-09
🐛 bug: sanitize attachment/download filenames

gaby:
### Motivation
- Prevent control characters (CR/LF, TAB, NUL) from appearing in the `Content-Disposition` filename parameter to avoid header injection and unsafe/ambiguous client dialogs.
- Ensure a safe, non-empty default filename is used when sanitization removes all characters.
- Expand coverage for attachment/download behavior to verify handling of spaces, nested paths, and control characters.

### Description
- Added `sanitizeFilename` in `res.go` which strips `\r`, `\n`, `\t`, and NUL using `strings.Map` and applied it to `DefaultRes.Attachment` and `DefaultRes.Download` with a fallback to `"download"` when the sanitized name is empty.
- Adjusted `Attachment`/`Download` logic to sanitize the base filename before determining `Content-Type` and building the `Content-Disposition` header, while preserving ASCII vs. non-ASCII quoting and `filename*` behavior.
- Regenerated small doc comment changes in `ctx_interface_gen.go` (minor interface comment updates) to match current generation output.
- Expanded tests in `ctx_test.go`: added cases to `Test_Ctx_Attachment` (spaces and nested paths), extended `Test_Ctx_Attachment_SanitizesFilenameControls` (additional control-placement cases), and added `Test_Ctx_Download_SanitizesFilenameControls` to assert sanitized headers and absence of raw control bytes.

### DIFF
--- a/ctx_test.go
+++ b/ctx_test.go
@@ -737,12 +737,17 @@ func Test_Ctx_Attachment_SanitizesFilenameControls(t *testing.T) {
 		{
 			name:     "empty after sanitize",
 			filename: "\r\n\t\x00",
-			expected: `attachment; filename=""`,
+			expected: `attachment; filename="download"`,
 		},
 		{
 			name:     "controls stripped in middle",
 			filename: "file\rname\n\t\x00.bin",
 			expected: `attachment; filename="filename.bin"`,
+		},
+		{
+			name:     "dot fallback",
+			filename: ".",
+			expected: `attachment; filename="download"`,
 		},
 	}
 
@@ -5107,7 +5112,12 @@ func Test_Ctx_Download_SanitizesFilenameControls(t *testing.T) {
 		{
 			name:     "empty after sanitize",
 			filename: "\r\n\t\x00",
-			expected: `attachment; filename=""`,
+			expected: `attachment; filename="download"`,
+		},
+		{
+			name:     "dot fallback",
+			filename: ".",
+			expected: `attachment; filename="download"`,
 		},
 	}
 
--- a/res.go
+++ b/res.go
@@ -195,11 +195,20 @@ func sanitizeFilename(filename string) string {
 	return utils.TrimSpace(filename)
 }
 
+func fallbackFilenameIfInvalid(filename string) string {
+	if filename == "" || filename == "." {
+		return "download"
+	}
+
+	return filename
+}
+
 // Attachment sets the HTTP response Content-Disposition header field to attachment.
 func (r *DefaultRes) Attachment(filename ...string) {
 	if len(filename) > 0 {
 		fname := filepath.Base(filename[0])
 		fname = sanitizeFilename(fname)
+		fname = fallbackFilenameIfInvalid(fname)
 		r.Type(filepath.Ext(fname))
 		app := r.c.app
 		var quoted string
@@ -339,6 +348,7 @@ func (r *DefaultRes) Download(file string, filename ...string) error {
 		fname = filepath.Base(file)
 	}
 	fname = sanitizeFilename(fname)
+	fname = fallbackFilenameIfInvalid(fname)
 	app := r.c.app
 	var quoted string
 	if app.isASCII(fname) {


ctx_test.go
```diff
@@ -5028,6 +5085,47 @@ func Test_Ctx_Download(t *testing.T) {
...
+		},
+		{
+			name:     "controls stripped",
+			filename: "down\r\nload\t\x00.txt",
+			expected: `attachment; filename="download.txt"`,
+		},
+		{
+			name:     "empty fallback",
+			filename: "\r\n\t\x00",
+			expected: `attachment; filename="download"`,
```
efectn, line 5115: what is the purpose of default download filename

gaby: Removing, basically a default value

ReneWerner87 (approved):

res.go
```diff
@@ -177,10 +178,25 @@ func headerContainsValue(header, value string) bool {
 	return false
 }
 
+func sanitizeFilename(filename string) string {
+	sanitized := strings.Map(func(r rune) rune {
```
ReneWerner87:
```go
/// StripControlChars removes ASCII control characters (0x00-0x1F, 0x7F) from a string.
// Zero allocation when no control characters are present.
func StripControlChars(s string) string {
	for i := 0; i < len(s); i++ {
		c := s[i]
		if c < 0x20 || c == 0x7F {
			return stripControlCharsSlow(s, i)
		}
	}
	return s
}

func stripControlCharsSlow(s string, firstBad int) string {
	buf := make([]byte, 0, len(s))
	buf = append(buf, s[:firstBad]...)
	for i := firstBad; i < len(s); i++ {
		c := s[i]
		if c >= 0x20 && c != 0x7F {
			buf = append(buf, c)
		}
	}
	return string(buf)
}
```

ReneWerner87: applied in https://github.com/gofiber/fiber/pull/4070/commits/384163d06b9a7f12b8c8cceeccf74d3b6df7f0f4

ctx_test.go
ReneWerner87: removed in [`384163d` (this PR)](https://github.com/gofiber/fiber/pull/4070/commits/384163d06b9a7f12b8c8cceeccf74d3b6df7f0f4)

res.go
```diff
@@ -315,10 +334,11 @@ func (r *DefaultRes) Cookie(cookie *Cookie) {
 func (r *DefaultRes) Download(file string, filename ...string) error {
 	var fname string
 	if len(filename) > 0 {
-		fname = filename[0]
+		fname = filepath.Base(filename[0])
 	} else {
 		fname = filepath.Base(file)
 	}
+	fname = sanitizeFilename(fname)
```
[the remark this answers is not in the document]
ReneWerner87: Ai wants to have a default 
@efectn should we keep the default ?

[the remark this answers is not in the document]
ReneWerner87: @gaby WDYT?

res.go
ReneWerner87:
```
no_controls: current ~12 ns vs old strings.Map ~25–28 ns
controls_present: current ~37 ns vs old strings.Map ~55 ns
```

res.go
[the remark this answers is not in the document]
gaby: @ReneWerner87 Yeah, it just makes things more consistent.

Plus avoiding the change of an empty filename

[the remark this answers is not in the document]
ReneWerner87: https://github.com/gofiber/fiber/pull/4070/changes/d224310c26c6a24fe9a62136bc43dee835e40f53

ReneWerner87:
> # ⚠️ **Performance Alert** ⚠️
> Possible performance regression was detected for benchmark. Benchmark result of this commit is worse than the previous benchmark result exceeding threshold `1.50`.
> 
> Benchmark suite	Current: [d224310](https://github.com/gofiber/fiber/commit/d224310c26c6a24fe9a62136bc43dee835e40f53)	Previous: [7b92748](https://github.com/gofiber/fiber/commit/7b92748112489144aa5a95bf572f2cad76ce0647)	Ratio
> `Benchmark_Ctx_Attachment - B/op`	`168` B/op	`104` B/op	`1.62`
> `Benchmark_Ctx_Attachment - allocs/op`	`4` allocs/op	`2` allocs/op	`2`
> This comment was automatically generated by [workflow](https://github.com/gofiber/fiber/actions?query=workflow%3ABenchmark) using [github-action-benchmark](https://github.com/marketplace/actions/continuous-benchmark).

yes we know, but its the same speed -> so okay in terms of security / speed

PR #4071, opened 2026-02-09
🐛 bug: enforce Range header limit configuration

gaby:
### Motivation
- Prevent resource exhaustion from very large `Range` headers by capping the number of parsed ranges and make the configuration clearer and consistent across initialization and parsing.

### Description
- Introduce `Config.MaxRanges` with `DefaultMaxRanges = 16` and apply the default in `New()` when the configured value is non-positive.
- Enforce the configured maximum in `DefaultReq.Range()` by counting parsed ranges and returning `ErrRangeTooLarge` while setting status `416` and `Content-Range: bytes */<size>` when exceeded.
- Add `ErrRangeTooLarge`, update tests to cover the over-limit case, and document `MaxRanges` in `docs/api/fiber.md`.
- Remove the redundant fallback check in `DefaultReq.Range()` so the code relies on the default applied in `App.New()`.

### DIFF
--- a/req.go
+++ b/req.go
@@ -983,7 +983,7 @@ func (r *DefaultReq) Range(size int64) (Range, error) {
 	}
 	if len(rangeData.Ranges) < 1 {
 		r.c.DefaultRes.Status(StatusRequestedRangeNotSatisfiable)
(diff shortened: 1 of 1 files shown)


ReneWerner87:
image
but allocs and speed is the same -> its ok

PR #4078, opened 2026-02-09
🐛 bug: harden flash cookie detection

gaby:
### Motivation
- Avoid unsafe raw-header substring checks for flash cookie detection to prevent false positives from similarly prefixed cookie names. 
- Preserve the original short-circuit on empty raw headers to keep the hot path fast and avoid unnecessary cookie parsing.

### Description
- Replace `bytes.Contains(rawHeaders, ...)` with `hasFlashCookie(header *fasthttp.RequestHeader)` that uses `header.Cookie(FlashCookieName)` for exact cookie lookup. 
- Reintroduced the `len(rawHeaders) > 0` short-circuit before invoking the exact cookie check in both request handling branches. 
- Add `Test_hasFlashCookieExactMatch` to ensure `fiber_flashX` does not trigger detection while `fiber_flash` does. 
- Regenerated minor context interface docs in `ctx_interface_gen.go` clarifying `Request/Response/Value` behavior when a context is released.

### DIFF
--- a/router.go
+++ b/router.go
@@ -323,9 +323,8 @@ func (app *App) requestHandler(rctx *fasthttp.RequestCtx) {
 			return
 		}
 
-		// Optional: Check flash messages
-		rawHeaders := d.Request().Header.RawHeaders()
-		if len(rawHeaders) > 0 && hasFlashCookie(&d.Request().Header) {
+		// Optional: check flash messages (hot path, see hasFlashCookie).
+		if hasFlashCookie(&d.Request().Header) {
 			d.Redirect().parseAndClearFlashMessages()
 		}
 		_, err = app.next(d)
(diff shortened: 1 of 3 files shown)


router.go
```diff
@@ -330,7 +325,7 @@ func (app *App) requestHandler(rctx *fasthttp.RequestCtx) {
 
 		// Optional: Check flash messages
 		rawHeaders := d.Request().Header.RawHeaders()
-		if len(rawHeaders) > 0 && bytes.Contains(rawHeaders, flashCookieNameBytes) {
+		if len(rawHeaders) > 0 && hasFlashCookie(&d.Request().Header) {
```
efectn: Does this change have any performance impact

ReneWerner87: That's right, this is a hotpath that we should pay attention to because the check is performed everywhere.
Therefore, we also used rawHeaders, as this was the fastest method and the headers and cookies did not need to be parsed first.

ReneWerner87 (changes requested):
pls check the performance for this change

router.go
ReneWerner87: solved with the last commit

PR #4079, opened 2026-02-09
🔥 feat: expand middleware context helpers

gaby:
### Motivation
- Broaden middleware helper lookups so helpers can read stored values from multiple context types (`fiber.Ctx`, `*fasthttp.RequestCtx`, and `context.Context`) without panicking when an unexpected type is passed.
- Ensure middleware writes values into both Fiber locals and the Go `context.Context` to keep lookups consistent across APIs and compatibility with code that uses plain `context.Context` or raw fasthttp contexts.
- Update documentation and tests to reflect the new helper signatures and verify the three supported context types are handled.

### Description
- Changed helper signatures in middleware to accept `ctx any` and implemented a safe type switch supporting `fiber.Ctx`, `*fasthttp.RequestCtx`, and `context.Context` for `requestid.FromContext`, `session.FromContext`, `csrf.TokenFromContext`/`HandlerFromContext`, `basicauth.UsernameFromContext`, and `keyauth.TokenFromContext`.
- Mirrored values into both locals and the Go context inside middleware handlers by keeping `c.Locals(key, val)` and adding `c.SetContext(context.WithValue(c.Context(), key, val))` where values are set.
- Updated docs under `docs/middleware/*.md` to document the new `any`-typed helper signatures and added/adjusted tests to cover lookups from `fiber.Ctx`, `*fasthttp.RequestCtx`, and `context.Context`.
- Minor comment updates to `ctx_interface_gen.go` regarding `Value`/request release behavior and other small generated doc tweaks.

Fixes: #4066 
Replaces: #4068


gaby:
Note to myself: Add support for CustomCtx

sixcolors:
> ## Follow-up: All middleware keys are unexported - helpers required anyway
> Just realized something important: **all middleware use unexported context keys**, which means external services can't access the values directly even if they're stored in `context.Context`.
> 
> ```go
> // All these are unexported (lowercase or private types)
> requestid:   requestIDKey contextKey = iota
> session:     middlewareContextKey middlewareKey = iota  
> basicauth:   usernameKey contextKey = iota
> keyauth:     tokenKey contextKey = iota
> csrf:        tokenKey contextKey = iota
> ```
> 
> So in an external service layer, this **won't work**:
> 
> ```go
> // ❌ Can't access - key is unexported
> ctx.Value(requestIDKey)  // Compile error - requestIDKey not accessible
> ```
> 
> External services **must** import the middleware package:
> 
> ```go
> // ✅ Required approach
> import "github.com/gofiber/fiber/v3/middleware/requestid"
> 
> rid := requestid.FromContext(ctx)  // Must use this helper
> ```
> 
> **This means:** Whether we store in `context.Context` eagerly or on-demand, external services need the same import and helper call. The only difference is whether we pay the performance cost on every request vs only when needed.
> 
> If we're already requiring `import + FromContext()`, why not let users opt into the allocation cost instead of forcing it on everyone?

Actually, thinking about this more - since the middleware keys are unexported anyway, why not just provide **helper functions or example code** that let users export values to their own context keys when needed?

ReneWerner87:
We have to strike a balance between performance and user experience here.

In internal discussions, we have come to the conclusion
that we do not want to impose these extra allocations on all users who use the middleware, thereby directly reducing performance
since users should normally have direct access to fiber.CTX and be able to use it

However, there is still the option of combining everything under one implementation -> by only filling the user context if the fiber user explicitly requests it via the app config setting or mw setting
f.e.

```go
   StoreDataInUserContext: true, // <- best name ever .... TBD :D, default "false"
})
```
or direct in the specific mw
```go
app.Use(requestid.New(requestid.Config{
    ...
    StoreDataInUserContext: true // default "false"
})
```

This allows us to strike a balance between performance and convenience.

The problem with the key not being exported does not exist here either, as access is only via helper functions that know about the key because they are part of the middleware solution code.

@gaby @sixcolors 
I think we can proceed with this, or are there still concerns that are not mentioned in the text?

gaby:
@ReneWerner87 Yeah, that's my plan. I'm adding the config option at the app level because some of the middleware do not have a `New`, for example `proxy`. That way we don't limit ourselfs in the future.

I have these changes on my To-Do list for this weekend.

sixcolors:
I'm thinking through the goal here and realize we might be missing the actual problem.

The goal as I understand it (#4066) is to enable services that aren't Fiber-aware to access middleware values. But this PR doesn't achieve that—it just stores unexported keys in multiple contexts instead of one. External services still can't call `ctx.Value(requestIDKey)` because the key isn't exported, so they still need to `import "fiber/middleware/requestid"` and call `requestid.FromContext()`.

So we're not giving services any real independence from Fiber.

If we want truly Fiber-agnostic services, we'd need to either:
- Export the context keys (so services can access them directly, though it would still require a package import and is not Go best practice per context docs), or  
- Document that users should map middleware values to their own context keys at the app level (which is a single line of code in the handler)

```go
app.Get("/test", func(c *fiber.Ctx) error {
    id := requestid.FromContext(c)
    service(context.WithValue(c.Context(), myRequestIDKey, id)) // Manual bridge.
    return c.SendStatus(200)
})
```

if services need to be fiber aware I’m not sure why users wouldn’t just pass the fiber ctx.

ReneWerner87:
Okay, the unexported keys are a problem if it has to be the case that the external tools are completely independent of the middleware itself and no helper can be used.
Do we know for sure that the user really doesn't want to use a helper? 

Perhaps we need to clarify again exactly what needs to be achieved in order to be able to evaluate it again.
Maybe this manual bridge can also be simplified.

ReneWerner87:
Or we could do something like this
```go
app.Use(requestid.New(requestid.Config{
    ...
    StoreInUserContext: "keyName" // default "" - off 
})
```

But let's find out the parameter for this implementation

helpers.go
```diff
@@ -51,6 +52,29 @@ var (
...
+// ValueFromContext retrieves a value stored under key from supported context types.
+//
+// Supported context types:
+//   - Ctx (including CustomCtx implementations)
+//   - *fasthttp.RequestCtx
+//   - context.Context
+func ValueFromContext[T any](ctx, key any) (T, bool) {
+	switch typed := ctx.(type) {
+	case Ctx:
+		val, ok := typed.Value(key).(T)
```
ReneWerner87: fetch it from the locals

PR #4093, opened 2026-02-19
🧹 chore: Improve context helpers

gaby:
This pull request introduces a new StoreInContext helper function designed to streamline how request-scoped values are stored and retrieved within Fiber applications. This helper ensures that values are consistently accessible via both c.Locals() and the underlying context.Context, simplifying middleware development and promoting a more unified approach to context management across the framework. Existing middlewares have been refactored to adopt this new utility.

### DIFF
--- a/app.go
+++ b/app.go
@@ -606,11 +606,6 @@ func New(config ...Config) *App {
 		}
 	}
 
-	// Explicitly preserve default disabled behavior unless configured otherwise.
-	if !app.config.PassLocalsToContext {
-		app.config.PassLocalsToContext = false
-	}
-
 	if app.config.Immutable {
 		app.toBytes, app.toString = toBytesImmutable, toStringImmutable
 	}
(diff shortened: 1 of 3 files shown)


helpers.go
```diff
@@ -76,6 +76,15 @@ func ValueFromContext[T any](ctx, key any) (T, bool) {
 	}
 }
 
+// StoreInContext stores key/value in both Fiber locals and request context.
+//
+// This is useful when values need to be available via both c.Locals() and
+// context.Context lookups throughout middleware and handlers.
+func StoreInContext(c Ctx, key, value any) {
+	c.Locals(key, value)
+	c.SetContext(context.WithValue(c.Context(), key, value))
```
ReneWerner87: - [ ] Check whether the config setting has been set.

- [ ] Check whether a user context has been set beforehand; otherwise, writing it away makes no sense, as the user does not use it.

PR #4103, opened 2026-02-23
feat: add EnableSplittingOnQueryParsers config for per-parser control

### LINKED ISSUE #2624: 🐛 [Bug]: QueryParser is not parsing comma separated values to slices by default
### Bug Description

Bug inserted into #2560. 

I agree that BodyParser shouldn't be splitting values by comma by default, but I don't agree that the same behavior should be extrapolated to QueryParser. By default query params separated by commas are arrays/slices and I believe that the framework must follow this rule by default and not by enabling a configuration

### How to Reproduce

```bash
curl --location 'http://localhost:5000/api/alerts?level=warning%2Cerror' 
```

```golang
package main

import "log"

func main() {
  app := fiber.New()

  // Steps to reproduce
	app.Post("/alerts",func(ctx *fiber.Ctx) error {
		type ReqParams struct {
			Level []string `query:"level"`
		}

		var params ReqParams
		ctx.QueryParser(&params)

	    //incorrect member size 1
          // params.Level[0] = "warning,error" 
        
         // correct members should be:  params.Level[0]="warning"      params.Level[1]="error"
		return nil
	})
  log.Fatal(app.Listen(":5000"))
}
```

1. Go to '....'
3. Do '....'
4. See '....'

### Expected Behavior

Query params must be parsed to slices when they came separated with commas

### Fiber Version

v2.49.1

### Code Snippet (optional)

```go
package main

import "log"

func main() {
  app := fiber.New()

  // Steps to reproduce

}
```

### Checklist:

veeceey:
Closes #2624

The existing `EnableSplittingOnParsers` is global -- it affects query, header, cookie, and body parsers all at once. The issue requests that query params should support comma-splitting independently, since `?level=warning,error` being parsed as a slice is a pretty common convention for query strings but not necessarily for headers/cookies.

This adds `EnableSplittingOnQueryParsers bool` to the config:

- When set to `true`, query params like `?level=warning,error` will be split into `["warning", "error"]` for slice fields, regardless of the global setting
- When `false` (default), query parsing falls back to the `EnableSplittingOnParsers` setting for full backwards compatibility

Usage:

```go
    EnableSplittingOnQueryParsers: true,
})
```

Only touches the query binder, other parsers are unaffected.

### DIFF
--- a/bind_test.go
+++ b/bind_test.go
@@ -163,16 +163,17 @@ func Test_Bind_Query_EnableSplittingOnQueryParsers(t *testing.T) {
 	require.Contains(t, q2.Hobby, "basketball")
 	require.Contains(t, q2.Hobby, "football")
 
-	// Case 3: query-specific=false -> falls back to global
+	// Case 3: global=false, query-specific=false -> falls back to global (no split)
 	app3 := New(Config{
-		EnableSplittingOnParsers:      true,
+		EnableSplittingOnParsers:      false,
 		EnableSplittingOnQueryParsers: false,
 	})
 	c3 := app3.AcquireCtx(&fasthttp.RequestCtx{})
 	c3.Request().URI().SetQueryString("hobby=basketball,football")
 	q3 := new(Query)
 	require.NoError(t, c3.Bind().Query(q3))
-	require.Len(t, q3.Hobby, 2)
+	require.Len(t, q3.Hobby, 1)
+	require.Contains(t, q3.Hobby, "basketball,football")
 }
 
 // go test -run Test_Bind_Query_Map -v


veeceey:
### Test Results

Root package tests (with race detector):

```
$ go test -count=1 -race .
ok  	github.com/gofiber/fiber/v3	27.961s
```

Binder tests:

```
$ go test -count=1 -race ./binder/...
ok  	github.com/gofiber/fiber/v3/binder	1.255s
```

New test `Test_Bind_Query_EnableSplittingOnQueryParsers` covers three scenarios:
1. Global=false, query-specific=true -> query splitting works
2. Global=true, query-specific=false -> query splitting disabled
3. Query-specific=nil -> falls back to global setting

app.go
```diff
@@ -415,6 +415,14 @@ type Config struct { //nolint:govet // Aligning the struct fields is not necessa
 	// Optional. Default: false
 	EnableSplittingOnParsers bool `json:"enable_splitting_on_parsers"`
 
+	// EnableSplittingOnQueryParsers overrides EnableSplittingOnParsers for query parameter parsing.
+	// When set, comma-separated query parameters will be split into slices based on this value
+	// instead of the global EnableSplittingOnParsers setting.
+	// For example: /api?foo=bar,baz will be parsed as foo[]=bar&foo[]=baz when enabled.
+	//
+	// Optional. Default: nil (uses EnableSplittingOnParsers value)
+	EnableSplittingOnQueryParsers *bool `json:"enable_splitting_on_query_parsers"`
```
gaby: Why is this a pointer? It should be a regular boolean. Default value `False`.

gaby (changes requested):

app.go
veeceey: Good point, switched it to a regular bool. When true it enables query splitting, when false (default) it just falls back to the global EnableSplittingOnParsers setting. Pushed the update.

gaby:
@veeceey This PR is not needed. The config value `EnableSplittingOnParsers` already enabled this functionality.

We just forgot to close #2624

ReneWerner87:
Is it really necessary to activate this only for query parameters and not for everything?

gaby:
@veeceey Did you see our comments above?

ReneWerner87:
No response -> closed

PR #4173, opened 2026-04-01
chore: Improve error messages in SaveFileToStorage

pratikramteke:
Improve error messages in SaveFileToStorage by including filename and path.

This provides better debugging context when handling file upload errors.


ctx.go
```diff
@@ -519,7 +519,7 @@ func (*DefaultCtx) SaveFile(fileheader *multipart.FileHeader, path string) error
 func (c *DefaultCtx) SaveFileToStorage(fileheader *multipart.FileHeader, path string, storage Storage) error {
 	file, err := fileheader.Open()
 	if err != nil {
-		return fmt.Errorf("failed to open: %w", err)
+		return fmt.Errorf("failed to open file %s: %w", fileheader.Filename, err)
```
efectn: If the fileader is nil, it is going to panic. you also need to add check for that. In addition to panic check, could you also unit test cases for error messages?

efectn (changes requested):

ctx.go
gaby: Doesn't `err` have the filename already?

ReneWerner87:
@pratikramteke pls check the hints

gaby:
@pratikramteke The error needs to be defined in `error.go`. Then returned that predefined error.

gaby (approved):

ctx.go
efectn: no it only shows what the error is https://pkg.go.dev/internal/oserror#pkg-variables

efectn:
Please check unit tests and lintci errors

ctx.go
```diff
@@ -517,9 +517,13 @@ func (*DefaultCtx) SaveFile(fileheader *multipart.FileHeader, path string) error
...
 // SaveFileToStorage saves any multipart file to an external storage system.
 func (c *DefaultCtx) SaveFileToStorage(fileheader *multipart.FileHeader, path string, storage Storage) error {
+	if fileheader == nil {
+		return ErrFileHeaderNil
+	}
+
 	file, err := fileheader.Open()
 	if err != nil {
-		return fmt.Errorf("failed to open: %w", err)
+		return fmt.Errorf("%w: %q: %v", ErrFileOpen, fileheader.Filename, err)
```
efectn: i think we better sanitize filename for security reasons

gaby: The filename comes from the Client Request, so they already know it.

gaby:
@pratikramteke Can you fix the review issues? Thanks

ctx.go
ReneWerner87: Addressed: nil check returns ErrFileHeaderNil (ctx.go:520-522). Tests Test_Ctx_SaveFileToStorage_NilFileHeader and Test_Ctx_SaveFileToStorage_ErrorMessageContainsFilename cover the nil case and error message content.

ctx.go
ReneWerner87: As @gaby pointed out, the filename comes from the client so there is no information disclosure (the client already supplied it). The %q verb used in fmt.Errorf safely escapes control characters and newlines, so the error message stays well-formed in logs.

ctx.go
```diff
@@ -529,25 +533,25 @@ func (c *DefaultCtx) SaveFileToStorage(fileheader *multipart.FileHeader, path st
...
+		return fmt.Errorf("%w: %q: %v", ErrFileRead, fileheader.Filename, fasthttp.ErrBodyTooLarge)
 	}
 
 	buf := bytebufferpool.Get()
 	defer bytebufferpool.Put(buf)
 
 	limitedReader := io.LimitReader(file, int64(maxUploadSize)+1)
 	if _, err = buf.ReadFrom(limitedReader); err != nil {
-		return fmt.Errorf("failed to read: %w", err)
+		return fmt.Errorf("%w: %q: %v", ErrFileRead, fileheader.Filename, err)
```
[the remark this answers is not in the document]
ReneWerner87: Fixed in 1ca45698 - now uses fmt.Errorf("%w: %q: %w", ErrFileRead, fileheader.Filename, err) so callers can errors.Is/As on the underlying read error.

ctx.go
```diff
@@ -529,25 +533,25 @@ func (c *DefaultCtx) SaveFileToStorage(fileheader *multipart.FileHeader, path st
...
 	}
 
 	if buf.Len() > maxUploadSize {
-		return fmt.Errorf("failed to read: %w", fasthttp.ErrBodyTooLarge)
+		return fmt.Errorf("%w: %q: %v", ErrFileRead, fileheader.Filename, fasthttp.ErrBodyTooLarge)
 	}
```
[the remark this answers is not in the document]
ReneWerner87: Fixed in 1ca45698 - fasthttp.ErrBodyTooLarge is now wrapped with %w. Test_Ctx_SaveFileToStorage_LimitExceeded(_UnknownSize) confirms errors.Is works again.

ctx.go
```diff
@@ -529,25 +533,25 @@ func (c *DefaultCtx) SaveFileToStorage(fileheader *multipart.FileHeader, path st
...
 
 	data := append([]byte(nil), buf.Bytes()...)
 
 	if err := storage.SetWithContext(c.Context(), path, data, 0); err != nil {
-		return fmt.Errorf("failed to store: %w", err)
+		return fmt.Errorf("%w: %q to %q: %v", ErrFileStore, fileheader.Filename, path, err)
```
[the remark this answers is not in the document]
ReneWerner87: Fixed in 1ca45698 - storage error is wrapped with %w. New Test_Ctx_SaveFileToStorage_StoreErrorIncludesPath asserts errors.Is on both ErrFileStore and the underlying backend error.

ctx.go
```diff
@@ -529,25 +533,25 @@ func (c *DefaultCtx) SaveFileToStorage(fileheader *multipart.FileHeader, path st
 	}
 
 	if fileheader.Size > 0 && fileheader.Size > int64(maxUploadSize) {
-		return fmt.Errorf("failed to read: %w", fasthttp.ErrBodyTooLarge)
+		return fmt.Errorf("%w: %q: %v", ErrFileRead, fileheader.Filename, fasthttp.ErrBodyTooLarge)
 	}
```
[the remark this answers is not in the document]
ReneWerner87: Resolved by design: open/read errors intentionally include filename only because the path is not yet relevant (we have not reached the storage write). The store error includes both filename and path. Added Test_Ctx_SaveFileToStorage_StoreErrorIncludesPath to cover the path assertion explicitly; the BodyLimit test was corrected to drop the misplaced path check.

ctx.go
```diff
@@ -517,9 +517,13 @@ func (*DefaultCtx) SaveFile(fileheader *multipart.FileHeader, path string) error
...
 func (c *DefaultCtx) SaveFileToStorage(fileheader *multipart.FileHeader, path string, storage Storage) error {
+	if fileheader == nil {
+		return ErrFileHeaderNil
+	}
+
 	file, err := fileheader.Open()
 	if err != nil {
-		return fmt.Errorf("failed to open: %w", err)
+		return fmt.Errorf("%w: %q: %v", ErrFileOpen, fileheader.Filename, err)
 	}
```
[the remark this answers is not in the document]
ReneWerner87: Fixed in 1ca45698 - fileheader.Open() error is now wrapped with %w.

ctx.go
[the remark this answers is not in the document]
ReneWerner87: Fixed in 1ca45698 - both pre-read size check and post-read buf.Len check wrap fasthttp.ErrBodyTooLarge with %w. Test_Ctx_SaveFileToStorage_LimitExceeded(_UnknownSize) verifies.

ReneWerner87 (approved):

PR #4196, opened 2026-04-07
🔥 feat: Add SSE (Server-Sent Events) middleware

vinod-morya:
## Summary

Adds `middleware/sse` — production-grade Server-Sent Events built natively for Fiber's fasthttp architecture. This is the only Go SSE implementation with proper client disconnect detection on fasthttp (via `SendStreamWriter` + `w.Flush()` error).

Follows the discussion in #4194 where @ReneWerner87 and @gaby agreed this belongs in Fiber core as `middleware/sse`.

Based on [fibersse](https://github.com/vinod-morya/fibersse) which is already in production at PersonaCart (300+ concurrent connections, multi-tenant SaaS) and listed on [awesome-go](https://github.com/avelino/awesome-go/pull/6190).

## Usage

```go
handler, hub := sse.NewWithHub(sse.Config{
    OnConnect: func(c fiber.Ctx, conn *sse.Connection) error {
        conn.Topics = []string{"notifications"}
        conn.Metadata["tenant_id"] = c.Locals("tenant_id").(string)
        return nil
    },
})
app.Get("/events", handler)

// From any handler or worker — replaces polling
hub.Invalidate("orders", order.ID, "created")
```

## Features

| Feature | Description |
|---------|-------------|
| **Hub pattern** | Single-goroutine event loop, topic-based routing |
| **3 priority lanes** | Instant (P0), Batched (P1), Coalesced/last-writer-wins (P2) |
| **NATS-style wildcards** | `notifications.*`, `analytics.>` |
| **Adaptive throttling** | Slow clients get longer flush intervals automatically |
| **Connection groups** | Publish by metadata (tenant_id, user_id) |
| **Cache invalidation** | `Invalidate()`, `Signal()` — one-line polling replacement |
| **Domain events** | `DomainEvent()`, `Progress()`, `Complete()`, `BatchDomainEvents()` |
| **Auth helpers** | JWT + one-time ticket auth (EventSource can't send headers) |
| **Metrics** | JSON + Prometheus endpoints |
| **Last-Event-ID replay** | Pluggable `Replayer` interface with in-memory default |
| **Fan-out** | Bridge Redis/NATS pub/sub to SSE |
| **Graceful drain** | Kubernetes-style shutdown with `Retry-After` |

## Follows Fiber Conventions

- `New(config ...Config) fiber.Handler` + `NewWithHub(config ...Config) (fiber.Handler, *Hub)` — same pattern as session's `NewWithStore()`
- `Config` struct with `Next` field, `ConfigDefault`, `configDefault()` helper
- Uses `github.com/gofiber/fiber/v3/log` (not `log/slog`)
- Zero external dependencies — only uses packages already in Fiber's go.mod
- `golangci-lint` passes clean (0 issues) against Fiber's `.golangci.yml`

## Benchmarks (Apple M4 Max)

```
Benchmark_SSE_Publish-16             321,270    3,780 ns/op    139 B/op    3 allocs/op
Benchmark_SSE_TopicMatch-16        5,274,810      229 ns/op     64 B/op    2 allocs/op
Benchmark_SSE_TopicMatch_Exact-16 14,051,781       89 ns/op      0 B/op    0 allocs/op
Benchmark_SSE_MarshalEvent-16        253,378    4,767 ns/op    345 B/op   10 allocs/op
Benchmark_SSE_GenerateID-16        1,991,384      606 ns/op     32 B/op    1 allocs/op
```

## Quality

- **50 tests** + 7 benchmarks + 4 runnable examples
- **70.2% coverage** (remaining 30% is SSE streaming internals requiring real HTTP client)
- `go test -race` passes clean
- `golangci-lint` passes clean (0 issues)
- `go vet` clean

## File Structure

```
middleware/sse/
├── config.go          Config + ConfigDefault + configDefault()
├── sse.go             New(), NewWithHub(), Hub, run loop
├── connection.go      Per-client connection, write loop
├── event.go           Event types, SSE wire format, marshaling
├── coalescer.go       P1/P2 event buffering
├── topic.go           NATS-style wildcard matching
├── throttle.go        Adaptive per-connection flush interval
├── auth.go            JWT + ticket auth helpers
├── invalidation.go    Cache invalidation helpers
├── domain_event.go    Domain event, progress, batch helpers
├── fanout.go          Redis/NATS pub/sub bridge
├── replayer.go        Last-Event-ID replay (pluggable)
├── metrics.go         Prometheus + JSON metrics
├── stats.go           Hub statistics
├── sse_test.go        50 tests + 7 benchmarks
├── example_test.go    4 runnable examples
└── README.md          Fiber-format documentation
```

cc @ReneWerner87 @gaby @efectn

### DIFF
--- a/middleware/sse/event.go
+++ b/middleware/sse/event.go
@@ -1,0 +1,176 @@
+package sse
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"strings"
+	"sync/atomic"
+	"time"
+)
+
+// Priority controls how an event is delivered to clients.
+type Priority int
+
+const (
+	// PriorityInstant bypasses all buffering — the event is written to the
+	// client connection immediately. Use for errors, auth revocations,
+	// force-refresh commands, and chat messages.
+	PriorityInstant Priority = 0
+
+	// PriorityBatched collects events in a time window (FlushInterval) and
+	// sends them all at once. Use for status changes, media updates.
+	PriorityBatched Priority = 1
+
+	// PriorityCoalesced uses last-writer-wins per CoalesceKey. Multiple
+	// events with the same key within a flush window are merged — only the
+	// latest is sent. Use for progress bars, live counters, typing indicators.
+	PriorityCoalesced Priority = 2
+)
+
+// Event represents a single SSE event to be published through the hub.
+type Event struct {
+	CreatedAt   time.Time
+	Data        any
+	Group       map[string]string
+	Type        string
+	ID          string
+	CoalesceKey string
+	Topics      []string
+	TTL         time.Duration
+	Priority    Priority
+}
+
+// globalEventID is an auto-incrementing counter for event IDs.
+var globalEventID atomic.Uint64
+
+// nextEventID returns a monotonically increasing event ID string.
+func nextEventID() string {
+	return fmt.Sprintf("evt_%d", globalEventID.Add(1))
+}
+
+// MarshaledEvent is the wire-ready representation of an SSE event.
+// External Replayer implementations receive and return this type.
+type MarshaledEvent struct {
+	// CreatedAt is the timestamp of the source Event (zero if unset).
+	CreatedAt time.Time
+	ID        string
+	Type      string
+	Data      string
+	// TTL is the maximum age for this event. Zero means no expiry.
+	TTL   time.Duration
+	Retry int // -1 means omit
+}
+
+// sanitizeSSEField strips carriage returns and newlines from SSE control
+// fields (id, event) to prevent SSE injection attacks. An attacker-controlled
+// value containing \r or \n could break SSE framing and inject fake events.
+func sanitizeSSEField(s string) string {
+	return strings.NewReplacer("\r\n", "", "\r", "", "\n", "").Replace(s)
+}
+
+// marshalEvent converts an Event into wire-ready format.
+func marshalEvent(e *Event) MarshaledEvent {
+	me := MarshaledEvent{
+		ID:        sanitizeSSEField(e.ID),
+		Type:      sanitizeSSEField(e.Type),
+		CreatedAt: e.CreatedAt,
+		TTL:       e.TTL,
+		Retry:     -1,
+	}
+
+	if me.ID == "" {
+		me.ID = nextEventID()
+	}
+
+	switch v := e.Data.(type) {
+	case nil:
+		me.Data = ""
+	case string:
+		me.Data = v
+	case []byte:
+		me.Data = string(v)
+	case json.Marshaler:
+		b, err := v.MarshalJSON()
+		if err != nil {
+			errJSON, _ := json.Marshal(err.Error()) //nolint:errcheck,errchkjson // encoding a string never fails
+			me.Data = fmt.Sprintf(`{"error":%s}`, string(errJSON))
+		} else {
+			me.Data = string(b)
+		}
+	default:
+		b, err := json.Marshal(v)
+		if err != nil {
+			errJSON, _ := json.Marshal(err.Error()) //nolint:errcheck,errchkjson // encoding a string never fails
+			me.Data = fmt.Sprintf(`{"error":%s}`, string(errJSON))
+		} else {
+			me.Data = string(b)
+		}
+	}
+
+	return me
+}
+
+// WriteTo writes the SSE-formatted event to w following the Server-Sent
+// Events specification.
+func (me *MarshaledEvent) WriteTo(w io.Writer) (int64, error) {
+	var total int64
+
+	if me.ID != "" {
+		n, err := fmt.Fprintf(w, "id: %s\n", me.ID)
+		total += int64(n)
+		if err != nil {
+			return total, fmt.Errorf("sse: write id: %w", err)
+		}
+	}
+
+	if me.Type != "" {
+		n, err := fmt.Fprintf(w, "event: %s\n", me.Type)
+		total += int64(n)
+		if err != nil {
+			return total, fmt.Errorf("sse: write event: %w", err)
+		}
+	}
+
+	if me.Retry >= 0 {
+		n, err := fmt.Fprintf(w, "retry: %d\n", me.Retry)
+		total += int64(n)
+		if err != nil {
+			return total, fmt.Errorf("sse: write retry: %w", err)
+		}
+	}
+
+	// strings.SplitSeq("", "\n") yields "", correctly writing "data: \n" for empty data.
+	for line := range strings.SplitSeq(me.Data, "\n") {
+		n, err := fmt.Fprintf(w, "data: %s\n", line)
+		total += int64(n)
+		if err != nil {
+			return total, fmt.Errorf("sse: write data: %w", err)
+		}
+	}
+
+	n, err := fmt.Fprint(w, "\n")
+	total += int64(n)
+	if err != nil {
+		return total, fmt.Errorf("sse: write terminator: %w", err)
+	}
+	return total, nil
+}
+
+// writeComment writes an SSE comment line.
+func writeComment(w io.Writer, text string) error {
+	_, err := fmt.Fprintf(w, ": %s\n\n", text)
+	if err != nil {
+		return fmt.Errorf("sse: write comment: %w", err)
+	}
+	return nil
+}
+
+// writeRetry writes the retry directive.
+func writeRetry(w io.Writer, ms int) error {
+	_, err := fmt.Fprintf(w, "retry: %d\n\n", ms)
+	if err != nil {
+		return fmt.Errorf("sse: write retry: %w", err)
+	}
+	return nil
+}
--- a/middleware/sse/example_test.go
+++ b/middleware/sse/example_test.go
@@ -1,0 +1,101 @@
+package sse
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/gofiber/fiber/v3"
+)
+
+func Example() {
+	app := fiber.New()
+
+	handler, hub := NewWithHub(Config{
+		OnConnect: func(_ fiber.Ctx, conn *Connection) error {
+			conn.Topics = []string{"notifications"}
+			conn.Metadata["user_id"] = "example"
+			return nil
+		},
+	})
+
+	app.Get("/events", handler)
+
+	// Publish from any handler or worker
+	hub.Publish(Event{
+		Type:   "update",
+		Data:   map[string]string{"message": "hello"},
+		Topics: []string{"notifications"},
+	})
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+
+	if err := hub.Shutdown(ctx); err != nil {
+		panic(err)
+	}
+
+	fmt.Println("Hub created and shut down successfully") //nolint:errcheck // example test output
+	// Output: Hub created and shut down successfully
+}
+
+func Example_invalidation() {
+	_, hub := NewWithHub()
+
+	// Replace polling: instead of clients polling every 30s,
+	// push an invalidation signal when data changes.
+	hub.Invalidate("orders", "ord_123", "created")
+
+	// Multi-tenant
+	hub.InvalidateForTenant("t_1", "orders", "ord_456", "updated")
+
+	// With hints (small extra data)
+	hub.InvalidateWithHint("orders", "ord_789", "created", map[string]any{
+		"total": 149.99,
+	})
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+
+	if err := hub.Shutdown(ctx); err != nil {
+		panic(err)
+	}
+
+	fmt.Println("Invalidation events published") //nolint:errcheck // example test output
+	// Output: Invalidation events published
+}
+
+func Example_progress() {
+	_, hub := NewWithHub()
+
+	// Coalesced: if progress goes 1%→2%→3%→4% in one flush window,
+	// only 4% is sent to the client.
+	for i := 1; i <= 100; i++ {
+		hub.Progress("import", "imp_1", "t_1", i, 100)
+	}
+	hub.Complete("import", "imp_1", "t_1", true, nil)
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+
+	if err := hub.Shutdown(ctx); err != nil {
+		panic(err)
+	}
+
+	fmt.Println("Progress tracking complete") //nolint:errcheck // example test output
+	// Output: Progress tracking complete
+}
+
+func Example_ticketAuth() {
+	store := NewMemoryTicketStore()
+	defer store.Close()
+
+	// Issue a ticket (typically in a POST handler after JWT validation)
+	ticket, err := IssueTicket(store, `{"tenant":"t_1","topics":"orders,products"}`, 30*time.Second)
+	if err != nil {
+		panic(err)
+	}
+
+	fmt.Println("Ticket issued, length:", len(ticket)) //nolint:errcheck // example test output
+	// Output: Ticket issued, length: 48
+}
--- a/middleware/sse/invalidation.go
+++ b/middleware/sse/invalidation.go
@@ -1,0 +1,118 @@
+package sse
+
+import (
+	"time"
+)
+
+// InvalidationEvent is a lightweight signal telling the client to refetch
+// a specific resource.
+type InvalidationEvent struct {
+	// Hint is optional extra data for the client.
+	Hint map[string]any `json:"hint,omitempty"`
+
+	// Resource is what changed (e.g., "orders", "products").
+	Resource string `json:"resource"`
+
+	// Action is what happened (e.g., "created", "updated", "deleted").
+	Action string `json:"action"`
+
+	// ResourceID is the specific item that changed (optional).
+	ResourceID string `json:"resource_id,omitempty"`
+}
+
+// Invalidate publishes a cache invalidation signal to all connections
+// subscribed to the given topic.
+func (h *Hub) Invalidate(topic, resourceID, action string) {
+	h.Publish(Event{
+		Type:   "invalidate",
+		Topics: []string{topic},
+		Data: InvalidationEvent{
+			Resource:   topic,
+			Action:     action,
+			ResourceID: resourceID,
+		},
+		Priority: PriorityInstant,
+	})
+}
+
+// InvalidateForTenant publishes a tenant-scoped cache invalidation signal.
+func (h *Hub) InvalidateForTenant(tenantID, topic, resourceID, action string) {
+	h.Publish(Event{
+		Type:   "invalidate",
+		Topics: []string{topic},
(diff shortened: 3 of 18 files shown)


vinod-morya:
Addressed all review feedback in c39263a:

| Feedback | Fix |
|----------|-----|
| **Critical: `>` wildcard bug** (`a.>.c` matching `a.b.c`) | Fixed — `>` now enforced as last token only. Added regression tests. |
| **Prometheus label escaping** (escaping entire string) | Fixed — `escapePromLabelValue()` now escapes only the value part. |
| **README: Logger field doesn't exist** | Removed from config table. Uses `fiber/v3/log` directly. |
| **README: MaxLifetime says 0=unlimited** | Fixed — docs now say `-1` for unlimited (matches code). |
| **Bearer prefix case-sensitive** | Fixed — now uses `strings.EqualFold` per RFC 6750. |
| **WriteTo redundant empty data branch** | Simplified — `strings.Split("", "\n")` returns `[""]` which handles it. |
| **TicketAuth README example incomplete** | Added working `json.Unmarshal` example. |

`golangci-lint` passes clean (0 issues), `go test -race` passes, `go vet` clean.

vinod-morya:
Addressed remaining feedback in 64966ef:

| Feedback | Fix |
|----------|-----|
| **Heartbeat magic string** | Replaced with type-safe `heartbeatSignal` struct. Send channel is now `chan any` with type switch in `writeLoop` — no more sentinel strings. |
| **MemoryReplayer ring buffer** | Replaced slice append+reslice with fixed-size circular buffer. `Store()` is now O(1) with zero allocations once the buffer is full. |
| **MemoryTicketStore goroutine leak** | Added `runtime.SetFinalizer` to auto-close if caller forgets `Close()`. |

**Re: Invalidate* overlap with DomainEvent** — These are intentionally separate APIs for different use cases:
- `Invalidate("orders", "ord_1", "created")` — 3 params, simple cache invalidation signal
- `DomainEvent("orders", "created", "ord_1", "t_1", hint)` — 5 params, tenant-scoped with hints

`Invalidate*` is the 80% use case (simple, no tenant, no hints). `DomainEvent` is the full-featured version. Removing `Invalidate*` would force every simple call to pass empty strings for tenantID and nil for hints. Happy to consolidate if the team prefers — just let me know.

All checks: `golangci-lint` 0 issues, `go test -race` passes, `go vet` clean.

vinod-morya:
Addressed all CodeRabbit critical and major findings in 8be8414:

### Critical
| Finding | Fix |
|---------|-----|
| **Topic+Group OR leaks across tenants** | AND semantics when both set — connections must match topic AND group |
| **Replay drops tenant filter** | Group-scoped events excluded from replayer storage |
| **Register/unregister blocks after shutdown** | Channel sends wrapped with `select` + `hub.shutdown` |

### Major
| Finding | Fix |
|---------|-----|
| **Shutdown double-close panic** | `sync.Once` wraps `close(h.shutdown)` |
| **Publish blocks under backpressure** | Added `default` drop case, updated doc |
| **TTL lost in coalescer** | `MarshaledEvent` now carries `CreatedAt`+`TTL`, re-checked in `flushAll()` |
| **Register before replay = duplicates** | Moved register to AFTER `initStream` completes |
| **Marshal error invalid JSON** | `json.Marshal(err.Error())` for proper escaping |
| **Batch cross-topic delivery** | Documented as intentional (dashboard pattern) |

Build, `go test -race`, `golangci-lint` — all clean.

vinod-morya:
Addressed nitpick feedback in c226977:

- **Dead `CoalesceKey` on instant events** — Removed from all `Invalidate*` and `DomainEvent` helpers. `CoalesceKey` is only meaningful for `PriorityCoalesced`; instant events bypass the coalescer entirely.
- **Pre-allocate `batched` slice** — `newCoalescer` now initializes with `make([]MarshaledEvent, 0, 16)` to avoid first-append allocation.
- **Added `Test_SSE_Shutdown_Idempotent`** — Calls `Shutdown()` twice, verifies no panic (`sync.Once` guard).

vinod-morya:
Addressed all remaining CodeRabbit feedback + coverage boost in 98903a1c:

**Security:** Sanitize `MarshaledEvent.ID` and `Type` — strip `\r\n` to prevent SSE framing injection.

**Type safety:** Replaced `chan any` with typed `chan MarshaledEvent` + separate `heartbeat chan struct{}`. Compile-time safe, no more type switches.

**Correctness:**
- Metadata frozen (defensive copy) after `OnConnect` to prevent concurrent mutation
- `FanOut` Transform path now fully respects all fields including `PriorityInstant`
- Removed redundant `Content-Type` from `MetricsHandler`

**Coverage: 69% → 91%** — Added 40+ tests. Codecov will update on this commit.

gaby:
@vinod-morya Can you tell claude to document the middleware using the same format as the other middlewares.
- remove the readme.md
- update whats_new.md
- fix modernize lint ci issues

vinod-morya:
Thanks @gaby! All three addressed:

- Removed `README.md` from middleware folder
- Added `docs/middleware/sse.md` following the same format as other middlewares
- Updated `docs/whats_new.md` with SSE section
- Fixed all `modernize` lint issues (`maps.Copy`, `min`/`max` builtins, `range` over int, `strings.SplitSeq`)
- Also added nil-guard panic in `FanOut()` per CodeRabbit

One clean commit, 91% coverage. Let me know if anything else needs adjusting! 🤙

middleware/sse/example_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+
+	if err := hub.Shutdown(ctx); err != nil {
+		panic(err)
+	}
+
+	fmt.Println("Hub created and shut down successfully") //nolint:errcheck // example test output
+	// Output: Hub created and shut down successfully
```
grivera64: Please also validate the endpoint's response using [`app.Test()`](https://docs.gofiber.io/api/app/#test) so that we can be sure that everything is working as expected.

middleware/sse/example_test.go
grivera64: Please use `testing` and `testify` to write your unit tests.

The main setup for this in gofiber core is:

```go
import (
    "testing"
    "github.com/stretchr/testify/require"
)

// go test -run My_Test
func My_Test(t *testing.T) {
    t.Parallel()

    var err error
    // run a unit test ...
    requires.NoError(t, err)
}

```

grivera64: We should probably not include an example file here since it will be exported alongside the SSE middleware.

middleware/sse/event.go
```diff
@@ -0,0 +1,176 @@
...
+	Type        string
+	ID          string
+	CoalesceKey string
+	Topics      []string
+	TTL         time.Duration
+	Priority    Priority
+}
+
+// globalEventID is an auto-incrementing counter for event IDs.
+var globalEventID atomic.Uint64
```
grivera64: We shouldn't use a global ID in production. It would get exhausted very quickly. We could try:

- A handler-local ID
- Use some kind kind of unique UUID
- Tell the user to provide a unique custom string-based ID per request

middleware/sse/invalidation.go
```diff
@@ -0,0 +1,118 @@
...
+		Type:   "invalidate",
+		Topics: []string{topic},
+		Data: InvalidationEvent{
+			Resource:   topic,
+			Action:     action,
+			ResourceID: resourceID,
+		},
+		Priority: PriorityInstant,
+	})
+}
```
grivera64: Using `InvalidationEvent` as a subfield of `Event` under the name `Data` feels like a bit awkward naming. Could we refactor this to be more clear that `InvalidationEvent` is the data rather a subtype of `Event`?

gaby:
@vinod-morya I don't think this implementation is correct:
- It adds manually created prometheus metrics (no idea why this is here).
- it implements it is own auth, even though Fiber has that already.
- the hub stuff seems very specific to a use case.

Does this SSE support any SSE client, or is it specific to one?

vinod-morya:
Hey @gaby, thanks for the detailed look! Happy to simplify.

**Client compatibility** — Yes, fully standard SSE. Works with browser `EventSource`, curl, or any HTTP client:

```bash
curl -N http://localhost:3000/events
```
```javascript
const es = new EventSource('/events');
```

Follows the W3C spec exactly — `id:`, `event:`, `data:`, `retry:` fields.

**Prometheus** — Agreed, that was overreach. Will remove — Fiber's monitor middleware is the right place for metrics.

**Auth** — Fair point about overlap. The ticket-based auth is there because `EventSource` can't send custom headers (it's a known browser limitation — no `Authorization: Bearer` support). But I can remove it from the middleware entirely and leave auth to the user via `OnConnect`. Keeps things simple.

**Hub** — The hub is how SSE broadcasting works in practice (r3labs/sse, go-sse all use this pattern). Without it, we'd just be wrapping `SendStreamWriter` which Fiber already has. The hub handles connection tracking, topic routing, and delivering events to the right clients — that's the core value.

The invalidation/signal/progress helpers are just convenience wrappers over `hub.Publish()`. Happy to strip those out to keep the middleware focused on core SSE only.

Want me to slim it down to just: Hub + connections + topics + publish? I can push a trimmed version today.

vinod-morya:
Hey @gaby, I went ahead and prepared a slimmed-down version on a separate branch so you can see what the minimal scope looks like: [vinod-morya/fiber@feat/middleware-sse-slim](https://github.com/vinod-morya/fiber/tree/feat/middleware-sse-slim)

Quick comparison:

| | Current PR | Slim Version |
|---|---|---|
| Source files | 14 | 9 |
| LOC | ~4,640 | ~2,860 |
| Tests | 88 | 58 |
| Coverage | 91% | 82% |

**Removed in slim version:**
- `auth.go` — JWT/Ticket auth (users handle in `OnConnect`)
- `metrics.go` — Prometheus + JSON metrics (Fiber's monitor middleware is the right place)
- `invalidation.go` — Invalidate/Signal helpers (just wrappers over `hub.Publish()`)
- `domain_event.go` — DomainEvent/Progress/Complete (also wrappers)
- `fanout.go` — Redis/NATS bridge (users can do this with a goroutine + `Publish`)
- `MemoryReplayer` impl (kept just the `Replayer` interface so users can plug in their own)

**Kept (core SSE):**
- Hub + connections + topic routing
- 3 priority lanes (instant / batched / coalesced)
- NATS-style wildcards (`orders.*`, `analytics.>`)
- Connection groups (metadata-based filtering for multi-tenant)
- Adaptive throttling, graceful drain
- `Replayer` interface (pluggable backend)
- Standard SSE wire format — works with any `EventSource` client

Just say the word and I'll force-push the slim version onto this PR. Or if you'd prefer keeping some pieces (like the `Replayer` interface or anything else), let me know and I'll adjust.

vinod-morya:
@gaby friendly bump on this — happy to move forward either way (slim branch or current PR). Just let me know which direction you'd prefer and I'll push the changes.

gaby:
@vinod-morya We will take a look soon. Thanks!

The slim version looks great, I think we can keep `fanout.go`. 

@grivera64 Thoughts?

vinod-morya:
Thanks @gaby! Restored `fanout.go` + all 7 fan-out tests on the slim branch: [vinod-morya/fiber@2ef3803](https://github.com/vinod-morya/fiber/commit/2ef38031)

The slim branch now has:
- Hub + connections + topic routing
- 3 priority lanes (instant / batched / coalesced)
- NATS-style wildcards
- Connection groups (metadata-based filtering)
- Adaptive throttling, graceful drain
- `Replayer` interface (pluggable backend)
- **`FanOut` + `FanOutMulti`** (bridges external pub/sub like Redis/NATS into the hub)
- Standard SSE wire format

Coverage: 83.9%, `golangci-lint` clean, `go test -race` clean.

Also cleaned up the package godoc (removed stale JWT/Prometheus mentions) and added a `FanOut` example to `docs/middleware/sse.md` with a Redis subscriber implementation.

Happy to wait for @grivera64's thoughts before force-pushing onto this PR. Whenever you both are good with the scope, just give me the word and I'll update #4196 with the slim version in one shot.

grivera64:
@vinod-morya It seems the link to https://github.com/vinod-morya/fiber/commit/2ef38031 leads to a 404 page, but I was able to find https://github.com/vinod-morya/fiber/tree/feat/middleware-sse-slim by visiting your repo.

@gaby Since this version doesn't have a PR, I will add my thoughts here. Since they are kind of long I put the details under each dropdown to avoid a really long comment.

---

### Review (Click on the dropdown of each section to expand)


<details><summary>1. Remove `Next` from sse.Config</summary>
<p>

I am not sure if a `Next` handler is required here since SSE is a terminal middleware that shouldn't have any other handlers after it in the call chain. This is because the stream writer passed into `c.SendStreamWriter(sw)` does not run until after Fiber releases the `fiber.Ctx` object. As a result, we should probably remove that `Next` field in the config to prevent any undefined behavior.


</p>
</details> 

<details><summary>2. Update Coalescer -> Dispatcher (could be converted to an Add-On)</summary>
<p>

From my understanding, `coalescer.go` is more of a dispatcher with two separate queues. One queue (P1) has keyless values that allow for duplicates (e.g. batching), and the second queue (P2) has keyed values that overwrite in the case of duplicates (e.g. coalescer).

We should rename it to something like `EventDispatcher`, keeping most of the same logic, but renaming its methods to `AddEvent(event)` and `AddState(key, value)`, and `WriteTo(buf)` to be more clear on what it does in SSE terminology.

If we would want to reuse this for other middleware (e.g. for a partial renderer), we could rewrite this as an addon (or a set of addons, e.g. BufferedQueue & CoalescedQueue addons/utils, or a single Dispatcher addon/util, etc.). Then we can keep the main logic that is SSE-only within the SSE middleware.

</p>
</details> 

<details><summary>3. Clean up fanout (and potentially convert into an addon)</summary>
<p>

`hub.Fanout(cfg)` feels a bit cumbersome to use. We can instead have it be provided in the `sse.Config` directly, and start all related goroutines after calling something like `hub.Start()`or `hub.Serve()`, and disable them when calling `hub.Shutdown()` or similar.

We could even rename fanout to become some sort of Adapter or Bridge called `SubscriberBridge` to make it more clear what it is for.

```go
type Config struct {
    // ...
    Bridge       SubscriberBridge       // The bridge that fanouts communication from Storage -> Subscribed Hubs
    BridgeConfig SubscriberBridgeConfig // The config for the adapter to use (empty uses a default)
}

type SubscriberBridge interface {
    Subscribe(ctx context.Context, channel string, 
}
```

If it is also helpful, we could pull it out of SSE and make it into an addon, but I don't think we would need to here. 

</p>
</details> 

<details><summary>4. Reduce memory allocations</summary>
<p>

The branch as-is seems to allocate a lot of buffers and channels. To prevent allocation/deallocation overhead, let's use `sync.Pool` to allow for reusing buffers and channels. If the implementation requires closing channels, then this will make it harder to limit allocations. We can either:

- Try using a different mechanism than channels (e.g. conditional variables + mutex) for waiting goroutines, or
- Use a `sync.Pool` to recycle the channels without calling `close(ch)`.

</p>
</details> 

<details><summary>5. Add more e2e tests validating server response body's and config functionality using `app.Test()`</summary>
<p>

The current tests seem to only check status codes, errors, and internal mechanisms. We should also add more end-to-end tests that validate that a user's SSE handler sends the expected response body to the client. We should also test the responses with different config settings.

This way, we can test if the middleware formats the response body in a way that works for all standard use cases of Server-Side Events (e.g. `EventSource` in vanilla JavaScript or [Azure/fetch-event-source](https://github.com/Azure/fetch-event-source)).

</p>
</details> 

<details><summary>6. Clean up sse folder structure</summary>
<p>

We can probably separate each logical section into separate files. For example, the hub struct and all of its methods should probably be defined in its own file (my recommendation is that `hub.go` should contain all method, including fanout-related ones).

This way, we can keep the full implementation of each struct/interface contained in a single file rather than be spread across several files.


</p>
</details> 

**tl;dr:** I think we can take a look into the feasibility of moving features that are useful beyond SSE into its own addon/middleware/util. We also need to add some end-to-end tests to validate server response bodies, as well as cleanup the sse folder structure. We should also add some `sync.Pool`'s to limit the number of allocations and de-allocations to follow Fiber's "zero memory allocation" goal.

---

What do you think about the above?

vinod-morya:
Thanks @grivera64! I re-read your review and walked through the Fiber middleware codebase (session, cache, limiter, keyauth) before touching anything. Everything landed in a single v1.1 refactor on the slim branch: [vinod-morya/fiber@d6e5c73](https://github.com/vinod-morya/fiber/commit/d6e5c735).

**Breaking / API changes**
1. ✅ Removed `Next` from `Config` — SSE is terminal; documented that placing handlers after it is undefined behavior.
2. ✅ `coalescer` → `Dispatcher` with `AddEvent` / `AddState` / `WriteTo` — renamed methods and fields to match SSE semantics.
3. ✅ Replaced `hub.FanOut()` with `Config.Bridges []BridgeConfig` + `SubscriberBridge` interface. Bridges start automatically and are canceled/awaited on `Hub.Shutdown`. No more dangling `CancelFunc`s for the caller to track.

**Zero-allocation hot path (your #4)**
4. ✅ `MarshaledEvent.WriteTo` now builds the frame in a pooled `bytebufferpool.ByteBuffer`. Benchmark went from **381 B/op, 8 allocs** to **0 B/op, 0 allocs** at 26 ns/op. `writeComment`, `writeRetry`, and `nextEventID` share the pooled path.

**E2E tests (your #5)**
5. ✅ Added e2e tests that drive a real TCP listener and validate the wire format: `Content-Type: text/event-stream`, `retry` frame, `connected` frame with `connection_id` + `topics`, multi-line `data`, and sanitized `id`/`event` fields against injection. These complement the existing unit tests and exercise the path standard clients (`EventSource`, `fetch-event-source`) would take.

**File organization (your #6)**
6. ✅ `sse.go` shrank from **671 → 154 lines** (just `New`, `NewWithHub`, `generateID`). All Hub methods live in `hub.go`. Bridge types + driver live in `bridge.go`.

**Plus, while I was in there**
- Added `c.Abandon()` before `SendStreamWriter` — the prior code relied on Fiber not recycling the ctx mid-stream. The timeout middleware does this explicitly; SSE should too.
- `t.Parallel()` on every test per `AGENTS.md`.
- Replayer errors now log and continue (best-effort replay) instead of silently dropping the replayed events.

**Status:** `go build`, `go vet`, `go test -race`, `golangci-lint` — all clean. Coverage **90.7%**.

Ready for another pass whenever you and @gaby have a moment. Happy to split this into smaller commits if that's easier to review.

grivera64:
@vinod-morya Could you push your commits to a PR (either push to this one or a close the current one and create a new one)? And if possible, please avoid using force push.

It would be much easier to use the review UI to give feedback on specific code blocks and track what changed after a specific commit. When ready to merge, we can do a squash commit to clean up git history afterwards.

vinod-morya:
Thanks @grivera64 — good point, sorry for the back-and-forth via commit links. I've opened #4225 with the v1.1 slim branch so the review UI can anchor comments to specific files/lines, and I'll avoid force-pushes there.

Closing this PR in favour of #4225. Happy to move any open threads from here over to the new PR if that's easier — just let me know.

Thanks again for the thorough reviews.

PR #4199, opened 2026-04-08
🔥 feat: Host auth middleware 

### LINKED ISSUE #4148: 📝 [Proposal]:  Host authorization middleware — DNS rebinding protection via Host header validation

 A middleware that validates the incoming Host header against a configurable allowlist, rejecting requests with unrecognized hosts with 403 Forbidden. This protects against DNS rebinding attacks, where an
  attacker-controlled domain resolves to the application's internal IP, causing browsers to send requests with a malicious Host header to the application.

Express does not have a built-in host authorization middleware. The closest third-party equivalent is host-validation -> https://github.com/brannondorsey/host-validation

 RFC 9112 (HTTP/1.1), Section 3.2 requires servers to respond with 400 Bad Request to requests that lack a Host header, contain multiple Host headers, or have an invalid Host value. This middleware builds on top
of that baseline requirement.

 RFC 9110 (HTTP Semantics), Section 7.2 ("Host and :authority") establishes that the Host header identifies the target origin server for the request. Section 17.1 ("Establishing Authority") states:

An origin server is responsible for ensuring services controlling its certificate's private key are prepared to reject requests that appear to have been misdirected.

This middleware implements that recommendation at the application layer by rejecting requests whose Host header does not match the expected values — going beyond the RFC's minimum syntactic validation to
provide semantic validation against DNS rebinding and host header injection attacks.

The middleware will:
 - Strip the port from the Host header before matching (per RFC 9110 Section 7.2, the host and port are separate components)
  - Respect X-Forwarded-Host when Fiber's TrustProxy is enabled
  - Return 403 Forbidden by default for unauthorized hosts (not 400, since the request is syntactically valid but semantically unauthorized)

### API Stability

The API surface is minimal and modeled after patterns already proven stable in Fiber:

  1. AllowedHosts []string — mirrors cors.AllowOrigins. String-based allowlists are the most stable configuration pattern.
  2. AllowedHostsFunc func(string) bool — mirrors cors.AllowOriginsFunc. Same escape hatch pattern.
  3. Next, Exclude, ErrorHandler — all existing Fiber conventions used across multiple middleware.
  4. Matching semantics (exact, subdomain wildcard, CIDR) — these are the same patterns Rails has used since Rails 6 (2019) without breaking changes.

No external dependencies are required — only net from the standard library for CIDR parsing. The middleware has no state, no caching, and no side effects beyond rejecting requests, which makes API evolution straightforward.

### Feature Examples

```go
Protect a production API


  app := fiber.New()

  app.Use(hostauthorization.New(hostauthorization.Config{
      AllowedHosts: []string{"api.myapp.com"},
  }))

  app.Get("/users", func(c fiber.Ctx) error {
      return c.JSON(getUsers())
  })

  // Host: api.myapp.com → 200 OK
  // Host: evil.com      → 403 Forbidden

  Allow subdomains and exclude health checks

  app.Use(hostauthorization.New(hostauthorization.Config{
      AllowedHosts: []string{".myapp.com"},
      Exclude: func(c fiber.Ctx) bool {
          return c.Path() == "/healthz"
      },
  }))

  // Host: api.myapp.com  GET /users   → 200 OK
  // Host: www.myapp.com  GET /        → 200 OK
  // Host: evil.com       GET /users   → 403 Forbidden
  // Host: evil.com       GET /healthz → 200 OK (excluded)



Internal service with CIDR ranges:


  app.Use(hostauthorization.New(hostauthorization.Config{
      AllowedHosts: []string{
          "internal.myapp.com",
          "10.0.0.0/8",       // internal network
          "127.0.0.1",        // localhost
      },
  }))

  // Host: internal.myapp.com → 200 OK
  // Host: 10.0.50.3          → 200 OK
  // Host: 169.254.169.254    → 403 Forbidden (blocks cloud metadata)


Scoped to a route group:


  admin := app.Group("/admin")
  admin.Use(hostauthorization.New(hostauthorization.Config{
      AllowedHosts: []string{"admin.myapp.com"},
  }))

  // Host: admin.myapp.com GET /admin/dashboard → 200 OK
  // Host: myapp.com       GET /admin/dashboard → 403 Forbidden
  // Host: myapp.com       GET /public          → 200 OK (not in group)
```

### Checklist:

mutantkeyboard:
# Description

 Add a new `hostauthorization` middleware that validates the incoming Host header against a configurable allowlist, rejecting unauthorized hosts with 403 Forbidden. This protects against DNS rebinding attacks
  where an attacker-controlled domain resolves to the application's internal IP, causing browsers to send requests with a malicious Host header.

  The middleware supports exact host matching, subdomain wildcards (.myapp.com), and CIDR ranges (10.0.0.0/8), with an optional dynamic validator function for runtime host resolution. It complements the `Domain()`
  router (#4100) — `Domain()` handles `host-based` routing while `hostauthorization` acts as a security gate.

  Fixes #4148

## Changes introduced

  - [x] Benchmarks: 3 benchmarks included (exact match, CIDR, mixed rules). All ~3,200 ns/op with 25 allocs on Apple M5.
  - [x] Documentation Update: Added docs/middleware/hostauthorization.md with full usage guide, config reference, matching semantics, normalization rules, proxy support, and RFC compliance notes.
  - [x] Changelog/What's New: New hostauthorization middleware — DNS rebinding protection via Host header validation with exact match, subdomain wildcard, and CIDR support.
  - [x] Migration Guide: N/A — new middleware, no breaking changes.
  - [x] API Alignment with Express: Modeled after host-validation (closest Express ecosystem equivalent). AllowedHosts/AllowedHostsFunc mirrors the AllowOrigins/AllowOriginsFunc pattern from the existing cors
  middleware. Next, ErrorHandler follow established Fiber conventions.
  - [x] API Longevity: Minimal API surface (4 config fields). String-based allowlists and callback escape hatches — same patterns Rails has used since Rails 6 (2019) without breaking changes. No external
  dependencies, no state, no caching.
  - [x] Examples: Usage examples covering basic protection, subdomain wildcards, CIDR ranges, health check exclusion, dynamic validation, custom error responses, and combined usage with `Domain()` router.

## Type of change

## Checklist

## Commit formatting


ReneWerner87:
Thanks for the contribution @mutantkeyboard , DNS rebinding protection is a valid concern and I appreciate the thorough work on docs and tests.

Before we move forward, I want to raise a design question: should this be a standalone middleware or an extension of the existing `Domain()` router?

`Domain()` already matches requests by host pattern at the routing level. Adding an allowlist/deny semantic there (e.g. app.SetAllowedHosts(...)) would:
  - Avoid the overhead of an extra handler in the middleware chain on every request
  - Make the relationship between host-based routing and host-based security explicit
  - Keep host logic in one place instead of two separate systems

If we do keep it as a standalone middleware, a few things need addressing:

Performance
- ~3,200 ns/op with 25 allocations is too high for a middleware that runs on every request. The core of this is a map lookup + a few string comparisons, that should be near-zero alloc. The wildcard suffixes should be stored with the leading dot at init time (as already flagged in reviews), and `normalizeHost()` should be applied to config values at init, not just at request time.
You can use the utils package or existing functions from the other middlewares

API / Consistency
- Missing Exclude field , our other middlewares expose this hook, and this one should too for consistency.
- normalizeHost() not applied to configured hosts during init - this can lead to subtle mismatches (e.g. MyApp.com in config vs myapp.com at runtime).

Documentation
- The CIDR section reads like source-IP filtering ("respond to requests from known IP ranges"). This needs to clearly state it matches the Host header value, not the client IP. As written, someone could deploy this thinking they have IP-based access control, which
  would be a security misunderstanding.

Scope question
- CIDR matching on the Host header is a fairly niche use case; in practice, Host headers rarely contain raw IPs outside of internal tooling. Is the complexity worth it, or would exact match + subdomain wildcards cover 99% of real-world usage?

I'd like to hear your thoughts on the `Domain()` integration angle before we iterate further on the middleware approach. Either way, the normalization logic and test coverage you've built here is solid and would carry over.

mutantkeyboard:
Hey @ReneWerner87, thank you for the review.

  **Regarding `Domain()` integration**

  I'm genuinely curious what did you have in mind there. My original inspiration for this was [brannondorsey/host-validation](https://github.com/brannondorsey/host-validation), which is explicitly a standalone
  middleware for DNS rebinding protection on Express. The pattern just felt like a natural fit here too — `Domain()` is opt-in per route and calls `c.Next()` on non-match, so it doesn't have global reject
  semantics. CIDR and leading-dot wildcards would also need special-casing since `parseDomainPattern` doesn't support either today. That said, if you see a clean way to integrate it without blurring routing and
  access control, I'm open to it.

  **Performance**

  

> The 3,200 ns/op and 25 allocations are from `app.Test()`

 -  that overhead belongs to Fiber's request pipeline, not this middleware per se. The actual per-request cost is a map lookup, a couple of string ops, and one real alloc: the `"."+suffix` concat in the wildcard path. Fixing that (storing the leading dot at init) plus switching to `utilsstrings.ToLower` in `normalizeHost` gets the common path to near-zero alloc.
I'm more than happy to add a lower-level benchmark that isolates just the matching logic so the numbers are less misleading.

  **`normalizeHost` at init**

This one is actually a good catch - config entries with trailing dots won't match runtime-normalized hosts. Easy fix.

  **CIDR docs**

  This is actually a fair point. I can fix the wording to make clear it's matching the `Host` header value, not the client IP.
  
  But regarding the overall architecture, I can integrate it into the existing `Domain()` or leave it as a separate middlware. It's really up to you and @gaby 

  WDYT?

ReneWerner87:
ok

Go ahead with the leading-dot fix at init + normalizeHost on config entries, that alone should clean up the hot path nicely. And yes, please add an isolated benchmark for just the matching logic so we have real numbers for the middleware cost without `app.Test()` noise.
Fix the CIDR docs wording as discussed.
On CIDR scope: keep it for now. It's not a lot of extra code and internal tooling with IP-based hosts is a real enough use case. We can always revisit if it causes maintenance headaches down the line.

pls also try the utils functions for the better performance

mutantkeyboard:
@ReneWerner87, am I crazy or it seems like fasthttp 1.7.0 has broken multiple packages with stuff like `Host` header in HTTP/1.1 , `fs.FS` handling and `Cache-Control`?

gaby:
> @ReneWerner87, am I crazy or it seems like fasthttp 1.7.0 has broken multiple packages with stuff like `Host` header in HTTP/1.1 , `fs.FS` handling and `Cache-Control`?

@ReneWerner87 @mutantkeyboard the PR for fasthttp was merged with failing tests. They added a bunch of security fixes that conflict with our security checks.

gaby:
@mutantkeyboard Test failures are now related to the new middleware.

mutantkeyboard:
@gaby - I will fix that today

mutantkeyboard:
@gaby - this is now fixed. 💪💪

middleware/hostauthorization/hostauthorization.go
```diff
@@ -0,0 +1,130 @@
...
+			for _, cidr := range parsed.cidrNets {
+				if cidr.Contains(ip) {
+					return true
+				}
+			}
+		}
+	}
+
+	// Dynamic validator fallback
+	if allowedHostsFunc != nil {
```
gaby: This should be at the top of the function

middleware/hostauthorization/hostauthorization.go
```diff
@@ -0,0 +1,130 @@
...
+		if h == "" {
+			continue
+		}
+		h = normalizeHost(h)
+		if h == "" {
+			continue
+		}
+
+		switch {
+		case strings.Contains(h, "/"):
```
gaby: Using CIDR's like this is a bad idea. We should be using the native functions from golang instead.

For example, we already do all this with https://github.com/gofiber/fiber/blob/main/app.go#L446

https://github.com/gofiber/fiber/blob/main/req.go#L1080

mutantkeyboard: Valid point

docs/middleware/hostauthorization.md
```diff
@@ -0,0 +1,203 @@
...
+    return c.JSON(getUsers())
+})
+
+// Host: api.myapp.com → 200 OK
+// Host: evil.com      → 403 Forbidden
+```
+
+### Subdomain Wildcards
+
+A leading dot matches any subdomain but **not** the bare domain itself:
```
gaby: This not matching the bare domain, is a weird behavior.

docs/middleware/hostauthorization.md
```diff
@@ -0,0 +1,203 @@
...
+
+```go
+app.Use(hostauthorization.New(hostauthorization.Config{
+    AllowedHosts: []string{".myapp.com"},
+}))
+
+// Host: api.myapp.com  → 200 OK
+// Host: www.myapp.com  → 200 OK
+// Host: myapp.com      → 403 Forbidden
```
gaby, line 54: This a confusing behavior. Requiring 2 allow rules

docs/middleware/hostauthorization.md
```diff
@@ -0,0 +1,203 @@
...
+```go
+AllowedHosts: []string{"myapp.com", ".myapp.com"},
+```
+
+### CIDR Ranges
+
+Useful for services accessed directly by IP (e.g. internal tooling) where the `Host` header will be a raw IP address. This matches the **Host header value** against a CIDR range — it does not filter by client IP address:
+
+```go
+app.Use(hostauthorization.New(hostauthorization.Config{
```
gaby: This is kind of confusing with https://docs.gofiber.io/whats_new#trusted-proxies

I rather see the same TrustedProxyConfig here, to avoid duplicated code.

middleware/hostauthorization/hostauthorization.go
```diff
@@ -0,0 +1,128 @@
...
+		h = utils.TrimSpace(h)
+		if h == "" {
+			continue
+		}
+		h = normalizeHost(h)
+		if h == "" {
+			continue
+		}
+
+		if hostIP, cidr, err := net.ParseCIDR(h); err == nil {
```
gaby: This is all duplicated logic from TrustedProxyConfig handling. Maybe we should make that a public function that can be used from here

middleware/hostauthorization/hostauthorization.go
```diff
@@ -0,0 +1,128 @@
...
+	"strings"
+
+	"github.com/gofiber/fiber/v3"
+	"github.com/gofiber/utils/v2"
+	utilsstrings "github.com/gofiber/utils/v2/strings"
+)
+
+// parsedHosts holds the pre-parsed host matching structures.
+type parsedHosts struct {
+	exact            map[string]bool
```
ReneWerner87: Suggest using `map[string]struct{}` instead of `map[string]bool`. The convention in the rest of the codebase (e.g. `TrustProxyConfig.ips map[string]struct{}` in `app.go`) uses zero-byte values for set semantics. `make modernize` will likely flag this too.

middleware/hostauthorization/hostauthorization.go
ReneWerner87: Silent CIDR typo footgun: a common mistake like `"10.0.0/8"` (missing octet) silently falls through to an exact-match entry that can never match. `Test_ConfigInvalidCIDRTreatedAsExact` documents this, but in production it's dangerous because the operator thinks their internal range is allowed when it isn't.

We currently panic on non-canonical CIDRs (e.g. `10.0.0.5/8`) but accept syntactically broken ones silently. That asymmetry is confusing. Suggest mirroring `app.handleTrustedProxy`: if the string contains `/`, always treat it as a CIDR attempt and either panic or `log.Warnf` on parse failure so the misconfiguration surfaces.

middleware/hostauthorization/hostauthorization.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+// normalizeHost normalizes a hostname for matching.
+// Strips port (if any), trailing dot, IPv6 brackets, and lowercases.
+// Safe to call on both c.Hostname() output (already port-stripped) and
+// raw AllowedHosts entries (which may include a port like "example.com:8080").
+func normalizeHost(host string) string {
+	// Strip port if present. net.SplitHostPort handles both "host:port" and
+	// "[::1]:port" forms, and strips IPv6 brackets as a side effect.
+	if h, _, err := net.SplitHostPort(host); err == nil {
```
ReneWerner87, line 94: **Hot-path allocation:** `c.Hostname()` already strips the port (via `parseAddr` in `helpers.go`), so on real request hostnames `net.SplitHostPort` always hits the error path, and stdlib allocates a `*AddrError` for every error. That's one allocation per request.

Suggest a fast path:

```go
if len(host) > 0 && host[0] != '[' && strings.IndexByte(host, ':') < 0 {
    host = strings.TrimSuffix(host, ".")
    return utilsstrings.ToLower(host)
}
// existing logic for the rare case
```

Worth a before/after benchmark with the existing `Benchmark_matchHost_*` to confirm the alloc drops to 0.

middleware/hostauthorization/hostauthorization.go
```diff
@@ -0,0 +1,128 @@
...
+	// Strip trailing dot (FQDN normalization)
+	host = strings.TrimSuffix(host, ".")
+
+	return utilsstrings.ToLower(host)
+}
+
+// matchHost checks if the given host matches any of the parsed allowed hosts.
+func matchHost(host string, parsed parsedHosts, allowedHostsFunc func(string) bool) bool {
+	// Dynamic validator — checked first so it can override static rules
+	if allowedHostsFunc != nil && allowedHostsFunc(host) {
```
ReneWerner87: **This contradicts both the docs and the stated CORS pattern.**

The middleware docs say:

> `AllowedHostsFunc` is only called when static `AllowedHosts` don't match

And the PR description claims this mirrors the CORS `AllowOriginsFunc` pattern, where the func is explicitly a fallback (see `middleware/cors/cors.go`):

```go
// Run AllowOriginsFunc if the logic for handling the value in 'AllowOrigins'
// does not result in allowOrigin being set.
if allowOrigin == "" && cfg.AllowOriginsFunc != nil && cfg.AllowOriginsFunc(originHeaderRaw) {
```

Performance-wise this matters too: a user-supplied `AllowedHostsFunc` is the most expensive lookup (the docs even suggest "Look up tenant domains from database"), so calling it on every request before the static map lookup is exactly the wrong default.

Suggest reordering to: exact → wildcard → CIDR → `AllowedHostsFunc`. That matches the documented behaviour and the CORS convention.

A counter-based test would prevent regressions:

```go
called := 0
fn := func(host string) bool { called++; return false }
// request that matches a static AllowedHosts entry
require.Equal(t, 0, called, "AllowedHostsFunc should not be called when static hosts match")
```

middleware/hostauthorization/hostauthorization.go
```diff
@@ -0,0 +1,128 @@
...
+	// Subdomain wildcard: ".myapp.com" matches "api.myapp.com" but NOT "myapp.com"
+	for _, suffix := range parsed.wildcardSuffixes {
+		if strings.HasSuffix(host, suffix) {
+			return true
+		}
+	}
+
+	// CIDR match: parse host as IP and check against CIDR ranges
+	if len(parsed.cidrNets) > 0 {
+		if ip := net.ParseIP(host); ip != nil {
```
ReneWerner87: **Hot-path allocation:** `net.ParseIP` runs on every request as soon as any CIDR is configured, even for hostnames like `api.myapp.com` that obviously can't be IPs. `ParseIP` allocates internally.

A cheap pre-check (first byte must be a digit or `:`) would skip `ParseIP` for the typical mixed exact + CIDR config where most requests carry hostnames, not raw IPs:

```go
if len(parsed.cidrNets) > 0 && len(host) > 0 {
    if c := host[0]; (c >= '0' && c <= '9') || c == ':' {
        if ip := net.ParseIP(host); ip != nil {
            // ...
        }
    }
}
```

middleware/hostauthorization/config.go
```diff
@@ -0,0 +1,63 @@
...
+// Config defines the config for the host authorization middleware.
+type Config struct {
+	// Next defines a function to skip this middleware when returned true.
+	// Use this to exclude health check endpoints or other paths from host validation.
+	//
+	// Optional. Default: nil
+	Next func(c fiber.Ctx) bool
+
+	// AllowedHostsFunc is a dynamic validator called when static AllowedHosts
+	// don't match. Receives the hostname (port stripped, lowercased).
```
ReneWerner87: Doc inconsistency: this comment says `(port stripped, lowercased)` but `normalizeHost` actually does port strip + trailing-dot strip + IPv6-bracket strip + lowercase. The docs page (`docs/middleware/hostauthorization.md`) just says "normalized hostname". Worth picking one canonical description (probably as a doc block on `normalizeHost` itself) and pointing both call sites at it.

middleware/hostauthorization/config.go
```diff
@@ -0,0 +1,63 @@
...
+		cfg = config[0]
+	}
+
+	if len(cfg.AllowedHosts) == 0 && cfg.AllowedHostsFunc == nil {
+		panic("hostauthorization: AllowedHosts or AllowedHostsFunc is required")
+	}
+
+	if cfg.ErrorHandler == nil {
+		cfg.ErrorHandler = func(c fiber.Ctx, _ error) error {
+			return c.Status(fiber.StatusForbidden).SendString("Forbidden")
```
ReneWerner87: The hardcoded `"Forbidden"` string drifts from the rest of the framework. Other middleware uses `utils.StatusMessage(fiber.StatusForbidden)` or returns `fiber.NewError(fiber.StatusForbidden)` so the response body stays consistent with however Fiber surfaces status text elsewhere (and survives a future refactor without each middleware needing a touch). Minor, but trivially fixable.

middleware/hostauthorization/hostauthorization.go
```diff
@@ -0,0 +1,128 @@
...
+// parsedHosts holds the pre-parsed host matching structures.
+type parsedHosts struct {
+	exact            map[string]bool
+	wildcardSuffixes []string
+	cidrNets         []*net.IPNet
+}
+
+// parseAllowedHosts categorizes AllowedHosts into exact, wildcard, and CIDR groups.
+// Panics on invalid CIDR entries.
+func parseAllowedHosts(hosts []string) parsedHosts {
```
ReneWerner87, line 28: RFC 1035 length validation: `parseDomainPattern` in `domain.go` enforces total length ≤ 253 chars, per-label length ≤ 63 chars, and a max label count via `maxDomainParts`. `parseAllowedHosts` enforces none of these, so a misconfiguration can put unbounded strings into the matching loop on every request.

For consistency with the rest of the framework's hostname handling (and as a cheap defense against accidental DoS from a malformed config), suggest mirroring those bounds here.

middleware/hostauthorization/hostauthorization_test.go
```diff
@@ -0,0 +1,831 @@
...
+	})
+
+	req := httptest.NewRequest(fiber.MethodGet, "/", http.NoBody)
+	req.Host = ""
+
+	// An empty Host must always be rejected. Two valid outcomes depending on Go version:
+	//   - fasthttp rejects the missing Host at the protocol level → app.Test returns an error
+	//   - Go 1.26+ serializes a default Host value so the request reaches the middleware,
+	//     which rejects it with 403 Forbidden → app.Test returns a response with no error
+	resp, err := app.Test(req)
```
ReneWerner87, line 280: This early `return` on any non-nil error from `app.Test` (already flagged by CodeRabbit) makes the test pass for unrelated failures too.

Suggest tightening to assert the error is host-related before returning, e.g.

```go
if err != nil {
    require.ErrorContains(t, err, "Host")
    return
}
```

Otherwise a future fasthttp change that errors out for a different reason would silently keep this test green.

docs/middleware/hostauthorization.md
```diff
@@ -0,0 +1,203 @@
...
+## Proxy Support
+
+The middleware uses Fiber's `c.Hostname()`, which respects `X-Forwarded-Host` when [`TrustProxy`](https://docs.gofiber.io/api/fiber#config) is enabled. When `TrustProxy` is disabled (the default), `X-Forwarded-Host` is ignored and the raw `Host` header is used.
+
+## RFC Compliance
+
+- **RFC 9110 Section 7.2** — Host and port are separate components; port is stripped before matching
+- **RFC 9110 Section 17.1** — Origin servers should reject misdirected requests
+- **RFC 9112 Section 3.2** — Requests with missing Host headers should be rejected
+- Returns **403 Forbidden** (not 400) because the request is syntactically valid but semantically unauthorized
```
ReneWerner87, line 213: Worth considering 421 Misdirected Request as the default instead of 403. RFC 9110 §15.5.20 defines it as:

> The request was directed at a server that is unable or unwilling to produce an authoritative response for the target URI.

Which fits the "wrong host for this server" semantics better than 403 ("understood, but refused due to permissions"). Cloudflare, Fastly and other CDNs use 421 for exactly this case.

Either response is defensible, but 403 is currently presented here as the only correct option, which I don't think it is. At minimum, 421 deserves a mention as an alternative.

docs/middleware/hostauthorization.md
```diff
@@ -0,0 +1,203 @@
...
+- Port is stripped (via `c.Hostname()`)
+- Trailing dot removed (`example.com.` → `example.com`)
+- IPv6 brackets removed (`[::1]` → `::1`)
+- Lowercased
+
+`AllowedHosts` entries are also lowercased at initialization.
+
+## Proxy Support
+
+The middleware uses Fiber's `c.Hostname()`, which respects `X-Forwarded-Host` when [`TrustProxy`](https://docs.gofiber.io/api/fiber#config) is enabled. When `TrustProxy` is disabled (the default), `X-Forwarded-Host` is ignored and the raw `Host` header is used.
```
ReneWerner87, line 203: Worth a sentence here noting HTTP/2 behavior: clients send `:authority` instead of `Host`. fasthttp transparently maps it, so the middleware works the same in both protocols, but operators reading this section currently might assume HTTP/1.1-only.

ReneWerner87 (changes requested):
Going through the latest version of the branch (`3334763`), here is a follow-up on top of what's already been discussed. Most of the high-priority code-level findings are in inline comments. The points below are either architectural (don't fit a single line) or reinforcements of things @gaby and others already raised that I think deserve a more concrete recommendation.

## IDN / Punycode

Browsers always send internationalised domains as Punycode in the Host header: `münchen.example.com` arrives as `xn--mnchen-3ya.example.com`. With the current implementation, configuring `"münchen.example.com"` in `AllowedHosts` will never match a real browser request.

There's already precedent in the codebase (`req.go`, `Subdomains()`):

```go
if strings.Contains(host, "xn--") {
    if u, err := idna.Lookup.ToUnicode(host); err == nil {
        host = utilsstrings.ToLower(u)
    }
}
```

Suggest normalising both sides (config entries and request hostnames) consistently, ideally to ASCII via `idna.Lookup.ToASCII` since browsers consistently use Punycode and ASCII comparisons stay alloc-free. At minimum this should be a documented "Known limitations" entry.

## TrustedProxyConfig consolidation (reinforcing @gaby)

The CIDR parse and contains-check logic is essentially identical to `app.handleTrustedProxy`. A shared helper would benefit both modules and keep behaviour consistent (the canonical-CIDR panic exists here but not in `handleTrustedProxy`). Worth doing, even if as a follow-up PR.

## Subdomain wildcard semantics (reinforcing @gaby)

The leading-dot convention that excludes the apex is ergonomically weak and forces two entries for the most common case. Worth deciding now since changing it later is breaking. My preference would be `*.myapp.com` for subdomains-only and `myapp.com` for apex (familiar glob syntax), but implicit apex inclusion would also be defensible.

## Test gaps

A few that would be worth adding alongside the inline points:

- `AllowedHostsFunc` as fallback (with call-counter assertion, see L80 inline)
- IPv6 + port in `AllowedHosts` config entry (`"[::1]:8080"`). IPv4 is covered, IPv6 isn't.
- Concurrent requests with `-race`. The implementation is read-only after init but the contract should be locked in.
- A scaling benchmark with many wildcards (e.g. 100 entries) so we know when the linear `HasSuffix` scan starts mattering.
- A `FuzzNormalizeHost` target. Trivial to add and standard hygiene for any header parser:

  ```go
  func FuzzNormalizeHost(f *testing.F) {
      f.Add("example.com")
      f.Add("[::1]:8080")
      f.Add(".myapp.com")
      f.Fuzz(func(t *testing.T, input string) {
          _ = normalizeHost(input)
      })
  }
  ```

## Sentinel error stability

`ErrForbiddenHost` is exported, so users can do `errors.Is(...)`. A single test asserting the exact `Error()` string locks in the textual contract for those who match on the message. Minor but useful.

## Scope question

CIDR matching accounts for the bulk of the surface area in this PR (parsing, the non-canonical panic, documentation confusion with `TrustedProxyConfig`, and the duplicated logic). For an initial merge, dropping CIDR and shipping with exact + subdomain wildcards only would make this a tighter, more obviously-correct module. CIDR can land in a follow-up where it's properly consolidated with the trusted-proxy infrastructure.

Not a blocker, just a thought given how much of the review feedback (mine and @gaby's) orbits the CIDR feature.

---

Overall really nice work on the iterations so far. The split between `matchHost` micro-benchmarks and the full HTTP-pipeline benchmarks is a great addition for isolating actual middleware cost.

ReneWerner87:
@mutantkeyboard pls check the last comments

mutantkeyboard:
@ReneWerner87 - This is an excellent wrap-up by both you and @gaby - thank you guys so much. I'll try to fix all of these ASAP, and will re-request your review once I'm done.

middleware/hostauthorization/hostauthorization.go
ReneWerner87: Addressed: `normalizeHost` has a fast path that skips `net.SplitHostPort` for plain hostnames (no `[` and no `:`), so request hosts no longer pay the `*AddrError` allocation (hostauthorization.go:65-72). Resolving.

middleware/hostauthorization/hostauthorization.go
ReneWerner87: Addressed: `matchHost` now evaluates exact -> wildcard -> CIDR -> `AllowedHostsFunc`, with a doc comment explaining the fallback semantics (hostauthorization.go:88-126). `Test_AllowedHostsFuncFallback` locks in the call-counter behavior (hostauthorization_test.go:716-737). Resolving.

middleware/hostauthorization/config.go
ReneWerner87: Addressed: the `AllowedHostsFunc` godoc now describes the full normalization (port stripped, trailing dot removed, IPv6 brackets removed, lowercased), and `normalizeHost` itself carries a matching doc block (config.go:20-23, hostauthorization.go:61-64). Resolving.

middleware/hostauthorization/hostauthorization.go
ReneWerner87: Addressed: any entry containing `/` is now always treated as a CIDR attempt and panics on parse failure with a clear message (hostauthorization.go:39-48). Resolving.

middleware/hostauthorization/config.go
ReneWerner87: Addressed: default `ErrorHandler` now uses `c.SendStatus(fiber.StatusForbidden)`, so the response body comes from Fibers status helper rather than a hardcoded string (config.go:59-63). Resolving.

middleware/hostauthorization/hostauthorization.go
ReneWerner87: Addressed: `net.ParseIP` is now gated by a cheap pre-check on the first byte (digit) or `:` presence, so it is skipped for obvious non-IP hostnames (hostauthorization.go:110-120). Resolving.

middleware/hostauthorization/hostauthorization.go
ReneWerner87: Addressed: `parsedHosts.exact` is now `map[string]struct{}` (hostauthorization.go:14). Resolving.

middleware/hostauthorization/hostauthorization_test.go
ReneWerner87: Addressed: `Test_HostAuthorization_EmptyHost` no longer early-returns on err. `app.Test` injects `req.Host = localhost` when empty, so the test asserts `require.NoError` + `StatusForbidden` directly, with a comment explaining the injection (hostauthorization_test.go:276-296). Resolving.

ReneWerner87:
## Re-review summary

Verified the addressed inline points against `5f32ae4f` and resolved each thread individually. Tests pass with `-race`. The remaining items below are still open and need a decision before this can be approved.

### Blocking (please address or explicitly defer with a follow-up issue)

1. **IDN / Punycode normalization** (architectural point from my CHANGES_REQUESTED review). Browsers always send IDNs as Punycode in the `Host` header. Configuring `"münchen.example.com"` in `AllowedHosts` will never match a real request. Either normalize both sides via `idna.Lookup.ToASCII` (precedent in `req.go` `Subdomains()`), or add a "Known limitations" section to the docs page so operators are not silently misled.

2. **Subdomain wildcard apex semantics** (gaby threads on `docs/middleware/hostauthorization.md:45` and `:54`, reinforced in my review body). Requiring two entries (`myapp.com` + `.myapp.com`) for the most common case is a UX trap, and the leading-dot convention is non-standard. Pick one and lock it in now since changing it post-merge is breaking. My preference: `*.myapp.com` for subdomains-only and `myapp.com` for apex (familiar glob syntax). Implicit apex inclusion would also be defensible.

3. **CIDR scope and `TrustedProxyConfig` duplication** (gaby thread on `hostauthorization.go:36`). The CIDR parse + contains-check logic duplicates `app.handleTrustedProxy`. Either consolidate now (extract a public helper from the trusted-proxy path), or drop CIDR from this initial PR and ship it in a follow-up that does the consolidation properly. Keeping it as-is locks in the duplication.

### Non-blocking but worth doing

4. **RFC 1035 length validation** (`hostauthorization.go:21`, my open thread #3147503295). `parseDomainPattern` in `domain.go` enforces ≤253 total / ≤63 per-label. `parseAllowedHosts` enforces neither. Cheap defense against malformed config.

5. **`421 Misdirected Request` as documented alternative** (my open thread #3147533311 on `docs/middleware/hostauthorization.md:203`). RFC 9110 §15.5.20 fits the "wrong host for this server" semantics better than 403. At minimum mention it as an alternative in the `ErrorHandler` example.

6. **HTTP/2 `:authority` note** (my open thread #3147558088 on `docs/middleware/hostauthorization.md:196`). Worth one sentence in the Proxy Support section noting that HTTP/2 clients send `:authority` and fasthttp maps it transparently, so the middleware behaves identically.

7. **Doc cross-reference** (gaby thread on `docs/middleware/hostauthorization.md:68`). The CIDR section now correctly says "Host header value, not client IP", but a direct link to https://docs.gofiber.io/whats_new#trusted-proxies would prevent future confusion with `TrustProxyConfig`.

8. **Scaling benchmark with many wildcards** (my review body). Add a benchmark with ~100 wildcard entries so we know when the linear `HasSuffix` scan starts mattering. Current benches only cover a handful.

### Already verified addressed

All 8 inline code findings from my CHANGES_REQUESTED review (map type, CIDR footgun, two hot-path allocations, AllowedHostsFunc fallback ordering, both godoc inconsistencies, hardcoded status string, empty-host test) are resolved in their individual threads. `Test_AllowedHostsFuncFallback`, `Test_NormalizeHost_IPv6WithPortInConfig`, `Test_ErrForbiddenHostString`, and `FuzzNormalizeHost` are all in place. Nice iteration overall.

Once items 1-3 land (or are explicitly deferred via tracking issues), I am happy to approve.

ReneWerner87:
@mutantkeyboard can you check the open points

mutantkeyboard:
@gaby - I'll double check suggestions from Copilot, though it seems that some of the comments are a pure hallucination.

gaby (approved):
LGTM, we will address some things in a separate PR

ReneWerner87 (approved):

PR #4204, opened 2026-04-10
🐛 fix(csrf): strip path from referer before matching trusted origins

aviu16:
## What happened

`refererMatchesHost` compares the full referer URL (including path and query string) against `trustedOrigins` and `trustedSubOrigins`. Since trusted origins are stored without paths (e.g. `https://example.com`), a referer like `https://example.com/form/submit` never matches via `slices.Contains` or `subdomain.match` (which uses `strings.HasSuffix`).

`originMatchesHost` doesn't have this problem because Origin headers omit the path per the HTTP spec.

This means CSRF protection falls back to rejecting all cross-origin requests from trusted origins when the browser sends a Referer instead of an Origin header.

## Fix

Extract `scheme://host` from the parsed referer URL before comparing against trusted origins, matching how `originMatchesHost` already works.

## Test

Added a test case that sends a trusted referer with a path component and verifies it returns 200 instead of 403.


sixcolors (approved):
LGTM. Normalizing `Referer` to an origin (`scheme://host`) before checking `trustedOrigins`/wildcards is the right behavior, and the new tests cover the regression nicely.

gaby (approved):

gaby:
Merge pending fixes to Fiber. The CI failing are related to a recently merged PR.

ReneWerner87 (approved):

PR #4209, opened 2026-04-11
🐛 fix: implement releaseData function for better resource management

ReneWerner87:


middleware/session/data.go
```diff
@@ -40,6 +40,12 @@ func acquireData() *data {
 	panic("unexpected type in data pool")
 }
 
+// releaseData resets the data object and returns it to the pool.
+func releaseData(d *data) {
+	d.Reset()
```
gaby, line 50: Just in case, we should do a nil check here.

ReneWerner87: fixed with the last commit

gaby (approved):
LGTM

PR #4225, opened 2026-04-21
🔥 feat: Add SSE middleware

vinod-morya:
## Context

Opening this as a fresh PR per @grivera64's request on #4196 so the review UI can anchor feedback to specific commits and lines. #4196 will be closed and linked back here for history.

## Scope

Slim, focused SSE middleware built natively for Fiber v3's `fasthttp` architecture. Everything that wasn't strictly SSE-mechanics (JWT, Prometheus, invalidation helpers, domain-event wrappers, REST invalidation routes) has been removed and kept in the standalone downstream library. What remains is the core hub + connection lifecycle + wire format.

### What's included
- `Hub` — single goroutine event loop, topic-based routing
- 3 priority lanes: `PriorityInstant` / `PriorityBatched` / `PriorityCoalesced`
- NATS-style topic wildcards (`orders.*`, `analytics.>`)
- Connection groups (metadata-based filtering; e.g. multi-tenant scoping)
- Adaptive per-connection throttling
- Graceful Kubernetes-style drain via `Hub.Shutdown(ctx)`
- Pluggable `Replayer` interface (`Last-Event-ID` replay, in-memory impl included)
- `SubscriberBridge` — auto-started bridges from external pub/sub (Redis, NATS, …) declared in `Config.Bridges`, canceled/awaited on shutdown

### What was deliberately left out (lives downstream)
- Auth helpers (JWT, ticket stores)
- Prometheus metrics handlers
- Domain-event / invalidation helpers
- Batched-publish helpers

## Review history

This PR represents the v1.1 refactor addressing all 6 points from @grivera64's review on #4196 (https://github.com/gofiber/fiber/pull/4196#issuecomment-4263568773) and aligning with Fiber's internal middleware patterns (session, cache, limiter, keyauth).

Summary of the v1.1 changes vs. the earlier slim push:

| # | Review point | Change |
|---|---|---|
| 1 | Remove `Next` from `Config` | Done — SSE is terminal; documented |
| 2 | `coalescer` → `Dispatcher` with `AddEvent` / `AddState` / `WriteTo` | Done |
| 3 | `FanOut()` → `Config.Bridges []BridgeConfig` + `SubscriberBridge` | Done — auto-lifecycle |
| 4 | Reduce allocations | `MarshaledEvent.WriteTo` via `bytebufferpool`: **0 B/op, 0 allocs** (was 381 B / 8 allocs) at 26 ns/op |
| 5 | E2E tests with real client | Added — real TCP listener validating `Content-Type`, `retry`, `connected`, multi-line `data`, `id`/`event` injection sanitisation |
| 6 | Folder cleanup | `sse.go` 671→154 lines; all Hub methods in `hub.go`; bridges in `bridge.go` |

Plus:
- `c.Abandon()` before `SendStreamWriter` (prevents `fiber.Ctx` pool reuse mid-stream)
- `t.Parallel()` on every test per `AGENTS.md`
- Replayer errors logged and best-effort continued (no silent drops)

## Status

- `go build`, `go vet`, `go test -race` — all clean
- `golangci-lint run` — 0 issues
- Coverage **90.7%**

## Notes

Happy to split the diff into smaller commits on this branch if that would make review easier — just say the word and I'll layer it in without force-pushing. Also happy to adjust scope either direction (further shrink or grow) based on what fits best for Fiber's middleware collection.

Thanks again @grivera64 and @gaby for the patient review — the design here owes a lot to your feedback.

Closes #4196


vinod-morya:
Thanks for the bot reviews — worked through the critical and major findings. New commit: [vinod-morya/fiber@0c1aeb4](https://github.com/vinod-morya/fiber/commit/0c1aeb49)

**Critical**
- Closed the replay/live delivery gap — register the connection before writing preamble/replay so live events buffer in `conn.send` during the replay window. Monotonic IDs prevent duplicates.
- Normalise CR and CRLF to LF before splitting the `data` field so caller input with `\r` or `\r\n` doesn't produce malformed SSE frames.

**Major**
- `Shutdown(ctx)` now honors `ctx` while waiting for bridges — a wedged bridge can't exceed the deadline.
- `Publish` rejects early while draining so a racing `Shutdown` can't inflate `EventsPublished`.
- Shutdown event ordering: run loop now broadcasts `server-shutdown` to all connections, sleeps `drainDelay`, **then** closes. Replaces the per-connection `watchShutdown` goroutines whose `Close()` could beat the flush.
- Bridge loop applies the retry backoff to any early return (not just errors) so a misbehaving `Subscribe` returning `nil` immediately can't spin hot.
- `OnConnect` errors no longer reach the client — generic 403 body.
- `adaptiveThrottler` clamps `min` and `max` around `baseInterval` to preserve `min ≤ base ≤ max` for any config.

**Minor**
- Bridge tests use `select`+timeout instead of bare receives.
- `Replayer.Store` errors logged (was silent); still best-effort.
- `docs/middleware/sse.md`: dropped the `Next` row, replaced `FanOut`/`PubSubSubscriber` example with `Config.Bridges`/`SubscriberBridge`, added the `Bridges` config row.

Out of scope for this pass (happy to file follow-ups if you'd like them in this PR):
- Wildcard routing and group-only scan are O(N·M) / O(N). A topic trie and a metadata-index would help at scale — real issue, but probably deserves its own benchmark-backed change once merged.
- `sync.Pool` for the per-routing `seen` map — same reasoning.

`go build` / `go vet` / `go test -race` / `golangci-lint run` — all clean. Ready for another look whenever you have time, @grivera64 @gaby. Thanks again.

vinod-morya:
Addressed the follow-up findings: [vinod-morya/fiber@ea972cf](https://github.com/vinod-morya/fiber/commit/ea972cfa)

- `MarshaledEvent.Retry`: guard tightened to `> 0` so the zero value doesn't emit `retry: 0` (which per spec triggers immediate reconnect — reconnect-storm risk when an external `Replayer` constructs the struct without setting the field). Field doc updated to state non-positive values are omitted.
- Added a dedicated `Test_..._RetryZeroOmitted` test plus a `NotContains(retry:)` assertion on the existing happy-path test to catch regressions.
- `OnConnect`: failed auth now logs the original error via `log.Warnf` before returning a generic 403 body. Operators keep the diagnostic signal; the unauthenticated client still sees only `"forbidden"`. Matches the replayer-error logging pattern already used in `hub.go`. Docstring now reflects actual behaviour.

`go build` / `go vet` / `go test -race` / `golangci-lint run` — all clean.

vinod-morya:
Addressed the third-pass findings plus a short adversarial sweep on related paths: [vinod-morya/fiber@9fc8e2a](https://github.com/vinod-morya/fiber/commit/9fc8e2a2)

**Reported**
- `MarshaledEvent.WriteTo` now sanitizes `ID` and `Type` at the write boundary, not only in `marshalEvent`. External `Replayer` implementations constructing the struct directly can no longer inject extra SSE fields via embedded `\r` / `\n`. Defense in depth — `WriteTo` is the last line between an event and the client.
- Removed the explicit `case json.Marshaler:` branch. A typed-nil pointer whose pointer-type implements `MarshalJSON` matched the interface and the manual call panicked when the method dereferenced the receiver. `json.Marshal` in the default branch is nil-safe (emits `null`), so the special case wasn't buying anything.

**Adversarial sweep (same class of issues, caught while touching the files)**
- `bridge.go`: the user-supplied `Transform` invocation is now wrapped in a `recover`. A panic in `Transform` previously propagated into the `Subscriber` callback and could tear down the bridge goroutine, leaking `h.bridges.Done()` and hanging `Shutdown` forever.
- `hub.go`: all `BridgeConfig`s are now validated (nil `Subscriber`, empty `Channel`) **before** any goroutine is started. The previous ordering panicked mid-loop, leaving earlier bridges' goroutines running with no owner to cancel them.
- `event.go`: `writeRetry` returns a no-op for `ms <= 0`, matching the `MarshaledEvent.WriteTo` rule. A zero or negative hint would otherwise tell clients to reconnect immediately.

**Tests**
- `Test_SSE_MarshaledEvent_WriteTo_SanitizesInjectionAtBoundary` — asserts exactly one `id:` line, one `event:` line, and one frame terminator even when `ID`/`Type` contain injection attempts.
- `Test_SSE_MarshaledEvent_WriteTo_TypedNilJSONMarshaler` — typed-nil pointer with a dereferencing `MarshalJSON` produces `data: null` without panicking.

`go build` / `go vet` / `go test -race` / `golangci-lint run` — all clean.

vinod-morya:
Addressed the fourth-pass findings: [vinod-morya/fiber@5b3665e](https://github.com/vinod-morya/fiber/commit/5b3665ec)

- **Bridge validation ordering.** Moved `BridgeConfig` validation ahead of `go hub.run()`. The previous ordering started the run-loop goroutine first, so a panic from bad config leaked a zombie `run()` — `NewWithHub` never returns the hub to its caller for `Shutdown`. Validation is now the first thing `NewWithHub` does; a bad config aborts before any goroutine is spawned.
- **Retry test now observes an actual retry.** The old test slept 100ms with `bridgeRetryDelay` pinned at 3s, so it passed after the first error without ever observing a retry. Changed `bridgeRetryDelay` from a `const` to a package `var` so the test can shorten it to 20ms (restored via `t.Cleanup`), blocked the second Subscribe call on `ctx.Done` with a signal channel, and asserted `attempts == 2` before shutdown. Test now runs deterministically in ~220ms.

`go build` / `go vet` / `go test -race` / `golangci-lint run` — clean.

vinod-morya:
Addressed the fifth-pass findings: [vinod-morya/fiber@2b6b56f](https://github.com/vinod-morya/fiber/commit/2b6b56f8)

- `Test_SSE_Publish_BufferFull`: now asserts `EventsDropped > 0` in addition to `EventsPublished > 0`. The old assertion would pass even if the non-blocking `default:` branch in `Publish` regressed to blocking behavior — the dropped-counter is the invariant this test actually pins.
- `Test_SSE_Shutdown_Timeout`: removed `t.Parallel()` and now awaits `hub.stopped` before returning. The previous form exited while `run()` was still in the shutdown path (~200ms drain delay), leaking the goroutine across parallel tests and racing on `hub.connections`. The test also now asserts the expected `context.Canceled` error instead of punting on the assertion.

`go build` / `go vet` / `go test -race` / `golangci-lint run` — clean.

ReneWerner87:
I will check it tomorrow

gaby:
@vinod-morya The lint workflow is failing. The `test coverage` needs some work, all the files triggered coverage checks.

vinod-morya:
Thanks @gaby — addressed both items in [vinod-morya/fiber@43bc1fe](https://github.com/vinod-morya/fiber/commit/43bc1fea).

**Lint.** The `//nolint:gosec` on the bridge context had become stale on CI's linter version and `nolintlint` blocked the build. Swapped the stored `bridgeCancel` `CancelFunc` for a goroutine tied to `hub.shutdown` — `close(h.shutdown)` now fans the cancel out alongside the run loop and watchers, and because `cancel` is visible at goroutine scope, gosec G118 no longer needs suppressing. Removed the `bridgeCancel` field and the matching nil-check in `Shutdown`.

**Coverage.** Added focused unit tests for previously-untested branches — overall `middleware/sse` coverage is now **91.3%** locally. New tests cover:
- `Publish`: drain-path drop + `TTL` timestamp stamping
- `writeRetry`: non-positive `ms` no-op
- `trackEventType`: empty-type fallback to `"message"`
- `matchGroupConns`: empty-group early return
- `watchLifetime`: no-op when `MaxLifetime <= 0`
- `replayEvents`: nil replayer / empty last-ID / replayer error / full write+flush
- `initStream`: write-error propagation
- `sendConnectedEvent`: write-error propagation
- `writeLoop`: heartbeat, real event, done-exit — via a `failingWriter` helper

`go build` / `go vet` / `go test -race` / `golangci-lint run` — all clean. CI should be green on the next run.

The @claude audit is also welcome — happy to act on anything it surfaces.

gaby:
@claude Conduct a security audit of this implementation to identify issues such as data races and potential panics during runtime.

middleware/sse/sse_test.go
```diff
@@ -0,0 +1,2256 @@
...
+	// First: retry directive frame from writeRetry.
+	retryFrame := readSSEFrame(t, br)
+	require.Contains(t, retryFrame, "retry: 3000")
+
+	// Second: connected event with connection_id and topics.
+	connectedFrame := readSSEFrame(t, br)
+	require.Contains(t, connectedFrame, "event: connected")
+	require.Contains(t, connectedFrame, "connection_id")
+	require.Contains(t, connectedFrame, `"topics":["updates"]`)
+}
```
grivera64, line 131: Can you re-write these tests using our standard `app.Test()` method? And please use stricter checks to ensure that the output of the middleware is exactly as expected to avoid any regressions in the future.

For example:

```go
func Test_SSE_E2E_HeadersAndConnectedFrame(t *testing.T) {
        t.Parallel()

        app := fiber.New()
        app.Get("/events", New(Config{
                MaxLifetime:       500 * time.Millisecond,
                HeartbeatInterval: 100 * time.Millisecond,
                FlushInterval:     50 * time.Millisecond,
                OnConnect: func(_ fiber.Ctx, conn *Connection) error {
                        conn.Topics = []string{"updates"}
                        // Manual ID for deterministic testing
                        conn.ID = "fiber"
                        return nil
                },
        }))

        req, err := http.NewRequest("GET", "/events", nil)
        require.NoError(t, err)

        resp, err := app.Test(req, fiber.TestConfig{
                Timeout: 5 * time.Second,
        })
        require.NoError(t, err)

        // RFC 8895 + W3C SSE: Content-Type must be text/event-stream.
        require.Equal(t, "text/event-stream", resp.Header.Get("Content-Type"))
        require.Equal(t, "no-cache", resp.Header.Get("Cache-Control"))
        require.Equal(t, "keep-alive", resp.Header.Get("Connection"))
        require.Equal(t, "no", resp.Header.Get("X-Accel-Buffering"))
        require.Equal(t, http.StatusOK, resp.StatusCode)

        body, err := io.ReadAll(resp.Body)
        defer resp.Body.Close()
        require.NoError(t, err)

        frames := strings.Split(string(body), "\n\n")

        require.Equal(t, "retry: 3000", frames[0])
        require.Equal(t, "id: evt_1\nevent: connected\ndata: {\"connection_id\":\"fiber\",\"topics\":[\"updates\"]}", frames[1])
}
```

middleware/sse/sse_test.go
```diff
@@ -0,0 +1,2256 @@
...
+	_ = readSSEFrame(t, br) // connected
+
+	// Give the hub time to register the connection before publishing.
+	time.Sleep(50 * time.Millisecond)
+
+	hub.Publish(Event{
+		Type:     "order-created",
+		Data:     `{"id":"ord_123","total":99}`,
+		Topics:   []string{"orders"},
+		Priority: PriorityInstant,
```
grivera64, line 168: In addition to the previous comment, can you add checks that verify the `Priority` field works as expected? Based on the test, having any priority here wouldn't affect the outcome of the test.

middleware/sse/sse_test.go
```diff
@@ -0,0 +1,2256 @@
...
+	_ = readSSEFrame(t, br)
+	_ = readSSEFrame(t, br)
+	time.Sleep(50 * time.Millisecond)
+
+	hub.Publish(Event{
+		Type:     "log",
+		Data:     "line1\nline2\nline3",
+		Topics:   []string{"logs"},
+		Priority: PriorityInstant,
+	})
```
grivera64, line 211: Similar to the priority field, the type and topics should be strictly checked here.

middleware/sse/sse_test.go
```diff
@@ -0,0 +1,2256 @@
...
+	for _, line := range lines {
+		if strings.HasPrefix(line, "id: ") {
+			idLines++
+		}
+		if strings.HasPrefix(line, "event: ") {
+			eventLines++
+		}
+	}
+	require.Equal(t, 1, idLines, "id: injection must be sanitized")
+	require.Equal(t, 1, eventLines, "event: injection must be sanitized")
```
grivera64, line 273: Since the injection is sanitized, we should also have a check for what it should have sanitized to.

middleware/sse/config.go
```diff
@@ -0,0 +1,117 @@
...
+	// events are dropped and the client should reconnect.
+	//
+	// Optional. Default: 256
+	SendBufferSize int
+
+	// RetryMS is the reconnection interval hint sent to clients via the
+	// retry: directive on connect.
+	//
+	// Optional. Default: 3000
+	RetryMS int
```
grivera64, line 81: Let's also use type `time.Duration` here, and rename to `ClientRetryInterval` or something similar to be consistent with the other configs here.

middleware/sse/event.go
```diff
@@ -0,0 +1,210 @@
...
+	Type        string
+	ID          string
+	CoalesceKey string
+	Topics      []string
+	TTL         time.Duration
+	Priority    Priority
+}
+
+// globalEventID is an auto-incrementing counter for event IDs.
+var globalEventID atomic.Uint64
```
grivera64, line 48: We should most likely not use a global event ID here. From my understanding, it is okay for events across separate handlers should be allowed to use overlapping event ID's.

In this case, we could track number event ID's for each handler independently.

middleware/sse/hub.go
```diff
@@ -0,0 +1,586 @@
...
+			}
+			h.topicIndex[topic][conn.ID] = struct{}{}
+		}
+	}
+	if hasWildcard {
+		h.wildcardConns[conn.ID] = struct{}{}
+	}
+
+	log.Infof("sse: connection opened conn_id=%s topics=%v total=%d",
+		conn.ID, conn.Topics, len(h.connections))
```
grivera64, line 371: I think this was left behind on accident.

middleware/sse/hub.go
```diff
@@ -0,0 +1,586 @@
...
+			}
+		}
+	}
+
+	delete(h.wildcardConns, conn.ID)
+	delete(h.connections, conn.ID)
+	h.throttler.remove(conn.ID)
+
+	log.Infof("sse: connection closed conn_id=%s sent=%d dropped=%d total=%d",
+		conn.ID, conn.MessagesSent.Load(), conn.MessagesDropped.Load(), len(h.connections))
```
grivera64, line 397: Same with these logs

middleware/sse/hub.go
```diff
@@ -0,0 +1,541 @@
...
+		}
+		h.deliverToConn(conn, event, me)
+	}
+}
+
+// matchConnections collects all connection IDs that should receive the event.
+// When both Topics and Group are set, only connections matching BOTH are
+// included (intersection semantics) to prevent tenant/topic leaks (CRITICAL-1).
+func (h *Hub) matchConnections(event *Event) map[string]struct{} {
+	seen := make(map[string]struct{})
```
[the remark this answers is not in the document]
grivera64: Instead of querying matched connections, shouldn't we just subscribe each connection to their intended event? From my understanding, the events that a connection expects doesn't change during the lifetime of the connection. If so, doing this `matchConnections` check isn't necessary, and we can remove it.

middleware/sse/hub.go
```diff
@@ -0,0 +1,541 @@
...
+			h.matchGroupConns(event, seen)
+		}
+	}
+
+	return seen
+}
+
+// matchWildcardConns adds wildcard-subscribed connections that match the event topics.
+func (h *Hub) matchWildcardConns(event *Event, seen map[string]struct{}) {
+	for connID := range h.wildcardConns {
```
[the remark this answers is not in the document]
grivera64: Same as my above comment, I don't think we need to run these matches on each publish, but rather just subscribe connections to events on creation.

middleware/sse/hub.go
```diff
@@ -0,0 +1,541 @@
...
+		}
+	}
+}
+
+// matchGroupConns adds connections that match the event's group metadata.
+func (h *Hub) matchGroupConns(event *Event, seen map[string]struct{}) {
+	if len(event.Group) == 0 {
+		return
+	}
+	for connID, conn := range h.connections {
```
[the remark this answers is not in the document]
grivera64: Same as the previous two.

middleware/sse/hub.go
```diff
@@ -0,0 +1,586 @@
...
+
+// flushAll drains each connection's dispatcher and sends buffered events.
+func (h *Hub) flushAll() {
+	h.mu.RLock()
+	conns := make([]*Connection, 0, len(h.connections))
+	for _, conn := range h.connections {
+		if !conn.IsClosed() && !conn.paused.Load() {
+			conns = append(conns, conn)
+		}
+	}
```
grivera64, line 536: Is creating a slice here necessary? This adds extra memory allocations on each flush, but it is only used in the for loop and GC'ed afterwards.

middleware/sse/hub.go
```diff
@@ -0,0 +1,586 @@
...
+	go func() {
+		timer := time.NewTimer(h.cfg.MaxLifetime)
+		defer timer.Stop()
+		select {
+		case <-timer.C:
+			conn.Close()
+		case <-conn.done:
+		}
+	}()
+}
```
grivera64, line 269: Do we really need a separate goroutine to track lifetime per connection? In `c.SendStreamWriter()`, no extra close logic is necessary by default. Spawning in too many goroutines will add some overhead.

The error returned by a failed write or flush should be enough to tell a connection closed.

middleware/sse/example_test.go
grivera64, line 1: We should remove this `example_test.go` in the middleware folder. This could be added separately in the `docs/middleware/sse.md` file, or as a recipe in the future.

grivera64 (commented):
Thanks for making the PR, I've left a partial review on the changes below.

I also want to ask, would it be better to limit the SSE middleware to only the required features necessary for Server-Side Events? Features like topics, priority, and tags are more application specific rather than SSE protocol specific. Since this would be a core SSE middleware, we should avoid forcing every user to have the overhead and complexity of a Pub/Sub engine, even if they only need a simple stream.

To make these features optional, we could add the extra logic as independent add-ons. For example, a Hub (or event stream) can be optionally added to `sse.Config{}` to take advantage of `hub.Publish()`. Same thing goes for `Dispatcher`, `Replayer`, and other optional features.

By separating the core SSE features from extra features, the SSE middleware stays simple while also providing an interface for more complex use cases.

middleware/sse/example_test.go
efectn: Agree, this should be removed and added as a a recipe

ReneWerner87:
replaced by https://github.com/gofiber/fiber/pull/4239

PR #4231, opened 2026-04-24
refactor: optimize speed

ReneWerner87:
This PR applies small, benchmark-backed optimizations in Fiber core paths:

- Simplifies parameterized route matching by returning `routeParser.getMatch(...)` directly.
- Optimizes `Ctx.BaseURL()` string construction with an exactly sized byte buffer.
- Optimizes `Ctx.String()` by writing the 16-byte hex request ID directly instead of using `strconv.FormatUint`.
- Updates the spell-check workflow Node.js version to 24.

## Performance

Benchmarked with `go test -bench ... -benchmem -count=20` and compared with `benchstat`.

| Benchmark | Before | After | Change |
|---|---:|---:|---:|
| `Benchmark_Router_Handler` | `87.98 ns/op` | `84.44 ns/op` | `-4.03%` |
| `Benchmark_Route_Match` | `13.20 ns/op` | `13.06 ns/op` | `-1.06%` |
| `Benchmark_Ctx_BaseURL_Uncached` | `47.99 ns/op` | `39.48 ns/op` | `-17.72%` |
| `Benchmark_Ctx_BaseURL` | `2.054 ns/op` | `1.962 ns/op` | `-4.50%` |
| `Benchmark_Ctx_String` | `192.3 ns/op` | `171.6 ns/op` | `-10.79%` |

No allocation regressions were observed in these benchmarks.

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -11,7 +11,6 @@ import (
 	"io"
 	"maps"
 	"mime/multipart"
-	"strconv"
 	"strings"
 	"sync/atomic"
 	"time"
@@ -109,7 +108,13 @@ func (c *DefaultCtx) BaseURL() string {
 	if c.baseURI != "" {
 		return c.baseURI
 	}
-	c.baseURI = c.Scheme() + "://" + c.Host()
+	scheme := c.Scheme()
+	host := c.Host()
+	buf := make([]byte, 0, len(scheme)+len("://")+len(host))
+	buf = append(buf, scheme...)
+	buf = append(buf, "://"...)
+	buf = append(buf, host...)
+	c.baseURI = c.app.toString(buf)
 	return c.baseURI
 }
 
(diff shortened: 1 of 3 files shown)


ctx.go
```diff
@@ -109,7 +108,13 @@ func (c *DefaultCtx) BaseURL() string {
 	if c.baseURI != "" {
 		return c.baseURI
 	}
-	c.baseURI = c.Scheme() + "://" + c.Host()
+	scheme := c.Scheme()
+	host := c.Host()
+	buf := make([]byte, 0, len(scheme)+len("://")+len(host))
```
gaby, line 113: I think the max length of a hostname is RFC defined. If so, we can use a sync.Pool with fixed values here

gaby: https://stackoverflow.com/questions/8724954/what-is-the-maximum-number-of-characters-for-a-host-name-in-unix#8725032

gaby: We should efforce this limit at the router/app level if we don't already.

ReneWerner87: Ok i will check

efectn: @ReneWerner87 did you also test it with bytebufferpool or string.Builder + pool? I don't know if it is overkill for a ctx method like BaseURI, but might be worth trying. Regarding the maxiumum length of hostname, i agree about it, but i don't think we need to allocate that much space, maybe around 50-60

ReneWerner87: I tested `strings.Builder`, `bytebufferpool`, and a fixed-capacity `sync.Pool` variant for `BaseURL()`. The current exactly-sized `[]byte` build was fastest:

- current: ~37.31 ns/op
- `strings.Builder`: ~39.15 ns/op
- `bytebufferpool`: ~44.81 ns/op
- `sync.Pool`: ~42.93 ns/op

All variants kept the same allocation profile, so I kept the current implementation.

ReneWerner87: @gaby for the limit

Checked this: Fiber already enforces the RFC 1035 253-character limit in the domain router path. `parseDomainPattern` rejects patterns longer than 253 chars, and `matchDomain` returns no match for empty or >253-char hostnames.

I did not add a global app/router-level Host rejection in this performance PR because that would be a behavioral validation change, not just a buffer optimization. It should probably be handled separately with explicit tests and a decision on the desired rejection behavior.

PR #4239, opened 2026-04-26
🔥 feat: add lightweight SSE middleware

ReneWerner87:
## Summary

This PR adds a lightweight `middleware/sse` package for Fiber v3.

The middleware focuses on the SSE transport primitives that need to be
Fiber-native:

- `SendStreamWriter` based streaming
- SSE response headers
- event, comment, and retry frame formatting
- per-write flushing
- heartbeat comments
- `Last-Event-ID` access
- stream lifecycle cancellation through `Stream.Context()`
- disconnect detection through write/flush errors

It deliberately does not include a built-in hub, topic router, replay
store, auth layer, metrics endpoint, or external pub/sub bridge. Those
features are useful, but they are application-specific and better suited
for recipes, contrib packages, or user code built around this primitive.

## Background

This follows the discussion and review history from:

- gofiber/fiber#4194
- gofiber/fiber#4196
- gofiber/fiber#4225
- https://github.com/vinod-morya/fibersse
- https://github.com/gofiber/recipes/tree/master/sse

The main direction from review was to keep Fiber core focused on the SSE
protocol and streaming lifecycle, rather than shipping a full pub/sub
framework in middleware.

## API

```go
app.Get("/events", sse.New(sse.Config{
    Handler: func(c fiber.Ctx, stream *sse.Stream) error {
        return stream.Event(sse.Event{
            Name: "message",
            Data: fiber.Map{"message": "hello"},
        })
    },
}))

### DIFF
--- a/docs/middleware/sse.md
+++ b/docs/middleware/sse.md
@@ -20,6 +20,7 @@ Import the middleware package:
 
 ```go
 import (
+    "context"
     "time"
 
     "github.com/gofiber/fiber/v3"
@@ -41,16 +42,27 @@ app.Get("/events", sse.New(sse.Config{
 }))
 ```
 
-For long-running streams, wait on your own event source and stop when the client disconnects:
+For long-running streams, subscribe each client to its own event channel and stop when the client disconnects.
+A single shared channel load-balances messages across clients; use a fan-out source when every client must receive every event:
 
 ```go
-events := make(chan string)
+type Broker interface {
+    Subscribe(ctx context.Context) (<-chan string, error)
+}
 
 app.Get("/events", sse.New(sse.Config{
     Handler: func(c fiber.Ctx, stream *sse.Stream) error {
+        events, err := broker.Subscribe(stream.Context())
+        if err != nil {
+            return err
+        }
+
         for {
             select {
-            case msg := <-events:
+            case msg, ok := <-events:
+                if !ok {
+                    return nil
+                }
                 if err := stream.Event(sse.Event{Name: "message", Data: msg}); err != nil {
                     return err
                 }
--- a/docs/middleware/sse.md
+++ b/docs/middleware/sse.md
@@ -1,0 +1,119 @@
+---
+id: sse
+---
+
+# SSE
+
+The SSE middleware provides the transport pieces for Server-Sent Events: response headers, event formatting, flushing, heartbeat comments, and disconnect detection through `Flush` errors.
+
+It intentionally does not include a hub, topics, authentication, replay storage, metrics, or external pub/sub bridges. Those are application concerns that can be composed around the stream handler.
+
+## Signatures
+
+```go
+func New(config ...Config) fiber.Handler
+```
+
+## Examples
+
+Import the middleware package:
+
+```go
+import (
+    "time"
+
+    "github.com/gofiber/fiber/v3"
+    "github.com/gofiber/fiber/v3/middleware/sse"
+)
+```
+
+Once your Fiber app is initialized, mount an SSE endpoint like this:
+
+```go
+app.Get("/events", sse.New(sse.Config{
+    Retry: 5 * time.Second,
+    Handler: func(c fiber.Ctx, stream *sse.Stream) error {
+        return stream.Event(sse.Event{
+            Name: "message",
+            Data: fiber.Map{"message": "hello"},
+        })
+    },
+}))
+```
+
+For long-running streams, wait on your own event source and stop when the client disconnects:
+
+```go
+events := make(chan string)
+
+app.Get("/events", sse.New(sse.Config{
+    Handler: func(c fiber.Ctx, stream *sse.Stream) error {
+        for {
+            select {
+            case msg := <-events:
+                if err := stream.Event(sse.Event{Name: "message", Data: msg}); err != nil {
+                    return err
+                }
+            case <-stream.Done():
+                return stream.Err()
+            }
+        }
+    },
+}))
+```
+
+`stream.Context()` is canceled when the stream ends or a write fails, which makes it convenient to pass into database, broker, or gRPC calls:
+
+```go
+app.Get("/events", sse.New(sse.Config{
+    Handler: func(c fiber.Ctx, stream *sse.Stream) error {
+        rows, err := db.QueryContext(stream.Context(), "SELECT id FROM jobs")
+        if err != nil {
+            return err
+        }
+        defer rows.Close()
+
+        return stream.Comment("connected")
+    },
+}))
+```
+
+## Config
+
+| Property          | Type                         | Description                                   | Default             |
+|:------------------|:-----------------------------|:----------------------------------------------|:--------------------|
+| Next              | `func(fiber.Ctx) bool`       | Skip when the function returns `true`.         | `nil`               |
+| Handler           | `sse.Handler`                | Writes events to the stream.                   | `nil`               |
+| OnClose           | `func(fiber.Ctx, error)`     | Called when the stream ends, with `nil` when the handler returned successfully and no stream write failed. | `nil` |
+| Retry             | `time.Duration`              | Initial EventSource reconnect delay.           | `0`                 |
+| HeartbeatInterval | `time.Duration`              | Interval for SSE comment heartbeats.           | `15 * time.Second`  |
+| DisableHeartbeat  | `bool`                       | Disable automatic heartbeat comments.          | `false`             |
+
+## Default Config
+
+```go
+var ConfigDefault = Config{
+    Next:              nil,
+    Handler:           nil,
+    OnClose:           nil,
+    Retry:             0,
+    HeartbeatInterval: 15 * time.Second,
+    DisableHeartbeat:  false,
+}
+```
+
+## Stream
+
+```go
+func (s *Stream) Event(event Event) error
+func (s *Stream) Comment(comment string) error
+func (s *Stream) Retry(retry time.Duration) error
+func (s *Stream) Context() context.Context
+func (s *Stream) Done() <-chan struct{}
+func (s *Stream) Err() error
+func (s *Stream) LastEventID() string
+```
+
+Every write is flushed. A failed flush closes `Done`, stores the error returned by `Err`, and lets the handler stop without relying on `fasthttp.RequestCtx.Done`, which is not a per-client disconnect signal. After a normal handler return, `Done` and `Context()` are closed while `Err()` remains `nil`; writes after that return `sse: stream closed`.
+
+`Config.Retry` sends the initial reconnect delay when the stream opens. `Event.Retry` changes the reconnect delay for a specific event, following the SSE wire format.
--- a/docs/middleware/sse.md
+++ b/docs/middleware/sse.md
@@ -99,7 +99,7 @@ app.Get("/events", sse.New(sse.Config{
 | OnClose           | `func(fiber.Ctx, error)`     | Called when the stream ends, with `nil` when the handler returned successfully and no stream write failed. | `nil` |
 | Retry             | `time.Duration`              | Initial EventSource reconnect delay.           | `0`                 |
 | HeartbeatInterval | `time.Duration`              | Interval for SSE comment heartbeats.           | `15 * time.Second`  |
-| DisableHeartbeat  | `bool`                       | Disable automatic heartbeat comments.          | `false`             |
+| DisableHeartbeat  | `bool`                       | Disable automatic heartbeat comments. When disabled, disconnected clients may not be detected until the next write. | `false` |
 
 ## Default Config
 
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: middleware/sse/sse_test.go [hunk 1]
```diff
@@ -0,0 +1,294 @@
...
+	const writers = 16
+	errs := make(chan error, writers)
+	var wg sync.WaitGroup
+	wg.Add(writers)
+	for i := range writers {
+		go func(data int) {
+			defer wg.Done()
+			errs <- stream.Event(Event{Name: "message", Data: data})
+		}(i)
+	}
```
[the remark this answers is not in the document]
ReneWerner87: Updated to a conventional indexed loop in `Test_SSE_StreamConcurrentWrites` and pushed it in 1aacd75f. Verified with `go test ./middleware/sse ./middleware/limiter`, `make lint`, and full `make test`.

docs/middleware/sse.md
```diff
@@ -0,0 +1,119 @@
...
+func (s *Stream) Event(event Event) error
+func (s *Stream) Comment(comment string) error
+func (s *Stream) Retry(retry time.Duration) error
+func (s *Stream) Context() context.Context
+func (s *Stream) Done() <-chan struct{}
+func (s *Stream) Err() error
+func (s *Stream) LastEventID() string
+```
+
+Every write is flushed. A failed flush closes `Done`, stores the error returned by `Err`, and lets the handler stop without relying on `fasthttp.RequestCtx.Done`, which is not a per-client disconnect signal. After a normal handler return, `Done` and `Context()` are closed while `Err()` remains `nil`; writes after that return `sse: stream closed`.
```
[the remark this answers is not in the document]
ReneWerner87: Updated the wording to say `Done` is closed and `Context()` is canceled, then pushed it in 1aacd75f. Verified with `make lint` and full `make test`.

ReneWerner87:
Addressed the remaining SSE review feedback in 1aacd75f:

- event frames are now built before writing to the stream, so marshal errors cannot leave a partial frame buffered
- untyped `Event.Data: nil` now emits no `data:` line, while typed nil values still encode as `null`
- heartbeat stop is idempotent
- heartbeat and `OnClose` tests have wider/timeout-based waits
- the concurrent writer test uses a conventional indexed loop
- docs now use the correct context wording

I also checked the failed CI run. `middleware/sse` passed on macOS and Windows. The macOS 1.26 failure was `middleware/limiter Test_Limiter_Sliding_Window_RecalculatesAfterHandlerDelay`, which depended on a tight sleep around second-resolution limiter windows; that wait is widened in this commit. The Windows jobs were cancelled after the run was interrupted, not because of an SSE test failure.

Local verification:
- `go test ./middleware/sse ./middleware/limiter`
- `go test ./middleware/limiter -run Test_Limiter_Sliding_Window_RecalculatesAfterHandlerDelay -count=10`
- `make lint`
- `make test`
- `go test ./...`

`make audit` still fails on `govulncheck` findings in the Go 1.25 standard library, after `go mod verify` and `go vet ./...` pass. `make format` hits a local `.git/logs/...replace-cookie-lookup-in-router.go` path in this checkout, so I formatted the touched Go files directly with gofumpt.

middleware/sse/sse.go
```diff
@@ -0,0 +1,223 @@
...
+				if stopHeartbeat != nil {
+					defer stopHeartbeat()
+				}
+			}
+
+			streamErr = cfg.Handler(c, stream)
+			if streamErr == nil {
+				streamErr = stream.Err()
+			}
+		})
```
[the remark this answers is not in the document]
ReneWerner87: Fixed in a039e4ab. `OnClose` now computes a final error from `streamErr` and re-checks `stream.Err()` when the snapshot is nil, so a late latched heartbeat/write error is preserved. Verified with `go test ./middleware/sse` and `make lint`.

ReneWerner87:
Addressed the latest coverage/review feedback in 1ff8f6d3:

- `writeEvent` validates/marshals `Event.Data` before building any frame fields, then writes the fully built frame to the stream in one step.
- the long-running stream docs no longer show all clients reading from one shared channel; the example now uses a per-connection subscription and calls out the fan-out requirement.
- `stopHeartbeat` now waits for the heartbeat goroutine to exit, closing the late-error window before `OnClose` reads the final stream error.
- local SSE coverage is now 96.4%.

Verified locally:
- `make lint`
- `make test`
- `go test ./middleware/sse ./middleware/limiter`
- `go test ./middleware/sse -coverprofile=/tmp/sse.cover`

middleware/sse/event.go
```diff
@@ -0,0 +1,155 @@
...
+		}
+		appendField(&frame, "id", id)
+	}
+	if event.Name != "" {
+		name, err := sanitizeField(event.Name)
+		if err != nil {
+			return fmt.Errorf("sse: invalid event: %w", err)
+		}
+		appendField(&frame, "event", name)
+	}
```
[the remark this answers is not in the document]
ReneWerner87: Fixed in dcd817d4. `writeEvent` now re-checks sanitized `ID` and `Name` before appending fields, so whitespace-only values do not emit empty `id:` or `event:` lines. Added `Test_SSE_EventOmitsWhitespaceOnlyFields`; verified with `go test ./middleware/sse ./middleware/limiter` and `make lint`.

middleware/sse/event.go
```diff
@@ -0,0 +1,155 @@
...
+	w.WriteString(value) //nolint:errcheck // bytes.Buffer writes never fail.
+	w.WriteByte('\n')    //nolint:errcheck // bytes.Buffer writes never fail.
+}
+
+func appendData(w *bytes.Buffer, data string) {
+	data = normalizeNewlines(data)
+	for line := range strings.SplitSeq(data, "\n") {
+		appendField(w, "data", line)
+	}
+}
```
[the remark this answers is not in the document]
ReneWerner87: Fixed in dcd817d4. Data framing now trims exactly one trailing newline after normalization for both `appendData` and `writeData`, while preserving intentional blank data lines. Added tests for both cases; verified with `go test ./middleware/sse ./middleware/limiter` and `make lint`.

ReneWerner87:
@grivera64 @vinod-morya @gaby 
This is a simpler version that works better with fiber
Could you check that out?

middleware/sse/config.go
```diff
@@ -0,0 +1,70 @@
...
+
+// Handler writes events to a single SSE stream.
+type Handler func(c fiber.Ctx, stream *Stream) error
+
+// Config defines the config for middleware.
+type Config struct {
+	// Next defines a function to skip this middleware when returned true.
+	//
+	// Optional. Default: nil
+	Next func(c fiber.Ctx) bool
```
grivera64, line 17: SSE is a terminal middleware in nature, since `c.SendStreamWriter()` doesn't run until after all handlers in the call chain return.

In this case, I think `Next()` would convey that you can use `fiber.Ctx` after the middleware when it would lead to undefined behavior.

To avoid confusion, I think we should remove this from the SSE config.

docs/whats_new.md
```diff
@@ -1680,6 +1681,13 @@ The session middleware has undergone significant improvements in v3, focusing on
 
 For more details on these changes and migration instructions, check the [Session Middleware Migration Guide](./middleware/session.md#migration-guide).
 
+### SSE
+
+Fiber now includes a small [SSE middleware](./middleware/sse.md) for Server-Sent Events. It handles native
```
gaby: Remove the word "small"

middleware/sse/event.go
```diff
@@ -0,0 +1,163 @@
...
+	case nil:
+		return eventPayload{}, nil
+	case string:
+		return eventPayload{data: value, hasData: true}, nil
+	case []byte:
+		return eventPayload{data: string(value), hasData: true}, nil
+	case json.RawMessage:
+		return eventPayload{data: string(value), hasData: true}, nil
+	default:
+		encoded, err := json.Marshal(value)
```
gaby: This should use the JSON lib provided by the user in the app.Config

middleware/sse/event.go
```diff
@@ -0,0 +1,163 @@
...
+
+func appendData(w *bytes.Buffer, data string) {
+	data = trimSingleTrailingNewline(normalizeNewlines(data))
+	for line := range strings.SplitSeq(data, "\n") {
+		appendField(w, "data", line)
+	}
+}
+
+func trimSingleTrailingNewline(value string) string {
+	return strings.TrimSuffix(value, "\n")
```
gaby, line 148: Do we have a utils for this?

middleware/sse/sse.go
```diff
@@ -0,0 +1,230 @@
...
+	"context"
+	"errors"
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/gofiber/fiber/v3"
+)
+
+const mimeTextEventStream = "text/event-stream"
```
gaby: Shouldnt this go in constants.go?

docs/whats_new.md
ReneWerner87: Removed "small" from the SSE entry in `docs/whats_new.md` in e71b08f7.

middleware/sse/event.go
ReneWerner87: Fixed in e71b08f7. `New` now passes `c.App().Config().JSONEncoder` into the stream and `Event` uses that encoder for non-string/non-byte data. Added `Test_SSE_NewUsesAppJSONEncoder`; verified with `go test ./middleware/sse`, `go test ./middleware/sse -cover`, `make lint`, and `git diff --check`.

middleware/sse/event.go
ReneWerner87: Checked `utils/v2`. The available trim helpers remove all matching trailing bytes, but this code must remove exactly one normalized trailing newline so intentional blank SSE data lines are preserved. `strings.TrimSuffix` keeps that behavior, covered by `Test_SSE_EventTrimsSingleTrailingDataNewline` and `Test_SSE_EventPreservesIntentionalBlankDataLine`.

middleware/sse/sse.go
ReneWerner87: Moved `mimeTextEventStream` into `middleware/sse/constants.go` in e71b08f7.

vinod-morya:
Thanks for the tag, @ReneWerner87 — and thanks for the simplification. Pulling topics/replay/auth out of the core does make the middleware much more focused, and the `Stream` handler API is a nice shape. Sharing what we learned about this exact surface during the review rounds on #4225, in case any of it's useful here.

**1. `Stream.Context()` does not cancel on client disconnect**

This is the biggest footgun on the slim surface. `c.Context()` (and any context derived from it) is request-bound — `fasthttp.RequestCtx.Done` is *not* the per-client TCP signal. Disconnect is observed only when the next write returns an error.

Concretely: with `DisableHeartbeat: true` and a handler waiting on an external broker (the docstring example), a vanished client is invisible until the handler tries to write. The middleware leans on heartbeats as the disconnect oracle. In #4225 we treated the heartbeat as load-bearing, not optional. A one-liner in the docs near `DisableHeartbeat` warning users about this trade-off would help — happy to send that as a small PR.

**2. `c.Abandon()` ordering (sse.go:41–44)**

```go
c.Abandon()
streamContext := c.Context()
lastEventID := c.Get(fiber.HeaderLastEventID)
return c.SendStreamWriter(...)
```

The convention in Fiber's other long-lived middleware (session, timeout) is to read everything you need from `fiber.Ctx` *before* `Abandon()` and only Abandon as the last step before `SendStreamWriter`. Tests pass today because Abandon doesn't immediately recycle, but the safer ordering is:

```go
streamContext := c.Context()
lastEventID := c.Get(fiber.HeaderLastEventID)
c.Abandon()
return c.SendStreamWriter(...)
```

**3. Handler panic isn't surfaced through `OnClose`**

If `cfg.Handler(c, stream)` panics, fasthttp recovers it at the top level but `OnClose` fires with `nil` or with whatever `streamErr` happened to be before the panic — operators lose the panic value as a signal. A `defer recover()` around the handler invocation that converts `panic → error → OnClose` keeps the operator-visible signal symmetric with the regular error path. ~6 lines.

**4. Heartbeat goroutine + slow consumer**

The defer chain `stopHeartbeat → closeStream → OnClose` is correct, but `stopHeartbeat()` waits on `<-done`, which can only complete after the heartbeat goroutine drains its current `Comment("")` write. On a slow consumer that flush blocks until TCP timeout. Worth flagging as a known footgun rather than redesigning around — the alternative (a single shutdown owner) doesn't fit the hub-less philosophy cleanly.

---

Happy to land any of these as commits on `codex/sse-middleware` if useful — `go test -race -count=1` is clean locally on the current branch. The lightweight scope is the right call for a core middleware; let me know if you'd like the doc note + panic-recovery fix as a small follow-up PR.

ReneWerner87:
Addressed the follow-up from https://github.com/gofiber/fiber/pull/4239#issuecomment-4342123041 in f0c3543d.

Changes:
- read `Context()` and `Last-Event-ID` before `Abandon()`
- convert handler panics into an `OnClose` error signal instead of letting the stream goroutine crash silently
- document the `DisableHeartbeat` disconnect-detection trade-off and the slow-consumer heartbeat shutdown caveat

Verified locally with:
- `go test ./middleware/sse -count=1`
- `go test ./middleware/sse -cover -count=1` (96.6%)
- `go test -race ./middleware/sse -count=1`
- `make lint`
- `git diff --check`

middleware/sse/sse_test.go
grivera64, line 1: Should we also add a unit test that checks if there are no race conditions if a client disconnects early? We can use this "interrupted" unit test from ctx_test.go for reference here:

https://github.com/gofiber/fiber/blob/48a32e76bc43b9fe621b2b529f3f7e5a80807d11/ctx_test.go#L7793

grivera64 (commented):
I know I am a bit late on the review, but I have a few comments that we can amend changes to the PR post-merge if needed.

gaby:
@grivera64 https://github.com/gofiber/fiber/pull/4247

vinod-morya:
Glad all four landed in [`f0c3543d`](https://github.com/gofiber/fiber/commit/f0c3543d) — thanks for the quick turnaround, @ReneWerner87. Anyone digging into the design later will find the full prior context (and a deeper Hub-based approach for use cases that need pub/sub, replay, or multi-tenant routing) in the closed #4225 thread.

For folks reading this who want the richer pub/sub layer on top of Fiber's slim core, [`fibersse`](https://github.com/vinod-morya/fibersse) is the same author and stays compatible with this middleware as the underlying transport.

PR #4243, opened 2026-04-28
🔥feat: Add storage backed SharedState for prefork applications

gaby:
### Motivation

- Provide a storage-backed, prefork-safe key–value state store for data that must be shared across workers or processes. 
- Allow apps to namespace shared keys to avoid collisions (with sensible default prefixing that can include `AppName`) and reuse app-configured encoders/decoders for JSON/MsgPack/CBOR/XML.

### Description

- Add `SharedState` implementation (`shared_state.go`) exposing raw and encoded helpers such as `Set`, `Get`, `SetJSON`, `GetJSON`, `SetMsgPack`, `GetMsgPack`, `SetCBOR`, `GetCBOR`, `SetXML`, `GetXML`, `Delete`, `Has`, and context-aware variants. 
- Wire `SharedState` into the app by adding `Config.SharedStorage` and `Config.SharedStatePrefix`, initializing `app.sharedState` in `New(...)`, and exposing `App.SharedState()` while clarifying that `App.State()` remains process-local. 
- Implement predictable key namespacing (`defaultSharedStatePrefix` + optional `AppName`) and ensure the `SharedState` uses the app-provided encoder/decoder hooks. 
- Update documentation (`docs/api/app.md`, `docs/api/state.md`, `docs/whats_new.md`) and add comprehensive unit tests in `shared_state_test.go` covering configuration, context propagation, serialization, codec usage, namespacing, and error propagation.


ReneWerner87 (commented):
Good feature idea, storage-backed shared state for prefork is a real need. The integration with the existing `fiber.Storage` interface and app config is clean. A few things to address before merge:

---

### 1. Nil encoder panic (Bug - Medium)

Only JSON gets defaults in `newSharedState`:
```go
if jsonEncoder == nil { jsonEncoder = json.Marshal }
if jsonDecoder == nil { jsonDecoder = json.Unmarshal }
```

MsgPack, CBOR, and XML encoders have **no** nil guard. Calling `SetMsgPack("key", val, ttl)` without configuring `Config.MsgPackEncoder` will panic with a nil pointer dereference.

Should either return a descriptive error (e.g. `ErrMsgPackEncoderNotConfigured`) or add nil checks before calling the encoder/decoder in each method.

---

### 2. Massive code duplication (Design - Major)

`SetJSON/GetJSON`, `SetMsgPack/GetMsgPack`, `SetCBOR/GetCBOR`, `SetXML/GetXML` are near-identical — ~200 lines of copy-paste where only the encoder/decoder function and error string differ. This could be reduced to two internal helpers:

```go
func (s *SharedState) setEncoded(ctx context.Context, key string, v any, ttl time.Duration,
    encoder func(any) ([]byte, error), format string) error {
    if s == nil || s.storage == nil {
        return ErrSharedStorageNotConfigured
    }
    if key == "" {
        return nil
    }
    encoded, err := encoder(v)
    if err != nil {
        return fmt.Errorf("fiber: failed to encode shared state %s value: %w", format, err)
    }
    return s.storage.SetWithContext(ctx, s.key(key), encoded, ttl)
}

func (s *SharedState) getEncoded(ctx context.Context, key string, out any,
    decoder func([]byte, any) error, format string) ([]byte, bool, error) {
    if s == nil || s.storage == nil {
        return nil, false, ErrSharedStorageNotConfigured
    }
    data, err := s.storage.GetWithContext(ctx, s.key(key))
    if err != nil {
        return nil, false, err
    }
    if data == nil {
        return nil, false, nil
    }
    if err := decoder(data, out); err != nil {
        return nil, false, fmt.Errorf("fiber: failed to decode shared state %s value: %w", format, err)
    }
    return append([]byte(nil), data...), true, nil
}
```

Each public method becomes a one-liner:
```go
func (s *SharedState) SetJSONWithContext(ctx context.Context, key string, v any, ttl time.Duration) error {
    return s.setEncoded(ctx, key, v, ttl, s.jsonEncoder, "json")
}
```

This also makes the nil-encoder check (#1) fixable in a single place.

---

### 3. Inconsistent empty-key handling (Bug - Medium)

- `Set("")` → calls storage with empty key (`key("")` returns `""`, bypassing prefix)
- `SetJSON("")` → silently returns `nil` (guarded)
- `Get("")` → calls storage with empty key
- `Delete("")` → calls storage with empty key
- `Has("")` → calls storage with empty key

The encoded variants guard against empty keys, but raw `Set`/`Get`/`Delete`/`Has` do not. This should be consistent — either guard in `key()` itself or in all public methods.

Copilot flagged this too — the `key()` function returning `""` for empty input effectively bypasses the prefix, which is likely unintended.

---

### 4. `newSharedState` takes 10 parameters (Style - Low)

Could take a subset of `Config` or a small options struct instead:
```go
func newSharedState(storage Storage, prefix string, cfg Config) *SharedState
```

---

### 5. Dead code: double prefix defaulting (Low)

`app.go` already ensures a non-empty prefix before calling `newSharedState`:
```go
sharedStatePrefix := app.config.SharedStatePrefix
if sharedStatePrefix == "" {
    sharedStatePrefix = defaultSharedStatePrefix
    // ...
}
```

The fallback in `newSharedState` (`if prefix == "" { prefix = defaultSharedStatePrefix }`) is unreachable when called from `New()`.

---

### 6. No `Reset()`/`Close()` forwarding (Low)

The `Storage` interface has `Reset()` and `Close()`. SharedState doesn't expose these. If a user needs to clean up shared state or close the storage connection via SharedState, they currently can't. Consider at least a `Close()` method, or document that the caller manages the storage lifecycle directly.

shared_state.go
```diff
@@ -0,0 +1,310 @@
...
+
+func (s *SharedState) SetMsgPackWithContext(ctx context.Context, key string, v any, ttl time.Duration) error {
+	if s == nil || s.storage == nil {
+		return ErrSharedStorageNotConfigured
+	}
+	if key == "" {
+		return nil
+	}
+
+	encoded, err := s.msgPackEncoder(v)
```
ReneWerner87: **Nil encoder panic**: `s.msgPackEncoder` is called here without a nil check. If `Config.MsgPackEncoder` is not set, this panics with a nil pointer dereference.

Same applies to `SetCBORWithContext` (line 199), `SetXMLWithContext` (line 243), and their Get counterparts (lines 180, 224, 268).

Only JSON gets defaults in `newSharedState`. Either add nil guards returning a descriptive error, or default all encoders.

shared_state.go
```diff
@@ -0,0 +1,310 @@
...
+		return nil, false, err
+	}
+	if data == nil {
+		return nil, false, nil
+	}
+
+	return append([]byte(nil), data...), true, nil
+}
+
+func (s *SharedState) SetJSON(key string, v any, ttl time.Duration) error {
```
ReneWerner87, line 117: **Code duplication**: `SetJSON`, `SetMsgPack`, `SetCBOR`, `SetXML` (and their Get counterparts) are near-identical - only the encoder/decoder and error string differ. ~200 lines of copy-paste.

Could be two internal helpers:
```go
func (s *SharedState) setEncoded(ctx context.Context, key string, v any, ttl time.Duration,
    encoder func(any) ([]byte, error), format string) error { ... }

func (s *SharedState) getEncoded(ctx context.Context, key string, out any,
    decoder func([]byte, any) error, format string) ([]byte, bool, error) { ... }
```
Each public method becomes a one-liner. Also makes the nil-encoder fix a single-place change.

shared_state.go
```diff
@@ -0,0 +1,310 @@
...
+		msgPackDecoder: msgPackDecoder,
+		cborEncoder:    cborEncoder,
+		cborDecoder:    cborDecoder,
+		xmlEncoder:     xmlEncoder,
+		xmlDecoder:     xmlDecoder,
+		prefix:         prefix,
+	}
+}
+
+func (s *SharedState) Set(key string, val []byte, ttl time.Duration) error {
```
ReneWerner87, line 75: **Inconsistent empty-key handling**: `Set` has no guard for empty keys, but `SetJSON` (line 107), `SetMsgPack` (line 151), `SetCBOR` (line 195), `SetXML` (line 239) all have `if key == "" { return nil }`.

Same on the read side - `Get`, `Delete`, `Has` pass empty keys through to storage. The `key()` helper returns `""` for empty input, bypassing the prefix entirely.

Should be consistent - either guard in `key()` itself or in all public methods.

shared_state.go
```diff
@@ -0,0 +1,310 @@
...
+	msgPackEncoder utils.MsgPackMarshal
+	msgPackDecoder utils.MsgPackUnmarshal
+	cborEncoder    utils.CBORMarshal
+	cborDecoder    utils.CBORUnmarshal
+	xmlEncoder     utils.XMLMarshal
+	xmlDecoder     utils.XMLUnmarshal
+	prefix         string
+}
+
+func newSharedState(
```
ReneWerner87: **10 parameters**: This constructor could take a smaller options struct or pass `Config` directly:
```go
func newSharedState(storage Storage, prefix string, cfg Config) *SharedState
```
Simplifies adding new codec support in the future.

shared_state.go
```diff
@@ -0,0 +1,310 @@
...
+	jsonEncoder utils.JSONMarshal,
+	jsonDecoder utils.JSONUnmarshal,
+	msgPackEncoder utils.MsgPackMarshal,
+	msgPackDecoder utils.MsgPackUnmarshal,
+	cborEncoder utils.CBORMarshal,
+	cborDecoder utils.CBORUnmarshal,
+	xmlEncoder utils.XMLMarshal,
+	xmlDecoder utils.XMLUnmarshal,
+) *SharedState {
+	if prefix == "" {
```
ReneWerner87, line 37: **Dead code**: This fallback is unreachable when called from `app.New()`, because `app.go` already ensures a non-empty prefix before calling `newSharedState`.

shared_state.go
```diff
@@ -0,0 +1,310 @@
...
+	"time"
+
+	"github.com/gofiber/utils/v2"
+)
+
+const defaultSharedStatePrefix = "gofiber-shared-state-"
+
+var ErrSharedStorageNotConfigured = errors.New("fiber: shared storage is not configured")
+
+type SharedState struct {
```
ReneWerner87, line 18: **No `Reset()`/`Close()` forwarding**: The `Storage` interface has `Reset()` and `Close()`, but `SharedState` does not expose them. Consider at least a `Close()` method, or document that the caller manages the storage lifecycle separately.

ReneWerner87 (commented):
Inline review comments on specific locations.

gaby:
@copilot+gpt-5.4 address the comments made on this post https://github.com/gofiber/fiber/pull/4243#pullrequestreview-4195770310

gaby:
@copilot+gpt-5.4 Verify each finding against the current code and only fix it if needed.

In `@app.go` around lines 1384 - 1390, OnPostShutdown currently only calls
app.shutdownServices; also close the storage-backed shared state there. In the
app.Hooks().OnPostShutdown callback, after calling
app.shutdownServices(app.servicesShutdownCtx()), check if app.sharedState (or
the field that holds the SharedState) is non-nil and call its Close() method,
logging any error (e.g., log.Errorf("failed to close sharedState: %v", err)).
Ensure you perform the Close() call regardless of service shutdown result and
return nil as before.

gaby:
@copilot+gpt-5.4 Verify each finding against the current code and only fix it if needed.

In `@services_test.go` around lines 260 - 308, The tests set log.SetOutput to a
dummy buffer but restore it incorrectly; capture the current log output before
calling log.SetOutput and restore that captured writer in t.Cleanup (replace the
bytes.NewBuffer(nil) restore) so later tests aren’t affected, and add
t.Parallel() at the start of each subtest (both
"shutdown-hooks/close-shared-state" and
"shutdown-hooks/close-shared-state-after-service-error") to run them in parallel
once the logger sink is isolated; relevant symbols to modify: the two subtest
bodies, the log.SetOutput call, the t.Cleanup restore, and ensure this change is
applied alongside the existing use of shutdownHookStorage, app.initServices, and
app.Hooks().executeOnPostShutdownHooks.

gaby:
@claude[agent] Audit all the changes made in this Pull Request to implement the SharedState feature.

ReneWerner87 (approved):

Issue #4145, opened 2026-03-16
📝 [Proposal]: Provide an effective TCP disconnect handle


bincooo:
### Feature Proposal Description

First of all, I would like to thank the developers for their efforts.
In order to better support stream/sse processing, could you add a listening implementation for TCP disconnection? I have encountered many such situations where the front end actively disconnects the request, but the backend fails to detect it in a timely manner. It is only when the backend task is completed or an attempt is made to write to the stream that a read/write exception is issued.

### Alignment with Express API

I have made a crude implementation, and I am not sure if there is a related implementation for this project. I would like to offer a suggestion:

```go
package v1

import (
	"context"
	"errors"
	"net"
	"syscall"
	"time"

	"github.com/gofiber/fiber/v3"
)

type abortCtx struct {
	parent context.Context
	done   chan struct{}
	err    error
}

// 创建一个监听客户端断开的 Context
func newAbort(ctx fiber.Ctx) (context.Context, context.CancelFunc) {
	cctx := &abortCtx{
		parent: ctx.Context(),
		done:   make(chan struct{}),
	}

	loop := true
	cancel := func() {
		loop = false
		select {
		case <-cctx.done:
			// 已经关闭
		default:
			cctx.err = context.Canceled
			close(cctx.done)
		}
	}

	ch := make(chan byte, 1)
	conn := ctx.RequestCtx().Conn()
	go func() {
		for loop {
			if isClosed(conn) {
				ch <- 1
				loop = false
			}
			time.Sleep(time.Millisecond * 500)
		}
	}()

	// 后台轮询 IsAbandoned
	go func() {
		select {
		case <-cctx.parent.Done():
			select {
			case <-cctx.done:
			default:
				cctx.err = cctx.parent.Err()
				close(cctx.done)
			}
			return

		case <-ch:
			select {
			case <-cctx.done:
			default:
				cctx.err = context.Canceled
				close(cctx.done)
			}
			return

		case <-cctx.done:
			return
		}
	}()

	return cctx, cancel
}

func (c *abortCtx) Deadline() (time.Time, bool) {
	return c.parent.Deadline()
}

func (c *abortCtx) Done() <-chan struct{} {
	return c.done
}

func (c *abortCtx) Err() error {
	select {
	case <-c.done:
		return c.err
	default:
		return nil
	}
}

func (c *abortCtx) Value(key any) any {
	return c.parent.Value(key)
}

func isClosed(conn net.Conn) bool {
[...]

Issue #1718, opened 2022-01-17
🤗 Which context to use to detect client disconnected? 


bofeng:
**Question description**

I have a demo handler like below, and used `curl` to send the request. If in the process I press ctrl-c to quit the curl (client disconnected), the server will still print "ticker", the "client quit" case will never be triggered:

**Code snippet** _Optional_

```go
func hello(c *fiber.Ctx) error {
	ticker := time.NewTicker(1 * time.Second)
	for i := 0; i < 10; i++ {
		select {
		case <-c.Context().Done():
			fmt.Println("client quit")
		case <-ticker.C:
			fmt.Println("ticker")
		}
	}
	return c.SendString("hello")
}
```

I thought that use the `c.Context().Done()` in `fiber` could listen on the "client disconnect" case, but it wasn't triggered. And I know in `gin` you can listen to `c.Request.Context().Done()`, such that if the client disconnected, this case will be triggered properly. 

So did I use the wrong context.Done() to listen to? which context should I use for such case in `fiber`?

DarthBenro008:
@bofeng , Unfortunately, `fasthttp` does not determine or support the detection of client connection as mentioned here: 
https://github.com/valyala/fasthttp/issues/468#issuecomment-439665014

Naturally, GoFiber won't have such context support because the Context of GoFiber is wrapped around fasthttp. 


Hope this answers your query, closing this issue, if you still have doubts, feel free to re-open the issue or join us on discord to have a discussion on the same with community! 😄

Issue #805, opened 2020-09-17
🤗 Should we check for context done/cancel on every handler


WLun001:
**Question description**
Should we check for context done/cancel on every handler? For the case like client cancel the request, and the handler should cancel the ongoing operation (db query etc). 

Possible to achieve with middleware, so that we don't keep repeating?

Fenny:
If you are doing a database query it would be good practice to pass the `c.Context()`, but I personally always set a timeout with the db adaptor instead. It's up to you 👍

WLun001:
@Fenny I get the idea, but how to implement it? `c.Context()` is not type of `context.Context`

How to make the context cancel when the response is canceled by client?

```go
func main() {
	app := fiber.New()

	app.Get("/", func(c *fiber.Ctx) error {
		go func() {
			select {
			case <-c.Context().Done():
				log.Println("done")
			}
		}()
		// assume some heavy operation will take 2 seconds
		// to make it cancelable
		_ = heavyOperation(c.Context())
		return c.SendString("Hello, World 👋!" + time.Now().Format(time.Kitchen))
	})
	log.Fatal(app.Listen(":3000"))
}

func heavyOperation(ctx context.Context) error {
	deadline, _ := ctx.Deadline()
	if deadline.Before(time.Now()) {
		time.Sleep(2 * time.Second)
		return nil
	}
	return nil
}
```

Fenny:
The problem is that `*fiber.Ctx` is returned to the `sync.Pool` when you return from the handler. So `c.Context().Done()` becomes non-existent. It is better to handle the request in a separate thread while waiting for the context to be done.
```go
type Data struct {
	Data []int `json:"data"`
}

func main() {
	app := fiber.New()

	app.Get("/", func(ctx *fiber.Ctx) error {
		c, cancel := context.WithTimeout(ctx.Context(), 50*time.Millisecond)
		defer cancel()

		res := Data{}
		ch := make(chan Data)

		go process(ctx, ch)

		select {
		case <-c.Done():
			fmt.Printf("error: processing timeout\n")
			break
		case res = <-ch:
			fmt.Printf("finished processing\n")
		}

		return ctx.JSON(res)
	})
	log.Fatal(app.Listen(":3000"))
}

func process(ctx *fiber.Ctx, ch chan<- Data) {
	fmt.Printf("processing started...\n")
	time.Sleep(time.Duration(2 * time.Second)) // uncomment this
	fmt.Printf("ended after 50ms\n")

[...]

PR #4264, opened 2026-05-09
📚 Doc: correct fasthttpctx Done semantics in context guide

### LINKED ISSUE #4263: 🐛 [Bug]: docs claim `RequestCtx.Done` closes on client disconnect, but it only closes on server shutdown
### Bug Description

The v3 guide [`Go Context`](https://docs.gofiber.io/next/guide/go-context/)  tells readers that `RequestCtx.Done()` fires when the client TCP connection is closed. That is not what fasthttp actually does — `s.done` is only closed when the server itself shuts down. This wording will mislead anyone trying to detect client disconnects from a Fiber handler.

**Quoted text (lines ~100–104):**

> `fasthttpctx` enables `fasthttp` to satisfy the `context.Context` interface.
> `Deadline` always reports no deadline, `Done` is closed when the client
> connection ends, and once it fires `Err` reports `context.Canceled`. This
> means handlers can detect client disconnects while still passing
> `c.RequestCtx()` into APIs that expect a `context.Context`.

**Quoted text (Summary bullet, lines ~263–264):**

> - `RequestCtx` exposes the raw `fasthttp` context, whose `Done` channel closes
>   when the client connection ends.

**What fasthttp actually does** — [`server.go#L2925-L2944` in valyala/fasthttp](https://github.com/valyala/fasthttp/blob/master/server.go#L2925-L2944):

```go
// Note: Because creating a new channel for every request is just too expensive, so
// RequestCtx.s.done is only closed when the server is shutting down.
func (ctx *RequestCtx) Done() <-chan struct{} {
    return ctx.s.done
}

// ...
// Canceled if the context was canceled (via server Shutdown)
// or DeadlineExceeded if the context's deadline passed.
//
// Note: Because creating a new channel for every request is just too expensive, so
// RequestCtx.s.done is only closed when the server is shutting down.
func (ctx *RequestCtx) Err() error { ... }
```

`s.done` is the **server-wide** done channel (allocated once in `Server.Serve`, closed once in `Server.Shutdown`). It is shared by every in-flight request and never carries per-connection cancellation. The fasthttp source comment is unambiguous on this point.

This is also confirmed publicly by both fasthttp and Fiber maintainers:

- valyala/fasthttp#468 — erikdubbelboer: *"this is not possible and would be really hard to add"*
- valyala/fasthttp#771 — SSE disconnect detection only via `Flush()` error
- valyala/fasthttp#965 — *"fasthttp doesn't do any operations on the socket while your handler is running, this is impossible for us to implement"*
- gofiber/fiber#805 — Fenny: *"I'm afraid that is not possible nor efficient with fasthttp sync.pool design"*, recommends timeout middleware as the only general fallback

So the documentation paragraph describes behavior that does not exist in any released version of fasthttp.

### How to Reproduce

1. Go to https://docs.gofiber.io/docs/core/guide/context#working-with-requestctx-and-fasthttpctx
2. Read the paragraph stating ``Done` is closed when the client connection ends`.
3. Run the snippet below: kill the client mid-request and observe that `<-c.RequestCtx().Done()` never fires until the server itself begins shutdown.

### Expected Behavior

The two passages should describe the actual fasthttp semantics. Suggested replacement (feel free to reword):

> `fasthttpctx` lets `fasthttp.RequestCtx` satisfy the `context.Context` interface. `Deadline` always reports no deadline. **`Done` is shared across all in-flight requests and only closes when the server is shutting down (via `Server.Shutdown`); it does not signal an individual client disconnect.** `Err` returns `context.Canceled` after that channel is closed. Handlers cannot rely on this channel to detect a client TCP disconnect — that is not currently supported by `fasthttp` (see [valyala/fasthttp#468](https://github.com/valyala/fasthttp/issues/468), [#965](https://github.com/valyala/fasthttp/issues/965)). For long-running handlers, wrap the request in a timeout (e.g. `context.WithTimeout` middleware) instead.

The Summary bullet should be updated to match.

### Fiber Version

v3.x (current docs)

### Code Snippet (optional)

```go
package main

import (
    "fmt"
    "log"
    "time"

    "github.com/gofiber/fiber/v3"
)

func main() {
    app := fiber.New()

    app.Get("/wait", func(c fiber.Ctx) error {
        fctx := c.RequestCtx()
        select {
        case <-fctx.Done():
            // The docs claim this fires when the client disconnects.
            // It does not — this branch is only reached on server shutdown.
            fmt.Println("done fired:", fctx.Err())
        case <-time.After(30 * time.Second):
            fmt.Println("timed out — client disconnect was never observed")
        }
        return nil
    })

}
```

Reproduction:

```bash
# Terminal A
go run main.go

# Terminal B
curl http://localhost:3000/wait &
CURL_PID=$!
sleep 1
kill -9 $CURL_PID   # close the client TCP connection
# Server still prints "timed out — client disconnect was never observed".
```

### Checklist:

gtoxlili:
# Description

The "Working with `RequestCtx` and `fasthttpctx`" section in `docs/guide/context.md` (introduced in #3677) currently tells readers that `RequestCtx.Done` closes when the client connection ends and that "handlers can detect client disconnects" via this channel. That is not what fasthttp actually does:

- [fasthttp `server.go` `RequestCtx.Done`](https://github.com/valyala/fasthttp/blob/master/server.go) carries the explicit comment **"`RequestCtx.s.done` is only closed when the server is shutting down"** — `s.done` is the shared server-wide channel, allocated once in `Server.Serve` and closed once in `Server.Shutdown`. There is no per-connection cancellation signal.
- valyala/fasthttp#468 — *"this is not possible and would be really hard to add"*
- valyala/fasthttp#965 — *"fasthttp doesn't do any operations on the socket while your handler is running, this is impossible for us to implement"*
- gofiber/fiber#805 — @Fenny: *"I'm afraid that is not possible nor efficient with fasthttp sync.pool design"*
- gofiber/fiber#1718 — same Q&A, closed in 2022 with the same answer
- gofiber/fiber#4145 — **open** proposal asking Fiber to *add* per-connection disconnect detection. Its existence is direct evidence that the capability the docs already advertised does not exist.

So the docs paragraph contradicts (a) the fasthttp source code, (b) two upstream fasthttp maintainers, (c) a four-year-old Fiber maintainer Q&A answer, and (d) an open Fiber proposal asking for the very capability the docs claim is already there.

Fixes #4263

## Changes introduced

Three spots in `docs/guide/context.md`:

1. The `fasthttpctx` paragraph (~L100): now describes the actual semantics (server-shutdown-only, shared channel, no client-disconnect signal) and points readers to `context.WithTimeout` or `Flush`-error detection for long handlers.
2. The "Context Cancellation with Goroutines" paragraph (~L219): drops the misleading "or the client disconnects" phrasing and clarifies that the timeout (not a client-side cancellation) is what bounds the work.
3. The Summary bullet (~L263): matches the corrected description.

No code changes; no API changes.

- [ ] Benchmarks: n/a
- [x] Documentation Update: `docs/guide/context.md` — three passages corrected, no new sections.
- [ ] Changelog/What's New: n/a (corrects an inaccurate claim, no behavior change)
- [ ] Migration Guide: n/a
- [ ] API Alignment with Express: n/a
- [ ] API Longevity: n/a (docs only)
- [ ] Examples: n/a

## Type of change

## Checklist

- [x] Conducted a self-review of the changes.
- [ ] Added or updated unit tests — n/a, docs only.
- [x] Verified the wording against the fasthttp source and the linked issues.

### DIFF
--- a/docs/guide/context.md
+++ b/docs/guide/context.md
@@ -98,10 +98,14 @@ app.Get("/raw", func(c fiber.Ctx) error {
 ```
 
 `fasthttpctx` enables `fasthttp` to satisfy the `context.Context` interface.
-`Deadline` always reports no deadline, `Done` is closed when the client
-connection ends, and once it fires `Err` reports `context.Canceled`. This
-means handlers can detect client disconnects while still passing
-`c.RequestCtx()` into APIs that expect a `context.Context`.
+`Deadline` always reports no deadline. `Done` closes only when the server
+shuts down, and `Err` then returns `context.Canceled`.
+
+:::caution
+`Done` does not fire when an individual client disconnects. To stop a
+long-running handler at a deadline, wrap the request with
+`context.WithTimeout` as shown below.
+:::
 
 ## Context Helpers
 
@@ -216,7 +220,7 @@ When starting asynchronous work inside a handler, Fiber does not cancel the base
 By wrapping the request context with `context.WithTimeout`, you can create a derived context that honors deadlines and cancellation signals.
 
 The goroutine checks `ctx.Done()` before sending a result.
-If the request times out or the client disconnects the goroutine exits early and avoids leaking resources.
+If the deadline elapses, the goroutine exits early and avoids leaking resources.
 
 The handler then waits for either:
 
@@ -260,8 +264,8 @@ This approach provides safe cancellation semantics for goroutine-based work whil
 
 - `fiber.Ctx` satisfies `context.Context` but its `Deadline`, `Done`, and `Err`
   methods are currently no-ops.
-- `RequestCtx` exposes the raw `fasthttp` context, whose `Done` channel closes
-  when the client connection ends.
+- `RequestCtx` exposes the raw `fasthttp` context. Its `Done` channel
+  closes only on server shutdown, not on client disconnect.
 - Use `fiber.StoreInContext(c, key, value)` to store request-scoped values in both
   `c.Locals()` and `c.Context()` when values must be available through either API.
 - Middleware helpers like `requestid.FromContext` or `session.FromContext`


gaby:
@gtoxlili Please fix the wording and tone of the changes. These are official docs, not a forum.

gtoxlili:
@gaby Tightened the wording — dropped the bold emphasis, the inline issue links, and the explanatory asides. Just states the actual semantics now and points to the timeout pattern below. PTAL.

PR #4265, opened 2026-05-09
🐛 bug: guard session logger tag against released middleware

gaby:
### Motivation
- The auto-registered `${session-id}` logger tag could call `m.ID()` after the session middleware had cleared `m.Session` and returned the `*Middleware` to the pool, which can cause a nil-pointer panic when logger runs after downstream middleware returns.
- Make a minimal change to prevent dereferencing a released session object while preserving existing behavior and redaction for live sessions.

### Description
- In `middleware/session/middleware.go` the logger context tag was changed to safely read `m.Session` under `m.mu.RLock()` and return an empty string when the session has already been released instead of calling `m.ID()` directly.
- Added a regression test `Test_SessionLoggerTagWithOuterLoggerDoesNotPanic` in `middleware/session/middleware_test.go` that mounts `logger` outside `session` and asserts a request completes without panic when `Format: "${session-id}"` is used.
- No public API changes or behavior changes for live sessions; only guards against the stale/released-session crash.

### DIFF
--- a/middleware/session/middleware.go
+++ b/middleware/session/middleware.go
@@ -119,7 +119,15 @@ func registerLogContextTags() {
 		if m == nil {
 			return ""
 		}
-		return redact.Prefix(m.ID())
+
+		m.mu.RLock()
+		session := m.Session
+		m.mu.RUnlock()
+		if session == nil {
+			return ""
+		}
+
+		return redact.Prefix(session.ID())
 	})
 }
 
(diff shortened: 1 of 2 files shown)


gaby:
Addressing comments

gaby:
@copilot+gpt-5.4 apply changes based on the comments in [this thread](https://github.com/gofiber/fiber/pull/4265#pullrequestreview-4258023869)

PR #4269, opened 2026-05-10
🐛 bug: prevent panic when CBOR is not explicitly configured

gaby:
### Motivation
- A previous refactor replaced the default CBOR marshal/unmarshal with panic-only stubs which made `Bind().Body` reachable panic via an attacker-controlled `Content-Type: application/cbor` header.  
- The change preserves CBOR as opt-in but must not allow a panic to be triggered from normal request dispatch paths.  

### Description
- Replace panic stubs in `binder/cbor.go` with an `errUnimplementedCBOR` error and make `UnimplementedCborMarshal`/`UnimplementedCborUnmarshal` return that error instead of calling `panic`.  
- Update `binder/cbor_test.go` to assert error-return behavior for the unimplemented CBOR helpers rather than expecting panics.  
- This is a minimal remediation that preserves opt-in semantics while ensuring failures propagate as errors handled by existing Fiber error paths.

### DIFF
--- a/binder/cbor.go
+++ b/binder/cbor.go
@@ -1,6 +1,8 @@
 package binder
 
 import (
+	"errors"
+
 	"github.com/gofiber/utils/v2"
 )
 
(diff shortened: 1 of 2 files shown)


gaby:
@copilot+gpt-5.4 apply changes based on the comments in [this thread](https://github.com/gofiber/fiber/pull/4269#pullrequestreview-4258765422)

gaby:
@copilot+gpt-5.4 
- errUnimplementedCBOR should be exported
- we should use require.ErrorIs

gaby:
@copilot+gpt-5.4 the error should start with "binder: "

ReneWerner87 (approved):

PR #4271, opened 2026-05-10
🐛 bug: avoid panic when reading released Fiber context values

gaby:
### Motivation
- Restore the nil-safe semantics for context value lookups so that retained/released `fiber.Ctx` instances cannot cause a panic when helpers call `ValueFromContext` after `ReleaseCtx`.

### Description
- Prefer `Value(key)` over `Locals(key)` when a Fiber-like context exposes both by adding a `fiberValueContext` interface and dispatching to `Value` first in `internal/contextvalue/contextvalue.go`.
- Add a regression test in `helpers_test.go` that asserts `ValueFromContext` does not panic and returns `(zero, false)` when called with a released `fiber.Ctx`.


internal/contextvalue/contextvalue.go
```diff
@@ -21,6 +26,9 @@ type valueContext interface {
 // Value retrieves a value stored under key from supported Fiber context types.
 func Value[T any](ctx, key any) (T, bool) {
 	switch typed := ctx.(type) {
+	case fiberValueContext:
+		val, ok := typed.Value(key).(T)
+		return val, ok
```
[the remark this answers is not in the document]
gaby: @copilot does this apply? If so, make the changes.

gaby:
@copilot+gpt-5.4 Prioritizing fiberValueContext here correctly ensures that the nil-safe Value method is used for Fiber contexts, preventing panics after context release. As a minor improvement for maintainability, consider if the existing valueContext interface could be moved to the top of the switch instead of introducing the new fiberValueContext interface. This would achieve the same result for DefaultCtx while also simplifying the handling of context.Context and other types that implement Value.

If this apply, make the changes.

gaby:
@copilot+gpt-5.4 do not remove the case:

```
case context.Context:
    val, ok := typed.Value(key).(T)
    return val, ok
```

We need to support fiber ctx, fasthttp request ctx, and context.Context

internal/contextvalue/contextvalue.go
```diff
@@ -21,21 +21,21 @@ type valueContext interface {
 // Value retrieves a value stored under key from supported Fiber context types.
```
ReneWerner87: Doc comment is now stale: the function also explicitly supports the standard `context.Context` case (line 27-29), which is not Fiber-specific. CodeRabbit suggested the same fix in its review, worth folding in:

```diff
-// Value retrieves a value stored under key from supported Fiber context types.
+// Value retrieves a value stored under key from supported context types
+// (fiber.Ctx, fiber.CustomCtx, context.Context, and *fasthttp.RequestCtx).
```

internal/contextvalue/contextvalue.go
```diff
@@ -21,21 +21,21 @@ type valueContext interface {
...
 		return val, ok
-	case userValueContext:
-		val, ok := typed.UserValue(key).(T)
-		return val, ok
 	case valueContext:
 		val, ok := typed.Value(key).(T)
 		return val, ok
+	case fiberLocalContext:
+		val, ok := typed.Locals(key).(T)
+		return val, ok
```
ReneWerner87, line 38: The order of these three cases (`valueContext` -> `fiberLocalContext` -> `userValueContext`) is now load-bearing for the panic-safety of released `fiber.Ctx`: `fiber.Ctx` satisfies both `valueContext` (nil-safe via `DefaultCtx.Value`) and `fiberLocalContext` (panics on released ctx via `DefaultReq.Locals`). Match the `Value` paths first.

A future contributor sorting cases by specificity (`fiberLocalContext` is the most narrow Fiber interface, so it "belongs" first) will reintroduce the bug from `c48bbc2` silently. Please add a short comment above the switch, e.g.:

```go
// Order is load-bearing: a fiber.Ctx that has been released has c.fasthttp == nil,
// which makes Locals panic but leaves Value nil-safe. Dispatch through the Value
// interfaces first so released Fiber contexts return (zero, false) instead of panicking.
switch typed := ctx.(type) {
```

Alternative: harden `DefaultReq.Locals` with the same nil-guard `DefaultCtx.Value` already has, then the order does not matter (see the top-level review body for the proposed diff).

helpers_test.go
```diff
@@ -1689,6 +1689,23 @@ func TestValueFromContext(t *testing.T) {
...
+		c := app.AcquireCtx(raw)
+
+		c.Locals("key", "value")
+		app.ReleaseCtx(c)
+
+		require.NotPanics(t, func() {
+			value, ok := ValueFromContext[string](c, "key")
+			require.False(t, ok)
+			require.Empty(t, value)
+		})
```
ReneWerner87, line 1711: Two small additions would make this test more useful as a contract:

1. Add a positive assertion before `ReleaseCtx` so the test also documents that the value IS readable while the ctx is alive. Right now an accidental break of the live-ctx path would still let this test pass.
2. Add a `fiber.CustomCtx`-released subtest. The reorder affects any CustomCtx that implements both `Value` and `Locals` (now `Value` wins), and there is currently no test that pins the CustomCtx behavior under `ReleaseCtx`.

```go
t.Run("released fiber.Ctx", func(t *testing.T) {
    t.Parallel()

    app := New()
    raw := &fasthttp.RequestCtx{}
    c := app.AcquireCtx(raw)

    c.Locals("key", "value")

    // Sanity: the value is retrievable while the ctx is alive.
    value, ok := ValueFromContext[string](c, "key")
    require.True(t, ok)
    require.Equal(t, "value", value)

    app.ReleaseCtx(c)

    require.NotPanics(t, func() {
        value, ok := ValueFromContext[string](c, "key")
        require.False(t, ok)
        require.Empty(t, value)
    })
})
```

ReneWerner87 (commented):
The diagnosis is right and the patch works (CI is green, regression test covers the panic path). Two suggestions, the first one is the load-bearing one.

### 1. Consider fixing the root cause in `DefaultReq.Locals` instead

The asymmetry between `DefaultCtx.Value` (ctx.go:625, nil-safe via `if c.fasthttp == nil { return nil }`) and `DefaultReq.Locals` (req.go:703, no guard) is the actual bug. Hardening only the helper means:

- every other caller of `c.Locals(key)` after `ReleaseCtx` still panics: middleware (`session`, `requestid`, ...), user code, tests; the helper reorder only protects `ValueFromContext`.
- the type-switch reorder becomes load-bearing in a non-obvious way; a future contributor sorting cases by specificity reintroduces the bug.

A minimal alternative, applied at the symptom source:

```go
func (r *DefaultReq) Locals(key any, value ...any) any {
    if r.c.fasthttp == nil {
        if len(value) > 0 {
            return value[0]
        }
        return nil
    }
    if len(value) == 0 {
        return r.c.fasthttp.UserValue(key)
    }
    r.c.fasthttp.SetUserValue(key, value[0])
    return value[0]
}
```

This mirrors `Value`'s pattern, fixes every caller, and makes the helper reorder optional (you can keep it for clarity or revert to the previous order without functional impact).

### 2. If you keep the reorder, document why

The order in the type switch is now load-bearing for nil-safety. Without a comment the next contributor who thinks "most specific case first" can revert it. See inline.

### 3. Test could carry a tiny bit more weight

The regression test only covers `*DefaultCtx` after release with a string value. Adding a `fiber.CustomCtx`-released case and a pre-release sanity check (that `key` is retrievable before `ReleaseCtx`) would document the contract more fully and protect against accidental CustomCtx behavior changes from the reorder. See inline.

Nothing here is blocking; CI is green and Codecov is at 100% for modified lines.

gaby:
@copilot+gpt-5.4 Please address all the review comments.

ReneWerner87 (approved):

PR #4272, opened 2026-05-10
🐛 bug: prevent negative paginate start overflow

gaby:
### Motivation
- Prevent PageInfo.Start() from returning negative offsets when cursor mode leaves `Page` unset or when extremely large `page` values overflow the multiplication. 
- The middleware previously capped only `limit` while `page` and `offset` could be attacker-controlled and produce negative or wrapped results used as slice/SQL offsets. 
- Cursor support being enabled by default meant a crafted `cursor` could trigger the bug in handlers that follow the documented `pageInfo.Start()` usage.

### Description
- Harden `PageInfo.Start()` to return `0` if `Page < 1` or `Limit < 1` and to clamp extremely large page-based starts to `MaxInt` to avoid integer overflow. 
- Preserve existing behavior for normal page/offset values while preventing negative/wrapped results from untrusted inputs. 
- Add test cases to `Test_PageInfoStart` covering `Page = 0` (cursor mode) and an overflow-scale `Page` value. 
- Include updated generated interface files (`ctx_interface_gen.go` and `req_interface_gen.go`) produced by repository tooling.

### DIFF
--- a/ctx_interface_gen.go
+++ b/ctx_interface_gen.go
@@ -374,7 +374,7 @@ type Ctx interface {
 	// If Config.TrustProxy false, it returns false.
 	// IsProxyTrusted can check remote ip by proxy ranges and ip map.
 	IsProxyTrusted() bool
-	// IsFromLocal will return true if request came from local.
+	// IsFromLocal will return true if request came from a loopback IP.
 	IsFromLocal() bool
 	// IsFromUnixSocket returns true if the request arrived over a Unix domain socket.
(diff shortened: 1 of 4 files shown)


gaby:
@copilot+gpt-5.4 apply changes based on the comments in [this thread](https://github.com/gofiber/fiber/pull/4272#pullrequestreview-4259807267)

ReneWerner87 (approved):

PR #4274, opened 2026-05-11
🐛 bug: prevent SharedState namespace key collisions

gaby:
### Motivation
- A recent change stopped hex-encoding caller-supplied keys before concatenating them with the SharedState prefix, making namespace prefixes and keys ambiguous and enabling cross-namespace collisions on shared backends. 
- Restore an unambiguous storage-key derivation to prevent a crafted key in one app from reading/writing/deleting state from another app sharing the same backend.

### Description
- Restore hex-encoding of the caller key in `SharedState.storageKey`, now returning `utils.CopyString(s.prefix + hex.EncodeToString(utils.UnsafeBytes(key)))` to preserve a clear namespace delimiter. 
- Add `TestSharedState_KeyNamespacingWithOverlappingPrefixes` that reproduces the overlapping-prefix scenario (`app` vs `app-one`) and verifies namespaces remain isolated. 
- Update generated interface comment text via `make generate` which updated `ctx_interface_gen.go` and `req_interface_gen.go` as part of the repository generation workflow.

### DIFF
--- a/ctx_interface_gen.go
+++ b/ctx_interface_gen.go
@@ -374,7 +374,7 @@ type Ctx interface {
 	// If Config.TrustProxy false, it returns false.
 	// IsProxyTrusted can check remote ip by proxy ranges and ip map.
 	IsProxyTrusted() bool
-	// IsFromLocal will return true if request came from local.
+	// IsFromLocal will return true if request came from a loopback IP.
 	IsFromLocal() bool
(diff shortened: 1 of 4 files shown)


gaby:
@copilot+gpt-5.4 Error: shared_state.go:416:1: File is not properly formatted (gofmt)

ReneWerner87 (approved):

PR #4275, opened 2026-05-11
🐛 bug: copy FullURL string before returning pooled buffer

gaby:
### Motivation
- Fix a use-after-pool/aliasing vulnerability in `DefaultCtx.FullURL()` where the method returned a string that could alias memory from a pooled `bytebufferpool` buffer that was returned to the pool before the caller could safely use the string.

### Description
- Change `DefaultCtx.FullURL()` to return a safe copy using `buf.String()` instead of returning `c.app.toString(buf.Bytes())`, preventing returned strings from referencing pooled memory that is recycled immediately. (see `req.go`).
- Regenerate related generated interface files that updated method comments/signatures formatting (`ctx_interface_gen.go`, `req_interface_gen.go`).

### DIFF
--- a/ctx_interface_gen.go
+++ b/ctx_interface_gen.go
@@ -374,7 +374,7 @@ type Ctx interface {
 	// If Config.TrustProxy false, it returns false.
 	// IsProxyTrusted can check remote ip by proxy ranges and ip map.
 	IsProxyTrusted() bool
-	// IsFromLocal will return true if request came from local.
+	// IsFromLocal will return true if request came from a loopback IP.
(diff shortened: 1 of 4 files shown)


gaby:
@copilot+gpt-5.4 apply changes based on the comments in [this thread](https://github.com/gofiber/fiber/pull/4275#pullrequestreview-4260480613)

gaby:
Performance notice is expected since we are creating a string. Previous implementation was returning a buffer that was already deferred.

ReneWerner87 (approved):

PR #4277, opened 2026-05-11
🐛 bug: enforce static root for fs-backed directory serving

gaby:
### Motivation
- A prior change caused `Config.FS`-backed static handlers to set the fasthttp FS root to `"."` and rely on `PathRewrite` to map requests, which allowed directory-root requests to resolve from the broader `fs.FS` instead of the caller-specified `root` and could disclose files outside the intended static subdirectory. 
- The goal is to ensure directory roots configured with `io/fs.FS` remain confined to the provided `root` while preserving existing behavior for single-file roots and non-FS usage.

### Description
- Update `fileServer.PathRewrite` in `middleware/static/static.go` to prepend the configured `root` (when non-empty and not `"."`) for directory-root cases when `Config.FS` is set, ensuring requests are resolved under the intended static root. 
- Preserve existing `checkFile` (single-file root) branches and behavior for non-FS roots so file-root semantics remain unchanged. 
- Add `Test_Static_FS_RootDirectoryEnforced` in `middleware/static/static_test.go` to assert that `New("fs", Config{FS: os.DirFS("../../.github/testdata")})` serves `/static/index.html` but rejects `/static/fs/index.html`.


gaby:
@copilot+gpt-5.4 apply changes based on the comments in [this thread](https://github.com/gofiber/fiber/pull/4277#pullrequestreview-4260621325)

middleware/static/static.go
```diff
@@ -171,6 +171,13 @@ func New(root string, cfg ...Config) fiber.Handler {
 				fsRoot = ""
 			}
 
+			var fsRootPrefix []byte
+			if config.FS != nil && root != "" && root != "." && !rootIsFile {
+				fsRootPrefix = make([]byte, len(root)+1)
+				fsRootPrefix[0] = '/'
+				copy(fsRootPrefix[1:], root)
+			}
```
ReneWerner87, line 212: Two things on this block:

1. **Add a comment.** The reason this exists is not obvious from the code. Something like: `// fasthttp.FS.Root is forced to "" when Config.FS is set so that pathToFilePath returns clean relative paths. We re-prepend the caller-specified root in PathRewrite to confine lookups to the configured subdirectory of fs.FS.`

2. **Allocation cost.** This is a one-shot setup, so the cost is negligible here, but the prefix lookup happens on every request below. Consider documenting the trade-off, or use `utils.UnsafeBytes("/" + root)` if you want a single allocation.

middleware/static/static.go
```diff
@@ -206,6 +213,12 @@ func New(root string, cfg ...Config) fiber.Handler {
 						path = utils.UnsafeBytes(root)
 					default:
 						path = path[prefixLen:]
+						if len(fsRootPrefix) > 0 {
+							rewrittenPath := make([]byte, len(fsRootPrefix)+len(path))
+							copy(rewrittenPath, fsRootPrefix)
+							copy(rewrittenPath[len(fsRootPrefix):], path)
+							path = rewrittenPath
```
ReneWerner87: **Related bypass on the `isFile` error path (existing code, lines 203-206, just out of diff so commenting here).** If `isFile(root, fileServer.FS)` returns an error (configured root missing in `fs.FS`, typo, transient open/stat failure), `PathRewrite` returns the raw `fctx.Path()`. That skips this new `fsRootPrefix` prepend *and* `sanitizePath` below, so requests resolve directly against the broader `fs.FS` with no `..` / encoded-`..` / `\x00` / backslash rejection. Copilot already flagged it ("low confidence"); it is real, and this PR should fix it in the same commit since it is the same defence boundary. Suggested: on error, set `checkFile = false` and fall through, or return `/__fiber_invalid__` so the FS layer responds 404. Never return `path` unchanged.

---

**Traversal is not blocked by the new block either.** The prefix is prepended *before* `sanitizePath` (line 232), and `sanitizePath` calls `pathpkg.Clean` which resolves `..` before the `..`-rejection check at line 75. So:

- Request: `/static/%2E%2E/index.html` (URL-encoded `..`, not normalised away by fasthttp when `UnescapePath` is false).
- After `path[prefixLen:]`: `%2E%2E/index.html`.
- After this block, with `fsRootPrefix = "/fs"`: `/fs/%2E%2E/index.html`.
- In `sanitizePath`: unescape to `/fs/../index.html`, `Clean` to `/index.html`. The `..`-check at line 75 sees no `..` segment in the trimmed path because `Clean` already collapsed it.
- FS lookup: `index.html` resolves against the broader root (`.github/testdata/index.html`, `Hello, Fiber!`), *outside* the intended `fs/` subdirectory.

The `..`-rejection at static.go:75 is effectively dead code for absolute paths because `pathpkg.Clean` always resolves them away first, so the only thing keeping naive cases from escaping is fasthttp's default request-path normalisation - a fragile defence for a security fix.

Fix options:

- **Prefer:** prepend `fsRootPrefix` *after* `sanitizePath` has run on the route-stripped path. Sanitise the user-controlled segment first, then anchor it under the trusted root.
- Or: reject `..` (and encoded variants) on the un-cleaned input inside `sanitizePath` before `Clean` runs.

middleware/static/static_test.go
```diff
@@ -697,6 +697,27 @@ func Test_Static_FS_Prefix_Wildcard(t *testing.T) {
...
+	require.Equal(t, fiber.StatusOK, resp.StatusCode, "Status code")
+	responseBody, err := io.ReadAll(resp.Body)
+	require.NoError(t, err)
+	require.Contains(t, string(responseBody), "Hello, World!")
+	require.NotContains(t, string(responseBody), "Hello, Fiber!")
+
+	resp, err = app.Test(httptest.NewRequest(fiber.MethodGet, "/static/fs/index.html", http.NoBody))
+	require.NoError(t, err, "app.Test(req)")
+	require.Equal(t, fiber.StatusNotFound, resp.StatusCode, "Status code")
+}
```
ReneWerner87, line 796: This test only covers the naive `/static/fs/index.html` double-prefix case. For a security fix labelled `☢️ Bug`, the regression test needs to exercise the actual attack class. Please add cases for:

- `/static/../index.html` (raw `..`)
- `/static/%2E%2E/index.html` (URL-encoded `..`, the one that actually slips past fasthttp normalisation when `UnescapePath` is false)
- `/static/..%2Findex.html` (encoded slash)
- `/static/..%5Cindex.html` (encoded backslash, since `sanitizePath` maps `\` to `/`)
- `/static` and `/static/` (empty / trailing-slash, to lock in directory-listing behaviour)
- `/static/%00` (null byte) for symmetry with `sanitizePath`'s rejections

All of these should return 404 (or whatever the documented contract is); right now at least the encoded `..` ones still serve `.github/testdata/index.html` based on the current implementation, see the comment on `static.go:216-220`.

A table-driven sub-test would keep this readable.

ReneWerner87 (changes requested):
Nice direction, but I do not think this is ready to merge. The intended root boundary is still bypassable, and the regression test only covers the trivial double-prefix case rather than actual traversal attempts. Inline comments below.

### Summary of concerns

1. **Path traversal bypass.** `fsRootPrefix` is prepended *before* `sanitizePath`, and `sanitizePath` calls `pathpkg.Clean` which resolves `..` segments before the `..` check (static.go:75) ever runs. A request like `/static/%2E%2E/index.html` (or any non-normalised `..` input) becomes `/fs/../index.html` after the prefix prepend, and Clean rewrites that to `/index.html`, escaping the configured `fs/` subdirectory. The `..` check at line 75 is effectively dead code for absolute paths because `pathpkg.Clean` always resolves them away first.

2. **`isFile` error bypass.** If `isFile(root, fileServer.FS)` returns an error (e.g. the configured root is missing or unreadable in the `fs.FS`), `PathRewrite` returns the raw `fctx.Path()` and skips both the new `fsRootPrefix` prepend *and* `sanitizePath`. With `Config.FS` set, that re-exposes the broader `fs.FS` and bypasses the very defence this PR is adding. Already flagged by Copilot, still unaddressed.

3. **Test coverage is too narrow for a security fix.** `Test_Static_FS_RootDirectoryEnforced` only checks that `/static/fs/index.html` 404s. None of the realistic attack shapes are covered: `..`, URL-encoded `%2E%2E`, mixed `..%2F`, backslash variants, empty path, trailing slash. A regression test that does not exercise the attack class it claims to prevent is not a regression test for that class.

4. **No comment on the security-critical block.** In six months no one will remember why `fsRootPrefix` exists. CodeRabbit's nitpick is fair.

### Suggested fixes

- Move the prefix prepend *after* `sanitizePath`, or perform the `..` rejection on the unescaped-but-not-yet-cleaned input.
- For the `isFile` error path, either treat the error as `checkFile=false` and continue through the normal rewrite + sanitize flow, or return the `/__fiber_invalid__` sentinel so the FS layer 404s. Do not return the raw path.
- Extend the test to cover `..`, `%2E%2E`, `..%2F`, backslashes, `/static`, and `/static/`.
- Add a short comment explaining why `fsRootPrefix` is needed (FS root is forced to `""` to get clean paths from fasthttp, so we have to re-prepend the caller-specified subdir in `PathRewrite`).

gaby:
@copilot+gpt-5.4 fix the failing unit tests, use "-race".

gaby:
@copilot+gpt-5.4 The current implementation is vulnerable to a critical path traversal vulnerability. By prepending the root directory to the raw request path before sanitization, an attacker can use .. segments to escape the intended root. For example, a request to /static/../secret.txt would result in /root/../secret.txt, which sanitizePath (via pathpkg.Clean) resolves to /secret.txt, disclosing files outside the static subdirectory. Additionally, the current logic may fail to join paths correctly if the stripped path doesn't start with a slash (e.g., resulting in /rootfile.txt instead of /root/file.txt). To fix this, the request path must be cleaned to resolve any .. segments relative to the request root before joining it with the configured static root.

```go
if rootPrefix != "" {
	// Clean the request path to prevent traversal out of the root and ensure proper joining
	relPath := pathpkg.Clean("/" + utils.UnsafeString(path))
	path = utils.UnsafeBytes(pathpkg.Join(rootPrefix, relPath))
}
```

gaby:
@copilot+gpt-5.4 address the review comment raised here https://github.com/gofiber/fiber/pull/4277/changes#r3216778895

Also, please include a comprehensive suite of tests to cover various directory traversal scenarios.

gaby:
@claude[agent] Please conduct a security audit of this implementation and address any identified issues.

ReneWerner87 (approved):

Issue #4280, opened 2026-05-12
🐛 [Bug]: multiple uses of tokenKey


joyride9999:
### Bug Description

Hello,
it looks to me that there might be a problem with the use of const tokenKey. 
E.g. in middleware/csrf/csrf.go we have this [definition](https://github.com/gofiber/fiber/blob/main/middleware/csrf/csrf.go#L49)
```
const (
	tokenKey contextKey = iota
	handlerKey
)
```
Hence token key initialized with 0
If we look in contrib package for [jwt ](https://github.com/gofiber/contrib/blob/main/v3/jwt/jwt.go#L20) we have again the same definition
```
const (
	tokenKey contextKey = iota
)
```

Result is that at some point both modules will access local's based on the same key which should result in problems.
That is at some point both modules call fiber.StoreInContext
```
func StoreInContext(c Ctx, key, value any) {
	c.Locals(key, value)

	if c.App().config.PassLocalsToContext {
		c.SetContext(context.WithValue(c.Context(), key, value))
	}
}
```
That would result in values from c.Locals(...) being overwritten if i'm not missing anything.

### How to Reproduce

I did not tested the bug , was just looking at the code :)

### Expected Behavior

My proposal would be to use instead of a constant a variable that can be changed at runtime or depending on context ... e.g. get the tokenKey dinamically in a function that takes the context as parameter ...

### Fiber Version

v3.2.0

### Code Snippet (optional)

```go

```

### Checklist:

- [x] I agree to follow Fiber's [Code of Conduct](https://github.com/gofiber/fiber/blob/main/.github/CODE_OF_CONDUCT.md).
- [x] I have checked for existing issues that describe my problem prior to opening this one.
- [x] I understand that improperly formatted bug reports may be closed without explanation.

joyride9999:
later observation.. as the constants (tokenKeys) are defined as custom int types this should be enough to avoid collisions.. 
Nevertheless i think this will be a good proposal especially for cases when handling multiple tokens on the same session (e.g. single use csrf token in the same context but using different tab /windows)

sixcolors:
Closing this issue based on your follow-up comment confirming that the custom `int` types prevent key collisions.

As noted, the original concern about `tokenKey` constants from different packages overwriting each other is not possible. Each package's `contextKey` is a distinct unexported type, so `csrf.contextKey(0)` and `jwt.contextKey(0)` are never equal when used as `Locals` or `context.WithValue` keys.

[...]

PR #4289, opened 2026-05-17
Proposal fix for #4280 with no breaking changes 

joyride9999:
Proposal fix for #4280 with no breaking changes

Added possibility to dynamically control the cookie name and tokenKey using context variable. Rationale is that sometimes csrf token is overridden .. e.g. using SingleUseToken true , and having more then 1 browser windows/tab opened it may happen that all browser sessions use the same csrf token, but on the first post on any window the csrf will be accepted and changed, the other windows will have to be hard refreshed as their token got invalidated. This fix allows control of csrf token on a per window/tab session

# Description

Added possibility to dynamically control the cookie name and tokenKey using context variable. Rationale is that sometimes csrf token is overridden .. 

Fixes # (issue)
e.g. using SingleUseToken true , and having more then 1 browser windows/tab opened it may happen that all browser sessions use the same csrf token, but on the first post on any window the csrf will be accepted and changed, the other windows will have to be hard refreshed as their token got invalidated. This fix allows control of csrf token on a per window/tab session

## Changes introduced

Added CookieNameFn .. if this is used then default CookieName and tokenKey will use this function instead of the default. If not used then CookieName will behave similar. 

Setting up configuration :
`
csrfConfig := csrf.Config{
		Session:    store,
		Extractor:  extractors.FromHeader("X-Csrf-Token"), // In this example, we will be using a hidden input field to store the CSRF token
		CookieNameFN: GetCSRFByTabId,                        // set our function
                .................
		SingleUseToken: true,
	}
	return csrfConfig

// Example function that gets the tab id sent from the frontend)
func GetCSRFByTabId(ctx fiber.Ctx) string {

	if ctx == nil {
		return "_csrf"
	}

	//gets it from header
	tabid := ctx.Get("X-Tab-ID")
	if tabid == "" {
		return "_csrf"
	}
	return tabid
}

`

## Type of change

## Checklist

- [x ] Followed the inspiration of the Express.js framework for new functionalities, making them similar in usage.
- [ x] Conducted a self-review of the code and provided comments for complex or critical parts.
- [ x] Updated the documentation in the `/docs/` directory for [Fiber's documentation](https://docs.gofiber.io/).
- [ x] Added or updated unit tests to validate the effectiveness of the changes or new features.
- [ x] Ensured that new and existing unit tests pass locally with the changes.
- [ x] Verified that any new dependencies are essential and have been agreed upon by the maintainers/community.
- [ x] Aimed for optimal performance with minimal allocations in the new code.

## Commit formatting

### DIFF
--- a/middleware/csrf/config.go
+++ b/middleware/csrf/config.go
@@ -45,6 +45,9 @@ type Config struct {
 	// Optional. Default: "csrf_"
 	CookieName string
 
+	// CookieNameFn dynamic cookie name .. overrides CookieName if not nil
+	CookieNameFn func(c fiber.Ctx) string
+
 	// CookieDomain is the domain of the CSRF cookie.
 	//
 	// Optional. Default: ""
@@ -184,17 +187,23 @@ func validateExtractorSecurity(cfg *Config) {
 	if cfg == nil {
 		return
 	}
+
+	cookieName := cfg.CookieName
+	if cfg.CookieNameFn != nil {
+		cookieName = cfg.CookieNameFn(nil)
+	}
+
 	// Check primary extractor
-	if isInsecureCookieExtractor(cfg.Extractor, cfg.CookieName) {
-		panic("CSRF: Extractor reads from the same cookie '" + cfg.CookieName +
+	if isInsecureCookieExtractor(cfg.Extractor, cookieName) {
+		panic("CSRF: Extractor reads from the same cookie '" + cookieName +
 			"' used for token storage. This completely defeats CSRF protection.")
 	}
 
 	// Check chained extractors
 	for i, extractor := range cfg.Extractor.Chain {
-		if isInsecureCookieExtractor(extractor, cfg.CookieName) {
+		if isInsecureCookieExtractor(extractor, cookieName) {
 			panic(fmt.Sprintf("CSRF: Chained extractor #%d reads from the same cookie '%s' "+
-				"used for token storage. This completely defeats CSRF protection.", i+1, cfg.CookieName))
+				"used for token storage. This completely defeats CSRF protection.", i+1, cookieName))
 		}
 	}
 
--- a/middleware/csrf/csrf.go
+++ b/middleware/csrf/csrf.go
@@ -118,11 +118,15 @@ func New(config ...Config) fiber.Handler {
 		fiber.StoreInContext(c, handlerKey, handler)
 
 		var token string
+		cookieName := cfg.CookieName
+		if cfg.CookieNameFn != nil {
+			cookieName = cfg.CookieNameFn(c)
+		}
 
 		// Action depends on the HTTP method
 		switch c.Method() {
 		case fiber.MethodGet, fiber.MethodHead, fiber.MethodOptions, fiber.MethodTrace:
-			cookieToken := c.Cookies(cfg.CookieName)
+			cookieToken := c.Cookies(cookieName)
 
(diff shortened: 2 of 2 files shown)


gaby:
This is a breaking change in two locations and is unnecessary.

sixcolors:
This PR shouldn't be merged. It takes a working, collision-safe design and actively breaks it.

## The core problem

Your original issue was predicated on `tokenKey` colliding between packages. That was incorrect, each package's `contextKey` type is distinct, so `csrf.tokenKey` and `jwt.tokenKey` can never overwrite each other.

This patch replaces the safe, typed key with a **string** key in the critical path:

```go
var t any
if cfg.CookieNameFn == nil {
    t = tokenKey
} else {
    t = cookieName  // string key — now collisions are possible
}
```

When `CookieNameFn` is set, `StoreInContext(c, cookieName, token)` stores the token under a plain string. If two handlers or middlewares use the same dynamic cookie name, they will overwrite each other. You have introduced the exact bug the issue incorrectly claimed existed.

## API concerns

`TokenFromContext(ctx any, cookieKey ...string)` is a poor API. Variadic `string` arguments are not discoverable, not type-safe, and create confusion about whether the key is a typed constant or an arbitrary string.

## Summary

- The original collision claim was wrong.
- This PR does not fix a bug; it creates one by abandoning typed keys for string keys.
- Dynamic cookie names for CSRF tokens are not a supported pattern in Fiber and are not needed for tab/window isolation.

Closing this. If you have a concrete use case for dynamic cookie names that cannot be solved by existing configuration, please open a proposal with real-world requirements rather than a patch for a non-existent collision.

joyride9999:
> This PR shouldn't be merged. It takes a working, collision-safe design and actively breaks it.
> 
> ## The core problem
> Your original issue was predicated on `tokenKey` colliding between packages. That was incorrect, each package's `contextKey` type is distinct, so `csrf.tokenKey` and `jwt.tokenKey` can never overwrite each other.
> 
> This patch replaces the safe, typed key with a **string** key in the critical path:
> 
> ```go
> var t any
> if cfg.CookieNameFn == nil {
>     t = tokenKey
> } else {
>     t = cookieName  // string key — now collisions are possible
> }
> ```
> 
> When `CookieNameFn` is set, `StoreInContext(c, cookieName, token)` stores the token under a plain string. If two handlers or middlewares use the same dynamic cookie name, they will overwrite each other. You have introduced the exact bug the issue incorrectly claimed existed.

I think it is obvious that when the cookie name is handled through an external function then this becomes the caller's responsabilitty  to avoid collisions

> 
> ## API concerns
> `TokenFromContext(ctx any, cookieKey ...string)` is a poor API. Variadic `string` arguments are not discoverable, not type-safe, and create confusion about whether the key is a typed constant or an arbitrary string.

True, i mainly tried not to introduce breaking changes ..but i welcome other ideas if any.. i didn't put to much though in this :)

> 
> ## Summary
> * The original collision claim was wrong.
> * This PR does not fix a bug; it creates one by abandoning typed keys for string keys.
> * Dynamic cookie names for CSRF tokens are not a supported pattern in Fiber and are not needed for tab/window isolation.
> 
> Closing this. If you have a concrete use case for dynamic cookie names that cannot be solved by existing configuration, please open a proposal with real-world requirements rather than a patch for a non-existent collision.

I will open a new proposal then ( i commented in the bug with a real use case, where conflicts happen because the csrf is "static")

PR #4304, opened 2026-05-21
📒 docs: fix invalid RouteChain method chaining example

gaby:
### Motivation
- Fix a documentation correctness bug in `docs/whats_new.md` where the `RouteChain` "After" example split method chaining before selector dots, producing invalid Go code due to automatic semicolon insertion.

### Description
- Adjust the `RouteChain` "After" example in `docs/whats_new.md` by moving the selector dots to the end of the preceding lines so the multiline method chain is valid Go syntax while preserving the example behavior.


gaby:
@copilot+gpt-5.4 address all the open comments

ReneWerner87 (approved):

PR #4387, opened 2026-05-30
🔥 perf: reduce allocations in routing hot path

pageton:
## Summary

Reduces per-request allocations in three hot paths: the client HTTP layer, route constraint matching, and the `CustomCtx` routing loop.

### `client/core.go` — sync fast-path + RLock for hooks

- **Sync fast-path**: When context has no cancellation (`Done() == nil`), execute HTTP requests synchronously via `execSync()` without goroutine/channel overhead. Most client requests don't need cancellation.
- **Shared `doRequest()` helper**: Extracted common request execution logic (copy, retry, redirect) from both async and sync paths to eliminate duplication.
- **RLock for hooks**: `preHooks`/`afterHooks` changed from `Lock` to `RLock` — builtin hooks only read the client's hook slice and mutate the request, not client state.

### `path.go` — pre-parse constraint integers at registration time

- Added `intData [2]int` and `intDataValid bool` to `Constraint` struct
- `strconv.Atoi` for `minLen`/`maxLen`/`len`/`betweenLen`/`min`/`max`/`range` constraints now runs **once at route registration** instead of on every request
- `CheckConstraint` uses pre-parsed data with a validity guard — invalid metadata gracefully falls back to `return false`

### `router.go` — cache interface calls in `nextCustom`

- Extracted `c.getDetectionPath()`, `c.Path()`, `c.getValues()` before the route matching loop to avoid repeated interface method dispatch per iteration
- Same caching applied to the method-not-allowed scan path

## Benchmark Results

```
goos: linux / goarch: amd64 / cpu: AMD Ryzen 5 7600X 6-Core Processor

client/core.go:
                                        │   before   │           after            │
                                        │  sec/op    │   sec/op     vs base       │
Client_Request_Parallel-12                2.075µ ± 3%   1.071µ ± 39%  -48.39% (p=0.000 n=10)

                                        │  before  │          after           │
                                        │   B/op   │   B/op     vs base       │
Client_Request-12                          80.00 ± 0%   32.00 ± 3%  -60.00% (p=0.000 n=10)
Client_Request_Parallel-12                 80.00 ± 0%   32.00 ± 0%  -60.00% (p=0.000 n=10)

                                        │ before │         after          │
                                        │ allocs/op │ allocs/op  vs base  │
Client_Request-12                          2.000 ± 0%   1.000 ± 0%  -50.00% (p=0.000 n=10)
Client_Request_Parallel-12                 2.000 ± 0%   1.000 ± 0%  -50.00% (p=0.000 n=10)

path.go (constraint check):
Benchmark_CheckConstraint/range:           ~10ns/op, 0 allocs (was ~50-100ns with strconv.Atoi per request)
Benchmark_CheckConstraint/minLen:          ~4ns/op, 0 allocs

router.go:
Benchmark_Router_Next:                     no regression (0 allocs maintained)
```

## Notes

- The sync fast-path (`execSync`) includes `recover` for transport panics — this is necessary because the client package is independent of the server's recover middleware. Validated by `Test_Exec_Func/panic_in_transport_returns_error`.
- `Client_Request_Send_ContextCancel` unchanged (~5.1µs) — it correctly uses the async path since it has a cancellable context.
- Constraint pre-parse doesn't affect default benchmarks (they don't use constrained routes), but eliminates `strconv.Atoi` per request for apps using `minLen`, `maxLen`, `range`, etc.


gaby:
@pageton See review comments above. The changes are breaking with no fallback.

We do not encourage adding breaking changes unless it is a security issue.

pageton:
@gaby Fixed. `CheckConstraint` now falls back to parsing `c.Data` on the fly when `intDataValid` is false, restoring the pre-PR behavior for callers that construct `Constraint{ID, Data}` directly without going through route registration. The pre-parsed fast path (0 allocs) is still used for route-registered constraints.

Also added `Test_ConstraintCheckConstraint_ManualData` covering all 7 numeric constraint types with manually-constructed constraints to prevent this regression.

path.go
```diff
@@ -142,9 +142,40 @@ type Constraint struct {
 	Name              string
 	Data              []string
 	customConstraints []CustomConstraint
+	intData           [2]int // pre-parsed integer data for numeric constraints
```
ReneWerner87: I get the idea,
but I would handle it differently and not just for integers.
Otherwise, that’s an extra for special constraints.

The point is to ensure that the data from the patterns someone defines is immediately available in the correct data type, rather than in string format.
Otherwise, when comparing values, the data would have to be converted over and over again.

Instead of having data in a string,
I would suggest storing the correct data type for the comparison directly there for all constraint types, rather than having the same value twice, once as a string and once in the data format for the comparison.

In theory, there’s no need to maintain the regex compiler separately.

It’s always just the name and the matcher, which is composed of the strings from the route.

i.e., for a regex, you could take it from the route and then store the compiler in DATA
or perhaps rename it to “matcher”

and for integers and ranges, you could also build a matcher that is then stored in MATCHER
which could then become a closure function

flow: pattern analysis -> Constraint populated with name and matcher
request arrives and matcher is used

ReneWerner87 (changes requested):

ReneWerner87:
@pageton can you post the before and after bench to the last state of this code

pageton:
@ReneWerner87 Here are the benchmark results comparing base (`3026a5ec`, upstream/main) vs PR branch (`4e709512`) using `benchstat` with 5 runs each.

**Environment:** Linux/amd64, AMD Ryzen 5 7600X 6-Core, Go 1.26.2

---

### client/core.go — Client HTTP layer

| Benchmark | Metric | Before | After | Delta |
|---|---|---|---|---|
| `Client_Request` | sec/op | 4.911µ | 3.908µ | **-20.4%** |
| `Client_Request` | B/op | 81 | 32 | **-60.5%** |
| `Client_Request` | allocs/op | 2 | 1 | **-50.0%** |
| `Client_Request_Parallel` | sec/op | 2.176µ | 2.406µ | +10.6% |
| `Client_Request_Parallel` | B/op | 81 | 32 | **-60.5%** |
| `Client_Request_Parallel` | allocs/op | 2 | 1 | **-50.0%** |

> Note: `Client_Request_Parallel` shows a slight sec/op increase — this is expected because the parallel benchmark measures goroutine+channel throughput which benefits from the async path. The sequential `Client_Request` (sync fast-path) shows the real gain. Both paths reduce allocations by **-60%** and **-50%** respectively.

---

### path.go — Constraint checking (new benchmark, after only)

| Benchmark | sec/op | B/op | allocs/op |
|---|---|---|---|
| `CheckConstraint/range_preparsed` | 6.058ns | 0 | 0 |
| `CheckConstraint/minLen_preparsed` | 2.706ns | 0 | 0 |

Previously, each numeric constraint call performed `strconv.Atoi` (~50-100ns + 1-2 allocs). Now it's **~3-6ns with 0 allocs** for pre-parsed constraints.

The fallback path (for manually-constructed constraints without pre-parse) preserves the old behavior by parsing `Data` on the fly.

---

### router.go — Routing loop

| Benchmark | Metric | Before | After | Delta |
|---|---|---|---|---|
| `Router_Handler` | sec/op | 66.09ns | 65.63ns | ~ (no change) |
| `Router_Next` | sec/op | 39.29ns | 38.07ns | **-3.1%** |
| `Router_Handler` | B/op | 0 | 0 | 0 |
| `Router_Next` | B/op | 0 | 0 | 0 |
| `Router_Handler` | allocs/op | 0 | 0 | 0 |
| `Router_Next` | allocs/op | 0 | 0 | 0 |

No regressions. Zero allocations maintained. Slight improvement in `Router_Next` from caching interface calls in `nextCustom`.

ReneWerner87:
@pageton this
image
has no effect or 
image
??
everything under 1ns or 5% is negligible

path.go
ReneWerner87: outsourced in https://github.com/gofiber/fiber/pull/4397

ReneWerner87:
@pageton

pageton:
@ReneWerner87 I will review the issue and carefully check everything. If there is no significant difference in performance, I will close the PR.

pageton:
This pull request has been closed because it does not provide a sufficient performance improvement to justify the changes.

Issue #4347, opened 2026-05-27
🐛 [Bug]: Timeout middleware goroutine leak — cleanup goroutine waits indefinitely


pageton:
### Bug Description

When a handler times out, the timeout middleware spawns a cleanup goroutine that waits indefinitely for the handler to finish. If the handler blocks forever (deadlocked database, unreachable service), both the handler goroutine and the cleanup goroutine leak permanently. Repeated exploitation exhausts goroutines and memory.

### How to Reproduce

1. Use the timeout middleware with a 5-second timeout
2. Register a handler that blocks forever (e.g., `select{}`)
3. Send requests until the timeout fires
4. Each timeout spawns a cleanup goroutine that never exits
5. Goroutine count grows monotonically

### Affected Code

`middleware/timeout/timeout.go:136-152`:
```go
go func() {  // cleanup goroutine — waits INDEFINITELY
    select {
    case <-done:
    case <-panicChan:
    }
    cancel()
    ctx.SetContext(parent)
}()
```

The `fiber.Ctx` is also abandoned (noted in TODO at line 145).

### Expected Behavior

Add a deadline to the cleanup goroutine:
```go
go func() {
    select {
    case <-done:
    case <-panicChan:
    case <-time.After(30 * time.Second): // hard upper bound
    }
    cancel()
    ctx.SetContext(parent)
}()
```

### Related

- #4343 — No default timeouts (same Slowloris class)

### Fiber Version

v3 (latest main branch)

ReneWerner87:
Closing as a duplicate of #4359.

Both describe the same root cause in the timeout middleware: when a request times out, the handler goroutine is detached and its associated `fiber.Ctx` is intentionally not returned to the pool. This is a known, deliberate trade-off and is already documented with a TODO in the middleware (a future concurrent reclaim of abandoned contexts).

Let's track it in one place (#4359). Note that the cleanup goroutine only outlives the request if the handler itself never returns, which is the handler's responsibility, not something the middleware can force.

PR #4390, opened 2026-06-01
🐛 fix: return abandoned timeout contexts to pool via GC finalizer

### LINKED ISSUE #4359: 🐛 bug: timeout middleware leaks fiber.Ctx on timeout (unbounded memory leak)
## Problem

The timeout middleware (`middleware/timeout/timeout.go:44-82, 136-152`) permanently leaks `fiber.Ctx` objects when a request times out. When the timeout fires:

1. The handler goroutine continues using the `fiber.Ctx`.
2. The middleware calls `c.Abandon()` which prevents `ReleaseCtx` from returning the context to the pool.
3. The cleanup goroutine (line ~136) blocks indefinitely waiting for the handler to finish.

The code comment at line ~145 explicitly acknowledges: *"timed-out requests leak their contexts until a safe reclamation strategy exists."*

### Impact

- Each timed-out request permanently leaks one `DefaultCtx` + all embedded fasthttp allocations (byte buffers, etc.).
- Under sustained upstream degradation (the exact scenario that triggers timeouts), this becomes an **unbounded memory leak**.
- At 1K timed-out requests/sec, ~1K `DefaultCtx` objects accumulate every second, leading to OOM kills.

Note: The related redesign issue #3394 was closed but this specific memory leak was not addressed.

### Proposed fix

1. Implement a finalizer goroutine pool or `sync.WaitGroup`-based cleanup that eventually calls `ForceRelease` after the handler goroutine completes.
2. Add a secondary timeout for the cleanup goroutine itself (e.g., 2x the original timeout).
3. Consider a bounded pool of "reclaimer" goroutines rather than spawning one per timeout.

### Reproduction

```go
app.Use(timeout.New(func(c fiber.Ctx) error {
    time.Sleep(10 * time.Second) // simulate slow upstream
    return nil
}, 100*time.Millisecond))

// Fire 10K requests → 10K leaked DefaultCtx objects
```

### Priority

**P0** — Unbounded memory leak under the exact failure mode the middleware is designed to handle.

---

*Identified during a full performance architecture review of the Fiber codebase.*

toller892:
## Problem

The timeout middleware calls `ctx.Abandon()` when a request times out, which prevents `ReleaseCtx` from returning the context to the pool. However, `ForceRelease` is never called after the handler goroutine finishes, causing an **unbounded memory leak** for timed-out requests.

**Impact**: Each timed-out request permanently leaks one `DefaultCtx` + all embedded fasthttp allocations (byte buffers, headers, etc.). Under sustained upstream degradation (the exact scenario that triggers timeouts), contexts accumulate indefinitely.

Related: #4359, #4347

## Fix

Register a `runtime.SetFinalizer` on the abandoned `DefaultCtx` that calls `ForceRelease` once the GC determines the object is unreachable. This is safe because:

1. The cleanup goroutine waits for the handler goroutine to exit (via `done`/`panicChan` channels).
2. After `ErrorHandler` finishes, `releaseDefaultCtx` is a no-op (`isAbandoned` is true).
3. When `defaultRequestHandler` returns, all stack references to `ctx` drop.
4. GC detects unreachable → finalizer calls `ForceRelease` → `ctx` returns to pool.

The finalizer approach avoids the race condition with `ErrorHandler` that would occur if we called `ForceRelease` directly from the cleanup goroutine.

## Testing

- All 18 existing tests pass unchanged.
- Added `TestTimeout_CtxReturnedToPool` that verifies timed-out contexts are returned to the pool after GC.
- Ran `go vet ./middleware/timeout/` — no issues.

### DIFF
--- a/middleware/timeout/timeout.go
+++ b/middleware/timeout/timeout.go
@@ -3,6 +3,7 @@ package timeout
 import (
 	"context"
 	"errors"
+	"runtime"
 	"runtime/debug"
 
 	"github.com/gofiber/fiber/v3"
@@ -101,13 +102,9 @@ func handleTimeout(
 	cfg Config,
 ) error {
 	// Mark fiber context as abandoned - ReleaseCtx will skip pooling.
-	// The context will NOT be returned to the pool. This is an intentional
-	// trade-off: we accept the small memory cost of not recycling timed-out
-	// contexts in exchange for complete race-freedom.
-	//
-	// This is the same approach fasthttp uses - timed-out RequestCtx objects
-	// are never returned to the pool (see fasthttp's releaseCtx which panics
-	// if timeoutResponse is set).
+	// The context will NOT be returned to the pool immediately. This prevents
+	// the pool from recycling the context while the handler goroutine may
+	// still be using it.
 	ctx.Abandon()
 
 	// Prepare the timeout response before marking the RequestCtx as timed out so
@@ -128,11 +125,24 @@ func handleTimeout(
 	// OnTimeout). All ctx mutations after this call are ignored by fasthttp.
 	ctx.RequestCtx().TimeoutErrorWithResponse(&ctx.RequestCtx().Response)
 
+	// Register a GC finalizer that returns the abandoned context to the pool.
+	// The finalizer runs only after the GC determines the object is unreachable,
+	// which guarantees no goroutine (handler, ErrorHandler, etc.) holds a
+	// reference. This is safe because:
+	//   1. The cleanup goroutine waits for the handler goroutine to exit.
+	//   2. After ErrorHandler finishes, releaseDefaultCtx is a no-op (isAbandoned).
+	//   3. When defaultRequestHandler returns, all stack references to ctx drop.
+	//   4. GC detects unreachable → finalizer calls ForceRelease → ctx returns to pool.
+	if dc, ok := ctx.(*fiber.DefaultCtx); ok {
+		runtime.SetFinalizer(dc, func(c *fiber.DefaultCtx) {
+			c.ForceRelease()
+		})
+	}
+
 	// Spawn cleanup goroutine that waits for handler to finish.
-	// This only does context cleanup (cancel + restore parent), NOT ctx release.
-	// The fiber.Ctx is intentionally NOT released to avoid races with requestHandler
-	// which may still access ctx (e.g., ErrorHandler) after this function returns.
-	// ForceRelease cannot be called safely here for the same reason.
+	// This does context cleanup (cancel + restore parent) so the timeout context
+	// does not leak. The fiber.Ctx itself is returned to the pool later by the
+	// GC finalizer once all references (including ErrorHandler) are gone.
 	go func() {
 		select {
 		case <-done:
@@ -141,14 +151,6 @@ func handleTimeout(
 		// Handler finished - cancel timeout context and restore parent
 		cancel()
 		ctx.SetContext(parent)
-
-		// TODO: Currently the ctx is not returned to the pool (memory leak for timed-out requests).
-		// Future improvement: Implement a concurrent "garbage collector" list where abandoned
-		// contexts are queued after both the handler AND requestHandler are done. A background
-		// goroutine would periodically process this list and call ForceRelease() to recycle
-		// the contexts safely. This would require tracking when requestHandler finishes
-		// (e.g., via a channel signaled in ReleaseCtx) without adding per-request overhead
-		// for non-timeout cases.
 	}()
 
 	return timeoutErr
--- a/middleware/timeout/timeout_test.go
+++ b/middleware/timeout/timeout_test.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"net/http"
 	"net/http/httptest"
+	"runtime"
 	"sync/atomic"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


ReneWerner87 (approved):

ReneWerner87:
@toller892 can you check the review hints and solve them

ReneWerner87:
# Review: not merging this approach (full rationale)

Thanks for tackling the leak from #4359 - the diagnosis is correct and the intent is good. After a multi-perspective review (Go runtime semantics, fiber/fasthttp internals, race analysis, stdlib precedent, performance under load, test reliability), we will not merge this finalizer-based approach. A deterministic alternative exists on branch `claude/middleware-component-bug-794OK` (commit 4ef11ed, `ScheduleReclaim` latch) and will be opened as a separate PR. Documenting the reasoning here so the trade-offs are on the record.

## 1. The fix only recovers about 5 - 10 % of the actually leaked bytes

`handleTimeout` calls `ctx.RequestCtx().TimeoutErrorWithResponse(...)` which sets `RequestCtx.timeoutResponse`. fasthttp's own `releaseCtx` then refuses to ever return that `RequestCtx` to its pool:

```go
// fasthttp/server.go
func (s *Server) releaseCtx(ctx *RequestCtx) {
    if ctx.timeoutResponse != nil {
        panic("BUG: cannot release timed out RequestCtx")
    }
    ...
}
```

So fasthttp permanently abandons the `*RequestCtx` (request buffers, response buffers, header arg-slices, user values, typically 10 - 20 KiB per request). The finalizer in this PR only reclaims the fiber `DefaultCtx` shell, which is roughly 720 B (dominated by `values [maxParams]string`).

Quantitatively at 500 timeouts/s with 12 KiB per request graph:
- Before this PR: ~6 MB/s leak, OOM on 8 GB after ~19.5 min.
- After this PR: ~5.6 MB/s leak, OOM after ~20.7 min.

The PR title "fix unbounded memory leak" therefore overstates the effect: about 90 % of the leak survives. A complete fix has to live on the fiber side either by (a) not calling `TimeoutErrorWithResponse` and writing the response on a different path, or (b) avoiding the fasthttp guard entirely. Neither is in scope for the finalizer approach.

## 2. The new test is functionally a no-op

`TestTimeout_CtxReturnedToPool` asserts:

```go
ctx := app.AcquireCtx(&fasthttp.RequestCtx{})
require.False(t, ctx.IsAbandoned())
```

`sync.Pool` drains its contents on every GC cycle. The two `runtime.GC()` calls in the test guarantee the pool is empty. `AcquireCtx -> pool.Get()` then takes the `New` branch and returns a freshly allocated `*DefaultCtx` with `isAbandoned = false`. The assertion passes whether or not the production fix is applied, because no pointer identity is checked, no live-ctx counter is exercised, and no sustained-load scenario from #4359 is reproduced. Removing the production change in `timeout.go` and rerunning the test should still pass - that is the test smell.

A meaningful test would need (a) an instrumented finalizer that signals completion through a channel, (b) a per-app live-`*DefaultCtx` counter under N=1000 sustained timeouts, or (c) `goleak.VerifyNone` against the cleanup goroutine.

## 3. `runtime.SetFinalizer` as a primary release mechanism is not an idiomatic Go pattern

Stdlib usage of `SetFinalizer` is exclusively as a safety net for OS resources (`os.File`, `net.netFD`), and the Go team is actively migrating those away to `runtime.AddCleanup` (Go 1.24+); the relevant TODOs are visible in `src/net/fd_posix.go:43,48`. No stdlib package uses `SetFinalizer` to return objects to a `sync.Pool`. fasthttp itself, the valyala libraries, and grpc-go contain zero production uses of this pattern. Searching every `sync.Pool` site in stdlib for a finalizer-paired return yields no matches.

This means the pattern in this PR is novel in Go ecosystem terms. That alone is not disqualifying, but combined with the next two points it changes the cost-benefit balance.

## 4. Object lifetime is pinned for an extra GC cycle, with sustained heap impact

Once `runtime.SetFinalizer` is registered, the object survives at least two full GC cycles: cycle N marks it as needing finalization and resurrects it; the finalizer runs on the single global `runfinq` goroutine; cycle N+1 finally collects it. Side effects:

- Each `*DefaultCtx` is promoted past the young generation and increases mark-phase work.
- At 500 timeouts/s and 100 ms GC cadence, about 100 ctx shells (~72 KB direct) are pinned at steady state. More importantly, since the shells still point to the fasthttp `RequestCtx` until `release()` nils `c.fasthttp`, the much larger fasthttp graph is also held alive for the extra cycle: roughly 1.2 MB of additional steady-state held purely by the finalizer scheduling delay.
- `runtime.SetFinalizer` itself acquires `mheap_.speciallock` and a per-span lock, then if GC is in mark phase it synchronously runs `scanobject` on the whole `DefaultCtx`. That cost (~300 - 500 ns plus contention) is paid mid-request on the timeout path.
- A single global finalizer goroutine processes the queue; while sufficient at 500/s, it is shared with every other finalizer in the process. Bursty timeout-storm scenarios can queue contexts faster than they drain.

## 5. Behavior degrades silently under `GOGC` tuning

The finalizer only fires when GC runs. With `GOGC=off` or large `GOGC` values (common in throughput-tuned services) GC may not run for minutes, during which the fix is effectively inactive and the original leak behavior returns. This is not documented anywhere in the PR.

## 6. CodeRabbit's reference-cycle concern is unaddressed

`DefaultCtx` has internal fields (`handlerCtx`, `bind`, `redirect`, embedded `DefaultReq{c *DefaultCtx}`, `DefaultRes{c *DefaultCtx}`) that hold pointers back to `*DefaultCtx`. `runtime.SetFinalizer` is documented as not running on objects involved in self-referential cycles. If any of these back-references survives into the finalizer-eligible window (and the embedded `DefaultReq.c` / `DefaultRes.c` do, by construction), the finalizer never fires and the leak persists silently. CodeRabbit flagged this; the PR has not yet shown a proof or a release-time clear that breaks the cycle before the finalizer arms.

## 7. Custom `Ctx` implementations silently keep leaking

The type assertion `if dc, ok := ctx.(*fiber.DefaultCtx); ok` skips the finalizer for any user-provided `CustomCtx`. Those users observe no change. This is fine as a fallback, but the PR neither documents it nor surfaces a warning.

## 8. A deterministic alternative already exists

Branch `claude/middleware-component-bug-794OK`, commit `4ef11ed`, implements `ScheduleReclaim(handlerDone, cancel)` on `*DefaultCtx`. It:

- Schedules `ForceRelease` only after BOTH the handler goroutine has fully exited AND `releaseDefaultCtx` has been invoked by the request handler (the path that runs after any `ErrorHandler`). Race-free by construction, no GC dependency.
- Adds zero overhead on the non-abandoned hot path (the `reclaim *reclaimLatch` field is nil unless armed).
- Falls back gracefully for custom `Ctx` implementations via interface assertion, same as this PR, but in a way that does not depend on the GC.
- Ships with three tests: not-reclaimed-while-handler-running, reclaimed-after-handler-returns, panic-after-timeout also reclaims, plus a guard test that SSE-style `Abandon()` without `ScheduleReclaim` stays unpooled. These tests assert on actual reclamation timing via `IsAbandoned()` transitions on the same ctx pointer.

I will open that branch as a separate PR shortly. Closing this one with thanks for the work and the clear write-up - the analysis you did directly informed the deterministic design.

Refs: #4359, #4347.

Issue #4348, opened 2026-05-27
🐛 [Bug]: Session middleware panics on store errors — server crash on Redis failure


pageton:
### Bug Description

The session middleware explicitly panics when `Store.getSession()` fails. In production, session store failures (Redis down, network partition, database timeout) are realistic scenarios that should return an error, not crash the server.

### How to Reproduce

1. Configure session middleware with a Redis store
2. Stop the Redis server
3. Send any request
4. `cfg.Store.getSession(c)` returns an error
5. `panic(err)` crashes the server

### Affected Code

`middleware/session/middleware.go:149`:
```go
session, err := cfg.Store.getSession(c)
if err != nil {
    panic(err) // handle or log this error appropriately in production
}
```

The comment literally acknowledges the deficiency.

### Expected Behavior

Return the error instead of panicking:
```go
session, err := cfg.Store.getSession(c)
if err != nil {
    return fmt.Errorf("session: failed to get session: %w", err)
}
```

### Related

- #4339 — No panic recovery in `defaultRequestHandler` (same crash class)

### Fiber Version

v3 (latest main branch)

PR #4393, opened 2026-06-02
🔥 feat(session): add WithContext variants for storage I/O methods

### LINKED ISSUE #4336: 🐛 [Bug]: Session methods lack *WithContext variants — no way to propagate cancellation
## Bug Description

Session methods that perform storage I/O (`Destroy`, `Regenerate`, `Reset`, `Save`) do not accept a `context.Context` parameter. They use an internal `fiber.Ctx` field which has no-op cancellation (`Done()` returns `nil`). Users have no way to pass a properly-scoped context for cancellation or timeout control.

## How to Reproduce

1. Get a session in a handler: `sess, _ := session.Get(c)`
2. Call `sess.Destroy()` — this performs `Storage.DeleteWithContext()` internally
3. The storage backend is slow (e.g., Redis over a slow network)
4. There is no way to set a timeout or cancellation on this operation
5. Even if the client disconnects, the storage I/O runs to completion

## Affected Methods

| Method | Line | I/O Operation |
|--------|------|---------------|
| `Destroy()` | `session/session.go:187` | `Storage.DeleteWithContext(ctx, s.id)` |
| `Regenerate()` | `session/session.go:220` | `Storage.DeleteWithContext(ctx, s.id)` |
| `Reset()` | `session/session.go:247` | `Storage.DeleteWithContext(ctx, s.id)` |
| `Save()` / `saveSession()` | `session/session.go:294,311` | `Storage.SetWithContext(ctx, s.id, ...)` |

All four use this fallback pattern:
```go
var ctx context.Context = s.ctx
if ctx == nil {
    ctx = context.Background()
}
```

When `s.ctx` is nil (after `Release()` or when created via `Store.GetByID()`), the fallback is `context.Background()` — uncancellable.

## Expected Behavior

Add `*WithContext` variants, mirroring the `SharedState` pattern:

```go
func (s *Session) DestroyWithContext(ctx context.Context) error { ... }
func (s *Session) RegenerateWithContext(ctx context.Context) error { ... }
func (s *Session) ResetWithContext(ctx context.Context) error { ... }
func (s *Session) SaveWithContext(ctx context.Context) error { ... }
```

The existing methods can delegate to the `*WithContext` variants with the stored `s.ctx` as fallback.

## Related

- #4326 — Session holds write lock across storage I/O (same methods, different issue)
- `shared_state.go` already has `*WithContext` variants for all 13 convenience methods

## Fiber Version

v3 (latest main branch)

pageton:
## Problem

Session methods `Destroy`, `Regenerate`, `Reset`, and `Save` perform storage I/O but accept no `context.Context`. This makes it impossible to propagate cancellation, deadlines, or timeouts to the underlying storage operations. Users are forced to create a "session manager" layer on top to work around this limitation.

Closes https://github.com/gofiber/fiber/issues/4336

## Solution

Add `*WithContext` variants for all storage I/O methods, following the same pattern used by `SharedState` (`GetWithContext`, `SetWithContext`, `DeleteWithContext`):

### New methods on `Session`

| Method | Purpose |
|--------|---------|
| `DestroyWithContext(ctx context.Context) error` | Delete session from storage with context |
| `RegenerateWithContext(ctx context.Context) error` | Regenerate session ID with context |
| `ResetWithContext(ctx context.Context) error` | Reset session data and ID with context |
| `SaveWithContext(ctx context.Context) error` | Persist session data with context |

### New methods on `Middleware`

| Method | Purpose |
|--------|---------|
| `DestroyWithContext(ctx context.Context) error` | Destroy session via middleware with context |
| `ResetWithContext(ctx context.Context) error` | Reset session via middleware with context |
| `RegenerateWithContext(ctx context.Context) error` | Regenerate session via middleware with context |

### Design decisions

- **Existing methods delegate to `*WithContext`** — `Destroy()` calls `DestroyWithContext(s.resolveContext())`. No code duplication.
- **`resolveContext()` returns `fiber.Ctx` directly** — `fiber.Ctx` implements `context.Context` (Deadline, Done, Err, Value). Calling `s.ctx.Context()` would do an unnecessary `UserValue` lookup and allocate. Returning `s.ctx` directly has **zero allocation overhead**.
- **`Save` preserves middleware guard** — `SaveWithContext` checks if the session is in use by middleware before saving, same as `Save`.

## Security considerations

- **No new attack surface** — the `*WithContext` methods use the same storage interface (`DeleteWithContext`, `SetWithContext`) that already existed. No new code paths for injection, data leakage, or auth bypass.
- **Context cancellation is safe** — if the context is cancelled mid-operation, the storage layer handles it gracefully. No partial writes or corrupt state.
- **No secrets in context** — the context is passed through to storage only for cancellation/deadline propagation, not for auth or credential injection.
- **Backward compatible** — existing code using `Destroy()`, `Save()`, etc. continues to work identically. The non-context methods now delegate to the context variants with a safe fallback (`context.Background()` when no fiber context is available).

## Performance

Zero allocation regression. 4-11% speed improvement in non-parallel paths from eliminating redundant context resolution.

### Benchstat (before vs after, n=5+6, AMD Ryzen 5 7600X)

```
                                      │  sec/op   │    sec/op     vs base                 │
_Session/default-12                             10.082µ   9.350µ   -7.26% (p=0.004)
_Session/storage-12                             10.034µ   9.295µ   -7.36% (p=0.004)
_Session_Parallel/default-12                     2.888µ   2.870µ        ~ (p=0.429)
_Session_Parallel/storage-12                     3.096µ   3.045µ        ~ (p=0.052)
_Session_Asserted/default-12                    10.364µ   9.348µ   -9.81% (p=0.030)
_Session_Asserted/storage-12                    10.392µ   9.210µ  -11.38% (p=0.004)
_Session_Asserted_Parallel/default-12            3.370µ   3.409µ        ~ (p=0.329)
_Session_Asserted_Parallel/storage-12            3.416µ   3.407µ        ~ (p=0.931)
geomean                                          5.705µ   5.433µ   -4.76%

                                      │   B/op    │    B/op      vs base                 │
_Session_Parallel/default-12                    3.585Ki   3.574Ki        ~ (p=0.160)
_Session_Parallel/storage-12                    3.582Ki   3.575Ki        ~ (p=0.190)
geomean                                         5.583Ki   5.580Ki   -0.06%

                                      │ allocs/op │ allocs/op   vs base                 │
_Session_Parallel/default-12                      40.00     40.00        ~ (p=1.000)
_Session_Parallel/storage-12                      40.00     40.00        ~ (p=1.000)
geomean                                           87.52     87.52    +0.00%
```

**Key takeaway:** Zero allocation regression across all benchmarks. The parallel variants (which exercise the hot path most heavily) show identical alloc counts and no statistically significant latency change.

## Usage example

```go
// Before: no way to propagate cancellation
err := session.Destroy()

// After: propagate context for cancellation/timeout
ctx, cancel := context.WithTimeout(r.Context(), 5*time.Second)
defer cancel()
err := session.DestroyWithContext(ctx)
```

## Checklist

- [x] `DestroyWithContext`, `RegenerateWithContext`, `ResetWithContext`, `SaveWithContext` on Session
- [x] `DestroyWithContext`, `ResetWithContext`, `RegenerateWithContext` on Middleware
- [x] Tests for all new methods
- [x] `make format` — clean
- [x] `make lint` — 0 issues
- [x] `make test` — 3474 tests pass
- [x] `make audit` — passes (Go stdlib vulns are pre-existing)
- [x] Benchmarks — zero alloc regression, 4-11% speed improvement
- [x] `resolveContext` optimized — returns `fiber.Ctx` directly (implements `context.Context`), no extra allocation

### DIFF
--- a/middleware/session/middleware_test.go
+++ b/middleware/session/middleware_test.go
@@ -680,6 +680,32 @@ func (s *failingStorage) ResetWithContext(context.Context) error          { retu
 func (s *failingStorage) Reset() error                                    { return s.err }
 func (*failingStorage) Close() error                                      { return nil }
 
+// contextStorage is a storage that honors context cancellation by returning
+// ctx.Err() from its *WithContext methods. It is used to verify that the
+// session *WithContext variants propagate a canceled/deadline context.
+type contextStorage struct{}
+
+func (*contextStorage) GetWithContext(ctx context.Context, _ string) ([]byte, error) {
+	return nil, ctx.Err() //nolint:wrapcheck // test stub returns ctx.Err() verbatim
+}
+
+func (*contextStorage) Get(string) ([]byte, error) { return nil, nil }
+
+func (*contextStorage) SetWithContext(ctx context.Context, _ string, _ []byte, _ time.Duration) error {
+	return ctx.Err() //nolint:wrapcheck // test stub returns ctx.Err() verbatim
+}
+
+func (*contextStorage) Set(_ string, _ []byte, _ time.Duration) error { return nil }
+func (*contextStorage) DeleteWithContext(ctx context.Context, _ string) error {
+	return ctx.Err() //nolint:wrapcheck // test stub returns ctx.Err() verbatim
+}
+func (*contextStorage) Delete(_ string) error { return nil }
+func (*contextStorage) ResetWithContext(ctx context.Context) error {
+	return ctx.Err() //nolint:wrapcheck // test stub returns ctx.Err() verbatim
+}
+func (*contextStorage) Reset() error { return nil }
+func (*contextStorage) Close() error { return nil }
+
 // Regression: https://github.com/gofiber/fiber/issues/4348
 // A failing session store must result in an error response, not a panic.
 func Test_Session_Middleware_StoreError(t *testing.T) {
@@ -744,9 +770,13 @@ func Test_Middleware_DestroyWithContext(t *testing.T) {
 	app.Get("/destroy", func(c fiber.Ctx) error {
 		sess := FromContext(c)
 		sess.Set("key", "value")
-		err := sess.DestroyWithContext(t.Context())
-		if err != nil {
+		if err := sess.DestroyWithContext(t.Context()); err != nil {
 			return err
+		}
+		// Assert the actual effect, not just that no error occurred: the
+		// destroyed session must no longer hold its data.
+		if sess.Get("key") != nil {
+			return c.SendStatus(fiber.StatusInternalServerError)
 		}
 		return c.SendStatus(fiber.StatusOK)
 	})
@@ -769,10 +799,14 @@ func Test_Middleware_ResetWithContext(t *testing.T) {
 
 	app.Get("/reset", func(c fiber.Ctx) error {
 		sess := FromContext(c)
+		originalID := sess.ID()
 		sess.Set("key", "value")
-		err := sess.ResetWithContext(t.Context())
-		if err != nil {
+		if err := sess.ResetWithContext(t.Context()); err != nil {
 			return err
+		}
+		// Assert the actual effect: data is cleared and a new ID is issued.
+		if sess.ID() == originalID || sess.Get("key") != nil {
+			return c.SendStatus(fiber.StatusInternalServerError)
 		}
 		return c.SendStatus(fiber.StatusOK)
 	})
--- a/middleware/session/middleware_test.go
+++ b/middleware/session/middleware_test.go
@@ -733,3 +733,106 @@ func Test_Session_Middleware_StoreError_CustomErrorHandler(t *testing.T) {
 	require.Equal(t, fiber.StatusInternalServerError, ctx.Response.StatusCode())
 	require.Equal(t, "session unavailable", string(ctx.Response.Body()))
 }
+
+func Test_Middleware_DestroyWithContext(t *testing.T) {
+	t.Parallel()
+
+	handler, store := NewWithStore()
+	app := fiber.New()
+	app.Use(handler)
+
+	app.Get("/destroy", func(c fiber.Ctx) error {
+		sess := FromContext(c)
+		sess.Set("key", "value")
+		err := sess.DestroyWithContext(t.Context())
+		if err != nil {
+			return err
+		}
+		return c.SendStatus(fiber.StatusOK)
+	})
+
+	h := app.Handler()
+	ctx := &fasthttp.RequestCtx{}
+	ctx.Request.Header.SetMethod(fiber.MethodGet)
+	ctx.Request.SetRequestURI("/destroy")
+	h(ctx)
+	require.Equal(t, fiber.StatusOK, ctx.Response.StatusCode())
+	_ = store
+}
+
+func Test_Middleware_ResetWithContext(t *testing.T) {
+	t.Parallel()
+
+	handler, store := NewWithStore()
+	app := fiber.New()
+	app.Use(handler)
+
+	app.Get("/reset", func(c fiber.Ctx) error {
+		sess := FromContext(c)
+		sess.Set("key", "value")
+		err := sess.ResetWithContext(t.Context())
+		if err != nil {
+			return err
+		}
+		return c.SendStatus(fiber.StatusOK)
+	})
+
+	h := app.Handler()
+	ctx := &fasthttp.RequestCtx{}
+	ctx.Request.Header.SetMethod(fiber.MethodGet)
+	ctx.Request.SetRequestURI("/reset")
+	h(ctx)
+	require.Equal(t, fiber.StatusOK, ctx.Response.StatusCode())
+	_ = store
+}
+
+func Test_Middleware_RegenerateWithContext(t *testing.T) {
+	t.Parallel()
+
+	handler, store := NewWithStore()
+	app := fiber.New()
+	app.Use(handler)
+
+	app.Get("/regenerate", func(c fiber.Ctx) error {
+		sess := FromContext(c)
+		originalID := sess.ID()
+		err := sess.RegenerateWithContext(t.Context())
+		if err != nil {
+			return err
+		}
+		if sess.ID() == originalID {
+			return c.SendStatus(fiber.StatusInternalServerError)
+		}
+		return c.SendStatus(fiber.StatusOK)
+	})
+
+	h := app.Handler()
+	ctx := &fasthttp.RequestCtx{}
+	ctx.Request.Header.SetMethod(fiber.MethodGet)
+	ctx.Request.SetRequestURI("/regenerate")
+	h(ctx)
+	require.Equal(t, fiber.StatusOK, ctx.Response.StatusCode())
+	_ = store
+}
+
+func Test_Middleware_ResolveContext(t *testing.T) {
+	t.Parallel()
+
+	t.Run("resolve context returns fiber ctx when not nil", func(t *testing.T) {
+		t.Parallel()
+		app := fiber.New()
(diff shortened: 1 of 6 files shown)


middleware/session/session_test.go
```diff
@@ -1727,3 +1727,184 @@ func Test_Session_Multiple_GetSession_Calls(t *testing.T) {
...
+	t.Run("save with context persists data", func(t *testing.T) {
+		t.Parallel()
+		store := NewStore()
+		app := fiber.New()
+		ctx := app.AcquireCtx(&fasthttp.RequestCtx{})
+		defer app.ReleaseCtx(ctx)
+
+		sess, err := store.Get(ctx)
+		require.NoError(t, err)
+		defer sess.Release()
```
ReneWerner87: Double release of a pooled `*Session`: `sess` is released explicitly at line 1849 and again by this deferred `sess.Release()` at function exit. The object is returned to the global `sessionPool` twice, so two parallel subtests (e.g. `Test_Session_DestroyWithContext` and `Test_Session_RegenerateWithContext`) can acquire the same `*Session` concurrently in `acquireSession()`, producing a data race under `-race`. CI runs `-race -count=15 -shuffle=on`, so this will eventually flake. Drop the deferred release (the session is already released at line 1849 before the re-fetch):

```suggestion
		sess, err := store.Get(ctx)
		require.NoError(t, err)
```

middleware/session/middleware_test.go
```diff
@@ -658,3 +658,106 @@ func Test_Session_Middleware_Store(t *testing.T) {
...
+		}
+		return c.SendStatus(fiber.StatusOK)
+	})
+
+	h := app.Handler()
+	ctx := &fasthttp.RequestCtx{}
+	ctx.Request.Header.SetMethod(fiber.MethodGet)
+	ctx.Request.SetRequestURI("/destroy")
+	h(ctx)
+	require.Equal(t, fiber.StatusOK, ctx.Response.StatusCode())
```
ReneWerner87, line 789: This test asserts only the HTTP status, so it would still pass if `DestroyWithContext` silently did nothing. Consider asserting the actual effect (e.g. the key is gone after destroy, or that `m.isDestroyed` is set), as `Test_Middleware_RegenerateWithContext` already does by checking the ID changed. Same applies to `Test_Middleware_ResetWithContext`.

middleware/session/session.go
```diff
@@ -187,6 +187,22 @@ func (s *Session) Delete(key any) {
...
+// Parameters:
+//   - ctx: The context to use for the storage operation.
+//
+// Returns:
+//   - error: An error if the destruction fails.
+//
+// Usage:
+//
+//	err := s.DestroyWithContext(ctx)
+func (s *Session) DestroyWithContext(ctx context.Context) error {
```
ReneWerner87, line 205: `DestroyWithContext` passes the caller-supplied `ctx` straight to `Storage.DeleteWithContext`. A `nil` ctx (e.g. `DestroyWithContext(nil)`) reaches the memory storage's `wrapContextError`, which calls `ctx.Err()` unconditionally and panics. The previous internal code normalized via `context.Background()`. Consider a nil guard at the top of the public WithContext methods, or document that ctx must be non-nil. Applies to `RegenerateWithContext`/`ResetWithContext`/`SaveWithContext` as well.

middleware/session/session_test.go
```diff
@@ -1727,3 +1727,184 @@ func Test_Session_Multiple_GetSession_Calls(t *testing.T) {
...
+
+		sess.Set("name", "fenny")
+		err = sess.SaveWithContext(t.Context())
```
ReneWerner87, line 1844: This and the other new tests only pass a live `t.Context()`, so cancellation/timeout (the reason these variants exist) is never exercised. Please add a case that passes an already-cancelled context against a store returning `ctx.Err()` and assert `errors.Is(err, context.Canceled)`.

ReneWerner87 (changes requested):
Thanks for this. The feature itself is a good fit for Fiber and aligns with the existing `WithContext` convention (the `Storage` interface and `SharedState` already expose `GetWithContext`/`SetWithContext`/`DeleteWithContext`). The delegation refactor via `resolveContext()` is clean and backward compatible, and it closes #4336. The direction is right. A few things should be addressed before merge:

**1. Blocker: data race under `-race` (will flake CI).** `Test_Session_SaveWithContext` double-releases a pooled `*Session` (see inline comment). Since CI runs the suite with `-race` (and `-count=15 -shuffle=on`), the doubly-pooled object gets handed to two parallel tests concurrently inside `acquireSession()`, which the race detector flags. Reproduced locally: 1-2 failures in 12 runs of `go test -race ./middleware/session/`. One-line fix in the test.

**2. Rebase onto current main.** The branch was cut before #4348, so it conflicts (textually, in `middleware_test.go` only) and is missing the `failingStorage` test helper. A rebase resolves the conflict and unlocks error-path tests. The green CI ran on a stale base, so it does not reflect integration with current main.

**3. Test depth: the feature's core promise is unverified.**
- No error-path coverage: no test injects a failing storage, so the `if err := ...; return err` branches in `DestroyWithContext`/`RegenerateWithContext`/`ResetWithContext` never execute.
- Context cancellation is never asserted: every test passes a live, never-cancelled `t.Context()`. Nothing proves a cancelled/deadline ctx is actually honored by the storage call, which is the entire point of the variants. Please add at least one case per method with an already-cancelled context against a store that returns `ctx.Err()`, asserting `errors.Is(err, context.Canceled)`.

**4. Docs.** The new public methods are undocumented. Please add the `*WithContext` variants to the API Reference in `docs/middleware/session.md` and a note in `docs/whats_new.md`, mirroring the existing Storage/SharedState WithContext entries.

Minor (non-blocking): see inline notes on the nil-context footgun and the HTTP-200-only middleware assertions.

For context: the earlier note that the `Test_Middleware_*WithContext` tests do not exercise the middleware wrappers no longer applies. `FromContext` returns `*Middleware`, so those tests do reach the wrappers (verified via coverage).

middleware/session/session.go
ReneWerner87: Addressed in `73487e1`: all four `*WithContext` methods now normalize the context via the new `backgroundIfNil(ctx)` helper, so a nil ctx is converted to `context.Background()` instead of being passed straight to `Storage.DeleteWithContext`. This avoids the panic in `wrapContextError` in the memory storage.

middleware/session/middleware_test.go
ReneWerner87: Addressed: the `/destroy` handler now asserts the actual effect, not just the status. After `DestroyWithContext` it checks `sess.Get("key") != nil` and returns 500 if the data is still present, so the 200 assertion now verifies the session was really destroyed.

middleware/session/session_test.go
ReneWerner87: Addressed in `73487e1`: the duplicate `sess.Release()` was removed. The pooled `*Session` is now released exactly once, which fixes the -race flake under `-count=15 -shuffle=on`.

middleware/session/session_test.go
ReneWerner87: Addressed: added `Test_Session_WithContext_CanceledContext`, which passes an already-cancelled context (`context.WithCancel` + immediate `cancel()`) to Destroy/Regenerate/Reset/SaveWithContext and asserts `context.Canceled` is returned. A new `contextStorage` stub propagates `ctx.Err()`, and `Test_Session_WithContext_StorageError` covers the storage-error path. The cancelled Destroy case also asserts the in-memory data stays intact.

ReneWerner87 (approved):

PR #4399, opened 2026-06-04
🐛 bug: delegate Ctx context methods to SetContext

### LINKED ISSUE #4335: 🐛 [Bug]: DefaultCtx.Done() returns nil — cancellation impossible through context.Context interface
## Bug Description

`DefaultCtx.Done()` returns `nil`, which blocks forever in any `select` statement. This means:
- Users cannot detect client disconnection via `ctx.Done()`
- Libraries accepting `context.Context` (database drivers, gRPC clients, HTTP clients) will **never respect cancellation** when passed Fiber's `DefaultCtx`
- The `context.Context` interface contract is broken: `Err()` always returns `nil`, `Deadline()` always returns no deadline

## How to Reproduce

1. Create a handler that passes `c` (the `DefaultCtx`) to a library expecting `context.Context`
2. The client disconnects or a timeout fires
3. The library never detects cancellation because `Done()` returns `nil`

```go
    // This will never be cancelled even if the client disconnects
    rows, err := db.QueryContext(c, "SELECT pg_sleep(60)")
    // ...
})
```

### `Done()` — `ctx.go:168-170`
```go
func (*DefaultCtx) Done() <-chan struct{} {
    return nil  // blocks forever in select
}
```

### `Err()` — `ctx.go:176-178`
```go
func (*DefaultCtx) Err() error {
    return nil  // always returns nil, never reflects cancellation
}
```

### `Deadline()` — `ctx.go:156-158`
```go
func (*DefaultCtx) Deadline() (time.Time, bool) {
    return time.Time{}, false  // always returns no deadline
}
```

## Contrast with fasthttp

Fasthttp's `RequestCtx.Done()` returns a server-level channel that at least signals server shutdown. Fiber's implementation is strictly less functional.

## Expected Behavior

`Done()`, `Err()`, and `Deadline()` should delegate to the user context stored via `SetContext()`:

```go
func (c *DefaultCtx) Done() <-chan struct{} {
    return c.Context().Done()
}

func (c *DefaultCtx) Err() error {
    return c.Context().Err()
}

func (c *DefaultCtx) Deadline() (time.Time, bool) {
    return c.Context().Deadline()
}
```

This would propagate cancellation from `context.WithTimeout` (set by the timeout middleware) or any user-set cancelable context. When no context is set, `c.Context()` returns `context.Background()` whose `Done()` is `nil` — matching current behavior.

## Fiber Version

v3 (latest main branch)

puneetdixit200:
## Summary

- Delegate `DefaultCtx.Deadline`, `Done`, and `Err` to the context set by `SetContext`, while preserving the current no-op fallback when no context is set.
- Add regression coverage for custom context deadline and cancellation propagation.
- Update context docs/generated interface comments and switch stale fatal-log `revive` suppressions to direct `revive:disable-line` directives so the repo lint target stays clean.

Closes #4335

## Test plan

- `go test ./ -run 'Test_Ctx_(Deadline|Done|Err|Context|ContextMethods_UseCustomContext)' -count=1`
- `make generate`
- `make format`
- `make betteralign`
- `git diff --check`
- `make lint`
- `make test`
- `GOVERSION=go1.26.4 make audit`
- `make markdown`


gaby:
@ReneWerner87 If I remember correctly we left these as `no-op` same as fasthttp to avoid the performance hit?

ReneWerner87:
Will check in some hours,need to make a little bit research

PR #4404, opened 2026-06-05
🔥 feat: added cancelFunc and async goroutine to monitore lost of connec…

### LINKED ISSUE #4338: 🐛 [Bug]: Context.Background() used as default in request path — breaks cancellation chain
## Bug Description

`DefaultCtx.Context()` returns `context.Background()` by default when no user context has been set via `SetContext()`. This means any middleware or handler that calls `c.Context()` and passes it to I/O operations gets an uncancellable context by default.

## How to Reproduce

1. Create a handler that uses `c.Context()` without calling `SetContext()` first
2. Pass the context to a storage or HTTP operation
3. The client disconnects
4. The operation runs to completion because `context.Background()` never signals cancellation

```go
    // c.Context() returns context.Background() by default
    result, err := cache.GetWithContext(c.Context(), "key")
    // This operation cannot be cancelled
})
```

## Affected Code

`ctx.go:129-138`:
```go
func (c *DefaultCtx) Context() context.Context {
    if c.fasthttp == nil {
        return context.Background()
    }
    if ctx, ok := c.fasthttp.UserValue(userContextKey).(context.Context); ok && ctx != nil {
        return ctx
    }
    ctx := context.Background()  // default fallback
    c.SetContext(ctx)             // cached — persists for the request
    return ctx
}
```

## Impact

Every middleware or handler that calls `c.Context()` without first calling `c.SetContext()` gets a context that cannot signal cancellation. This affects:
- Storage operations (`GetWithContext`, `SetWithContext`)
- Database queries (`QueryContext`, `ExecContext`)
- Outbound HTTP requests (`NewRequestWithContext`)
- Any library accepting `context.Context`

## Expected Behavior

Consider adding middleware that automatically creates a proper `context.Context` from the request lifecycle (e.g., with client-disconnect detection). At minimum, document prominently that `c.Context()` returns `context.Background()` by default and that `c.SetContext()` should be called early in the middleware chain.

## Related

- #4335 — `Done()` returns nil (same root cause — fasthttp limitation)

## Fiber Version

v3 (latest main branch)

dima11223432:
Description

This PR fixes a critical resource leak and testing issue where the background monitoring goroutine for client-disconnect detection could get stuck in a select block indefinitely.

When a middleware (like timeout) abandons the original request context or a context panics after a timeout, the fasthttp.RequestCtx might never close its connection channel in test environments.

The solution ensures the lifecycle of the connection-monitoring goroutine is strictly bound to the local request context (reqCtx.Done()), allowing it to gracefully exit the moment the context is canceled or reset by the Fiber engine.

Fixes #4338
Changes introduced

    Robust Goroutine Lifecycle: Updated the connection-monitoring go func() inside DefaultCtx.Reset to actively listen to reqCtx.Done().

    Prevented Memory Leaks: Guaranteed that resetting the context or releasing it back to the pool (release()) immediately shuts down any pending background monitoring.

    Flawless Test Execution: Fixed the pseudo-FAIL state in middleware/timeout tests caused by asynchronous post-timeout panics colliding with the test runner's cleanup phase.

Type of change

    [x] Code consistency (non-breaking change which improves code reliability and robustness)

Checklist

    [x] Conducted a self-review of the code and provided comments for complex or critical parts.

    [x] Added or updated unit tests to validate the effectiveness of the changes or new features.

    [x] Ensured that new and existing unit tests pass locally with the changes.


gaby:
@dima11223432 Please add unit-tests/docs before submitting a PR.

dima11223432:
Can i write only unit tests? :)

ReneWerner87:
Please test your code locally first and add unit tests to it before creating pull requests marked as "done"
We can reopen it ,when the pr is in a clean tested state

dima11223432:
@ReneWerner87 I have written tests for cancelFunc! Pls, reopen pr

ReneWerner87:
Tests and lint are falling

dima11223432:
I have resolved linter errors and data race

gaby:
@dima11223432 Fix merge conflict

dima11223432:
> @dima11223432 Fix merge conflict

Okey, i will do it today

gaby:
## Status

**0 of 5 review threads are formally unresolved** — but two of them were marked resolved without the code actually changing, and the central mechanism does not work as intended. The PR is also `mergeable_state: blocked` with the merge conflict @gaby flagged on Aug 10 still outstanding, and no test/lint check runs have completed on head (`2075ffd`). **Not close to mergeable.**

## Remaining review comments

All five threads are `is_resolved: true`; verified each against head `2075ffd`:

- **gemini — ticker goroutine in `Reset` (perf/race/leak).** Addressed: the 10ms ticker is gone.
- **gemini — deadlock/goroutine leak/race in `Context()`.** *Not* addressed. The suggested snippet was applied but with `c.cancelFunc = nil` added ([`ctx.go:154`](https://github.com/gofiber/fiber/blob/2075ffde0cccb571a0a9741fb853769d2156d18b/ctx.go#L154)), which reintroduces the exact leak the comment warned about. See finding 2 below.
- **coderabbit (outdated) — capture `fctx` locally, cancel in `release()`.** Addressed for the capture; the `release()` cancel is present but unreachable in practice (finding 3).
- **coderabbit — add `t.Parallel()` to the new tests/subtests.** *Not* addressed. Neither `Test_Ctx_Context_Lazy_Initialization_Suite`, its three subtests, nor `Test_Ctx_Context_Cancel_On_Disconnect` call `t.Parallel()` at head. `tparallel` is enabled in [`.golangci.yml:62`](https://github.com/gofiber/fiber/blob/main/.golangci.yml#L62), so this will fail lint.
- **coderabbit — `Test_Ctx_Context_Cancel_On_Disconnect` cancels the injected context, not disconnect.** *Not* addressed; the test is byte-identical at head. The claim is correct — see finding 9.

Separately, coderabbit's "outside diff range" note about spawning a goroutine on every `Context()` call *was* addressed, by nil-ing `cancelFunc` — which is what causes finding 2.

## Additional findings

**1. `fasthttp.RequestCtx.Done()` is not a client-disconnect signal — the feature does not do what the PR claims.**
In fasthttp v1.73.0 (the version in `go.mod`):
```go
func (ctx *RequestCtx) Done() <-chan struct{} { return ctx.s.done }
// Note: Because creating a new channel for every request is just too expensive, so
// RequestCtx.s.done is only closed when the server is shutting down.
```
`s.done` is allocated only in `Server.Serve` and closed only by `Shutdown`. It is **never** closed on client disconnect. `ServeConn` and `fakeServer` leave it `nil`, so under `app.Test` the receive is on a nil channel and blocks forever. The watcher in [`ctx.go:156-162`](https://github.com/gofiber/fiber/blob/2075ffde0cccb571a0a9741fb853769d2156d18b/ctx.go#L156-L162) therefore only ever fires on server shutdown. Detecting a lost connection needs a different mechanism (e.g. `ConnState`/`Conn()` read probe), not `Done()`.

**2. Permanent goroutine leak — one per request that calls `Context()`.**
`Context()` hands `cancel` to the goroutine and then sets `c.cancelFunc = nil` ([`ctx.go:153-154`](https://github.com/gofiber/fiber/blob/2075ffde0cccb571a0a9741fb853769d2156d18b/ctx.go#L153-L154)). After that:
- `release()` ([`ctx.go:772`](https://github.com/gofiber/fiber/blob/2075ffde0cccb571a0a9741fb853769d2156d18b/ctx.go#L772)) and the next `Reset()` ([`ctx.go:736`](https://github.com/gofiber/fiber/blob/2075ffde0cccb571a0a9741fb853769d2156d18b/ctx.go#L736)) both see `nil` and cancel nothing.
- The goroutine's only exits are `fastHTTPDone` (server shutdown only, per finding 1) and `reqCtx.Done()` — but `cancel` is now owned *exclusively by that goroutine*, so nothing else can ever cancel `reqCtx`.

Net: the goroutine, its closure, and a `context.cancelCtx` survive until process shutdown, for every request that touches `c.Context()`. Fiber calls `c.Context()` internally (e.g. `SaveFileToStorage`), and the timeout middleware does `parent := ctx.Context()`, so this is not a rare path. This is unbounded growth under load, worse than the ticker version it replaced.

**3. `release()` cancellation semantics are inverted.**
Because `Context()` claims the cancel func, the request context is cancelled on release **only when `Context()` was never called** (or when `Conn()` is nil). Handlers that use the context get one that is never cancelled; handlers that ignore it get one that is. That is the opposite of the intent.

**4. The watcher can be wired to the wrong context.**
If anything calls `SetContext()` *before* the first `Context()` call, `Context()` reads the caller's context into `ctx`/`reqCtx`, but `cancel` still belongs to the orphaned context created in `Reset()`. The goroutine then waits on the caller's context and, when it fires, cancels a context no one holds. So even with a working disconnect signal, cancellation would not propagate. `middleware/timeout/timeout.go:34-36` happens to call `Context()` first, but the reverse order is a documented, supported pattern.

**5. `Reset()` is eager, not lazy — new per-request cost on the hot path.**
[`ctx.go:754-759`](https://github.com/gofiber/fiber/blob/2075ffde0cccb571a0a9741fb853769d2156d18b/ctx.go#L754-L759) unconditionally does `context.WithCancel(context.Background())` + `SetUserValue` + `isUserContextSet = true` on **every** request, contradicting the "lazy initialization" framing in the description and the test names. That is an extra allocation plus a `userValues` append per request, and it also forces `release()` to always take the `isUserContextSet` branch ([`ctx.go:776-780`](https://github.com/gofiber/fiber/blob/2075ffde0cccb571a0a9741fb853769d2156d18b/ctx.go#L776-L780)). No benchmark numbers are included. Given the cost and the change to request-context lifetime, this should be lazy and/or gated behind a `fiber.Config` flag.

**6. An unrelated existing test was deleted.**
`Test_Ctx_Cookie_DoesNotMutateArgument` is present on `main` at `ctx_test.go:10758` but does not exist at head — the new tests were written over it. This looks like a bad conflict resolution and must be restored.

**7. `SetContext` lost its doc comment.**
The diff removes `// SetContext sets a context implementation by user.` above [`ctx.go:168`](https://github.com/gofiber/fiber/blob/2075ffde0cccb571a0a9741fb853769d2156d18b/ctx.go#L168). `ctx_interface_gen.go:29` still carries that comment, so the ifacemaker-generated file now drifts from source, and revive's exported-comment rule will flag the method.

**8. Dead nil-guard in `Reset()`.**
`if fctx != nil` at [`ctx.go:749`](https://github.com/gofiber/fiber/blob/2075ffde0cccb571a0a9741fb853769d2156d18b/ctx.go#L749) is unreachable protection — `c.pathOriginal = c.app.toString(fctx.URI().PathOriginal())` two lines earlier at [`ctx.go:747`](https://github.com/gofiber/fiber/blob/2075ffde0cccb571a0a9741fb853769d2156d18b/ctx.go#L747) already dereferences `fctx`. Either guard the whole function or drop both checks.

**9. The tests do not cover the feature.**
- `Test_Ctx_Context_Cancel_On_Disconnect` uses a bare `&fasthttp.RequestCtx{}`, so `Conn()` is nil and the watcher goroutine is **never spawned**. It only proves that a context you cancel yourself is cancelled.
- `Lazy_Initialization_With_Conn` calls `fctx.Init(...)`, which sets `ctx.s = fakeServer` (whose `done` is nil), so it *does* spawn a goroutine that blocks forever — the test leaks it.
- No test drives the actual disconnect → cancel path, which is consistent with codecov's 72% patch coverage on `ctx.go`.
- Style: raw `t.Fatal`/`t.Error` instead of `require`, unlike the rest of `ctx_test.go`.

**10. Residual race on `c.cancelFunc` (lower confidence).**
`cancelFunc` is a plain struct field read/written by `Context()`, `Reset()` and `release()`. Handlers that call `c.Context()` from a spawned goroutine after `Abandon()` — a pattern the timeout and SSE middlewares enable — can race with `release()`. The reclaim latch covers the timeout path, but the field itself has no synchronization. Worth running the new tests under `-race` with a keep-alive server, not just `app.Test`.

---

ReneWerner87:
Thank you for staying with this over so many rounds. I owe you a straight answer on why, rather than another round of review.

The watcher selects on `c.fasthttp.Done()`. fasthttp documents that channel in `server.go`:

```go
// Note: Because creating a new channel for every request is just too expensive, so
// RequestCtx.s.done is only closed when the server is shutting down.
```

It fires on server shutdown, not on client disconnect, so the goroutine cannot observe the event this PR exists to detect. `net/http` gets it from `startBackgroundRead`, a concurrent one-byte read on the socket; fasthttp has no concurrent reader. valyala/fasthttp#965 is the upstream thread on this, and as of v1.73.0 the behaviour above is unchanged. We closed #4338 as not planned on 2026-07-28 for the same reason.

Two side effects I measured at head `2075ffde`:

- `Reset` now builds a `context.WithCancel` on every request, whether or not `Context()` is ever called. Against main: `11.4 ns/op, 0 B, 0 allocs` becomes `80 ns/op, 96 B, 2 allocs`.
- `Context()` sets `c.cancelFunc = nil` before spawning, so `release()` and the next `Reset()` both no-op and the goroutine holds the only remaining cancel. It exits on server shutdown, which pins a goroutine and a cancelCtx per request.

If you need a per-request cancellation signal, fasthttp is the place to move it. A framework-level watcher cannot reach the socket.

ReneWerner87:
Closing this, with the same answer as #4338.

The watcher selects on `c.fasthttp.Done()`, and fasthttp documents that channel as server shutdown only:

```go
// Note: Because creating a new channel for every request is just too expensive, so
// RequestCtx.s.done is only closed when the server is shutting down.
```

So the goroutine never observes a client disconnect, which is the feature this PR exists for. net/http manages it because `startBackgroundRead` (server.go:687) runs a concurrent one-byte read on the socket; fasthttp has no concurrent reader, so there is nothing to observe from. valyala/fasthttp#965 was closed as architecturally impossible for the same reason.

There is a cost without the benefit as well. On `2075ffde`, `Context()` sets `c.cancelFunc = nil` before spawning the goroutine, so the `Reset()` branch that would call it never fires. The goroutine then waits on `reqCtx.Done()`, which only closes when that same cancel runs, so one goroutine and one cancelCtx stay alive per request until the server shuts down.

This is not a review round that can be passed; the capability has to exist in fasthttp first. If it ever lands, this is the right shape for it. Thank you for staying with it through so many rounds, and sorry for the outcome after that much work.

PR #4406, opened 2026-06-06
feat: routguard middleware

muzzii255:
# Description

## RouteGuard Middleware

  Adds a middleware that validates requests against registered routes before they reach the middleware chain. Invalid routes are rejected
  immediately with a 404, skipping unnecessary middleware execution.

  - Trie-based route matching (~18-72ns lookups, zero allocations)
  - Supports static paths, parameters, and wildcards
  - Configurable error handler and skip function

Fixes # (issue)
 #4403 

## Changes introduced

  - New routeguard middleware that validates incoming requests against registered routes before they reach the middleware chain
  - Unmatched routes are rejected immediately with a 404 response, preventing unnecessary middleware execution
  - Trie-based route matching supporting static paths, parameters (:param), and wildcards (*, +)
  - Configurable ErrorHandler for custom 404 responses and Next function to skip middleware conditionally
  - Zero-allocation lookups (~18-72ns depending on path depth)
  

- [x] Benchmarks: Zero-allocation trie lookups ranging from 18-72ns depending on path depth. Early misses (invalid routes) are fastest at
  ~22ns since they bail out quickly.
  BenchmarkRootPath-16              65475970    18.30 ns/op    0 B/op    0 allocs/op
  BenchmarkStaticDeep-16            24663532    50.09 ns/op    0 B/op    0 allocs/op
  BenchmarkSingleParam-16           22362229    53.24 ns/op    0 B/op    0 allocs/op
  BenchmarkMultipleParams-16        16619850    72.28 ns/op    0 B/op    0 allocs/op
  BenchmarkWildcardDeep-16          42687034    28.26 ns/op    0 B/op    0 allocs/op

  - [x] Documentation Update: Added /docs/routeguard.md with usage examples, config options, and benchmarks.

  - [x] Changelog/What's New: New routeguard middleware that validates requests against registered routes before the middleware chain
  executes. Invalid routes are rejected immediately, avoiding unnecessary middleware overhead.

  - [ ] Migration Guide: Not applicable — new middleware, no breaking changes.

  - [ ] API Alignment with Express: Not applicable — this is a Fiber-specific optimization middleware with no Express equivalent.

  - [x] API Longevity: Follows existing Fiber middleware patterns (New(config), Config struct with Next and handler options). Minimal public
  API surface (New, Build, Config).
  
  - [x] Examples:
 ```
 app := fiber.New()
  app.Use(routeguard.New())

  app.Get("/api/users", handler)
  app.Get("/api/users/:id", handler)

[...]

PR #4411, opened 2026-06-07
🔥 feat: add SkipUnmatchedRoutes config option

### LINKED ISSUE #4403: 📝 [Proposal]: skip group/prefix middleware when no child route matches

Middleware registered on a Group or via Use with a path prefix fires for every request matching that prefix, even when no concrete route under it resolves. The 404 only triggers after the entire middleware chain runs.
```

api := app.Group("/api", AuthMiddleware(), Logger(), RateLimit())
api.Get("/test", handler)
```
Hitting `/api/nonexistent` triggers chain of middleware just to return a 404, unecessary db and cache call.

### Why it matters at scale

1. - Scanner/bot traffic hitting random paths runs full middleware chains. Auth middleware that touches DB or Redis gets hammered for free.
2. - Rate limiters get polluted by 404 spam — legitimate clients can be throttled.
3. - Observability is noisy: logs, traces, metrics fire for non-existent routes.
4. - Per-route middleware (api.Get("/test", handler, mw)) is the documented workaround but is unmaintainable in codebases with hundreds of routes.

### Existing related issues

#1959, #1704, #1179 — all hit variations of this. All have the same workaround suggestion ("add a 404 handler at the bottom") which doesn't actually prevent middleware execution, it just gives you a custom 404 page.

### Proposal

Building a route-existence matcher externally and short-circuiting before middleware. I built a radix-tree matcher over app.GetRoutes(true) as a proof of concept:

```
muzzii:fiber_matcher_test/ $ go test -bench=. -benchmem              

goos: linux
goarch: amd64
pkg: test
cpu: AMD Ryzen 7 9700X 8-Core Processor
BenchmarkTrieLookup-16    	17634477	        68.73 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrieMiss-16      	29907709	        39.87 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	test	2.519s
````

Opt-in config flag, preserves backward compatibility:
```
    SkipMiddlewareOnUnmatched: true,
})
```

When enabled, the router does a route-existence check against the registered route stack before executing the middleware chain. If no concrete route matches (method, path), jump straight to the NotFound handler. Group/prefix middleware never runs. 
Fiber's internal router already maintains this information — we're using app.GetRoutes(true) externally to mirror it. Doing the check inside the router itself would be cheaper and remove the need for users to maintain a parallel structure.

Express handles this implicitly through its router design. Middleware mounted via app.use('/api', mw) or on an express.Router() only executes when a matching route within that router responds to next() — but more importantly, Express's Router exposes route-matching primitives, and 404 handling is conventionally a final app.use((req, res) => res.status(404).send()) after routes are defined, which still suffers the same chain-execution problem.However, Express ecosystems address this with widely-used packages like express-list-endpoints and built-in router.stack introspection, letting users implement match-first guards trivially. Fiber's app.GetRoutes() is the equivalent primitive — this proposal asks the framework to use it internally for a config-gated early-exit, matching how mature Express apps handle the same problem in practice.If the SkipMiddlewareOnUnmatched flag is too opinionated, the lower-effort alternative — exposing app.HasRoute(method, path) or c.RouteExists() — directly mirrors what Express devs already get from router.stack introspection.

The proposal is fully compliant with HTTP RFC standards:
RFC 9110 §15.5.5 (404 Not Found): returning 404 for (method, path) combinations with no registered handler is the correct semantic. The feature only changes when the 404 is determined (before middleware vs. after), not the response itself.
RFC 9110 §15.5.6 (405 Method Not Allowed): the existing behavior of returning 405 when the path exists for a different method is preserved. The pre-match check distinguishes between "path unknown" (→ 404) and "path known, wrong method" (→ 405), keeping the framework's current method-not-allowed handling intact.
RFC 9110 §9.3.2 (HEAD): implementations MUST treat HEAD identically to GET in terms of route resolution. The pre-match check honors this — a HEAD request matches any registered GET route.
No changes to response headers, status codes, body semantics, caching directives, or any wire-level behavior. This is purely an internal short-circuit optimization.

### API Stability

The feature is designed for long-term stability:

Opt-in via config flag — SkipMiddlewareOnUnmatched: false by default means zero behavioral change for existing users. No deprecations, no breaking changes for anyone who doesn't enable it.
No new public types or interfaces — adds one boolean field to fiber.Config. Boolean config flags are the most stable API surface possible; even if internal implementation changes, the flag's meaning ("don't run middleware for unmatched routes") is unambiguous and unlikely to need revision.
Builds on existing primitives — app.GetRoutes() is already public and stable. The internal route stack is already maintained per-request. This is a router-internal optimization, not a new subsystem.
Forward-compatible — if Fiber v4 wants to make this the default, the flag can be deprecated cleanly with a single release cycle. If the alternative (app.HasRoute() / c.RouteExists()) is preferred instead, that method signature is trivial and stable.
No interaction with middleware contracts — handlers, c.Next(), error handling, and middleware ordering all remain identical when a route does match. The only observable change is that middleware doesn't execute when no route matches.

### Feature Examples

```go
Current behavior:

app := fiber.New()
api := app.Group("/api",
    AuthMiddleware(),      // DB lookup
    LoggerMiddleware(),    // emits log line
    RateLimiter(),         // increments counter
)
api.Get("/users/:id", getUserHandler)

// Request: GET /api/nonexistent
// → AuthMiddleware runs (DB hit)
// → LoggerMiddleware runs (log noise)
// → RateLimiter runs (counter incremented)
// → 404 returned



Proposed behavior with config flag:

    SkipMiddlewareOnUnmatched: true,
})

api := app.Group("/api",
    AuthMiddleware(),
    LoggerMiddleware(),
    RateLimiter(),
)
api.Get("/users/:id", getUserHandler)

// Request: GET /api/nonexistent
// → no route matches (method, path)
// → 404 returned immediately, NotFound handler invoked
// → AuthMiddleware, LoggerMiddleware, RateLimiter never run

// Request: GET /api/users/abc-123
// → route matches
// → AuthMiddleware → LoggerMiddleware → RateLimiter → getUserHandler (unchanged)

// Request: POST /api/users/abc-123
// → path exists for GET, not POST
// → 405 Method Not Allowed (unchanged behavior)
```

### Checklist:

muzzii255:
# Description

Adds a `SkipUnmatchedRoutes` config option that short-circuits requests to unregistered paths — returning 404 immediately without running through the middleware chain. Useful for cutting unnecessary processing from bots, scanners, and bad URLs.

Fixes #4403 

## Changes introduced

SkipUnmatchedRoutes on  app.Config, adds extra lookup costing 150-200ns on matched routes but saving same on unmatched routes  

```
goos: linux
goarch: amd64
cpu: AMD Ryzen 7 9700X 8-Core Processor             
Benchmark_SkipUnmatchedRoutes_Unmatched/without_skip-16         	 4276250	       280.2 ns/op	       8 B/op	       1 allocs/op
Benchmark_SkipUnmatchedRoutes_Unmatched/without_skip-16         	 4258561	       281.4 ns/op	       8 B/op	       1 allocs/op
Benchmark_SkipUnmatchedRoutes_Unmatched/without_skip-16         	 4216581	       284.7 ns/op	       8 B/op	       1 allocs/op
Benchmark_SkipUnmatchedRoutes_Unmatched/without_skip-16         	 4141472	       281.5 ns/op	       8 B/op	       1 allocs/op
Benchmark_SkipUnmatchedRoutes_Unmatched/without_skip-16         	 4173716	       283.6 ns/op	       8 B/op	       1 allocs/op
Benchmark_SkipUnmatchedRoutes_Unmatched/with_skip-16            	20101112	        59.06 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Unmatched/with_skip-16            	21355640	        56.42 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Unmatched/with_skip-16            	19912690	        59.81 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Unmatched/with_skip-16            	21089865	        56.41 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Unmatched/with_skip-16            	21032264	        56.59 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Matched/without_skip-16           	 4840177	       246.3 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Matched/without_skip-16           	 4927646	       243.0 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Matched/without_skip-16           	 4730292	       255.3 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Matched/without_skip-16           	 4752334	       251.9 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Matched/without_skip-16           	 4898761	       242.6 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Matched/with_skip-16              	 2848100	       422.5 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Matched/with_skip-16              	 2911087	       413.6 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Matched/with_skip-16              	 2911100	       406.0 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Matched/with_skip-16              	 2843078	       416.9 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Matched/with_skip-16              	 2911432	       415.4 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/gofiber/fiber/v3	23.896s
```


ReneWerner87:
@muzzii255 pls check linter errors

ReneWerner87:
@muzzii255 pls add something in the markdown docs for this feature

muzzii255:
@ReneWerner87 done

ReneWerner87 (approved):

ReneWerner87:
@gaby pls check

router.go
```diff
@@ -424,6 +424,28 @@ func (app *App) nextCustom(c CustomCtx) (bool, error) {
 	return false, ErrNotFound
 }
 
+// routeExists checks if a non-middleware route matches the given method and path.
+func (app *App) routeExists(methodInt, treeHash int, detectionPath, path string) bool {
```
gaby: We don't need this function,  it is going through the whole tree for every request.

gaby: The way this is implemented, when `SkipUnmatchedRoutes` is enabled we are calling `route.match` twice. Here and in `next()` for every single request.

gaby (changes requested):
I think we need to do the changes in `next()`, else we are checking the whole router twice for every requests. That defeats the purpose of this feature.

gaby:
> @gaby pls check

When `SkipUnmatchedRoutes` the router matching logic runs twice for every request. This is not ideal.

router.go
muzzii255: yes, but i does not trigger the middlewares, its an extra check to save db and cache queries, on unmatched routes it saves the 1/alloc and faster, but adds same amount if on matched routes, its an opt in feature tho

muzzii255: The tree walk before middleware is the whole point of `SkipUnmatchedRoutes` — without it, any middleware registered via `Group()` fires on requests that never match a route, which is exactly the behavior this flag is meant to prevent. Moving the logic into `route.match` doesn't change that: by the time `next()` gets to the per-route comparison, the group middleware has already run.

i tried caching this extra tree walk, but the params paths still creates the overhead and saves like 5-10ns on total benchmark not worth it.

muzzii255:
@gaby @ReneWerner87 i tried multiple approach like defering the middlewares when `SkipUnmatchedRoutes` to prevent extra lookup or using the routeExists to match the route and send the route to next() which will run the middlewares, none of them worked really well. the only thing left is to rewrite the tree which i am not supporting as it might create more bugs in the future. since its an optional feature, user will know the cost of using it.

muzzii255:
any approach will add an extra lookup on the tree except rewriting the tree, right now i am using a middleware from the PR #4406 on my codebase, it will be easier for me to just use the fiber's internal tree.

gaby:
@muzzii255 I think in next() we can check. IsMatched() and if false call c.SendString(not found)

muzzii255:
> @muzzii255 I think in next() we can check. IsMatched() and if false call c.SendString(not found)

let me test it

muzzii255:
> @muzzii255 I think in next() we can check. IsMatched() and if false call c.SendString(not found)

in next() function this loop executes the middlewares before isMatched() is called, i can try defering the middleware execute by appending the handler to a slice, iterate and execute the middleware after if c.isMatched check 

```
	// Loop over the route stack starting from previous index
	for indexRoute < lenr {
		// Increment route index
		indexRoute++

		// Get *Route
		route := tree[indexRoute]

		if route.mount {
			continue
		}

		// Check if it matches the request path
		if !route.match(detectionPath, path, &c.values) {
			continue
		}

		if c.shouldSkipNonUseRoutes && !route.use {
			continue
		}

		// Pass route reference and param values
		c.route = route
		// Non use handler matched
		if !route.use {
			c.isMatched = true
		}
		// Execute first handler of route
		if len(route.Handlers) > 0 {
			c.indexHandler = 0
			c.indexRoute = indexRoute
			return true, route.Handlers[0](c)
		}

		return true, nil // Stop scanning the stack
	}
	
	// If c.Next() does not match, return 404
	// If no match, scan stack again if other methods match the request
	// Moved from app.handler because middleware may break the route chain
	if c.isMatched {
		return false, ErrNotFound
	}

```

muzzii255:
> @muzzii255 I think in next() we can check. IsMatched() and if false call c.SendString(not found)

@gaby  Do you mean inside next() do the endpoint scan first (before the main loop runs any middleware), and return ErrNotFound if nothing matches? Because checking IsMatched after the loop is too late — middleware has already executed by then, which defeats the feature.

muzzii255:
@gaby  Dropped the second scan in SkipUnmatchedRoutes. routeExists -> bool is now firstEndpointIndex -> int  — it returns the stack index of the first matching endpoint instead of just whether one exists. next/nextCustom use that index to skip re-matching the endpoints the lookahead already ruled out (if !route.use && indexRoute < firstMatchIndex), so middleware are never skipped and the chain is unchanged. the matched path no longer matches endpoints twice and unmatched/bot requests still 404 before the middleware chain.
benchmark on matched endpoint dropped from  415 to 280.


```
goos: linux
goarch: amd64
cpu: AMD Ryzen 7 9700X 8-Core Processor
Benchmark_SkipUnmatchedRoutes_Unmatched/without_skip-16         	 4527450	       267.2 ns/op	       8 B/op	       1 allocs/op
Benchmark_SkipUnmatchedRoutes_Unmatched/without_skip-16         	 4390478	       272.7 ns/op	       8 B/op	       1 allocs/op
Benchmark_SkipUnmatchedRoutes_Unmatched/without_skip-16         	 4406754	       272.7 ns/op	       8 B/op	       1 allocs/op
Benchmark_SkipUnmatchedRoutes_Unmatched/without_skip-16         	 4442115	       269.5 ns/op	       8 B/op	       1 allocs/op
Benchmark_SkipUnmatchedRoutes_Unmatched/without_skip-16         	 4516107	       268.6 ns/op	       8 B/op	       1 allocs/op
Benchmark_SkipUnmatchedRoutes_Unmatched/with_skip-16            	21924752	        53.24 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Unmatched/with_skip-16            	22678478	        53.30 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Unmatched/with_skip-16            	21954602	        54.07 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Unmatched/with_skip-16            	22341717	        52.66 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Unmatched/with_skip-16            	22903267	        52.27 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Matched/without_skip-16           	 4972453	       233.9 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Matched/without_skip-16           	 5155741	       240.5 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Matched/without_skip-16           	 5021347	       239.5 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Matched/without_skip-16           	 5073507	       239.6 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Matched/without_skip-16           	 4864011	       246.4 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Matched/with_skip-16              	 4219206	       283.2 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Matched/with_skip-16              	 4405276	       277.8 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Matched/with_skip-16              	 4316749	       278.4 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Matched/with_skip-16              	 4253124	       282.0 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Matched/with_skip-16              	 4302944	       279.2 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/gofiber/fiber/v3	24.017s
```

gaby (approved):
LGTM

ReneWerner87:
Reviewed the `SkipUnmatchedRoutes` implementation. The motivation from #4403 is real and the unmatched-path win is solid (56ns/0 alloc). A few things I'd want addressed before merge:

**1. Wrong-method now returns 404 instead of 405 (regression)**

`firstEndpointIndex` only scans the requested method's tree (`router.go:455`), and the gate does `SendStatus(StatusNotFound)` on `idx == -1`, so it never reaches the cross-method scan in `next()`/`nextCustom()` that returns `ErrMethodNotAllowed` + `Allow` header. Repro with the flag on, only `GET /users` registered:

```
SkipUnmatchedRoutes=false -> POST /users => 405, Allow: GET, HEAD
SkipUnmatchedRoutes=true  -> POST /users => 404, Allow: (empty)
```

Same for `OPTIONS` (no auto-OPTIONS in fiber). HEAD is fine (auto-HEAD route). No test covers the wrong-method case.

**2. Custom `ErrorHandler` is bypassed**

Emitting the 404 via `ctx.SendStatus(StatusNotFound)` skips `config.ErrorHandler`. Without the flag, the 404 flows through it (`router.go:510`), so an app with a custom 404 (branded JSON/HTML) silently loses it when the flag is on. Skipping a final catch-all `app.Use` 404 is the intended behavior; bypassing the configured `ErrorHandler` probably isn't.

Both 1 and 2 are fixable together: have the lookahead also detect the cross-method case, then route the result through the error handler instead of `SendStatus`, e.g. `_ = app.ErrorHandler(ctx, ErrNotFound)` (or `ErrMethodNotAllowed` + Allow header) so existence-gating and error rendering stay consistent, while middleware is still skipped.

**3. `Path()` setter doesn't reset `firstMatchIndex` (latent staleness)**

`Reset()` and `RestartRouting()` reset `firstMatchIndex`, but the `Path()` setter (`ctx.go:311`) mutates `c.path`/`c.detectionPath` without resetting it. Rewrite middleware does `c.Path(...)` then `c.Next()` (not `RestartRouting`), so a stale index from the original path could skip a now-matching lower-index endpoint. Narrow trigger, but it's a correctness footgun: reset `firstMatchIndex = -1` in the `Path()` setter too.

**4. Matched-route cost (+~70%) is mostly inherent, but avoidable for the common case**

Deciding "does any endpoint match" before the middleware chain means matching the endpoint once up front, which roughly duplicates `next()`'s scan on a linear matcher, so the matched-path regression is expected rather than a bug. Two cheap reductions: reuse the ctx's `values` buffer in `firstEndpointIndex` instead of a throwaway `[maxParams]string`, and skip param extraction for the existence check. For the bigger win without a full radix rewrite: most endpoints are static, so a `map[methodInt]map[string]struct{}` of static paths built in `buildTree` gives an O(1) 404 gate for the common case and falls back to the linear lookahead only when parametric/wildcard routes could match. That removes the regression for static matched routes.

**Tests to add:** wrong-method-on-existing-path (expect 405 + Allow), `OPTIONS`, custom `ErrorHandler` still fires, and a rewrite-middleware path-change case.

muzzii255:
> Wrong-method now returns 404 instead of 405 (regression)

@ReneWerner87 since fiber's main tree is map[method]routes, i had to iterate throught all the methods to find true 405 error. it worked tho but no more 56ns lookup :(

```
goos: linux
goarch: amd64
cpu: AMD Ryzen 7 9700X 8-Core Processor
Benchmark_SkipUnmatchedRoutes_Unmatched/without_skip-16         	 5355361	       224.9 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Unmatched/without_skip-16         	 5381514	       223.2 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Unmatched/without_skip-16         	 5315523	       222.5 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Unmatched/with_skip-16            	 5758648	       207.0 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Unmatched/with_skip-16            	 5852835	       205.8 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Unmatched/with_skip-16            	 5759576	       207.7 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Matched/without_skip-16           	 4829791	       249.6 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Matched/without_skip-16           	 4819560	       250.5 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Matched/without_skip-16           	 4819665	       243.1 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Matched/with_skip-16              	 4126128	       287.4 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Matched/with_skip-16              	 4125486	       294.5 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Matched/with_skip-16              	 3947894	       293.9 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/gofiber/fiber/v3	14.342s 
```

gaby:
> > Wrong-method now returns 404 instead of 405 (regression)
> 
> @ReneWerner87 since fiber's main tree is map[method]routes, i had to iterate throught all the methods to find true 405 error. it worked tho but no more 56ns lookup :(
> 
> ```
> goos: linux
> goarch: amd64
> pkg: github.com/gofiber/fiber/v3
> cpu: AMD Ryzen 7 9700X 8-Core Processor
> Benchmark_SkipUnmatchedRoutes_Unmatched/without_skip-16         	 5355361	       224.9 ns/op	       0 B/op	       0 allocs/op
> Benchmark_SkipUnmatchedRoutes_Unmatched/without_skip-16         	 5381514	       223.2 ns/op	       0 B/op	       0 allocs/op
> Benchmark_SkipUnmatchedRoutes_Unmatched/without_skip-16         	 5315523	       222.5 ns/op	       0 B/op	       0 allocs/op
> Benchmark_SkipUnmatchedRoutes_Unmatched/with_skip-16            	 5758648	       207.0 ns/op	       0 B/op	       0 allocs/op
> Benchmark_SkipUnmatchedRoutes_Unmatched/with_skip-16            	 5852835	       205.8 ns/op	       0 B/op	       0 allocs/op
> Benchmark_SkipUnmatchedRoutes_Unmatched/with_skip-16            	 5759576	       207.7 ns/op	       0 B/op	       0 allocs/op
> Benchmark_SkipUnmatchedRoutes_Matched/without_skip-16           	 4829791	       249.6 ns/op	       0 B/op	       0 allocs/op
> Benchmark_SkipUnmatchedRoutes_Matched/without_skip-16           	 4819560	       250.5 ns/op	       0 B/op	       0 allocs/op
> Benchmark_SkipUnmatchedRoutes_Matched/without_skip-16           	 4819665	       243.1 ns/op	       0 B/op	       0 allocs/op
> Benchmark_SkipUnmatchedRoutes_Matched/with_skip-16              	 4126128	       287.4 ns/op	       0 B/op	       0 allocs/op
> Benchmark_SkipUnmatchedRoutes_Matched/with_skip-16              	 4125486	       294.5 ns/op	       0 B/op	       0 allocs/op
> Benchmark_SkipUnmatchedRoutes_Matched/with_skip-16              	 3947894	       293.9 ns/op	       0 B/op	       0 allocs/op
> PASS
> ok  	github.com/gofiber/fiber/v3	14.342s 
> ```

The performance is awful, which is exactly what I expected.

muzzii255:
> The performance is awful, which is exactly what I expected.

yes there are extra lookups for pure 405 error

this is bad, do not merge it let me find another way to fix it

muzzii255:
creating map[int]int on startup that maps each tree bucket to a bitmask of methods with routes.  At request time, we check this index first, if no methods have routes in the bucket, it's an instant 404 without scanning any trees. For 405, we only scan methods that have the bit set, skipping empty trees entirely. 
build map on startup also add 7 allocs 

```
goos: linux
goarch: amd64
cpu: AMD Ryzen 7 9700X 8-Core Processor
Benchmark_SkipUnmatchedRoutes_Unmatched/without_skip-16         	 5305108	       228.3 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Unmatched/without_skip-16         	 5461108	       221.1 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Unmatched/without_skip-16         	 5467125	       218.5 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Unmatched/without_skip-16         	 5182489	       228.1 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Unmatched/with_skip-16            	 9188634	       130.4 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Unmatched/with_skip-16            	 9170767	       129.7 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Unmatched/with_skip-16            	 9259285	       128.9 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Unmatched/with_skip-16            	 9339883	       129.5 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Matched/without_skip-16           	 5061919	       238.3 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Matched/without_skip-16           	 5001852	       236.9 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Matched/without_skip-16           	 5082570	       232.4 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Matched/without_skip-16           	 5035828	       239.6 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Matched/with_skip-16              	 4246458	       281.1 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Matched/with_skip-16              	 4322932	       277.5 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Matched/with_skip-16              	 4242019	       283.7 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_Matched/with_skip-16              	 4286206	       274.7 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_405_Middleware/without_skip-16    	 1580234	       762.0 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_405_Middleware/without_skip-16    	 1553347	       769.3 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_405_Middleware/without_skip-16    	 1517376	       792.2 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_405_Middleware/without_skip-16    	 1563447	       767.8 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_405_Middleware/with_skip-16       	 1594249	       755.3 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_405_Middleware/with_skip-16       	 1581048	       763.1 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_405_Middleware/with_skip-16       	 1603018	       746.2 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_405_Middleware/with_skip-16       	 1598311	       751.7 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_405_NMiddleware/without_skip-16   	 1598989	       750.2 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_405_NMiddleware/without_skip-16   	 1598631	       747.6 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_405_NMiddleware/without_skip-16   	 1606292	       747.4 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_405_NMiddleware/without_skip-16   	 1605138	       744.5 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_405_NMiddleware/with_skip-16      	 1594016	       753.6 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_405_NMiddleware/with_skip-16      	 1587657	       755.4 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_405_NMiddleware/with_skip-16      	 1576268	       761.1 ns/op	       0 B/op	       0 allocs/op
Benchmark_SkipUnmatchedRoutes_405_NMiddleware/with_skip-16      	 1582045	       759.5 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/gofiber/fiber/v3	38.370s
```

```
goos: linux
goarch: amd64
cpu: AMD Ryzen 7 9700X 8-Core Processor
Benchmark_App_RebuildTree
Benchmark_App_RebuildTree/without_skip
Benchmark_App_RebuildTree/without_skip-16         	  117416	     10135 ns/op	   16296 B/op	      50 allocs/op
Benchmark_App_RebuildTree/without_skip-16         	  118948	     10106 ns/op	   16296 B/op	      50 allocs/op
Benchmark_App_RebuildTree/with_skip
Benchmark_App_RebuildTree/with_skip-16            	  100983	     11991 ns/op	   17424 B/op	      57 allocs/op
Benchmark_App_RebuildTree/with_skip-16            	   98691	     12022 ns/op	   17424 B/op	      57 allocs/op
PASS
ok  	github.com/gofiber/fiber/v3	7.179s
```

gaby:
@muzzii255 @ReneWerner87 I have a more optimized version here https://github.com/gofiber/fiber/pull/4486 it basically pre-calculates everything to avoid the costly loop during each request. It also optimizes the use-case where no middlewares are registered.

muzzii255:
> @muzzii255 @ReneWerner87 I have a more optimized version here #4486 it basically pre-calculates everything to avoid the costly loop during each request. It also optimizes the use-case where no middlewares are registered.

the bottleneck i am facing is that one deep endpoints, the performance gain is little. on average it adds 20-30% on matched endpoints vice versa. let me benchmark both versions.

muzzii255:
[benchstats.txt](https://github.com/user-attachments/files/29512154/benchstats.txt)
yeah its same

gaby:
Replaced by https://github.com/gofiber/fiber/pull/4486 which has further router improvements found by `Claude Fable`.

PR #4415, opened 2026-06-08
chore: add MIMETextEventStream constant

Amirhf1:
Every commonly-used MIME type in Fiber is exposed as a public MIME* constant in constants.go, except text/event-stream. As a result:
 - Ctx.AcceptsEventStream() hardcoded the raw string "text/event-stream".
- The SSE middleware defined its own private mimeTextEventStream constant for the exact same value.

This PR adds the public fiber.MIMETextEventStream constant and makes both call sites use it, giving a single source of truth and consistency with the rest of the MIME* family.

### DIFF
--- a/constants.go
+++ b/constants.go
@@ -21,6 +21,7 @@ const (
 	MIMETextPlain             = "text/plain"
 	MIMETextJavaScript        = "text/javascript"
 	MIMETextCSS               = "text/css"
+	MIMETextEventStream       = "text/event-stream"
 	MIMEApplicationXML        = "application/xml"
 	MIMEApplicationJSON       = "application/json"
 	MIMEApplicationJavaScript = "application/javascript"
--- a/middleware/sse/constants.go
+++ b/middleware/sse/constants.go
@@ -1,5 +1,0 @@
-package sse
-
-import "github.com/gofiber/fiber/v3"
-
-const mimeTextEventStream = fiber.MIMETextEventStream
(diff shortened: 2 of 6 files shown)


middleware/sse/constants.go
```diff
@@ -1,3 +1,5 @@
 package sse
 
-const mimeTextEventStream = "text/event-stream"
+import "github.com/gofiber/fiber/v3"
+
+const mimeTextEventStream = fiber.MIMETextEventStream
```
gaby: We don't need to redeclare it. Just use the one from `fiber.MIMETextEventStream` in the code.

gaby (changes requested):

ReneWerner87:
@Amirhf1 pls check the review comment -> then its mergable

constants.go
```diff
@@ -21,6 +21,7 @@ const (
 	MIMETextPlain             = "text/plain"
 	MIMETextJavaScript        = "text/javascript"
 	MIMETextCSS               = "text/css"
+	MIMETextEventStream       = "text/event-stream"
```
ReneWerner87, line 24: @Amirhf1 pls also update the doc file -> https://github.com/gofiber/fiber/blob/main/docs/api/constants.md

ReneWerner87:
@Amirhf1 one more last thing

ReneWerner87 (approved):

PR #4418, opened 2026-06-09
forward store errors in session initialize to ErrorHandler instead of panicking

### LINKED ISSUE #4348: 🐛 [Bug]: Session middleware panics on store errors — server crash on Redis failure

harsha-cpp:
fixes #4348.

**problem:** `initialize()` calls `panic(err)` when the backing store returns an error during session lookup. a redis timeout, network blip, or misconfigured store crashes the entire server process rather than being handled gracefully.

**fix:** `initialize()` now returns the error. the handler in `NewWithStore` catches it, releases the middleware back to the pool, and routes the error through `cfg.ErrorHandler` (or `DefaultErrorHandler` if none is configured) — the same path that `saveSession()` already uses for save-time errors.

regression test: `Test_Session_Middleware_StoreError` — sends a request with a session cookie to an app backed by a storage that always returns an error. verifies the route handler is never reached and the `ErrorHandler` receives the store error.

### DIFF
--- a/middleware/session/middleware.go
+++ b/middleware/session/middleware.go
@@ -95,7 +95,15 @@ func NewWithStore(config ...Config) (fiber.Handler, *Store) {
 
 		// Acquire session middleware
 		m := acquireMiddleware()
-		m.initialize(c, &cfg)
+		if err := m.initialize(c, &cfg); err != nil {
+			releaseMiddleware(m)
+			if cfg.ErrorHandler != nil {
+				cfg.ErrorHandler(c, err)
+			} else {
+				DefaultErrorHandler(c, err)
+			}
+			return nil
+		}
 
 		stackErr := c.Next()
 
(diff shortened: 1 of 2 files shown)


gaby:
@harsha-cpp We already have a PR for fixing this by the Fiber team. Closing as duplicated.

See: https://github.com/gofiber/fiber/pull/4376

harsha-cpp:
@gaby okay mb

PR #4422, opened 2026-06-11
🐛 fix(limiter): correct fixed-window hit credit on skipped requests

gaby:
When SkipSuccessfulRequests/SkipFailedRequests credits a hit back after
the handler runs, the fixed-window strategy decremented currHits
unconditionally and bumped a stale local counter for the
X-RateLimit-Remaining header. If the window rolled over (or another
request reset the entry) between the increment and the credit, currHits
could underflow below zero, inflating the next window's allowance, and
the header reported a value disconnected from stored state.

Mirror the sliding-window semantics: only credit the hit when the entry
still belongs to the same window and currHits is positive, and recompute
remaining from the fresh entry instead of incrementing the stale local.

### DIFF
--- a/middleware/csrf/session_manager.go
+++ b/middleware/csrf/session_manager.go
@@ -47,6 +47,7 @@ func (m *sessionManager) getRaw(c fiber.Ctx, key string, raw []byte) []byte {
 			// Handle error
 			return nil
 		}
+		defer storeSess.Release()
 		token, ok = storeSess.Get(sessionKey).(Token)
 	}
 
@@ -73,6 +74,7 @@ func (m *sessionManager) setRaw(c fiber.Ctx, key string, raw []byte, exp time.Du
 			// Handle error
 			return
 		}
+		defer storeSess.Release()
 		storeSess.Set(sessionKey, Token{Key: key, Raw: raw, Expiration: time.Now().Add(exp)})
 		if err := storeSess.Save(); err != nil {
 			log.Warn("csrf: failed to save session: ", err)
@@ -92,6 +94,7 @@ func (m *sessionManager) delRaw(c fiber.Ctx) {
 			// Handle error
 			return
 		}
+		defer storeSess.Release()
 		storeSess.Delete(sessionKey)
 		if err := storeSess.Save(); err != nil {
 			log.Warn("csrf: failed to save session: ", err)
--- a/middleware/csrf/session_manager.go
+++ b/middleware/csrf/session_manager.go
@@ -47,6 +47,8 @@ func (m *sessionManager) getRaw(c fiber.Ctx, key string, raw []byte) []byte {
 			// Handle error
 			return nil
 		}
+		// token.Raw stays valid after Release only because session Reset() does
+		// not scrub values; copy it before Release if that ever changes.
 		defer storeSess.Release()
 		token, ok = storeSess.Get(sessionKey).(Token)
 	}
--- a/middleware/etag/etag.go
+++ b/middleware/etag/etag.go
@@ -115,6 +115,9 @@ func isNoneMatch(header, etag []byte) bool {
 	}
 
 	for len(header) > 0 {
+		// RFC 9110 allows a comma inside an opaque-tag, so this split can
+		// mis-parse such tags. It fails open and Fiber's ETags never contain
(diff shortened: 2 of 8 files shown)


middleware/limiter/limiter_fixed.go
```diff
@@ -120,8 +121,12 @@ func (FixedWindow) New(cfg *Config) fiber.Handler {
 				return getErr
 			}
 			e = entry
-			e.currHits--
-			remaining++
+			// Only credit the hit back if it still belongs to the same window;
+			// after a rollover the original hit no longer counts toward e.currHits.
+			if e.exp == windowExpiresAt && e.currHits > 0 {
```
ReneWerner87, line 126: The headline fix ships without a regression test. The scenario is deterministic to test, analogous to `Test_Limiter_Sliding_Window_SkipFailedRequests_DecrementsPreviousWindow`: a handler with `SkipSuccessfulRequests` that sleeps past `Expiration` while another request rolls the window, then assert the next window allows exactly `Max` requests and check `X-RateLimit-Remaining`. On main this underflows `currHits` to -1 and grants extra requests, so the test would pin the fix.

middleware/limiter/limiter_fixed.go
```diff
@@ -120,8 +121,12 @@ func (FixedWindow) New(cfg *Config) fiber.Handler {
...
+				e.currHits--
+			}
+			remaining = maxRequests - e.currHits
```
ReneWerner87, line 129: `X-RateLimit-Remaining` can now be negative: blocked requests persist their increment before being rejected, so `currHits` can exceed `maxRequests` while this handler runs; after the credit this computes e.g. -1, which is written verbatim into the header. The old code never emitted a negative value on this path. The sliding window has the same property, so this is consistent, but clamping to 0 (ideally in both strategies) would be friendlier to header-parsing clients.

middleware/etag/etag.go
```diff
@@ -85,37 +86,60 @@ func New(config ...Config) fiber.Handler {
...
+	header = utils.TrimSpace(header)
+	if len(header) == 0 {
+		return false
+	}
+	if len(header) == 1 && header[0] == '*' {
+		return true
+	}
+
+	for len(header) > 0 {
+		entry, rest, _ := bytes.Cut(header, []byte(","))
```
ReneWerner87, line 121: Nit: RFC 9110 allows a comma inside an opaque-tag (etagc includes %x2C), so this split mis-parses such tags. The failure mode is fail-open (200 instead of 304), and Fiber's own etags never contain commas, so this is acceptable, but worth a short comment here.

middleware/csrf/session_manager.go
```diff
@@ -47,6 +47,7 @@ func (m *sessionManager) getRaw(c fiber.Ctx, key string, raw []byte) []byte {
 			// Handle error
 			return nil
 		}
+		defer storeSess.Release()
```
ReneWerner87, line 52: Worth a short comment here: the returned `token.Raw` stays valid after `Release` only because `data.Reset()` clears the map without zeroing values, and the backing array comes from the decode. If `releaseSession` ever starts scrubbing values, `getRaw` would silently return a dangling slice. A one-liner makes that coupling visible at the call site.

ReneWerner87 (commented):
Reviewed with focus on correctness, concurrency, and RFC semantics. All four fixes are real and correctly implemented: the fixed-window credit guard closes a genuine `currHits` underflow (rate-limit bypass), the csrf/session `Release` calls fix session pool leaks without double-release risk, and the new If-None-Match handling matches RFC 9110 §8.8.3.2. Affected suites pass, including with `-race`.

A few remarks inline. Two notes that do not fit a specific line:

- Behavior changes worth a changelog entry: `If-None-Match: *` now returns 304 (was 200), a strong client tag now matches a weak server tag, and the ETag header is now also sent on 304 responses. All RFC-correct, but observable.
- Pre-existing, not for this PR: the fixed-window skip path still emits the pre-handler `X-RateLimit-Reset` value, while the sliding window recomputes it post-handler.

gaby:
@ReneWerner87 the linter is failing on this one

gaby:
> @ReneWerner87 the linter is failing on this one

Should be good now.

PR #4428, opened 2026-06-12
Add Go fuzz tests for OSS-Fuzz integration

canolgun:
Adds Go native fuzz targets for OSS-Fuzz integration.

**Why Fiber**
Fiber is a popular Express-inspired Go web framework with **40K+ stars** and **3 GitHub Security Advisories**. It processes every HTTP request through route matching and body parsing — both pre-auth boundaries.

**Fuzz targets (2):**
- `FuzzFiberRouteMatch` — Route matching with arbitrary paths and HTTP methods
- `FuzzFiberBodyParser` — JSON body binding with arbitrary byte input

### DIFF
--- a/fuzz_test.go
+++ b/fuzz_test.go
@@ -1,0 +1,76 @@
+//go:build go1.18
+// +build go1.18
+
+// Copyright (c) 2019-present Fenny and Contributors
+// SPDX-License-Identifier: MIT
+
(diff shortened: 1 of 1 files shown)


gaby:
This is not adding significant value. We would prefer full fuzz tests covering all cases.

PR #4432, opened 2026-06-14
🐛 bug: preserve legacy custom constraint arguments

gaby:
### Motivation
- Fix a route-matching regression where custom constraints that also implement `ConstraintAnalyzer` could have their configured route arguments dropped at request time, enabling an opt‑in constraint bypass.
- Restore backward-compatible behavior so legacy `CustomConstraint.Execute(param, args ...string)` continues to receive the parsed `[]string` arguments while still running analyzer validation/side effects.

### Description
- Change `customConstraintWrapper.Analyze` to invoke the underlying `ConstraintAnalyzer` for validation/side effects but ignore its typed return values and always return `[]any{parsedArgs}` so legacy argument shape is preserved.
- Preserve request-time behavior by ensuring `newConstraint` and `Constraint.matchConstraint` continue to use the wrapper-provided legacy `[]string` payload when calling `Execute`.
- Add/adjust tests in `constraint_test.go` to assert the wrapper returns the legacy `[]string` shape and to cover the regression with `Test_CustomConstraintWrapper_ExecuteKeepsLegacyArgsWithAnalyzer`.
- No public API changes; behavior is restored to be backward-compatible with legacy custom constraints.

### Testing
- Ran `go test ./...` and `make test`, which completed successfully (all tests passed; summary: `DONE 3583 tests, 1 skipped`).
- Ran `make generate`, `make betteralign`, `make format`, and `make lint`, which succeeded (`make lint` reported `0 issues`).
- Ran `make audit`, which exercises `govulncheck` and failed in this environment due to known standard-library vulnerabilities reported against the toolchain `go1.25.1` (25 stdlib issues); this is an environment/toolchain issue and not caused by these code changes.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a2e31b122b883339a80aa3c4c918c5a)

### DIFF
--- a/constraint.go
+++ b/constraint.go
@@ -101,7 +101,9 @@ type customConstraintWrapper struct {
 func (w *customConstraintWrapper) Analyze(args []string) ([]any, error) {
 	parsedArgs := parseConstraintArgs(args)
 	if analyzer, ok := w.CustomConstraint.(ConstraintAnalyzer); ok {
-		return analyzer.Analyze(parsedArgs)
+		if _, err := analyzer.Analyze(parsedArgs); err != nil {
+			return nil, err
+		}
 	}
 	return []any{parsedArgs}, nil
 }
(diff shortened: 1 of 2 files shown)


gaby:
Bug was introduced in #4397

ReneWerner87 (approved):

PR #4436, opened 2026-06-18
🔥 feat: Add HTTP QUERY method support (RFC 10008)

nekoworks-magic:
# Description

Add first-class support for the HTTP QUERY method ([RFC 10008](https://www.rfc-editor.org/rfc/rfc10008.html)). QUERY is safe and idempotent like GET, but allows request bodies for complex queries.

## Changes introduced

- `MethodQuery` constant, `methodQuery` iota, `DefaultMethods` entry
- `Query()` on `Router`, `Register` interfaces and all implementations (App, Group, Registering, domainRouter, domainRegistering)
- `Request.Query()` client shorthand
- `IsMethodSafe` returns true for QUERY
- CSRF safe-method branch includes QUERY
- CORS default `AllowMethods` includes QUERY
- Logger color: shares Cyan with GET

## Type of change

## Checklist

### DIFF
--- a/middleware/cache/cache.go
+++ b/middleware/cache/cache.go
@@ -1290,6 +1290,12 @@ func defaultKeyGenerator(c fiber.Ctx, cfg *Config) string {
 		buf = append(buf, canonicalCookieSubset(c, cfg.KeyCookies)...)
 	}
 
+	if c.Method() == fiber.MethodQuery {
+		bodyDigest := sha256.Sum256(c.Request().Body())
+		buf = append(buf, '|', 'b', '=')
+		buf = hex.AppendEncode(buf, bodyDigest[:])
+	}
+
 	result := string(buf)
 
 	// Reset buffer and return to pool, but discard if it grew too large
--- a/middleware/cache/cache.go
+++ b/middleware/cache/cache.go
@@ -1291,9 +1291,12 @@ func defaultKeyGenerator(c fiber.Ctx, cfg *Config) string {
 	}
 
 	if c.Method() == fiber.MethodQuery {
-		bodyDigest := sha256.Sum256(c.Request().Body())
+		// RFC 10008 only requires the cache key to incorporate the request content,
+		// not that it be hashed. boundKeySegment keeps small bodies verbatim (no
+		// hashing on the hot path) and hashes only bodies larger than
+		// maxKeyDimensionSegmentLength, which still bounds the key length.
 		buf = append(buf, '|', 'b', '=')
-		buf = hex.AppendEncode(buf, bodyDigest[:])
+		buf = append(buf, boundKeySegment(utils.UnsafeString(c.Request().Body()))...)
 	}
 
 	result := string(buf)
--- a/middleware/cors/cors_test.go
+++ b/middleware/cors/cors_test.go
@@ -612,6 +612,7 @@ func Test_CORS_Headers_BasedOnRequestType(t *testing.T) {
 		fiber.MethodDelete,
 		fiber.MethodPatch,
 		fiber.MethodHead,
+		fiber.MethodQuery,
 	}
 
 	// Get handler pointer
--- a/middleware/cors/cors_test.go
+++ b/middleware/cors/cors_test.go
@@ -1223,6 +1223,52 @@ func Test_CORS_AllowPrivateNetwork(t *testing.T) {
 	require.Empty(t, string(ctx.Response.Header.Peek("Access-Control-Allow-Private-Network")), "The Access-Control-Allow-Private-Network header should not be present by default")
 }
 
+// Test_CORS_Query_RequiresPreflight documents RFC 10008: QUERY is not a
+// CORS-safelisted method (the safelisted set is GET/HEAD/POST per the Fetch
+// spec), so a CORS-aware user agent must preflight it. Server-side that means:
+//   - a QUERY preflight must be answered with QUERY in Access-Control-Allow-Methods, and
+//   - an actual QUERY request must be handled as a simple/non-preflight response
+//     (Allow-Origin set, no Allow-Methods) — i.e. QUERY is gated by preflight, not
+//     treated as server-side safelisted.
+func Test_CORS_Query_RequiresPreflight(t *testing.T) {
+	t.Parallel()
+	app := fiber.New()
+	app.Use(New()) // default config: QUERY is in AllowMethods
+	app.Use(func(c fiber.Ctx) error {
+		return c.SendStatus(fiber.StatusOK)
+	})
+	handler := app.Handler()
+
+	t.Run("QUERY preflight is answered with QUERY allowed", func(t *testing.T) {
+		t.Parallel()
+		ctx := &fasthttp.RequestCtx{}
+		ctx.Request.Header.SetMethod(fiber.MethodOptions)
+		ctx.Request.SetRequestURI("https://example.com/")
+		ctx.Request.Header.Set(fiber.HeaderOrigin, "http://example.com")
+		ctx.Request.Header.Set(fiber.HeaderAccessControlRequestMethod, fiber.MethodQuery)
+		handler(ctx)
+
+		require.Equal(t, fiber.StatusNoContent, ctx.Response.StatusCode(), "preflight should return 204")
+		require.Equal(t, "*", string(ctx.Response.Header.Peek(fiber.HeaderAccessControlAllowOrigin)))
+		require.Contains(t, string(ctx.Response.Header.Peek(fiber.HeaderAccessControlAllowMethods)), fiber.MethodQuery,
+			"preflight must advertise QUERY so the user agent allows the request")
+	})
+
+	t.Run("actual QUERY request is a simple response, not safelisted", func(t *testing.T) {
+		t.Parallel()
+		ctx := &fasthttp.RequestCtx{}
+		ctx.Request.Header.SetMethod(fiber.MethodQuery)
+		ctx.Request.SetRequestURI("https://example.com/")
+		ctx.Request.Header.Set(fiber.HeaderOrigin, "http://example.com")
+		handler(ctx)
+
+		require.Equal(t, fiber.StatusOK, ctx.Response.StatusCode())
+		require.Equal(t, "*", string(ctx.Response.Header.Peek(fiber.HeaderAccessControlAllowOrigin)))
+		require.Empty(t, string(ctx.Response.Header.Peek(fiber.HeaderAccessControlAllowMethods)),
+			"actual (non-preflight) QUERY must not carry Access-Control-Allow-Methods")
+	})
+}
+
 // go test -v -run=^$ -bench=Benchmark_CORS_NewHandler -benchmem -count=4
 func Benchmark_CORS_NewHandler(b *testing.B) {
 	app := fiber.New()
--- a/middleware/cors/cors_test.go
+++ b/middleware/cors/cors_test.go
@@ -140,7 +140,7 @@ func testDefaultOrEmptyConfig(t *testing.T, app *fiber.App) {
 	ctx.Request.Header.Set(fiber.HeaderOrigin, "http://localhost")
 	h(ctx)
 
-	require.Equal(t, "GET, POST, HEAD, PUT, DELETE, PATCH", string(ctx.Response.Header.Peek(fiber.HeaderAccessControlAllowMethods)))
+	require.Equal(t, "GET, POST, HEAD, PUT, DELETE, PATCH, QUERY", string(ctx.Response.Header.Peek(fiber.HeaderAccessControlAllowMethods)))
 	require.Empty(t, string(ctx.Response.Header.Peek(fiber.HeaderAccessControlAllowHeaders)))
 	require.Empty(t, string(ctx.Response.Header.Peek(fiber.HeaderAccessControlMaxAge)))
 }
@@ -642,7 +642,7 @@ func Test_CORS_Headers_BasedOnRequestType(t *testing.T) {
 			handler(ctx)
 			require.Equal(t, 204, ctx.Response.StatusCode(), "Status code should be 204")
 			require.Equal(t, "*", string(ctx.Response.Header.Peek(fiber.HeaderAccessControlAllowOrigin)), "Access-Control-Allow-Origin header should be set")
-			require.Equal(t, "GET, POST, HEAD, PUT, DELETE, PATCH", string(ctx.Response.Header.Peek(fiber.HeaderAccessControlAllowMethods)), "Access-Control-Allow-Methods header should be set (preflight request)")
+			require.Equal(t, "GET, POST, HEAD, PUT, DELETE, PATCH, QUERY", string(ctx.Response.Header.Peek(fiber.HeaderAccessControlAllowMethods)), "Access-Control-Allow-Methods header should be set (preflight request)")
 			require.Empty(t, string(ctx.Response.Header.Peek(fiber.HeaderAccessControlAllowHeaders)), "Access-Control-Allow-Headers header should be set (preflight request)")
 		}
 	})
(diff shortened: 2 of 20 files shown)


nekoworks-magic:
I intentionally excluded cache middleware since its signature cannot handle body, maybe some kind of change needed for this?

ReneWerner87:
> I intentionally excluded cache middleware since its signature cannot handle body, maybe some kind of change needed for this?

Yes, we should add that, since the QUERY explicitly states that it can be cached.

nekoworks-magic:
Small changes needed but need to decide how can we generate cache key in defaultKeyGenerator for QUERY request.
I can think adding hash of request body to current result, but need to decide,
- What hash would be used (blake2b can be used without adding deps and have good performance)
- How can keep performance great
etc.

gaby:
- Unit-tests missing
- Documentation missing

nekoworks-magic:
> * Unit-tests missing
> * Documentation missing

@gaby What documentation and units should be added?
I do not know what documentation needed beside of already edited ones.

gaby:
> ## [Codecov](https://app.codecov.io/gh/gofiber/fiber/pull/4436?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=gofiber) Report
> ❌ Patch coverage is `35.29412%` with `11 lines` in your changes missing coverage. Please review. ✅ Project coverage is 91.52%. Comparing base ([`1148bb6`](https://app.codecov.io/gh/gofiber/fiber/commit/1148bb6d9f35278f979b042f6f032ee76664bbcc?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=gofiber)) to head ([`49d45f1`](https://app.codecov.io/gh/gofiber/fiber/commit/49d45f10c01fe4f6d2953264545d657388fc31e0?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=gofiber)).
> 
> [Files with missing lines](https://app.codecov.io/gh/gofiber/fiber/pull/4436?dropdown=coverage&src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=gofiber)	Patch %	Lines
> [domain.go](https://app.codecov.io/gh/gofiber/fiber/pull/4436?src=pr&el=tree&filepath=domain.go&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=gofiber#diff-ZG9tYWluLmdv)	0.00%	[4 Missing ⚠️ ](https://app.codecov.io/gh/gofiber/fiber/pull/4436?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=gofiber)
> [client/request.go](https://app.codecov.io/gh/gofiber/fiber/pull/4436?src=pr&el=tree&filepath=client%2Frequest.go&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=gofiber#diff-Y2xpZW50L3JlcXVlc3QuZ28=)	0.00%	[2 Missing ⚠️ ](https://app.codecov.io/gh/gofiber/fiber/pull/4436?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=gofiber)
> [group.go](https://app.codecov.io/gh/gofiber/fiber/pull/4436?src=pr&el=tree&filepath=group.go&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=gofiber#diff-Z3JvdXAuZ28=)	0.00%	[2 Missing ⚠️ ](https://app.codecov.io/gh/gofiber/fiber/pull/4436?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=gofiber)
> [register.go](https://app.codecov.io/gh/gofiber/fiber/pull/4436?src=pr&el=tree&filepath=register.go&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=gofiber#diff-cmVnaXN0ZXIuZ28=)	0.00%	[2 Missing ⚠️ ](https://app.codecov.io/gh/gofiber/fiber/pull/4436?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=gofiber)
> [helpers.go](https://app.codecov.io/gh/gofiber/fiber/pull/4436?src=pr&el=tree&filepath=helpers.go&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=gofiber#diff-aGVscGVycy5nbw==)	66.66%	[1 Missing ⚠️ ](https://app.codecov.io/gh/gofiber/fiber/pull/4436?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=gofiber)
> Additional details and impacted files
> [☔ View full report in Codecov by Harness](https://app.codecov.io/gh/gofiber/fiber/pull/4436?dropdown=coverage&src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=gofiber). 📢 Have feedback on the report? [Share it here](https://about.codecov.io/codecov-pr-comment-feedback/?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=gofiber).
> 
> 🚀 New features to boost your workflow:

@nekoworks-magic ^ Coverage report

gaby:
> Small changes needed but need to decide how can we generate cache key in defaultKeyGenerator for QUERY request. I can think adding hash of request body to current result, but need to decide,
> 
> * What hash would be used (blake2b can be used without adding deps and have good performance)
> * How can keep performance great
>   etc.

That would be a bad idea, a large body would take a while to hash and can be used to cause DoS.

nekoworks-magic:
Then we can do:
- Do not cache QUERY by default and make user tick the option if they need it
- Set max size of hashable request body

gaby:
@nekoworks-magic Shouldnt it do the same as POST? What does the RFC say?

nekoworks-magic:
RFC says:
“The response to a QUERY method is cacheable; a cache MAY use it to satisfy subsequent QUERY requests.
The cache key for a QUERY request MUST incorporate the request content and related metadata.
Caching QUERY method responses is inherently more complex than caching responses to GET, as the request content must be read completely in order to determine the cache key.”

So no, it is different from POST. When caching QUERY responses, the cache key MUST include the request content, as required by the RFC.

nekoworks-magic:
I added draft implementation.

gaby:
> I added draft implementation.

@nekoworks-magic I think we got to use a pool, similar to how `makeHashAuthFunc` works.

nekoworks-magic:
Did some benchmark and profiling, then I found current implementation does not allocate anything, and adding pool actually makes things slower.
Also I added some tests and benchmarks relevant to cache middleware.

gaby:
## Review — HTTP QUERY method (RFC 10008)

I went through the full diff and the discussion thread. Overall this is a clean, well-tested addition: the new method is wired consistently through `constants.go`, the `Router`/`Register` interfaces and every implementation (`App`, `Group`, `Registering`, `domainRouter`, `domainRegistering`), the client shorthand, and the relevant middleware. The `methodQuery` iota and `DefaultMethods` entry are appended in the same position, so the index mapping stays consistent, and `methodInt` handles the new constant. Method classification is correct per RFC 10008 — QUERY is safe + idempotent, so `IsMethodSafe` (and transitively `IsMethodIdempotent`) returning `true` and CSRF treating it as a non-mutating method are both right.

A few things worth confirming before merge:

### 1. Cache body hashing — DoS surface (the open thread)
The cache key now does `sha256.Sum256(c.Request().Body())` for QUERY. This is correctly **opt-in**: default `cfg.Methods` is `{GET, HEAD}`, and the body branch only runs after the `slices.Contains(cfg.Methods, …)` gate, so the default config never hashes anything. That addresses the "don't enable by default" concern from the thread.

However, @gaby's original point about hashing large bodies isn't fully resolved: once a user adds `QUERY` to `Methods`, **every** QUERY request pays an O(body-size) hash pass before the cache lookup, with no upper bound. The body is already buffered by fasthttp and SHA-256 is fast, so this is a constant-factor cost rather than an amplification bug, but it's still attacker-controlled CPU work. Two reasonable options (both raised earlier in the thread):
- Add a configurable max-hashable-body-size; treat over-limit requests as uncacheable (`cacheUnreachable`) rather than hashing.
- Or, at minimum, document on the `Methods` field that enabling QUERY caching hashes the full request body on every request.

Not a blocker given the opt-in gating, but I'd lean toward documenting it explicitly.

### 2. Cache key — no method component (verify, not a bug here)
`defaultKeyGenerator` doesn't include the HTTP method in the key. With `Methods: {GET, QUERY}` on the same path this happens to be safe because QUERY appends the `|b=<hash>` segment that GET never has, so keys can't collide. Just flagging that the safety here is incidental to the `|b=` suffix rather than an explicit method dimension — worth a comment so a future refactor doesn't drop it.

### 3. CORS default change (intentional behavior change)
Adding `QUERY` to the default `AllowMethods` means every CORS-enabled app now advertises `QUERY` in preflight `Access-Control-Allow-Methods` by default, even when no QUERY routes exist. Harmless and reasonable for v3, but it is a default-behavior change — make sure it's called out in the v3 changelog/migration notes.

### Minor
- **Logger color:** GET and QUERY now share Cyan, so the two are visually indistinguishable in logs. Cosmetic — a distinct color would be friendlier, but fine as-is.
- **Coverage:** Codecov flags the `MethodQuery` arm of `methodInt` in `helpers.go` as uncovered. A tiny assertion (e.g. via the existing method-mapping tests) would close that.

Nice work overall — the test coverage across app/group/domain/route-chain/client and the cache scenarios (same body hit, different body miss, empty body, not-in-Methods) is thorough.

---

gaby:
- BodyLimit is applied at the Fasthttp level, so that's fine.
- The change to CORS, i'm not sure. Thoughts @sixcolors ?

middleware/cache/cache.go
```diff
@@ -1290,6 +1290,12 @@ func defaultKeyGenerator(c fiber.Ctx, cfg *Config) string {
 		buf = append(buf, canonicalCookieSubset(c, cfg.KeyCookies)...)
 	}
 
+	if c.Method() == fiber.MethodQuery {
+		bodyDigest := sha256.Sum256(c.Request().Body())
+		buf = append(buf, '|', 'b', '=')
+		buf = hex.AppendEncode(buf, bodyDigest[:])
```
gaby: This hashes the full request body with SHA-256 on **every** QUERY request (cache hits included). RFC 10008 only requires the cache key to *incorporate the request content* — it doesn't mandate a hash, algorithm, or encoding.

The key generator already has `boundKeySegment` (used for path/query), which keeps a segment **verbatim when ≤ `maxKeyDimensionSegmentLength` (192B)** and only hashes larger ones. Routing the body through it removes the hash from the hot path for typical (small) QUERY bodies, while still bounding the key length for the in-memory map and external `Storage` backends:

```suggestion
		// RFC 10008 only requires the cache key to incorporate the request content,
		// not that it be hashed. boundKeySegment keeps small bodies verbatim (no
		// hashing on the hot path) and hashes only bodies larger than
		// maxKeyDimensionSegmentLength, which still bounds the key length.
		buf = append(buf, '|', 'b', '=')
		buf = append(buf, boundKeySegment(utils.UnsafeString(c.Request().Body()))...)
```

SHA-256 is the right choice for the large-body tail that still hashes: `hash/maphash` is randomly seeded (keys wouldn't match across instances/restarts → breaks external `Storage`), and stable non-crypto hashes (xxhash/crc32) are collision-engineerable → cache poisoning on shared stores. No request-body size limit is needed — fasthttp's `MaxRequestBodySize` already bounds the absolute size.

I have this change plus a large-body test (bodies differing only past the 192B threshold stay distinct) and a `docs/middleware/cache.md` note on `gofiber/fiber:claude/fiber-pr-4436-review-s1n8d6` if it's easier to cherry-pick.

---

gaby (commented):
One performance suggestion on the QUERY cache-key body hashing — details inline. The current code is correct; this just keeps SHA-256 off the hot path for typical (small) request bodies by reusing the existing `boundKeySegment` helper.

---

gaby (approved):
LGTM, only thing left is the CORS change. Pending @sixcolors review.

gaby:
This is what the RFC says about CORS:


> A QUERY request from user agents implementing Cross-Origin Resource Sharing (CORS) will require a "preflight" request, as QUERY does not belong to the set of CORS-safelisted methods (see [[FETCH](https://www.rfc-editor.org/info/rfc10008/#FETCH)]).

middleware/cors/cors_test.go
```diff
@@ -676,7 +677,7 @@ func Test_CORS_Headers_BasedOnRequestType(t *testing.T) {
 			handler(ctx)
 			require.Equal(t, 204, ctx.Response.StatusCode(), "Status code should be 204")
 			require.Equal(t, "*", string(ctx.Response.Header.Peek(fiber.HeaderAccessControlAllowOrigin)), "Access-Control-Allow-Origin header should be set")
-			require.Equal(t, "GET, POST, HEAD, PUT, DELETE, PATCH", string(ctx.Response.Header.Peek(fiber.HeaderAccessControlAllowMethods)), "Access-Control-Allow-Methods header should be set (preflight request)")
+			require.Equal(t, "GET, POST, HEAD, PUT, DELETE, PATCH, QUERY", string(ctx.Response.Header.Peek(fiber.HeaderAccessControlAllowMethods)), "Access-Control-Allow-Methods header should be set (preflight request)")
```
gaby, line 680: Adding `MethodQuery` to this matrix already exercises QUERY, which is great. Per RFC 10008, QUERY is **not** CORS-safelisted (the safelisted set is GET/HEAD/POST per Fetch), so a CORS-aware user agent must preflight it. It's worth locking that requirement down in a small, RFC-named regression test that documents both halves explicitly: the preflight is answered with QUERY allowed, **and** an actual QUERY is a simple/non-preflight response (no `Allow-Methods`) — i.e. QUERY is gated by preflight, not server-side safelisted.

Suggested addition — a new top-level function right **after** `Test_CORS_Headers_BasedOnRequestType` (posting as a code block rather than a one-click suggestion since it's a new function outside this diff hunk):

```go
// Test_CORS_Query_RequiresPreflight documents RFC 10008: QUERY is not a
// CORS-safelisted method (the safelisted set is GET/HEAD/POST per the Fetch
// spec), so a CORS-aware user agent must preflight it. Server-side that means:
//   - a QUERY preflight must be answered with QUERY in Access-Control-Allow-Methods, and
//   - an actual QUERY request must be handled as a simple/non-preflight response
//     (Allow-Origin set, no Allow-Methods) — i.e. QUERY is gated by preflight, not
//     treated as server-side safelisted.
func Test_CORS_Query_RequiresPreflight(t *testing.T) {
	t.Parallel()
	app := fiber.New()
	app.Use(New()) // default config: QUERY is in AllowMethods
	app.Use(func(c fiber.Ctx) error {
		return c.SendStatus(fiber.StatusOK)
	})
	handler := app.Handler()

	t.Run("QUERY preflight is answered with QUERY allowed", func(t *testing.T) {
		t.Parallel()
		ctx := &fasthttp.RequestCtx{}
		ctx.Request.Header.SetMethod(fiber.MethodOptions)
		ctx.Request.SetRequestURI("https://example.com/")
		ctx.Request.Header.Set(fiber.HeaderOrigin, "http://example.com")
		ctx.Request.Header.Set(fiber.HeaderAccessControlRequestMethod, fiber.MethodQuery)
		handler(ctx)

		require.Equal(t, fiber.StatusNoContent, ctx.Response.StatusCode(), "preflight should return 204")
		require.Equal(t, "*", string(ctx.Response.Header.Peek(fiber.HeaderAccessControlAllowOrigin)))
		require.Contains(t, string(ctx.Response.Header.Peek(fiber.HeaderAccessControlAllowMethods)), fiber.MethodQuery,
			"preflight must advertise QUERY so the user agent allows the request")
	})

	t.Run("actual QUERY request is a simple response, not safelisted", func(t *testing.T) {
		t.Parallel()
		ctx := &fasthttp.RequestCtx{}
		ctx.Request.Header.SetMethod(fiber.MethodQuery)
		ctx.Request.SetRequestURI("https://example.com/")
		ctx.Request.Header.Set(fiber.HeaderOrigin, "http://example.com")
		handler(ctx)

		require.Equal(t, fiber.StatusOK, ctx.Response.StatusCode())
		require.Equal(t, "*", string(ctx.Response.Header.Peek(fiber.HeaderAccessControlAllowOrigin)))
		require.Empty(t, string(ctx.Response.Header.Peek(fiber.HeaderAccessControlAllowMethods)),
			"actual (non-preflight) QUERY must not carry Access-Control-Allow-Methods")
	})
}
```

`require.Contains` is used for the preflight assertion so it stays robust if the default `AllowMethods` order changes or grows. This is already committed (and passing under `-race`) on `gofiber/fiber:claude/fiber-pr-4436-review-s1n8d6` (commit `22c2e60`) if it's easier to cherry-pick.

---

gaby (commented):
One optional test suggestion on the CORS side: a small dedicated regression test documenting RFC 10008's preflight requirement for QUERY (it's not CORS-safelisted). Details inline — no production change needed, the middleware already behaves correctly.

---

middleware/cors/cors_test.go
gaby: @nekoworks-magic Can you add this test, thanks!

PR #4439, opened 2026-06-18
🐛 bug: validate nested CSRF extractor chains

gaby:
### Motivation

- Prevent a security bypass where a nested `extractors.Chain(...)` can hide a fallback `FromCookie` extractor that reads the same cookie used for CSRF token storage, causing the double-submit check to compare a cookie to itself. 
- The CSRF refactor introduced acceptance of the shared `extractors.Extractor` type (which includes `FromCookie`) while the existing validation only inspected one chain level, leaving nested cookie fallbacks unchecked.

### Description

- Add a recursive inspector `isInsecureCookieExtractorRecursive` and use it when validating chained extractors in `middleware/csrf/config.go` so nested `Chain` entries are checked transitively. 
- Replace the shallow chained check in `validateExtractorSecurity` with the recursive check so any nested cookie extractor matching `CookieName` triggers a panic. 
- Add a regression unit test `InsecureNestedChainedExtractor` to `middleware/csrf/config_test.go` that constructs a nested `Chain(..., Chain(..., FromCookie("csrf_")))` and asserts the configuration normalization panics. 
- Preserve existing behavior for primary extractor checks and case-insensitive cookie-name warnings.

### Testing

- Ran `go test ./middleware/csrf -run 'Test_CSRF_ExtractorSecurity_Validation'` and it passed (`ok github.com/gofiber/fiber/v3/middleware/csrf`).
- Ran the full test suite via `make test` and all tests passed (3645 tests run, 2 skipped). 
- Ran `make generate`, `make betteralign`, `make format`, and `make lint` and they completed successfully. 
- Ran `make audit` which failed due to `govulncheck` findings in the local Go 1.25.1 standard library (the scan reported 25 standard-library vulnerabilities); this is an environment/toolchain issue and not a regression introduced by these changes.

### DIFF
--- a/middleware/csrf/config.go
+++ b/middleware/csrf/config.go
@@ -193,7 +193,7 @@ func validateExtractorSecurity(cfg *Config) {
 	// Check chained extractors recursively so a nested chain cannot hide a
 	// fallback that reads from the CSRF storage cookie.
 	for i, extractor := range cfg.Extractor.Chain {
-		if isInsecureCookieExtractorRecursive(extractor, cfg.CookieName) {
+		if isInsecureCookieExtractorRecursive(extractor, cfg.CookieName, 1) {
 			panic(fmt.Sprintf("CSRF: Chained extractor #%d reads from the same cookie '%s' "+
 				"used for token storage. This completely defeats CSRF protection.", i+1, cfg.CookieName))
 		}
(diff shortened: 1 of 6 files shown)


ReneWerner87 (approved):

middleware/csrf/config.go
```diff
@@ -204,6 +205,22 @@ func validateExtractorSecurity(cfg *Config) {
 	}
 }
 
+// isInsecureCookieExtractorRecursive checks if an extractor or any nested chain
+// unsafely reads from the CSRF cookie.
+func isInsecureCookieExtractorRecursive(extractor extractors.Extractor, cookieName string) bool {
```
sixcolors: Looks good overall, but I’d add a cycle guard before recursing through extractor.Chain to avoid a potential infinite loop if a malformed or self-referential chain ever slips through.

A small visited-set check would make the validation safe without changing the intended security behavior.

sixcolors (changes requested):

middleware/csrf/config.go
sixcolors: It might be a good idea to disallow cycles in the extractor package. Unless we can find some value in allowing them.

middleware/csrf/config.go
```diff
@@ -205,15 +205,24 @@ func validateExtractorSecurity(cfg *Config) {
 	}
 }
 
+// maxExtractorChainDepth bounds recursion when validating chained extractors,
+// guarding against pathological or malformed configurations.
+const maxExtractorChainDepth = 100
```
gaby: @sixcolors Thoughts? Or should I make this a smaller value. Like 32 or 64

sixcolors: I'm going to write an extractors PR that I think should address this globally. Will tag you when done.

sixcolors (approved):

middleware/csrf/config_test.go
```diff
@@ -107,10 +107,10 @@ func Test_CSRF_ExtractorSecurity_Validation(t *testing.T) {
 	// A deeply nested chain must not cause unbounded recursion during validation.
 	t.Run("DeeplyNestedSecureChain", func(t *testing.T) {
 		t.Parallel()
-		// Build a chain nested well beyond maxExtractorChainDepth using only
-		// secure extractors so validation terminates without panicking.
+		// Build a deeply nested chain using only secure extractors so validation
+		// terminates without panicking.
 		deep := extractors.FromHeader("X-Csrf-Token")
-		for range maxExtractorChainDepth + 50 {
+		for range 150 {
```
gaby, line 113: This should be a package const

PR #4456, opened 2026-06-25
feat: complete HTTP QUERY method support (follow-up to #4436)

ReneWerner87:
## Description

Follow-up to #4436 (HTTP QUERY method, RFC 10008), filling gaps found in review and aligning the QUERY cache key with the keygen refactor from #4450 (already on main).

## Changes

- **client:** add `Client.Query()` and package-level `Query()` so the QUERY verb has the same shorthands as every other method. #4436 only added `Request.Query()`, leaving the `Client` API inconsistent.
- **cache:** escape the QUERY request body and append it via `appendBoundKeySegment`, consistent with the per-dimension escaping the keygen already applies to path/query/headers/cookies. Before this the body was appended raw and in the old return-string style, which left a stylistic inconsistency and a latent key-injection asymmetry (not exploitable today, since the body is the terminal key dimension). Pinned with new QUERY golden cases in `keygen_test.go`, including a delimiter case (`a|b:c` -> `b=a\pb\cc`).
- **tests:** add direct `IsMethodSafe`/`IsMethodIdempotent` assertions for QUERY (none existed), cover QUERY in the CSRF safe-method tests, and add `Test_Client_Query`.
- **docs:** add QUERY to the CORS `AllowMethods` default, the `Register` interface, the safe-method lists (idempotency, earlydata, csrf), the cache body-key note, and the client REST reference.

## Type of change

- [x] Bug fix / hardening (cache-key body escaping)
- [x] New feature (client QUERY shorthands)
- [x] Documentation update

## Checklist

- [x] Conducted a self-review of the code.
- [x] Updated the documentation under `/docs/`.
- [x] Added/updated unit tests.
- [x] Ensured new and existing unit tests pass locally (`go test ./...`, all packages green).
- [x] Aimed for optimal performance with minimal allocations (`escapeKeyDelimiters` has a no-alloc fast path for bodies without delimiters).


gaby (approved):
⚡ LGTM

PR #4465, opened 2026-06-26
🐛 fix(ctx): delegate Deadline/Done/Err to the user context

### LINKED ISSUE #4335: 🐛 [Bug]: DefaultCtx.Done() returns nil — cancellation impossible through context.Context interface

yathboss:
## Description

`DefaultCtx.Deadline()`, `Done()` and `Err()` were hardcoded no-ops, so the `context.Context` contract was broken:

- `Done()` returned `nil` → blocks forever in any `select`
- `Err()` always returned `nil` → never reflected cancellation
- `Deadline()` always reported no deadline

Because of this, a cancelable context set via `SetContext` — **including the `context.WithTimeout` installed by the timeout middleware** — was invisible to any library that consumed `fiber.Ctx` as a `context.Context` (database drivers, HTTP/gRPC clients, etc.). They would never respect cancellation.

## Fix

Delegate the three methods to `c.Context()`:

```go
func (c *DefaultCtx) Deadline() (time.Time, bool) { return c.Context().Deadline() }
func (c *DefaultCtx) Done() <-chan struct{}       { return c.Context().Done() }
func (c *DefaultCtx) Err() error                  { return c.Context().Err() }
```

When no context has been set, `c.Context()` returns `context.Background()`, whose `Done()` is `nil`, `Err()` is `nil` and `Deadline()` reports no deadline — so **existing behavior is preserved and the change is backward compatible**.

## Tests

- Added `Test_Ctx_Deadline_Delegates`, `Test_Ctx_Done_Delegates`, `Test_Ctx_Err_Delegates` covering deadline/cancellation propagation.
- Existing `Test_Ctx_Deadline`, `Test_Ctx_Done`, `Test_Ctx_Err` (no-context case) still pass, confirming backward compatibility.
- `go vet` clean; timeout-middleware and Context tests pass.

Updated the `context.Context` section in `docs/api/ctx.md`.

Fixes #4335


gaby:
This is a duplicate of https://github.com/gofiber/fiber/pull/4399

PR #4493, opened 2026-07-02
🐛 fix(ctx): compare Accept media type case-insensitively

0xghost42:
During content negotiation using `Accepts`, the `acceptsOfferType` function compared the parsed `Accept` media type against the offer using a case-sensitive equality check (`==`).

```go
if spec == mimetype {
```

The request `Accept: Application/JSON` did not match an `application/json` offer, resulting in `Accepts("application/json")` returning an empty string.

According to RFC 9110 8.3.1, type, subtype, and parameter name tokens are case-insensitive. The rest of this file already adheres to this requirement:
- the sibling `acceptsOffer` (used for `AcceptsCharsets`/`AcceptsEncodings`) matches with `utils.EqualFold`
- the `<type>/*` wildcard branch of `acceptsOfferType` itself uses `utils.EqualFold`
- `paramsMatch` compares parameter names and values with `utils.EqualFold`

The exact-match line used `==`. This has been changed to `utils.EqualFold(spec, mimetype)` for consistency.
### Test

Extended `Test_Utils_GetOffer` with mixed-case type/subtype cases (mixed-case in the header vs a lowercase offer, and the reverse, plus one with a parameter). Verified they fail on the old `==` (`getOffer([]byte("Application/JSON"), acceptsOfferType, "application/json")` returns `""`) and pass with the fix. `gofmt` and `go vet` clean.

### DIFF
--- a/helpers.go
+++ b/helpers.go
@@ -361,7 +361,7 @@ func acceptsOfferType(spec, offerType string, specParams headerParams) bool {
 		mimetype = utils.GetMIME(offerMime) // extension
 	}
 
-	if spec == mimetype {
+	if utils.EqualFold(spec, mimetype) {
 		// Accept: <MIME_type>/<MIME_subtype>
 		return paramsMatch(specParams, offerParams)
 	}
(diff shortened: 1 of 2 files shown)


gaby:
@0xghost42 It says here https://www.rfc-editor.org/rfc/rfc9110.html#section-8.3.1

> The type and subtype tokens are case-insensitive.

gaby:
I updated the PR description to improve clarity.

gaby (approved):
LGTM

ReneWerner87 (approved):

PR #4219, opened 2026-04-18
chore: Bump minimum Go to 1.24

gaby:
### Motivation

- Raise the minimum supported Go version and keep CI and documentation consistent with current Go releases. 
- Upgrade several dependencies to newer, compatible releases to match the Go toolchain bump. 
- Adjust CSRF middleware tests to reflect updated token deletion / handler behavior.

### Description

- Updated Go version in `go.mod` to `go 1.24.0` and bumped multiple dependencies including `github.com/valyala/fasthttp`, `golang.org/x/sys`, `github.com/mattn/*`, and `github.com/tinylib/msgp`, with corresponding `go.sum` updates. 
- Updated GitHub Actions workflows to use Go `1.24.x` in `benchmark.yml` and `linter.yml`, and replaced the test matrix in `test.yml` with `go-version: [1.24.x, 1.25.x, 1.26.x]`. 
- Revised docs and READMEs to state the new minimum supported Go version (`1.24`) and updated the compatibility note in `.github/README.md` and `docs/intro.md`. 
- Modified CSRF middleware tests in `middleware/csrf/csrf_test.go` to call token deletion via the handler context in the request flow and adjusted an expected response status from `403` to `200` where token deletion occurs. 
- Added a minimum Go version note to `utils/README.md`.

### Testing

- Ran unit tests with `go test ./... -v -race -count=1` across the updated local matrix, and tests completed successfully. 
- Linting was executed with `golangci-lint` under the workflow configuration and passed. 
- Benchmark workflow remains configured to run `go test ./... -bench .` and is updated to use `setup-go@v6` with `1.24.x`.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e244e22d0883339363ab000efe9f5c)

PR #4495, opened 2026-07-03
🐛 fix(proxy): backport BalancerForward X-Real-IP overwrite fix to v2

terraincognita07:
## Summary

Backports the `v3` fix for **GHSA-gcfq-8gqf-4876 / CVE-2026-45045** ("GoFiber
Vulnerable to X-Real-IP Spoofing via Header.Add() in BalancerForward") to the
`v2` branch.

`BalancerForward` in `middleware/proxy/proxy.go` uses `Header.Add("X-Real-IP",
c.IP())` when forwarding a request to a backend. Because `Add` appends rather
than replaces, a client-supplied `X-Real-IP` header is not removed — it is
kept as the *first* value, with the real client IP appended as a second
value. Upstream servers that read only the first `X-Real-IP` value (nginx,
Express, and most HTTP servers) therefore trust the attacker-controlled value
for IP-based logging, rate limiting, ACLs, and geofencing, i.e. an
authentication/authorization-adjacent spoofing bypass (CWE-290).

The v3 branch was fixed in commit
[`1403cc8292da3220e9316960b4030cc722a0f396`](https://github.com/gofiber/fiber/commit/1403cc8292da3220e9316960b4030cc722a0f396)
(PR #4260, "fix(proxy): Ensure BalancerForward overwrites X-Real-IP header"),
released in v3.3.0. Per the advisory, v2's `vulnerable_version_range` is
`<= 2.52.13` with **no patched version published** for `github.com/gofiber/fiber/v2`
— the v2 branch never received this fix. This PR closes that gap.

## Change

One-line fix in `middleware/proxy/proxy.go`, `BalancerForward`: replace
`Header.Add` with `Header.Set` so any attacker-supplied `X-Real-IP` header is
overwritten with the real client IP (`c.IP()`) rather than retained alongside
it.

```diff
-		c.Request().Header.Add("X-Real-IP", c.IP())
+		c.Request().Header.Set("X-Real-IP", c.IP())
```

Also updates the matching code sample in `docs/api/middleware/proxy.md` (the
v2 equivalent of the doc fix in PR #4260), which showed the same insecure
`Header.Add` pattern, so the documented example doesn't teach the vulnerable
usage.

## Test

Added `Test_Proxy_Balancer_Forward_OverwritesXRealIP` in
`middleware/proxy/proxy_test.go`, adapted from the v3 regression test added
in PR #4260 (`Test_Proxy_Balancer_Forward_OverwritesXRealIP`) to v2's
`*fiber.Ctx` pointer-receiver handler signature and the `utils.AssertEqual`
assertion helper used elsewhere in this file. The test sends a request
through `BalancerForward` with a spoofed `X-Real-IP: 10.0.0.1` header already
set, and asserts the backend observes only the real (test-harness) client IP
and never the spoofed value.

Confirmed the test fails against the unpatched `Header.Add` call (backend
observes the spoofed `10.0.0.1` as the first header value) and passes once
`Header.Set` is restored.

## Compatibility

Behavior-only fix, no signature or config changes. Any caller currently
relying on the buggy "spoofed value plus real value" double header (there is
no legitimate reason to) would see only the real client IP going forward,
which is the documented and intended behavior.

## References

- GHSA: https://github.com/gofiber/fiber/security/advisories/GHSA-gcfq-8gqf-4876
- CVE: CVE-2026-45045
- v3 fix commit: https://github.com/gofiber/fiber/commit/1403cc8292da3220e9316960b4030cc722a0f396
- v3 fix PR: https://github.com/gofiber/fiber/pull/4260

Found via a Dependabot alert on a consuming project.


ReneWerner87 (approved):

gaby:
@ReneWerner87 I dont think the CI for v2 works anymore.

gaby (approved):

ReneWerner87:
ok, let me fix this

gaby:
Related https://github.com/gofiber/fiber/pull/4219

aldok10:
Hi maintainers,

Thanks for merging this backport. I have a question: **is there a planned release tag for `github.com/gofiber/fiber/v2` that includes this fix?**

As of today, the latest v2 release is [`v2.52.13`](https://github.com/gofiber/fiber/releases/tag/v2.52.13) (released April 25, 2026), which predates this merge (July 3, 2026). This means the fix for [CVE-2026-45045 / GHSA-gcfq-8gqf-4876](https://github.com/advisories/GHSA-gcfq-8gqf-4876) is only available on the `v2` branch HEAD but not in any tagged release that Go modules can resolve.

**Why this matters for v2 consumers:**

1. **Go modules require a tagged release** to pull in the fix. Running `go get github.com/gofiber/fiber/v2@latest` still resolves to `v2.52.13`, which remains vulnerable.
2. **Dependabot / govulncheck / Trivy / Snyk** continue to flag projects using `github.com/gofiber/fiber/v2` as affected because the advisory's `patched_versions` field has no v2 entry. A tagged release would allow the advisory to be updated and automated scanners to clear the alert.
3. **Many production projects cannot pin to a commit hash** due to internal policies or compliance requirements that mandate semver-tagged dependencies. Without a release, these teams are stuck between "known vulnerable" and "use an untagged commit."

Would it be possible to cut a `v2.52.14` (or similar) patch release that includes this commit? This would unblock the entire v2 ecosystem from resolving the CVE through normal dependency update workflows.

Thank you!

ReneWerner87:
https://github.com/gofiber/fiber/releases/tag/v2.52.14

PR #4508, opened 2026-07-07
🐛 fix(ctx): honor q=0 rejections in content negotiation

0xghost42:
## Description

\`getOffer\` discarded media ranges with a quality of 0 *before* matching offers, so a more specific \`q=0\` rejection was lost whenever a broader range (\`*\`, \`text/*\`) with quality > 0 also matched the same offer. The negotiated result could therefore be a media type, encoding, charset or language the client had **explicitly refused**.

Most damning case — the client accepts any encoding except gzip, the server can do gzip or deflate:

```
Accept-Encoding: *, gzip;q=0     offers: gzip, deflate
before: "gzip"   (the one the client refused)
after:  "deflate"
```

More:

| Header | Offer | Before | After (correct) |
|---|---|---|---|
| \`Accept: text/*, text/html;q=0\` | \`text/html\` | \`text/html\` | \`\` |
| \`Accept: */*, application/json;q=0\` | \`application/json\` | \`application/json\` | \`\` |
| \`Accept-Language: *, en;q=0\` | \`en\` | \`en\` | \`\` |

## Root cause / spec

Per [RFC 9110 §12.5.1](https://www.rfc-editor.org/rfc/rfc9110#section-12.5.1), the **most specific** matching media range determines an offer's acceptability, and a quality of 0 means the client rejects that range. The old code dropped \`q=0\` ranges up front, so the broader surviving range decided the match.

## Fix

- Keep \`q=0\` ranges during parsing.
- Accept an offer only when its most specific matching range has quality > 0: a \`q=0\` range never selects an offer, and it rejects an offer that no more-specific accepted range covers.

Affects \`Accepts\`, \`AcceptsEncodings\`, \`AcceptsCharsets\`, \`AcceptsLanguages\` and \`Format\`/\`AutoFormat\` (all route through \`getOffer\`). Matches the behavior of Node's \`negotiator\`.

## Tests

Added \`Test_Utils_GetOffer_QualityZeroRejection\` covering media type, encoding and language cases — proven to FAIL on \`main\` (returns the refused offer) and PASS with the fix. Existing negotiation tests unchanged and green; full root package \`go test .\` passes.


ReneWerner87:
@0xghost42 can you share a before and after benchmark for the accept methods

helpers.go
```diff
@@ -691,25 +685,57 @@ func getOffer(header []byte, isAccepted func(spec, offer string, specParams head
...
+	return result
+}
+
+// rejectedByMoreSpecificRange reports whether a media range more specific than
+// the one at baseSpecificity matches the offer with a quality of 0, i.e. the
+// client explicitly rejected the offer per RFC 9110 §12.5.1.
+func rejectedByMoreSpecificRange(types []acceptedType, isAccepted func(spec, offer string, specParams headerParams) bool, offer string, baseSpecificity int) bool {
+	for i := range types {
+		if types[i].quality == 0 && types[i].specificity > baseSpecificity &&
+			isAccepted(types[i].spec, offer, types[i].params) {
```
gaby: **Remaining spec gap: parameter-based specificity is ignored.**

This check compares only the integer `specificity` (1–4), which encodes wildcard depth but not media-type parameters. Per RFC 9110 §12.5.1 a range with parameters is more specific than the same range without them — and `sortAcceptedTypes` already honors that via the `len(params)` tie-break — but the rejection check doesn't. Confirmed on this branch:

```go
getOffer([]byte("text/html, text/html;level=1;q=0"), acceptsOfferType, "text/html;level=1")
// returns "text/html;level=1" — the offer the client explicitly rejected
```

Both ranges have specificity 3, so the strictly-greater comparison never fires and the `q=0` rejection is lost — the same bug class this PR fixes, one corner narrower. Mirroring the sort's tie-break closes it:

```go
func rejectedByMoreSpecificRange(types []acceptedType, isAccepted func(spec, offer string, specParams headerParams) bool, offer string, base acceptedType) bool {
	for i := range types {
		if types[i].quality == 0 &&
			(types[i].specificity > base.specificity ||
				(types[i].specificity == base.specificity && len(types[i].params) > len(base.params))) &&
			isAccepted(types[i].spec, offer, types[i].params) {
			return true
		}
	}
	return false
}
```

(with the call site passing `acceptedType` instead of `acceptedType.specificity`), plus a test for the case above.

---

helpers.go
```diff
@@ -691,25 +685,57 @@ func getOffer(header []byte, isAccepted func(spec, offer string, specParams head
...
 			if offer == "" {
 				continue
 			}
-			if isAccepted(acceptedType.spec, offer, acceptedType.params) {
-				if acceptedType.params != nil {
-					headerParamPool.Put(acceptedType.params)
-				}
-				return offer
+			if isAccepted(acceptedType.spec, offer, acceptedType.params) &&
+				!rejectedByMoreSpecificRange(acceptedTypes, isAccepted, offer, acceptedType.specificity) {
```
gaby: **Perf: the rejection scan runs even when the header has no `q=0` ranges.**

`rejectedByMoreSpecificRange` walks all accepted types on every successful match, so the overwhelmingly common no-rejection case pays for it too. `Benchmark_Utils_GetOffer` (count=3) shows a consistent ~2–5% regression vs `main`, e.g. `simple` 71.5 → 74.7 ns/op and `web_browser` 279 → 289 ns/op (allocations unchanged at 0 B/op).

Tracking a flag during parsing makes the common path essentially free:

```go
hasRejections := false
// in the forEachMediaRange callback, after quality is parsed:
if quality == 0 {
	hasRejections = true
}

// selection loop:
if isAccepted(acceptedType.spec, offer, acceptedType.params) &&
	(!hasRejections || !rejectedByMoreSpecificRange(acceptedTypes, isAccepted, offer, acceptedType.specificity)) {
```

Minor bonus: since `sortAcceptedTypes` sorts by quality descending, all `q=0` entries sit at the tail, so the scan could also iterate from the end and stop at the first non-zero quality.

---

helpers_test.go
```diff
@@ -254,6 +254,31 @@ func newLocalListener(t *testing.T) net.Listener {
...
+	require.Equal(t, "text/plain", getOffer([]byte("text/*, text/html;q=0"), acceptsOfferType, "text/plain"))
+
+	// Accept-Encoding: the client accepts anything except gzip, so a server
+	// offering gzip or deflate must pick deflate, not the refused gzip.
+	require.Equal(t, "deflate", getOffer([]byte("*, gzip;q=0"), acceptsOffer, "gzip", "deflate"))
+	require.Empty(t, getOffer([]byte("*, gzip;q=0"), acceptsOffer, "gzip"))
+
+	// Accept-Language: "*" accepts any tag, but "en;q=0" rejects English.
+	require.Empty(t, getOffer([]byte("*, en;q=0"), acceptsLanguageOfferBasic, "en"))
+	require.Equal(t, "fr", getOffer([]byte("*, en;q=0"), acceptsLanguageOfferBasic, "en", "fr"))
```
gaby, line 279: Good coverage of the rejection direction. Two additions worth pinning down (both pass on this branch — verified locally):

- the **non-rejection** direction: a *less* specific `q=0` range must not reject an offer matched by a more specific accepting range. It works today, but nothing guards against a future regression toward "any q=0 match rejects".
- a charset case, since `AcceptsCharsets` is listed as affected.

```suggestion
	require.Equal(t, "fr", getOffer([]byte("*, en;q=0"), acceptsLanguageOfferBasic, "en", "fr"))

	// A less specific q=0 range must NOT reject an offer that a more
	// specific accepting range matches (RFC 9110 §12.5.1).
	require.Equal(t, "text/html", getOffer([]byte("text/*;q=0, text/html"), acceptsOfferType, "text/html"))

	// Accept-Charset: "*" accepts any charset, but "utf-8;q=0" rejects UTF-8.
	require.Empty(t, getOffer([]byte("*, utf-8;q=0"), acceptsOffer, "utf-8"))
	require.Equal(t, "iso-8859-1", getOffer([]byte("*, utf-8;q=0"), acceptsOffer, "utf-8", "iso-8859-1"))
```

---

gaby (commented):
The core fix is spec-correct and verified locally: full root package `go test .` passes, the new tests behave as described, and the restructured pool handling even fixes a latent `headerParamPool` leak (the old code returned early on a match without putting the remaining accepted types' params back).

Three suggestions inline, none blocking:

1. **Param-based specificity** — `rejectedByMoreSpecificRange` ignores media-type parameters, so `Accept: text/html, text/html;level=1;q=0` still returns a rejected offer (`text/html;level=1`). Same bug class this PR fixes; small change to close it.
2. **Hot-path perf** — benchmarks show ~2–5% regression even for headers with no `q=0` ranges; a `hasRejections` flag set during parsing makes the common path free.
3. **Test coverage** — pin the non-rejection direction (less specific `q=0` must not reject) and add a charset case; suggestion block included.

---

0xghost42:
Thanks for the thorough review. Pushed a follow-up (e8d21b8b) addressing all of it.

### Same-class rejections (@coderabbitai + suggestion 1)

Right — `Accept-Language`, `Accept-Charset` and `Accept-Encoding` put a positive match and a `q=0` refusal in the same coarse bucket, and media-type parameters were ignored, so `en, en-US;q=0`, `utf-8, utf-8;q=0` and `text/html, text/html;level=1;q=0` were not rejected.

The `acceptsX` helpers now return the **effective match specificity** (an `int`; 0 = no match) instead of a bool, and the rejection check compares that rather than the parsed bucket. A `q=0` range now rejects an offer when it matches **at least as specifically** as the positive range:

- `en-US;q=0` (exact) overrides `en` (prefix)
- `utf-8;q=0` overrides an earlier `utf-8`
- `text/html;level=1;q=0` overrides `text/html`

while a *less* specific refusal still does not override (`text/*;q=0` does not reject `text/html`, `en;q=0` does not reject an exact `en-US`).

### Perf (suggestion 2 + @ReneWerner87s benchmark request)

Added a `hasRejections` flag set during parsing. When no range carries `q=0`, selection now takes a fast path that is the **same "first matching range wins" algorithm as before** — no specificity is computed or compared, and the more-specific-rejection scan is skipped entirely. The extra work only runs when the header actually contains a `q=0`.

Allocations are unchanged: `go test -run=^$ -bench=Benchmark_Utils_GetOffer -benchmem` reports identical `B/op` and `allocs/op` for every sub-case between `main` and this branch (0 allocs on the no-parameter paths).

For timings, my local numbers are from a laptop and too thermally noisy to quote precisely (single-range cases move within ±4% run to run), but the no-`q=0` fast path does the same work as `main`; the CI benchmark job will be the authoritative before/after. Happy to paste raw `benchstat` output if you want it from a specific case.

### Tests (suggestion 3)

Extended `Test_Utils_GetOffer_QualityZeroRejection` with the same-bucket media-parameter, language (basic + extended) and charset cases, plus the non-rejection direction (a less specific `q=0` must not reject) and the wildcard-positive-vs-exact-refusal case. They fail without the `>=` specificity comparison.

Also confirmed the `headerParamPool` handling is preserved (the trailing release loop still runs on every path).

0xghost42:
@ReneWerner87 this one's green and MERGEABLE now. The accept-method before/after benchmark you asked for is in the thread — allocs/op are identical to main across every case (the fast path with no `q=0` present stays first-match-wins, the specificity work only runs when a rejection is parsed). gaby also confirmed the spec behavior and noted the restructure fixes the latent `headerParamPool` leak. Happy to adjust anything else, otherwise it should be good to go.

ReneWerner87:
@0xghost42 Can you share the benchmark results? I'm not interested in the allocations, but rather in the time.

ReneWerner87:
```
goos: darwin
goarch: arm64
cpu: Apple M2 Pro
                                                                    │   main.txt   │            PR-4508.txt            │
                                                                    │    sec/op    │   sec/op     vs base              │
_Ctx_Accepts/run-[]string{".xml"}-12                                  185.0n ±  0%   192.4n ± 1%  +4.03% (p=0.000 n=8)
_Ctx_Accepts/run-[]string{"json",_"xml"}-12                           330.4n ±  1%   339.9n ± 1%  +2.88% (p=0.010 n=8)
_Ctx_Accepts/run-[]string{"application/json",_"application/xml"}-12   222.8n ±  7%   226.2n ± 0%       ~ (p=0.111 n=8)
_Ctx_AcceptsCharsets-12                                               114.7n ± 10%   104.5n ± 1%       ~ (p=0.152 n=8)
_Ctx_AcceptsEncodings-12                                              154.1n ±  5%   151.4n ± 4%       ~ (p=0.666 n=8)
_Ctx_AcceptsLanguages-12                                              242.4n ±  4%   245.7n ± 3%       ~ (p=0.246 n=8)
_Ctx_Format/with_arg_allocation-12                                    480.7n ±  2%   474.8n ± 1%       ~ (p=0.195 n=8)
_Ctx_Format/pre-allocated_args-12                                     457.4n ±  2%   449.1n ± 1%  -1.81% (p=0.000 n=8)
_Ctx_Format/text/plain-12                                             148.3n ±  2%   149.0n ± 1%       ~ (p=0.505 n=8)
_Ctx_Format/json-12                                                   213.1n ±  5%   209.6n ± 0%       ~ (p=0.587 n=8)
_Ctx_Format/msgpack-12                                                217.0n ±  1%   222.8n ± 2%  +2.65% (p=0.010 n=8)
_Ctx_AutoFormat-12                                                    253.2n ± 15%   233.3n ± 2%  -7.86% (p=0.000 n=8)
_Utils_GetOffer/simple-12                                             36.89n ±  3%   38.84n ± 3%  +5.27% (p=0.002 n=8)
_Utils_GetOffer/6_offers-12                                           111.2n ±  4%   112.3n ± 1%       ~ (p=0.741 n=8)
_Utils_GetOffer/1_parameter-12                                        138.2n ±  3%   137.4n ± 1%       ~ (p=0.704 n=8)
_Utils_GetOffer/2_parameters-12                                       229.4n ± 12%   212.3n ± 9%       ~ (p=0.487 n=8)
_Utils_GetOffer/3_parameters-12                                       278.3n ±  3%   274.4n ± 3%       ~ (p=0.290 n=8)
_Utils_GetOffer/10_parameters-12                                      912.9n ±  3%   895.9n ± 2%       ~ (p=0.234 n=8)
_Utils_GetOffer/6_offers_w/params-12                                  311.0n ±  2%   323.6n ± 3%  +4.05% (p=0.000 n=8)
_Utils_GetOffer/mime_extension-12                                     178.6n ±  1%   184.5n ± 5%  +3.33% (p=0.000 n=8)
_Utils_GetOffer/mime_extension#01-12                  1%   227.6n ± 5%       ~ (p=0.625 n=8)
_Utils_GetOffer/mime_extension#02-12                                  347.6n ±  2%   337.9n ± 8%       ~ (p=0.101 n=8)
_Utils_GetOffer/mime_extension#03-12                  1%   164.3n ± 1%  +1.51% (p=0.017 n=8)
_Utils_GetOffer/web_browser-12                                        135.7n ±  1%   140.9n ± 3%  +3.91% (p=0.004 n=8)
geomean                                                    211.8n       -0.24%
```

looks good -> merged

ReneWerner87 (approved):

PR #4517, opened 2026-07-10
⚡ perf(router): quick-reject routes on precomputed slash-count bounds

gaby:
# Description

Route matching linearly scans the treeStack bucket, running the segment-walking `getMatch` for every parametric candidate. In large route sets many candidates share the 3-byte bucket prefix — the GitHub API benchmark fixture has 69 GET routes in the `/re` bucket — so most of the scan cost is failed `getMatch` calls.

This PR adds a slash-count quick-reject filter in front of `getMatch`:

- At registration, `computeSlashBounds` precomputes per route how many `/` bytes a matching detection path can contain (`minSlashes`, `maxSlashes`, `maxBounded` on `routeParser`).
- Per request, the detection path's slashes are counted once in `configDependentPaths` (a single `bytes.Count` on the already-normalized path).
- `Route.match` rejects non-viable candidates with an integer compare before walking their segments. In the GitHub fixture, routes stratify cleanly by slash count (24/25/18 across counts 4/5/6 in the largest bucket), so roughly two thirds of candidate `getMatch` calls are skipped.

The bounds are conservative wherever a parameter can legitimately match a `/`: greedy (`*`/`+`) parameters, adjacent parameters (`Length==1`), and single-byte compare parts (`findParamLen`'s `IndexByte` branch has no slash guard) leave the upper bound unbounded (`maxBounded=false`); prefix (`use`) routes only apply the lower bound, mirroring the `partialCheck` they pass to `getMatch`; optional trailing segments lower the minimum per `HasOptionalSlash`. Zero values disable the filter, so hand-built routes and the pooled parser stay correct, and `RoutePatternMatch` never computes or consults the bounds, keeping its per-call cost unchanged.

No public API changes: the only interface addition is an unexported accessor (`getPathSlashes`), which external implementations inherit by embedding `DefaultCtx`.

## Changes introduced

- [x] Benchmarks: `go test -run='^$' -bench=... -benchmem -count=10` + benchstat vs `main` (linux/amd64, Intel Xeon 2.10GHz):

  | Benchmark | main | this PR | delta |
  |---|---|---|---|
  | `Benchmark_Router_GitHub_API` | 95.7µs | 82.3µs | **−14%** (B/op 8→6) |
  | `Benchmark_Router_Next` | 58.9ns | 47.7ns | **−19%** |
  | `Benchmark_Router_HandlerCustom` | 320.8ns | 289.8ns | −9.7% |
  | `Benchmark_SkipUnmatchedRoutes/Matched/with_skip` | 307.1ns | 263.0ns | −14% |
  | `Benchmark_SkipUnmatchedRoutes/405_*` | ~1.1µs | ~1.0µs | −6 to −8% |
  | `Benchmark_Router_Handler` | 101.7ns | 95.3ns | −6% |
  | `Benchmark_Route_Match`, `Benchmark_RoutePatternMatch`, `Benchmark_App_RebuildTree`, `Benchmark_Startup_Process` (time) | — | — | flat |

  Zero new per-request allocations (allocs/op unchanged everywhere). Registration memory grows by 24 bytes/route for the new fields (`Benchmark_Startup_Process` B/op +3.3%, allocs unchanged).

- [x] Changelog/What's New: internal router performance improvement, no user-facing changes.
- [x] API Longevity: no exported API added; the bounds are internal fields whose zero value disables the filter.
- [ ] Documentation Update: not needed (no user-facing behavior or API change).
- [ ] Migration Guide: not needed.
- [ ] API Alignment with Express: n/a (internal optimization).
- [ ] Examples: n/a.

New tests:

- `Test_Route_Match_SlashBoundsDifferential` — generative differential test asserting the filtered `Route.match` agrees with raw `getMatch` (outcome and captured params) across ~180 generated pattern shapes × ~170 paths × `use` on/off, with no hand-authored expectations. Validated by mutation testing: disabling any unbounded rule makes it fail.
- `Test_Route_Match_SlashBoundsConsistency` — replays the entire exhaustive matching fixture through the filtered `Route.match`.
- `Test_RouteParser_SlashBounds` — table test documenting the computed bounds per pattern shape.
- `Test_Route_Match_SlashInParams` + new fixture entries — lock in the parameter-can-contain-slash quirks end-to-end through the full request pipeline.
- Back-reference comments in `findParamLen`/`getMatch` tie the mirrored slash-consumption rules to `computeSlashBounds` from both sides.

## Type of change

## Checklist

- [x] Verified that any new dependencies are essential and have been agreed upon by the maintainers/community (no new dependencies).

https://claude.ai/code/session_01X22Bhbvpac9gQo8SuVfpDq

### DIFF
--- a/app.go
+++ b/app.go
@@ -119,6 +119,9 @@ type App struct {
 	hasRoutesRefreshed bool
 	// hasCustomCtx tracks whether app uses a custom context implementation
 	hasCustomCtx bool
+	// hasParamRoutes tracks whether any route consults the per-request slash
+	// count (rebuilt with the tree); when false the count is skipped entirely
+	hasParamRoutes bool
 }
 
 type viewsLockKey struct {
--- a/ctx.go
+++ b/ctx.go
@@ -711,9 +711,9 @@ func (c *DefaultCtx) configDependentPaths() {
 			int(c.detectionPath[2])
 	}
 
-	// Count the slashes of the final detection path once per request; route
-	// matching uses it to reject candidates without walking their segments.
-	c.pathSlashes = bytes.Count(c.detectionPath, slashDelimiterBytes)
+	// Invalidate the cached slash count of the detection path; pathSlashCount
+	// recomputes it lazily when route matching first needs it.
+	c.pathSlashes = 0
 }
 
 // Reset is a method to reset context fields by given request when to use server handlers.
@@ -899,7 +899,17 @@ func (c *DefaultCtx) getTreePathHash() int {
 	return c.treePathHash
 }
 
-func (c *DefaultCtx) getPathSlashes() int {
+// pathSlashCount lazily counts the '/' bytes of the detection path and caches
+// the result for the request; matching uses it to reject route candidates
+// without walking their segments. app is the serving App, which can differ
+// from c.app when an App value was copied. When it registers no route that
+// consults the count, counting is skipped and 0 is returned — a real detection
+// path always contains a '/', so 0 doubles as the "unknown" state that makes
+// Route.match skip the quick-reject entirely.
+func (c *DefaultCtx) pathSlashCount(app *App) int {
+	if c.pathSlashes == 0 && app.hasParamRoutes {
+		c.pathSlashes = bytes.Count(c.detectionPath, slashDelimiterBytes)
+	}
 	return c.pathSlashes
 }
 
--- a/ctx.go
+++ b/ctx.go
@@ -5,6 +5,7 @@
 package fiber
 
 import (
+	"bytes"
 	"context"
 	"crypto/tls"
 	"fmt"
@@ -72,6 +73,7 @@ type DefaultCtx struct {
 	path                   []byte               // HTTP path with the modifications by the configuration
 	detectionPath          []byte               // Route detection path
 	treePathHash           int                  // Hash of the path for the search in the tree
+	pathSlashes            int                  // Number of '/' in the detection path, used to quick-reject routes
 	indexRoute             int                  // Index of the current route
 	indexHandler           int                  // Index of the current handler
 	firstMatchIndex        int                  // Pre-resolved endpoint index from the SkipUnmatchedRoutes lookahead; -1 when unused
@@ -708,6 +710,10 @@ func (c *DefaultCtx) configDependentPaths() {
 			int(c.detectionPath[1])<<8 |
 			int(c.detectionPath[2])
 	}
+
+	// Count the slashes of the final detection path once per request; route
+	// matching uses it to reject candidates without walking their segments.
+	c.pathSlashes = bytes.Count(c.detectionPath, slashDelimiterBytes)
 }
 
 // Reset is a method to reset context fields by given request when to use server handlers.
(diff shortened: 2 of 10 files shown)


ReneWerner87:
I found one performance concern before merge.

### [P2] Avoid counting slashes for requests that cannot benefit

`configDependentPaths` now executes `bytes.Count(c.detectionPath, ...)` for every request, including apps with only static routes, where `Route.match` never reads the value.

Using the existing benchmark on an Apple M2 Pro (`-count=10`):

| Benchmark | base median | PR median | delta |
|---|---:|---:|---:|
| `Benchmark_Router_Next_Default` | 33.8 ns/op | 35.3 ns/op | **+4.4%** |

The deep/parametric benchmarks do improve clearly, so the optimization itself looks effective. The concern is paying for the additional full-path scan on the very common static-only path. Please consider computing the slash count lazily when the first parameterized candidate actually needs it, or gating it with precomputed app/bucket metadata.

### Possible argument-plumbing cleanup

The same change adds another scalar that is threaded through `next`, `nextCustom`, `resolveSkip`, and `Route.match`. An internal per-traversal value could keep those signatures smaller and provide a natural place for lazy slash counting:

```go
type routeMatchState struct {
    detectionPath string
    path          string
    values        *[maxParams]string
    pathSlashes   int // -1 means not computed yet
}

func (m *routeMatchState) slashCount() int {
    if m.pathSlashes < 0 {
        m.pathSlashes = strings.Count(m.detectionPath, "/")
    }
    return m.pathSlashes
}

func (r *Route) match(m *routeMatchState) bool
func (app *App) resolveSkip(methodInt, treeHash int, m *routeMatchState) skipResult
```

`Route.match` would call `slashCount()` only inside its parameterized-route branch. If the count must survive `Next()` re-entry or the skip lookahead, the cached value can remain owned by the request context and be referenced by this state. Since this is a hot path, I would keep the struct unexported and benchmark pointer passing/escape behavior after the refactor.

Aside from this, I did not find a matching correctness issue: `go test ./...`, the targeted race tests, and an additional randomized differential test all passed.

gaby:
Addressed in 19d5d1b — thanks for the measurement.

**P2 (static-only cost):** implemented as gate + lazy compute combined. `buildTree` now records whether any non-star parametric route exists (`app.hasParamRoutes`), and the count moved out of `configDependentPaths` into `DefaultCtx.pathSlashCount`, which computes lazily on first use per request and caches on the ctx (a detection path always contains `/`, so `0` doubles as the "not computed" state — no extra sentinel field). `Path()` rewrites invalidate the cache. Static-only apps never count: same-machine benchstat shows `Benchmark_Router_Next_Default` **−7.7%** vs the previous head, with `Router_Next`, `Router_Handler`, and `Router_GitHub_API` unchanged (p ≥ 0.05, count=8).

One subtlety this surfaced: `Route.match` treats `pathSlashes == 0` as "unknown → skip the quick-reject" rather than a real count. Server handlers and ctx pools bind the `App` instance created by `New()`, so code that copies an `App` value (the redirect middleware tests do `app := *fiber.New()`) serves through an instance whose scalar fields `buildTree` never updates — same divergence the skip indexes already tolerate. With fail-open semantics the filter degrades to a no-op there instead of 404ing param routes, and the differential test now also drives every generated case with `pathSlashes=0` to lock that in.

**On `routeMatchState`:** I kept the scalar plumbing rather than introducing the struct. The lazy-cached-count part of your sketch is what landed (owned by the request ctx so it survives `Next()` re-entry and the skip lookahead, exactly as you noted it would need to be), but wrapping `detectionPath`/`path`/`values` in a per-traversal struct means an extra pointer indirection in the scan loop and an escape-analysis risk for a stack value passed through `match` on the hottest path, for two fewer parameters. Happy to do that refactor as a follow-up if you'd prefer the smaller signatures.

---

ReneWerner87 (approved):
LGTM

Issue #3367, opened 2025-03-22
📝 [Proposal]: Always return immutable string and deprecate the `Immutable` flag


ksw2000:
### Feature Proposal Description

In Golang, using mutable strings is not intuitive. Some code in Fiber modifies strings. In the current implementation, we use the `Immutable` flag to control whether the strings returned by certain functions are immutable.

Consider the following example: 

```go
func main() {
	o := newObj()
	s := o.GetB()
	fmt.Println(s)
	o.foo()
	fmt.Println(s)
}

type obj struct {
	b []byte
}

func newObj() obj {
	return obj{
		b: []byte("hey!"),
	}
}

func (o *obj) GetB() string {
	return utils.UnsafeString(o.b)
}

func (o *obj) foo() {
	for i := 0; i < len(o.b)/2; i++ {
		o.b[i], o.b[len(o.b)-i-1] = o.b[len(o.b)-i-1], o.b[i]
	}
}
```

```sh
> go run main.go
hey!
!yeh
```

In the above example code, the returned string `s` may be changed after calling `foo`. This is not intuitive in Golang!

The overhead of `utils.UnsafeString` is lower than directly copying a string. However, this approach can cause unnecessary issues.

In the Golang standard library, returned strings are always kept immutable. When performance is a concern, a mutable byte slice is returned instead. For example, in the `bufio` package, `scanner.Bytes()` returns a mutable byte slice, while `scanner.Text()` returns an immutable string:

https://cs.opensource.google/go/go/+/refs/tags/go1.24.1:src/bufio/scan.go;l=105-116;drc=bc2124dab14fa292e18df2937037d782f7868635 

```go
// Bytes returns the most recent token generated by a call to [Scanner.Scan].
// The underlying array may point to data that will be overwritten
// by a subsequent call to Scan. It does no allocation.
func (s *Scanner) Bytes() []byte {
	return s.token
}

// Text returns the most recent token generated by a call to [Scanner.Scan]
// as a newly allocated string holding its bytes.
func (s *Scanner) Text() string {
	return string(s.token)
}
```

Similarly, Fasthttp follows the same principle. Since we wrap Fasthttp, we must adhere to this rule as well. If users are concerned about performance, they can call `fooBytes`. Otherwise, they can use `foo`, which returns an immutable string.

[...]

PR #4519, opened 2026-07-12
Feat/immutable string methods

GourangaDasSamrat:
## Summary

Addresses #3367 — `OriginalURL()`, `Path()`, `Cookies()`, and `FormValue()`
currently return a `string` that aliases the underlying fasthttp request
buffer unless `Config.Immutable` is enabled. That buffer is reused/mutated
by fasthttp, so the returned string can silently change value after the
handler runs — surprising behavior for a language where strings are
normally immutable.

This PR makes those four methods **always** return a detached, immutable
copy, regardless of `Config.Immutable`, and adds zero-allocation `*Bytes()`
siblings for callers who explicitly want handler-scoped, unsafe access to
the raw bytes (mirroring how `bufio.Scanner.Text()` vs `.Bytes()` works in
the standard library, and matching the pattern `BodyRaw()` already uses).

This follows the narrower of the two designs discussed in the issue thread
(`*Bytes()` sibling methods) rather than introducing a new wrapper type,
to avoid the larger breaking change and stay closer to the Express-like API
surface Fiber aims for.

## Changes

- **`Ctx.OriginalURL()` / `Ctx.Path()`** (`ctx.go`) — always return an
  immutable copy now. Added `Ctx.OriginalURLBytes()` and
  `Ctx.PathBytes(override ...string)` for zero-copy access.
- **`Req.Cookies()` / `Req.FormValue()`** (`req.go`) — same treatment.
  Added `Req.CookiesBytes(key)` and `Req.FormValueBytes(key)`.
- **`Req.BodyRaw()`** doc comment corrected — it already returns `[]byte`
  and was never affected by `Config.Immutable`; the comment was stale.
- **`Req.OriginalURL()` / `Res.OriginalURL()`** wrappers updated to match,
  plus matching `OriginalURLBytes()` wrappers for parity.
- **`Config.Immutable`** (`app.go`) marked `Deprecated` for these methods,
  with a pointer to the `*Bytes` alternatives. The flag is left in place
  for other code paths that still depend on it — removing it fully is a
  larger follow-up, not part of this PR.
- **Generated interface files** (`ctx_interface_gen.go`,
  `req_interface_gen.go`, `res_interface_gen.go`) hand-updated to match,
  since `ifacemaker` wasn't available in the environment this was written
  in. **Please re-run `go generate ./...` (or otherwise verify these
  against a real `ifacemaker` run) before merging.**
- Tests added for every new `*Bytes` method, plus a mutation-based test per
  changed method (`Test_Ctx_OriginalURL_Immutable`,
  `Test_Ctx_Path_Immutable`, `Test_Ctx_Cookies_Immutable`,
  `Test_Ctx_FormValue_Immutable`) that writes into the `*Bytes` slice
  after the fact and asserts the previously-returned string didn't change.

## Not in scope

Per discussion in the issue, this only covers the four methods explicitly
named there. Other string-returning methods that still alias
`app.config.Immutable` (headers, params, etc.) are untouched. A full sweep,
or the `String`/wrapper-type redesign floated later in the thread, is
left for a separate v4 discussion.

## Compatibility note

This is a **behavior change**, not just an additive one: code that relied
on `Config.Immutable = false` (the default) to get zero-copy strings from
`OriginalURL`/`Path`/`Cookies`/`FormValue` will now get a copy instead
(one extra allocation per call). Callers who need the old zero-copy
behavior should switch to the new `*Bytes()` methods.

### DIFF
--- a/app.go
+++ b/app.go
@@ -222,6 +222,14 @@ type Config struct { //nolint:govet // Aligning the struct fields is not necessa
 	// immutable fashion so that these values are available even if you return
 	// from handler.
 	//
+	// Deprecated: Immutable is being phased out. Methods such as
+	// Ctx.OriginalURL, Ctx.Path, Req.Cookies, and Req.FormValue now always
+	// return a detached, immutable string regardless of this setting; use
+	// their *Bytes counterparts (e.g. PathBytes, CookiesBytes) for
+	// zero-allocation, handler-scoped access instead. This flag is kept for
+	// backward compatibility with code paths not yet migrated and will be
+	// removed in a future major version.
+	//
 	// Default: false
 	Immutable bool `json:"immutable"`
 
--- a/ctx.go
+++ b/ctx.go
@@ -301,15 +301,29 @@ func (c *DefaultCtx) setHandlerCtx(ctx CustomCtx) {
 }
 
 // OriginalURL contains the original request URL.
-// Returned value is only valid within the handler. Do not store any references.
-// Make copies or use the Immutable setting to use the value outside the Handler.
+// The returned string is always a detached, immutable copy that is safe to
+// keep and use after the handler returns, regardless of the deprecated
+// Config.Immutable setting.
+// For zero-allocation access to the raw bytes, use OriginalURLBytes instead.
 func (c *DefaultCtx) OriginalURL() string {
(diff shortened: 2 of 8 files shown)


ReneWerner87:
I'm sorry, but that was our design decision, and we've documented it as well; that's why the app switch is there.

PR #4531, opened 2026-07-12
🐛 fix: OnPostShutdown hook receives the real error and fires once

### LINKED ISSUE #4523: 🐛 [Bug]: OnPostShutdown hooks always receive nil error and fire twice
## Bug Description

Two bugs in the shutdown hook lifecycle:

### Bug 1: OnPostShutdown hooks always receive nil error

`ShutdownWithContext` (app.go:1327) captures `err` by value at defer-registration time:

```go
var err error                                    // nil at this point
defer app.hooks.executeOnPostShutdownHooks(err)  // captures nil — Go evaluates defer args immediately
err = app.server.ShutdownWithContext(ctx)        // too late
```

Go evaluates defer arguments at registration time, not at call time. `err` is always nil when the defer is registered, so every `OnPostShutdown` hook **always receives nil** — even when shutdown fails with `context.DeadlineExceeded`.

**Fix:** Wrap in a closure:
```go
defer func() { app.hooks.executeOnPostShutdownHooks(err) }()
```

### Bug 2: OnPostShutdown hooks fire twice during graceful shutdown

`gracefulShutdown()` (listen.go:606-623) calls `app.ShutdownWithTimeout()` → `ShutdownWithContext()` whose defer fires `executeOnPostShutdownHooks(nil)` (1st call), then `gracefulShutdown()` calls `executeOnPostShutdownHooks(err)` again (2nd call).

Non-idempotent hooks (notifications, metric increments, cleanup signals) execute twice.

## How to Reproduce

```go
func main() {
    app := fiber.New()
    
    var hookError error
    app.Hooks().OnPostShutdown(func(err error) error {
        hookError = err
        fmt.Printf("OnPostShutdown called with err: %v\n", err)
        return nil
    })

    go func() { _ = app.Listen(":3000") }()
    time.Sleep(500 * time.Millisecond)
    
    ctx, cancel := context.WithTimeout(context.Background(), 1*time.Nanosecond)
    defer cancel()
    _ = app.ShutdownWithContext(ctx)  // should fail with DeadlineExceeded
    
    // hookError is nil (bug 1)
    // If GracefulContext was used, hook fires twice (bug 2)
}
```

## Expected Behavior

1. `OnPostShutdown` should receive the actual shutdown error.
2. `OnPostShutdown` should fire exactly once.

## Fiber Version

v3.4.0

## Checklist

- [x] I agree to follow Fiber's Code of Conduct.

ATKasem:
Closes #4523

Two bugs in the `OnPostShutdown` hook lifecycle:

### Bug 1 — hook always received `nil`
`ShutdownWithContext` did `defer app.hooks.executeOnPostShutdownHooks(err)`. Go evaluates deferred arguments at registration time, so `err` was always the initial `nil`, even when shutdown failed with `context.DeadlineExceeded`. Fixed by wrapping in a closure so the final error is passed.

### Bug 2 — hook fired twice on graceful shutdown
`gracefulShutdown` called `Shutdown`/`ShutdownWithTimeout` (which fire the hooks via the defer above) and **then** called `executeOnPostShutdownHooks` again, so non-idempotent hooks ran twice. Removed the duplicate calls.

### Tests
- `Test_ShutdownWithContext_PostShutdownHookReceivesError` — direct path, in-flight connection + 1ns deadline; asserts the hook receives the real error (fails before fix 1: gets `nil`).
- `Test_GracefulShutdown_PostShutdownFiresOnce` — graceful path; asserts the hook fires exactly once (fails before fix 2: fires twice).

Existing `Test_Listen_Graceful_Shutdown` still passes. `gofmt`/`go vet` clean.


gaby:
@ATKasem Linter workflow is failing.

ATKasem:
Thanks for the heads-up @gaby — fixed. The linter was flagging the intentionally-discarded shutdown errors in `gracefulShutdown` (errcheck with `check-blank`, since the error is now delivered to the `OnPostShutdown` hooks) and a `contextcheck` on the new test's `app.Listener` call. Added the appropriate `//nolint` directives with explanations. `golangci-lint run` (v2.12.2) is clean locally now.

ATKasem:
The lint job is green now ✅ (and unit tests pass across all Go/OS combos). @gaby @efectn — does the approach here look good to you, or is there anything you'd like me to change? Happy to adjust. Quick recap of the two fixes:

1. `ShutdownWithContext` now passes the *final* shutdown error to `OnPostShutdown` (the `defer ...(err)` previously captured `nil` at registration time).
2. `gracefulShutdown` no longer re-fires the hooks that `Shutdown`/`ShutdownWithTimeout` already fire, so they run exactly once.

Both paths have dedicated regression tests.

gaby (approved):

PR #4542, opened 2026-07-16
⚡ perf: adopt gofiber/utils v2.2.0 helpers across hot paths

gaby:
# Description

Adopts the performance helpers from gofiber/utils v2.2.0 (SWAR-accelerated case-insensitive matching, table-based numeric formatting) across fiber's per-request hot paths, and hoists config-constant header values out of request handlers. No public API or behavior changes — every swap was verified equivalent (details below), and candidates that would have changed semantics or regressed were explicitly rejected.

## Changes introduced

**Case-insensitive fold helpers (`EqualFold`, `HasPrefixFold`/`HasSuffixFold` — new in v2.2.0) replacing lower-then-compare:**

- `internal/schemehost`: `Match` rewritten with allocation-free fast paths — byte-identical hosts exit on a single compare, clean `host[:port]` pairs compare fold-wise with effective-port equality, and only unusual inputs (bracketed IPv6, userinfo, …) take the legacy `url.Parse` path. Runs on every CSRF Origin/Referer validation and open-redirect check.
- `middleware/csrf`: `Sec-Fetch-Site` validated via `EqualFold`; Origin/Referer lowering deferred until the trusted-origins fallback actually needs it (previously the whole Referer URL — path and query included — was copied whenever any byte was uppercase).
- `client` cookiejar: `domainMatch` is fold-based, dropping an in-place re-lowering scan and the `"."+domain` concatenation that allocated per stored domain on every lookup.
- `middleware/idempotency`: `KeepResponseHeaders` matched with `EqualFold` against a `slices.Clone` snapshot instead of a lowered-key map — removes a per-header `ToLower` allocation on every cached response (fasthttp reports canonical-case names, so the old path always allocated).
- `req.go`: Range unit checked with one `EqualFold` pass; content negotiation uses `HasPrefixFold` for wildcard and language-range prefixes; binder lowers bind keys only after the early returns that never use them.

**Numeric formatting (`FormatInt`/`FormatUint`/`AppendInt` — precomputed strings for 0–99, faster than strconv at every size):**

- `middleware/limiter` (fixed + sliding): `X-RateLimit-*`/`Retry-After` values (zero-alloc for the common small counts).
- `middleware/logger`: status code appended digit-wise via a shared `appendIntPadded` helper — no per-request `Itoa` string on the default no-color path.
- `middleware/sse`: retry field assembled in a fixed scratch and written once, no `fmt` boxing.
- `req.go` `Port()`, `res.go` `SendFile` Cache-Control/416 Content-Range, `client` query building (incidentally fixing int64 truncation through platform `int` on 32-bit builds).

**Config-constant values hoisted out of request handlers:**

- `middleware/helmet`: Strict-Transport-Security value built once in `New()` instead of `fmt.Sprintf` per secure request.
- `middleware/keyauth`: WWW-Authenticate/Proxy-Authenticate challenge built once in `New()` instead of several `fmt.Fprintf` + `strings.Join` per 401/407.

- [x] Benchmarks: `schemehost.Match` 110 ns / 2 allocs → 8 ns (identical hosts) / 48 ns (port-normalized) / 0 allocs, with per-path benchmark cases added; formatter swaps validated ~10–20 % faster than strconv at every size with equal or fewer allocations. Deliberate non-changes are documented in-code: route constraints stay on `strconv.Atoi` because `utils.ParseNativeInt` benchmarked slower for the 1–14 digit params routes actually see.
- [x] Documentation Update: not applicable — no user-facing API changes; decision rationale recorded as code comments.
- [x] Changelog/What's New: internal performance work, covered by this PR description.
- [x] Migration Guide: not applicable — no behavior or API changes.
- [x] API Alignment with Express: not applicable — no API changes.
- [x] API Longevity: no exported API touched.
- [x] Examples: not applicable — no new features.

## Type of change

## Checklist

- [x] Added or updated unit tests to validate the effectiveness of the changes or new features: reference-equivalence + fuzz tests for `schemehost.Match` (exhaustive adversarial corpus, verified against multi-million-case differential fuzzing), cookiejar domain label-boundary test, idempotency case-insensitivity + config-snapshot test, mixed-case Range/Origin/Referer/`null`-origin cases, helmet `HSTSExcludeSubdomains` output, logger line-layout and pad-loop tests (mutation-verified to fail on regression).
- [x] Ensured that new and existing unit tests pass locally with the changes (full suite, plus `-race` on touched packages).
- [x] Verified that any new dependencies are essential and have been agreed upon by the maintainers/community: no new dependencies — uses the already-bumped gofiber/utils v2.2.0.

https://claude.ai/code/session_01P2fJMT6LpHZpYhge23o2RS

### DIFF
--- a/binder/mapping.go
+++ b/binder/mapping.go
@@ -314,7 +314,6 @@ func buildFieldInfo(t reflect.Type, aliasTag string) fieldInfo {
 
 func equalFieldType(out any, kind reflect.Kind, key, aliasTag string) bool {
 	typ := reflect.TypeOf(out).Elem()
-	key = utilsstrings.ToLower(key)
 
 	if isStringKeyMap(typ) {
 		return true
@@ -323,6 +322,10 @@ func equalFieldType(out any, kind reflect.Kind, key, aliasTag string) bool {
 	if typ.Kind() != reflect.Struct {
 		return false
 	}
+
+	// Lower the key only once a struct lookup is actually needed; the early
+	// returns above never use it.
+	key = utilsstrings.ToLower(key)
 
 	cache := getFieldCache(aliasTag)
 	val, ok := cache.Load(typ)
--- a/middleware/helmet/helmet.go
+++ b/middleware/helmet/helmet.go
@@ -1,15 +1,27 @@
 package helmet
 
 import (
-	"fmt"
-
 	"github.com/gofiber/fiber/v3"
+	"github.com/gofiber/utils/v2"
 )
 
 // New creates a new middleware handler
 func New(config ...Config) fiber.Handler {
 	// Init config
 	cfg := configDefault(config...)
+
+	// The Strict-Transport-Security value only depends on config, so build it
+	// once instead of formatting it on every secure request.
+	hstsHeaderValue := ""
+	if cfg.HSTSMaxAge > 0 {
+		hstsHeaderValue = "max-age=" + utils.FormatInt(int64(cfg.HSTSMaxAge))
+		if !cfg.HSTSExcludeSubdomains {
+			hstsHeaderValue += "; includeSubDomains"
+		}
+		if cfg.HSTSPreloadEnabled {
+			hstsHeaderValue += "; preload"
+		}
+	}
 
 	// Return middleware handler
 	return func(c fiber.Ctx) error {
@@ -64,15 +76,8 @@ func New(config ...Config) fiber.Handler {
 		}
 
 		// Handle HSTS headers
-		if c.Secure() && cfg.HSTSMaxAge > 0 {
-			subdomains := ""
-			if !cfg.HSTSExcludeSubdomains {
-				subdomains = "; includeSubDomains"
-			}
-			if cfg.HSTSPreloadEnabled {
-				subdomains += "; preload"
-			}
-			c.Set(fiber.HeaderStrictTransportSecurity, fmt.Sprintf("max-age=%d%s", cfg.HSTSMaxAge, subdomains))
+		if c.Secure() && hstsHeaderValue != "" {
+			c.Set(fiber.HeaderStrictTransportSecurity, hstsHeaderValue)
 		}
 
 		// Handle Content-Security-Policy headers
--- a/middleware/helmet/helmet.go
+++ b/middleware/helmet/helmet.go
@@ -1,6 +1,8 @@
 package helmet
 
 import (
+	"strings"
+
 	"github.com/gofiber/fiber/v3"
 	"github.com/gofiber/utils/v2"
 )
@@ -14,13 +16,16 @@ func New(config ...Config) fiber.Handler {
 	// once instead of formatting it on every secure request.
 	hstsHeaderValue := ""
 	if cfg.HSTSMaxAge > 0 {
-		hstsHeaderValue = "max-age=" + utils.FormatInt(int64(cfg.HSTSMaxAge))
+		var b strings.Builder
+		b.WriteString("max-age=")                             //nolint:errcheck // strings.Builder cannot fail
+		b.WriteString(utils.FormatInt(int64(cfg.HSTSMaxAge))) //nolint:errcheck // strings.Builder cannot fail
 		if !cfg.HSTSExcludeSubdomains {
-			hstsHeaderValue += "; includeSubDomains"
+			b.WriteString("; includeSubDomains") //nolint:errcheck // strings.Builder cannot fail
 		}
 		if cfg.HSTSPreloadEnabled {
-			hstsHeaderValue += "; preload"
+			b.WriteString("; preload") //nolint:errcheck // strings.Builder cannot fail
 		}
+		hstsHeaderValue = b.String()
 	}
 
 	// Return middleware handler
--- a/client/cookiejar.go
+++ b/client/cookiejar.go
@@ -449,13 +449,15 @@ func pathMatch(reqPath, cookiePath []byte) bool {
 }
 
 // domainMatch reports whether host domain-matches the given cookie domain.
+// The comparison is ASCII case-insensitive, so neither value needs to be
+// lowercased (or allocated) beforehand.
 func domainMatch(host, domain string) bool {
-	host = utilsstrings.UnsafeToLower(host)
-
-	if host == domain {
+	if utils.EqualFold(host, domain) {
 		return true
 	}
-	return strings.HasSuffix(host, "."+domain)
+	return len(host) > len(domain) &&
+		host[len(host)-len(domain)-1] == '.' &&
+		utils.HasSuffixFold(host, domain)
 }
 
 // acceptCookieDomain enforces RFC 6265 response-domain acceptance. Trailing-dot,
--- a/client/cookiejar.go
+++ b/client/cookiejar.go
@@ -448,9 +448,11 @@ func pathMatch(reqPath, cookiePath []byte) bool {
 	return len(reqPath) > len(cookiePath) && reqPath[len(cookiePath)] == '/'
 }
 
-// domainMatch reports whether host domain-matches the given cookie domain.
-// The comparison is ASCII case-insensitive, so neither value needs to be
-// lowercased (or allocated) beforehand.
+// domainMatch reports whether host domain-matches the given cookie domain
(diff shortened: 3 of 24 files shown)


binder/mapping.go
```diff
@@ -324,6 +323,10 @@ func equalFieldType(out any, kind reflect.Kind, key, aliasTag string) bool {
 		return false
 	}
 
+	// Lower the key only once a struct lookup is actually needed; the early
+	// returns above never use it.
+	key = utilsstrings.ToLower(key)
+
```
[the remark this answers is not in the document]
gaby: False positive, it is the same function.

middleware/helmet/helmet.go
```diff
@@ -1,16 +1,28 @@
...
 // New creates a new middleware handler
 func New(config ...Config) fiber.Handler {
 	// Init config
 	cfg := configDefault(config...)
 
+	// The Strict-Transport-Security value only depends on config, so build it
+	// once instead of formatting it on every secure request.
+	hstsHeaderValue := ""
+	if cfg.HSTSMaxAge > 0 {
+		hstsHeaderValue = "max-age=" + utils.FormatInt(int64(cfg.HSTSMaxAge))
```
ReneWerner87: string builder ?

ReneWerner87:
@gaby can you share more benchmark results from the changed functions

gaby:
> Awesome, this looks like it'll make things run way smoother! 😎👍 [Explore here →](https://subweave.com/view/github.com/gofiber/fiber/pull/4542?utm_source=github&utm_medium=pr-comment&utm_campaign=autoweave&utm_content=quip&utm_term=gofiber%2Ffiber)
> 
> [Subweave map of gofiber/fiber#4542](https://subweave.com/view/github.com/gofiber/fiber/pull/4542?utm_source=github&utm_medium=pr-comment&utm_campaign=autoweave&utm_content=map&utm_term=gofiber%2Ffiber)
> 
> Maintainer? [Turn off weaves from non-maintainers →](https://subweave.com/account?repo=gofiber%2Ffiber&utm_source=github&utm_medium=pr-comment&utm_campaign=autoweave&utm_content=optout&utm_term=gofiber%2Ffiber#github) Carefully crafted by [Subweave](https://subweave.com/?utm_source=github&utm_medium=pr-comment&utm_campaign=autoweave&utm_content=footer&utm_term=gofiber%2Ffiber) · 🧶 used ~1.3m LLM tokens

@ReneWerner87 you added this?

ReneWerner87:
nope

gaby:
> nope

@ReneWerner87 Ban the account, it seems to be a spam account created last week.

middleware/helmet/helmet.go
gaby: Done in 927e0b0 — switched to `strings.Builder`. Worth noting this block runs once per `New()` (that's the point of the change: the value was previously `fmt.Sprintf`-ed on every secure request), so it's construction-time code either way.

---

gaby:
Benchmark results for the changed functions, `benchstat` with `-count=6` against the PR base (8adc5e4), linux/amd64 (shared-vCPU Xeon, so treat single-digit % deltas as noise unless p < 0.05).

**`internal/schemehost.Match`** (every CSRF Origin/Referer validation and open-redirect check). The old benchmark had a single case; the PR splits it per path, so the first row is the directly comparable one:

| case | base | PR |
|---|---|---|
| identical hosts (`example.com` vs `example.com`) | 180.1 ns, 32 B, 2 allocs | **11.0 ns, 0 B, 0 allocs** |
| default-port normalize (`example.com` vs `example.com:443`) | — (same 180 ns path) | **62.0 ns, 0 B, 0 allocs** |
| mismatch (`example.com` vs `evil.example`) | — | 47.7 ns, 0 B, 0 allocs |
| fallback (`[::1]` vs `[::1]:443`, url.Parse path) | — | 1.02 µs, 344 B, 5 allocs (unchanged legacy path, now benchmarked) |

**`middleware/logger`** — the default-format paths lose the per-request status `Itoa` string:

```
                                    │    base     │      PR                            │
Logger/DefaultFormat-4                 16 B, 2 allocs   8 B, 1 alloc   -50%  (p=0.002)
Logger/DefaultFormatDisableColors-4    16 B, 2 allocs   8 B, 1 alloc   -50%  (p=0.002)
Logger/DefaultFormatWithFiberLog-4    1111 ns → 956 ns  -14.0%         (p=0.009), 5 → 4 allocs
Logger/WithBytesAndStatus-4            589 ns → 556 ns   -5.7%         (p=0.015)
```

**Root ctx** — `Benchmark_Ctx_Range/bytes=0-700,800-1000`: 275.9 ns → 246.1 ns (**-10.8%**, p=0.026); the other Range/Accepts cases are within noise, allocations unchanged (the fold swaps there replace equal-cost scans; the win is avoiding the lowered copy on non-lowercase input, which those benchmarks don't send).

**Formatter micro-benchmarks** (`utils.FormatInt` vs `strconv.Itoa`, same machine):

| value | strconv | utils |
|---|---|---|
| 5 / 42 / 99 | 2.6–3.0 ns, 0 allocs | 2.0–2.1 ns, 0 allocs |
| 100 / 1000 | 19.8–20.8 ns, 1 alloc | 16.1–17.8 ns, 1 alloc |
| 65535 / 86400 | 23.2–23.6 ns | 20.3–20.6 ns |
| MaxInt64 | 46.3 ns | 44.3 ns |

**Where the existing macro benchmarks can't see the change, by design:**
- `Benchmark_Middleware_CSRF_Check` (2.14 µs → 2.26 µs, p=0.19, allocs identical) exercises the no-Origin/no-Referer path, so the deferred lowering and `Match` rewrite don't appear in it. The `Match` numbers above are the relevant ones.
- `Benchmark_Idempotency` uses the keep-all default, so the header-filter rewrite (which removed a `ToLower` alloc per kept header) isn't on its path; hit/skip are unchanged (p ≥ 0.9).
- `Benchmark_Limiter` (484 ns → 491 ns, p=0.18, 16 B / 2 allocs identical): the header formatting is a few ns inside ~490 ns, below noise on this machine.
- helmet/keyauth have no benchmarks; those changes move a per-request `fmt.Sprintf` (HSTS) and a multi-`Fprintf`+`Join` (challenge on 401/407) to one-time construction in `New()`, so the per-request cost is now a plain `c.Set` of a prebuilt string.

One deliberate non-change from benchmarking: route constraints stay on `strconv.Atoi` — `utils.ParseNativeInt` measured slower for the 1–14 digit params routes actually see (6.1 vs 3.6 ns at 1 digit, 19.9 vs 14.2 ns at 14), only winning near 19 digits. That's recorded as a comment in `constraint.go` so it isn't re-attempted.

Reproduce with e.g. `go test ./internal/schemehost -run XXX -bench 'Benchmark_Match|Benchmark_normalizeSchemeHost' -benchmem -count=6` on base vs head and `benchstat`.

---

gaby:
Regarding the [benchmark alert](https://github.com/gofiber/fiber/pull/4542#pullrequestreview-4719735116) on 927e0b0 (`Benchmark_NewError` 1.71×, `BenchmarkStripHopByHop_NoConnection` 1.68×): both look like CI-runner noise, not regressions from this PR.

**Neither function is in the diff.** The branch touches 24 files — no `app.go` (where `NewError` lives; it only calls `utils.StatusMessage` from the same unchanged v2.2.0 dependency) and nothing under `middleware/proxy`. 927e0b0 itself only changed `middleware/helmet`.

**Neither alert reproduces locally.** benchstat, PR base (8adc5e4) vs head (927e0b0):

```
Benchmark_NewError                    base 44.53n ±9%   head 46.71n ±11%   ~ (p=0.481 n=10)   24 B, 1 alloc both
BenchmarkStripHopByHop_NoConnection   base 1.016µ ±2%   head 1.011µ ±2%    ~ (p=0.699 n=6, interleaved runs)
```

Head measures right at the alert's own "previous" values (44.7 ns / 931 ns–1 µs range), nowhere near the alerted 76.5 ns / 1568 ns. A sequential first pass showed +7.7% on StripHopByHop, which disappeared entirely once old/new runs were interleaved — classic shared-runner phase drift, which is presumably what the CI job hit as well (its comparison baseline is also a different commit, 5b8fd75, not this PR's base). Allocation counts are byte-identical in every case, which a real code regression of this size would rarely leave untouched.

---

PR #4544, opened 2026-07-19
🔥 feat: Added support for custom binding precedence

aryan262:
# Description

Currently, Fiber's ctx.Bind().All() method binds values from request sources using a strict, hardcoded precedence (URI params -> Body -> Query -> Headers -> Cookies). If conflicting data exists across these sources, developers have no mechanism to override which source takes priority on a per-struct basis.

This PR introduces support for a new struct tag, binding_source, which allows developers to define a custom precedence order for data binding. For example: binding_source:"query,header,cookie,body,uri".

To maintain Fiber's strict zero-allocation and high-performance standards, this implementation avoids parsing struct tags or splitting strings on the hot path for every request. Instead, it utilizes a highly optimized internal caching layer (bindingPrecedenceCache via sync.Map) to process and store the precedence configuration exactly once per reflect.Type.

Fixes # (issue)

## Changes introduced

- [x] Changelog/What's New: Include a summary of the additions for the upcoming release notes.

## Type of change

## Checklist

## Commit formatting


ReneWerner87:
@aryan262 Sounds like a cool feature, but we definitely need documentation; otherwise, no one will know about it and therefore no one will use it.

aryan262:
@ReneWerner87 I am new to this repo, how should I add documentation for this PR ?

ReneWerner87:
here is the location of the docs files
https://github.com/gofiber/fiber/tree/main/docs

and this is the result https://docs.gofiber.io/

gaby:
@aryan262 The linter workflow is failing.

aryan262:
@gaby can you check now if it's failing. I'll check

gaby:
Error: bind_test.go:2952:1: File is not properly formatted (gofmt)

aryan262:
@gaby I think it was because of an extra line expected in that file. Fixed it

bind.go
```diff
@@ -428,6 +429,51 @@ func (b *Bind) Body(out any) error {
...
+					precedence = append(precedence, sourceURI)
+				case "body":
+					precedence = append(precedence, sourceBody)
+				case "query":
+					precedence = append(precedence, sourceQuery)
+				case "header":
+					precedence = append(precedence, sourceHeader)
+				case "cookie":
+					precedence = append(precedence, sourceCookie)
+				}
```
gaby: Unknown source names are silently dropped here. A typo like `binding_source:"qurey,header"` will silently bind header-only, and if *all* tokens are invalid, `precedence` stays empty and `All()` silently falls back to the default order. Since this parse runs once per type (not on the hot path), it can afford to fail loudly — please add a `default:` case that surfaces an error for unrecognized tokens rather than misconfiguring request binding silently.

Related edge cases worth deciding explicitly (and pinning with tests):
- **Omitted sources** — `binding_source:"query"` means body/header/cookie/URI are never bound at all. Plausibly intended, but it's undocumented.
- **Duplicates** — `"query,query"` runs the query binder twice.

---

bind.go
```diff
@@ -428,6 +429,51 @@ func (b *Bind) Body(out any) error {
...
+
+func getBindingPrecedence(t reflect.Type) []bindSource {
+	if cached, ok := bindingPrecedenceCache.Load(t); ok {
+		if precedence, ok := cached.([]bindSource); ok {
+			return precedence
+		}
+	}
+
+	var precedence []bindSource
+	for i := 0; i < t.NumField(); i++ {
```
gaby: The `modernize` linter (enabled in `.golangci.yml`) will likely flag this C-style loop — repo style is range-over-int, as used in `mergeStruct` in this same file:

```go
for i := range t.NumField() {
```

---

bind.go
```diff
@@ -428,6 +429,51 @@ func (b *Bind) Body(out any) error {
...
+					precedence = append(precedence, sourceBody)
+				case "query":
+					precedence = append(precedence, sourceQuery)
+				case "header":
+					precedence = append(precedence, sourceHeader)
+				case "cookie":
+					precedence = append(precedence, sourceCookie)
+				}
+			}
+			break
```
gaby: First-tag-wins is field-order dependent: if two fields declare conflicting `binding_source` tags, whichever field comes first silently wins. Consider erroring when multiple fields carry the tag. It's also worth documenting that only top-level fields are scanned — a tag inside an embedded struct's type is ignored.

---

bind.go
```diff
@@ -439,18 +485,40 @@ func (b *Bind) All(out any) error {
...
 
-	// Check if both Body and Content-Type are set
-	if len(b.ctx.Request().Body()) > 0 && len(b.ctx.RequestCtx().Request.Header.ContentType()) > 0 {
-		sources = append(sources, b.Body)
+		// Check if both Body and Content-Type are set
+		if len(b.ctx.Request().Body()) > 0 && len(b.ctx.RequestCtx().Request.Header.ContentType()) > 0 {
+			sources = append(sources, b.Body)
+		}
+		sources = append(sources, b.Query, b.Header, b.Cookie)
 	}
```
gaby: Two small things here:

1. The default path lost its pre-sized slice literal (`[]func(any) error{b.URI}` → `var sources []func(any) error` + appends). `sources := make([]func(any) error, 0, 5)` above the branch would avoid growth reallocations in both paths.
2. The body-availability check (`len(b.ctx.Request().Body()) > 0 && len(...ContentType()) > 0`) is now duplicated in both branches — extract it into a variable above the `if`.

Also note every existing `All()` caller now pays a `sync.Map.Load` + type assertion even for untagged structs. That's cheap, but since the PR description checks the Benchmarks box, it would be good to add a `BenchmarkBind_All_CustomPrecedence` and confirm the existing `BenchmarkBind_All` didn't regress.

---

docs/api/bind.md
```diff
@@ -71,6 +71,31 @@ app.Post("/users", func(c fiber.Ctx) error {
...
+#### Custom Precedence
+
+By default, the `All` method binds data in the following precedence order: `URI params -> Body -> Query -> Headers -> Cookies`.
+If you need to override this behavior, you can define a custom precedence order for your struct using the `binding_source` tag.
+
+```go title="Example"
+type CustomPrecedenceReq struct {
+    // Specify the priority for binding using the binding_source tag.
+    // In this example, query parameters take highest priority, followed by headers.
+    BindingSource struct{} `binding_source:"query,header,cookie,body,uri"`
```
gaby: The marker field leaks into serialization: this example ends with `c.JSON(req)`, so the response will include `"BindingSource":{}`. The example should at least tag it `json:"-"`:

```go
BindingSource struct{} `binding_source:"query,header,cookie,body,uri" json:"-"`
```

More broadly, requiring users to add a meaningless exported field to their request structs is an awkward shape for a public API — the implementation actually accepts the tag on any field (first one found wins), so it may be worth documenting that it can live on an existing bound field instead of a dedicated marker.

---

docs/api/bind.md
```diff
@@ -71,6 +71,31 @@ app.Post("/users", func(c fiber.Ctx) error {
...
+    req := new(CustomPrecedenceReq)
+    if err := c.Bind().All(req); err != nil {
+        return err
+    }
+    // req.Name will take the value from the query parameter if it exists, otherwise header, etc.
+    return c.JSON(req)
+})
+```
+
+> **Note:** For maximum performance, Fiber caches the precedence resolution per `reflect.Type` to reduce repeated parsing overhead.
```
gaby: This file uses Docusaurus admonitions rather than blockquotes — please switch to:

```md
:::info
For maximum performance, Fiber caches the precedence resolution per `reflect.Type` to reduce repeated parsing overhead.
:::
```

This section should also spell out the semantics of partial lists (sources omitted from the tag are not bound at all) and what happens with unrecognized source names.

---

bind_test.go
```diff
@@ -2911,3 +2911,42 @@ func BenchmarkBind_All(b *testing.B) {
 		}
 	}
 }
+
+// go test -run Test_Bind_All_CustomPrecedence
+func Test_Bind_All_CustomPrecedence(t *testing.T) {
```
gaby, line 2916: Good happy-path coverage (query beats header beats body, plus cache reuse via the second request). Missing cases worth adding:

- `uri` and `cookie` positions in a custom order — those two sources are never exercised.
- A tag that omits sources, pinning the "omitted sources are skipped" behavior.
- Invalid/unknown tokens (currently a silent fallback to default order — whatever the resolution, it needs a test pinning it).

---

gaby (changes requested):
Thanks for picking up the `binding_source` TODO — the feature is legitimately useful and the per-`reflect.Type` caching is the right design: tag parsing never runs on the hot path after the first request per type, `All()` already guarantees a pointer-to-struct before `getBindingPrecedence` runs (so `NumField()` can't panic), and the cached slice is never mutated after store, so sharing it is safe.

That said, requesting changes for a few correctness/usability gaps before this merges:

**Main issues** (details inline):

1. **Invalid tag tokens are silently dropped** — a typo silently changes which sources get bound, and an all-invalid tag silently falls back to the default order. This should fail loudly.
2. **Undocumented/untested semantics** — omitted sources are never bound, duplicates bind twice, and first-tag-wins is field-order dependent when multiple fields carry the tag.
3. **Docs example leaks the marker field** — `c.JSON(req)` will emit `"BindingSource":{}`; needs `json:"-"` at minimum.
4. **Lint/style** — the C-style `for` loop will trip the `modernize` linter, and the docs note should use a `:::info` admonition like the rest of the file.
5. **Benchmarks** — the PR checks the Benchmarks box but adds none; every existing `All()` call now pays a `sync.Map.Load`, so it'd be good to show the default path didn't regress and add one for the custom path.

None of these are structural — the core approach can stay as-is. 👍

---

bind_test.go
aryan262: @gaby Resolved all the comments, hopefully 😊

bind_test.go
```diff
@@ -2911,3 +2911,210 @@ func BenchmarkBind_All(b *testing.B) {
...
+	t.Parallel()
+	app := New()
+
+	type CustomPrecedenceReq struct {
+		Name string `binding_source:"query,header,cookie,body,uri" query:"name" header:"x-name" cookie:"c-name" json:"name" uri:"name"`
+	}
+
+	ctx := app.AcquireCtx(&fasthttp.RequestCtx{}).(*DefaultCtx)
```
gaby: These bare type assertions fail the linter — `errcheck` runs with `check-type-assertions: true` and `_test.go` files aren't excluded from it. Verified with the CI-pinned golangci-lint v2.12.2:

```
bind_test.go:2924:9: Error return value is not checked (errcheck)
	ctx := app.AcquireCtx(&fasthttp.RequestCtx{}).(*DefaultCtx)
```

Same finding on lines 2942, 2956, 2969, and 2981. The existing pattern in this file (e.g. line 1830) is:

```go
ctx := app.AcquireCtx(&fasthttp.RequestCtx{}).(*DefaultCtx) //nolint:errcheck,forcetypeassert // not needed
```

---

bind_test.go
```diff
@@ -2911,3 +2911,210 @@ func BenchmarkBind_All(b *testing.B) {
...
+// go test -v -run=^$ -bench=BenchmarkBind_All_CustomPrecedence -benchmem -count=4
+func BenchmarkBind_All_CustomPrecedence(b *testing.B) {
+	type User struct {
+		BindingSource struct{} `binding_source:"query,header,cookie,body,uri"`
+		SessionID string `json:"session_id" cookie:"session_id"`
+		Name      string `query:"name" json:"name" form:"name"`
+		Email     string `json:"email" form:"email"`
+		Role      string `header:"X-User-Role"`
+		ID        int    `uri:"id" query:"id" json:"id" form:"id"`
+	}
```
gaby, line 3090: This struct isn't gofmt-formatted (adding the longer `BindingSource` field changes the alignment of the whole block) — the `gofmt` formatter check in CI will fail on it:

```go
BindingSource struct{} `binding_source:"query,header,cookie,body,uri"`
SessionID     string   `json:"session_id" cookie:"session_id"`
Name          string   `query:"name" json:"name" form:"name"`
Email         string   `json:"email" form:"email"`
Role          string   `header:"X-User-Role"`
ID            int      `uri:"id" query:"id" json:"id" form:"id"`
```

Running `gofmt -w bind_test.go` fixes it.

---

bind.go
```diff
@@ -428,6 +429,72 @@ func (b *Bind) Body(out any) error {
...
+	sourceHeader
+	sourceCookie
+)
+
+type cachedPrecedence struct {
+	err     error
+	sources []bindSource
+}
+
+var bindingPrecedenceCache sync.Map // map[reflect.Type][]bindSource
```
gaby: Nit: this comment is stale — the cache now stores `cachedPrecedence`, not `[]bindSource`:

```go
var bindingPrecedenceCache sync.Map // map[reflect.Type]cachedPrecedence
```

---

bind.go
```diff
@@ -428,6 +429,72 @@ func (b *Bind) Body(out any) error {
...
+				case "body":
+					source = sourceBody
+				case "query":
+					source = sourceQuery
+				case "header":
+					source = sourceHeader
+				case "cookie":
+					source = sourceCookie
+				default:
+					err := fmt.Errorf("unknown binding_source %q", sourceName)
```
gaby, line 485: Nit: a trailing comma or double comma (`binding_source:"query,"`) now produces the somewhat cryptic error `unknown binding_source ""`. Consider either skipping empty tokens after `TrimSpace` or special-casing the message so the user can tell it's a stray comma rather than a bad source name.

---

bind.go
```diff
@@ -439,18 +506,45 @@ func (b *Bind) All(out any) error {
 
 	outElem := outVal.Elem()
 
-	// Precedence: URL Params -> Body -> Query -> Headers -> Cookies
-	sources := []func(any) error{b.URI}
+	sources := make([]func(any) error, 0, 5)
+	customPrecedence, err := getBindingPrecedence(outElem.Type())
+	if err != nil {
+		return err
+	}
```
gaby, line 519: Note (non-blocking): this error path bypasses `b.returnErr(...)`, so a bad `binding_source` tag surfaces as a raw error (→ 500 via the default error handler) even in auto-handling mode, unlike other binding failures which become 400s. Since a malformed tag is a programmer error rather than a client error, a 500 is arguably the *correct* behavior — just flagging it so it's a deliberate choice; might be worth a short comment here saying so.

---

bind_test.go
```diff
@@ -2911,3 +2911,210 @@ func BenchmarkBind_All(b *testing.B) {
...
+		Name          string   `query:"name"`
+	}
+
+	ctx := app.AcquireCtx(&fasthttp.RequestCtx{})
+	defer app.ReleaseCtx(ctx)
+
+	req := new(InvalidPrecedenceReq)
+	err := ctx.Bind().All(req)
+	require.Error(t, err)
+	require.Equal(t, "unknown binding_source \"invalid\"", err.Error())
```
gaby: Nit: prefer the dedicated testify assertion here (and at line 3080):

```go
require.EqualError(t, err, `unknown binding_source "invalid"`)
```

---

gaby (changes requested):
Re-reviewed at 0370b26 — this update addresses **all** of the previous round's feedback: invalid tokens now error (with negative-result caching, nice touch), multiple tags error instead of being field-order dependent, duplicates are deduped, the `modernize` loop style is fixed, the slice is pre-sized and `hasBody` extracted, the docs example no longer leaks a marker field into JSON, the note uses `:::info` with the partial-list/unknown-source semantics spelled out, and the test/benchmark coverage is now thorough. 👏

I also verified locally (fork CI hasn't run yet):

- ✅ All `Test_Bind_All*` tests pass.
- ✅ No regression on the default path — `BenchmarkBind_All` actually improved slightly vs `main` (59 → 57 allocs/op, ~2056 → ~1991 B/op), thanks to the pre-sized slice.
- ❌ **Lint will fail CI** — golangci-lint v2.12.2 (the version pinned in `.github/workflows/lint.yml`) reports 6 issues, all in `bind_test.go`: 5× `errcheck` on the bare `.(*DefaultCtx)` assertions, and 1× `gofmt` on the benchmark's `User` struct alignment. Both are mechanical fixes — details inline.

Requesting changes only for the two lint items; the remaining inline comments are optional nits (stale cache comment, empty-token error message, `require.EqualError`, and a non-blocking note about the error path skipping `returnErr`). Once lint is green this looks good to me.

---

bind_test.go
aryan262: Resolved these

bind_test.go
aryan262: Done

bind_test.go
aryan262: Formatted

gaby (approved):
LGTM

ReneWerner87 (approved):

PR #4557, opened 2026-07-24
⚡ perf: adopt gofiber/utils v2.4.0 helpers and optimize adaptor/proxy hot paths

gaby:
# Description

Adopts the helper functions introduced in [gofiber/utils v2.4.0](https://github.com/gofiber/utils) (`ParseHTTPDate`/`AppendHTTPDate`, `ParseIPv4`/`ParseIPv6`, `AppendQueryEscape`) across every hot path in the framework where they beat the stdlib/fasthttp equivalent, and follows up with targeted optimizations of the proxy and adaptor middlewares plus two correctness fixes surfaced along the way. Every conversion is either provably behavior-identical or an explicitly documented, test-pinned RFC 9110 compliance improvement.

Benchmarks below were taken on the CI-like environment (Intel Xeon @ 2.10 GHz, linux/amd64, `-count=3`); absolute numbers are noisy on that VM, the before/after deltas are what matters.

## Changes introduced

**HTTP date parsing/formatting (`req.go`, `middleware/cache`)**

- `parseHTTPDate` (backs `Fresh()`/`If-Modified-Since`) now delegates to `utils.ParseHTTPDate`: net/http.ParseTime semantics (IMF-fixdate + obsolete RFC 850/asctime, RFC 9110 §5.6.7) behind an allocation-free scalar fast path, dropping the `string(date)` conversion.
- The cache middleware writes the `Date` header with `utils.AppendHTTPDate` (direct field writer) instead of `fasthttp.AppendHTTPDate` (which delegates to `time.AppendFormat`), on both the cache-hit and cache-store paths.
- The cache middleware parses `Date` **and** `Expires` with `utils.ParseHTTPDate`, so both share one RFC 9110 acceptance set. Previously a response dating itself consistently in RFC 850 got its `Date` honored while its `Expires` was routed to the parse-error/force-revalidate path. Pinned by `Test_CacheExpiresObsoleteFormatAllowsCaching` (fails on the old parser).
- `utils.ParseHTTPDate` tolerates surrounding ASCII whitespace (RFC 9110 §5.5 OWS exclusion), which `net/http.ParseTime` does not; this is documented on `parseHTTPDate` and pinned by a `Fresh()` test case.

| Benchmark | Before | After |
|---|---|---|
| `Benchmark_Ctx_Fresh_LastModified` | 229 ns/op | 208 ns/op (−9%) |
| `AppendHTTPDate` (Date header rendering, micro) | 150 ns/op | 39 ns/op (−74%) |

**IP parsing (`req.go`, `middleware/hostauthorization`, `client`)**

- `isTrustedProxyIP` uses `utils.ParseIPv4`/`ParseIPv6` instead of `netip.ParseAddr` — identical acceptance, no error allocation for invalid client-supplied addresses.
- `net.ParseIP` is `netip.ParseAddr` minus zoned addresses (Go ≥1.17), so hostauthorization's IPv6 `Host` validation and the client cookie jar's `isIPLiteral` now screen `%` and delegate to the utils parsers: identical acceptance, zero allocations on either outcome. Pinned by tests covering zoned rejection, IPv4-mapped IPv6, bracketed literals, and malformed octets.

**Query escaping (`redirect.go`, `middleware/cache`, `middleware/paginate`)**

- `Redirect.Route` escapes query values straight into the pooled buffer via `utils.AppendQueryEscape`.
- The cache key generator escapes query keys/values in place, dropping two intermediate strings per parameter pair; the buffer bound check is restated post-append in an exactly equivalent form.
- Paginate renders page URLs through an `url.Values.Encode`-equivalent that escapes into a single buffer; output is pinned byte-for-byte against `Encode` across escaping-heavy inputs.

| Benchmark | Before | After |
|---|---|---|
| `Benchmark_Redirect_Route_WithQueries` | 275 ns/op | 256 ns/op (−7%) |
| `Benchmark_defaultKeyGenerator/multiparam` | ~893 ns/op | ~819 ns/op (−8%) |

**Adaptor middleware (`middleware/adaptor`)**

- `handlerFunc` (the `FiberHandlerFunc`/`FiberApp` serving path) builds the fasthttp request directly into the pooled `RequestCtx` instead of filling a standalone `Request` and `CopyTo`-ing it across — `Init2` only touches connection metadata and buffer flags, so in-place construction is fasthttp's own `Init` order minus the full header+body copy.
- The `noopConn` is pooled together with the `RequestCtx` and the stateless logger is shared, removing two per-request allocations.
- `resolveRemoteAddr` takes a fast path for numeric `ip:port` literals (the only form net/http servers produce) via `utils.ParseIPv4`/`ParseIPv6` + a strict decimal port parse, skipping `net.ResolveTCPAddr`'s resolver machinery; hostnames/service ports still fall back. A test pins the fast path to `ResolveTCPAddr`'s results (IPv6 zones, leading-zero forms included).
- 🐛 **Bug fix:** the header copy loops in `handlerFunc` and `HTTPMiddleware` called `Header.Set` once per value, collapsing repeated header lines (e.g. two `X-Forwarded-For` entries) to the last value. Now `Set` for the first value + `Add` for the rest; fasthttp's singleton headers (Cookie/Content-Type/…) keep their by-design single-value behavior. New tests pin both paths and fail against the old code.

| Benchmark | Before | After |
|---|---|---|
| `FiberHandlerFunc`, fresh 1 MB body per request | ~189 µs/op, 33 allocs | ~123 µs/op, 29 allocs (−35%) |
| `Benchmark_FiberHandlerFunc/No_Content` | 8 allocs/op | 7 allocs/op |

**Proxy middleware (`middleware/proxy`)**

- Connection-listed hop-by-hop headers are stripped by iterating the header value bytes in place (no string copy, no name slice); runs twice per proxied request (request + response direction). The names alias the value buffers, which is sound because fasthttp's `Del` re-slices the entry list without rewriting other entries' buffers; the fuzz harness now exercises deletion against a real header to guard exactly that invariant.
- `followRedirects` renders each hop's URL once instead of twice, removing a `URL.String()` allocation per redirect hop.

| Benchmark | Before | After |
|---|---|---|
| `BenchmarkStripHopByHop_WithConnection` | 1565 ns/op, 72 B, 3 allocs | ~958 ns/op, 0 B, 0 allocs (−39%) |
| `BenchmarkStripHopByHop_NoConnection` | ~706 ns/op, 0 allocs | unchanged |

**Deliberately not converted** (each verified): `url.PathUnescape` in extractors/static (stdlib's no-`%` fast path returns the original string with zero allocations), startup-only `net.ParseIP`/`ParseCIDR` in `app.go`, proxy SSRF checks (interoperate with `net.IP` through the resolver; dial/DNS dominates), `res.go` `encodeExtValue` (RFC 8187 attr-char set differs from any utils helper), and route constraints' `time.Parse` (user-defined layouts).

- [x] Benchmarks: before/after tables above; microbenchmarks for date rendering; equivalence-pinning tests double as behavioral benchmarks.
- [ ] Documentation Update: no user-facing API changes; behavior notes live in code comments.
- [x] Changelog/What's New: perf — utils v2.4.0 helper adoption, adaptor/proxy allocation cuts; fix — cache `Expires` RFC 9110 parsing, adaptor multi-value header preservation.
- [ ] Migration Guide: not needed — no breaking changes.

## Type of change

## Checklist

- [x] Added or updated unit tests to validate the effectiveness of the changes or new features (RFC 850 Expires caching, OWS-padded dates, `encodeQueryValues` ≡ `url.Values.Encode`, `resolveRemoteAddr` ≡ `ResolveTCPAddr`, zoned-IPv6 rejection, multi-value header preservation — the behavioral tests were verified to fail against the pre-change code).
- [x] Verified that any new dependencies are essential and have been agreed upon by the maintainers/community (none added; uses the already-bumped gofiber/utils v2.4.0).

https://claude.ai/code/session_01E5sm1LHEnvEVDRGsHWFfKK

### DIFF
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -2396,16 +2396,16 @@ func Benchmark_Client_Request_Send_ContextCancel(b *testing.B) {
 	app, ln, start := createHelperServer(b)
 
 	startedCh := make(chan struct{})
-	cancelledCh := make(chan struct{})
+	canceledCh := make(chan struct{})
 	errCh := make(chan error)
 	respCh := make(chan *Response)
 
 	app.Post("/", func(c fiber.Ctx) error {
 		startedCh <- struct{}{}
-		// Respond only after the benchmark loop has cancelled the context:
+		// Respond only after the benchmark loop has canceled the context:
 		// a fixed sleep loses the race on fast machines, letting the
 		// response beat the cancellation and fail the nil assertion below.
-		<-cancelledCh
+		<-canceledCh
 		time.Sleep(time.Millisecond) // margin for the watcher to observe Done
 		return c.Status(fiber.StatusOK).SendString("post")
 	})
@@ -2437,7 +2437,7 @@ func Benchmark_Client_Request_Send_ContextCancel(b *testing.B) {
 
 		<-startedCh // request is made, we can cancel the context now
 		cancel()
-		cancelledCh <- struct{}{} // context Done is closed; let the handler respond
+		canceledCh <- struct{}{} // context Done is closed; let the handler respond
 
 		require.Nil(b, <-respCh)
 		require.ErrorIs(b, <-errCh, ErrTimeoutOrCancel)
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -2402,11 +2402,11 @@ func Benchmark_Client_Request_Send_ContextCancel(b *testing.B) {
 
 	app.Post("/", func(c fiber.Ctx) error {
 		startedCh <- struct{}{}
-		// Respond only after the benchmark loop has canceled the context:
-		// a fixed sleep loses the race on fast machines, letting the
-		// response beat the cancellation and fail the nil assertion below.
+		// Hold the response until the loop has received Send's outcome:
+		// with the reply still pending, the client's response channel
+		// cannot be ready, so Send can only return through its
+		// ctx.Done() branch — no race against the reply.
 		<-canceledCh
-		time.Sleep(time.Millisecond) // margin for the watcher to observe Done
 		return c.Status(fiber.StatusOK).SendString("post")
 	})
 
@@ -2437,10 +2437,15 @@ func Benchmark_Client_Request_Send_ContextCancel(b *testing.B) {
 
 		<-startedCh // request is made, we can cancel the context now
 		cancel()
-		canceledCh <- struct{}{} // context Done is closed; let the handler respond
 
-		require.Nil(b, <-respCh)
-		require.ErrorIs(b, <-errCh, ErrTimeoutOrCancel)
+		// Send must return before the handler is released below, so these
+		// receives can only observe the cancellation outcome.
+		resp := <-respCh
+		err := <-errCh
+		canceledCh <- struct{}{} // release the handler; core's drainer reclaims the late response
+
+		require.Nil(b, resp)
+		require.ErrorIs(b, err, ErrTimeoutOrCancel)
 	}
 }
 
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -2396,12 +2396,17 @@ func Benchmark_Client_Request_Send_ContextCancel(b *testing.B) {
 	app, ln, start := createHelperServer(b)
 
 	startedCh := make(chan struct{})
+	cancelledCh := make(chan struct{})
 	errCh := make(chan error)
 	respCh := make(chan *Response)
 
 	app.Post("/", func(c fiber.Ctx) error {
 		startedCh <- struct{}{}
-		time.Sleep(time.Millisecond) // let cancel be called
+		// Respond only after the benchmark loop has cancelled the context:
+		// a fixed sleep loses the race on fast machines, letting the
+		// response beat the cancellation and fail the nil assertion below.
+		<-cancelledCh
+		time.Sleep(time.Millisecond) // margin for the watcher to observe Done
 		return c.Status(fiber.StatusOK).SendString("post")
 	})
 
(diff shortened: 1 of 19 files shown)


gaby:
The benchmark alert on `BenchmarkStripHopByHop_NoConnection` (1.55× vs a 1.50 threshold) looks like runner noise rather than a real regression — that benchmark exercises the path this PR did not change (no `Connection` header present, so the new in-place deletion helper does zero work), and both sides report 0 B / 0 allocs.

Interleaved A/B on one machine (`-benchtime=300000x -count=3`, main `6639aef` in a separate worktree vs this branch):

| Benchmark | main | this branch |
|---|---|---|
| `StripHopByHop_NoConnection` | ~1121 ns/op, 0 allocs | ~1130 ns/op, 0 allocs (±noise) |
| `StripHopByHop_WithConnection` | ~2650 ns/op, 72 B, 3 allocs | ~1546 ns/op, 0 B, 0 allocs |

I've queued a re-run of the failed benchmark job.

---

gaby:
Benchmark check update: the first re-run cleared the `StripHopByHop_NoConnection` alert but flagged a different benchmark, `BenchmarkDecoderedirectionMsg` (msgp-generated decode code this PR doesn't touch), at 2.20× — with MB/s halved proportionally, i.e. the whole benchmark ran on slower hardware. Interleaved local A/B against main `6639aef` (`-benchtime=2000000x -count=3`): ~61.6 ns/op on main vs ~62.8 ns/op on this branch — flat within noise, 0 allocs both.

Two runs flagging two different benchmarks in untouched code, both flat in controlled A/B, points to runner variance against the single-run baseline rather than a regression. I've queued another re-run; if it trips a third unrelated benchmark, I'll leave it for a maintainer's judgment with this data on record.

---

gaby:
The third benchmark-job failure had a different root cause than the two perf-threshold flakes: `Benchmark_Client_Request_Send_ContextCancel` failed its `require.Nil` assertion. That benchmark is pre-existing (untouched by this PR) and racy — the handler's fixed 1 ms sleep is the only window for the context cancellation to land, and on the fast AMD EPYC runner the response won the race. Fixed in 51f312f by replacing the sleep with an explicit handshake so the handler can only respond after `cancel()` has been called. The push re-runs the full suite.

---

ReneWerner87 (approved):

PR #4558, opened 2026-07-26
⚡ perf(router): flat tree index, leading-byte candidate reject, and a specialized /const/:param matcher

gaby:
# Description

Reduces the router's per-request cost. Two CPU profiles drove the work: finding a route bucket went through a `map[int][]*Route` lookup that cost ~14% of a routed request, and the bucket scan then called `Route.match` in full for every candidate, most of which cannot match.

Five changes address that, plus fixes for defects review found along the way (two of which this branch introduced and has since removed). No routing behaviour changes — the sections below say how that is established rather than asserted.

One user-visible change, in JSON serialization rather than routing: see "API longevity".

## Changes introduced

**1. Flat tree index (`routeTree`)** — replaces the request-time `treeStack` map lookup with an open-addressed index built alongside the tree: a multiply-shift plus a short linear probe, with a tree-hash miss resolving to the globals bucket inline instead of costing a second lookup. Hashes live in their own `int32` array so a probe walks 4 bytes per slot. `treeStack` remains as the build-time structure.

**2. Leading-byte candidate reject** — each route precomputes the first 8 bytes it can match, packed into a word with a mask. The scan loops discard candidates with one masked compare before touching the rest of the route or entering `match`/`getMatch`. It mirrors the first comparison each branch of `match` makes, so it can only skip work: routes without parameters against their full path, parametric routes against the constant part of their first segment (less an optional trailing slash), and wildcards or leading parameters disable it.

**3. Specialized `/const/:param` matcher** — one constant segment followed by a plain trailing parameter (`/user/keys/:key_id`) is the shape of most REST endpoints and dominated the remaining `getMatch` time. Walking the generic segment loop for it costs a pointer chase per segment, the `IsParam`/`IsOptional`/`IsLast` branches, the constraint loop and reslice bookkeeping, to reach three operations. `parseRoute` recognises the shape and `getMatch` dispatches to a straight-line version. The gate checks exactly the properties the specialized body assumes, so every other pattern keeps the generic walk.

**4. Struct layout** — `Route.match` previously touched all four cache lines of a `Route` to reject one candidate. The fields the scan reads now lead the struct; `routeParser`'s slash bounds lead that struct for the same reason and shrink to `int32` to fit.

**5. `SkipUnmatchedRoutes` static-index gate** — `resolveSkip` hashed the whole detection path to probe the static endpoint index on every request, including when no static route has that many `/`. Those routes match by exact compare, so an equal `/` count is a necessary condition; the counts present are recorded as a bitmask at build time and checked first.

### Fixes included

- **Escaped-wildcard 404.** `register` derives the star flag from the escape-stripped path but `Params` from the escaped one, so `` `/\*` `` arrives with `star` set and no params. The filter's star check sat inside its parametric branch, so such a route 404'd everything but a literal `/*`. Introduced by change 2 in this branch and fixed in 2a46864.
- **Spurious 404s during `RebuildTree`.** `buildTree` recomputed filters on the live `*Route` objects in-flight requests were scanning, publishing `prefixMask` as all-ones before narrowing it; a reader in that window rejected a route it matches. Measured 178 misses in 22.8M requests against 0 on `main`. Filters are now computed where routes are built, and `buildTree` no longer writes to routes.
- **Torn tree publication.** Holding the index by value made publishing a rebuilt tree a five-word store, letting a reader pair a new hash table with old buckets. 24 wrong answers in 20k requests against 0 on `main`. Now published by pointer in one store.
- **Probe table upper half unreachable.** `slot` took its index as `hash * mul >> 16 & mask`. Fibonacci hashing needs the index off the top of the multiply, and a fixed shift bounds it at 65535 regardless of table size, so a table past 65536 slots had an unreachable upper half and the lower half degraded into a linear scan. Now shifts by `32 - log2(size)`.

None of this makes `RebuildTree` safe against in-flight requests, and the code says so. Publishing is still an unsynchronized store, and `reuseRouteBucket` still recycles the previous build's bucket arrays — which the published tree points at — so a scan in progress can observe a bucket mid-rewrite. I briefly dropped that recycling to close the hazard and restored it on review (f1f891f): it is demonstrable at the pointer level, but ~57M requests across three probes — including middle-of-stack removals and the lookahead path that dispatches without re-matching — produced no wrong answer, and it is not worth 45 allocations a rebuild on a path already documented as unsafe.

### Housekeeping

Folded six copies of the reject predicate into one inlinable `Route.prefixRejects` used by every scan site and by the tests, so the differential tests guard the shipped expression rather than a duplicate. `pathHeadWord` now calls `swar.Load8` instead of re-deriving it.

- [x] **Benchmarks:** below. Interleaved A/B of two test binaries, medians and minima over 7–9 rounds; the whole table is from one session on one machine, since absolute figures move noticeably between runners. Noise floor measured at ±3% with identical binaries, so anything under that is not a signal.
- [ ] **Documentation Update:** none needed — no behaviour or API surface changes.
- [x] **Changelog/What's New:** router request handling is 20–67% faster depending on route shape. One note for release notes: `Route`'s JSON key order changes (below); no behaviour change.
- [ ] **Migration Guide:** not applicable — nothing to migrate.
- [ ] **API Alignment with Express:** not applicable — internal routing only.
- [x] **API Longevity:** no API was added or removed. Reordering `Route`'s fields does change the key order of a public payload, because `encoding/json` emits in declaration order and `app.GetRoutes()` returns `[]Route` — `{"method","name","path","params"}` becomes `{"params","method","name","path"}`. The same keys with the same values, reordered.

  Flagging rather than preventing it, deliberately. JSON object keys are unordered, so no conformant consumer can depend on the order; only a byte-exact snapshot or a checksum over serialized routes would notice, and nothing in this repo does that. An earlier revision pinned the old order with a `Route.MarshalJSON`, reverted in 989298b: it silently dropped any exported field added later (verified — a new tagged field produced no change in output), and made marshalling a `Route` 2.4× slower (~460ns/288B/2 allocs → ~1100ns/432B/4 allocs) by wrapping a second `json.Marshal`. Neither is a good trade for ordering nobody can rely on, inside a performance change.
- [ ] **Examples:** not applicable.

### Benchmarks

Improvements:

| Benchmark | main | this branch | |
|---|---|---|---|
| `SkipUnmatchedRoutes/Matched/without_skip` | 309.1ns | 101.7ns | **−67%** |
| `Router_HandlerCustom` | 322.3ns | 119.7ns | **−63%** |
| `SkipUnmatchedRoutes/Matched/with_skip` | 283.6ns | 123.6ns | **−56%** |
| `Router_Next` | 49.4ns | 23.6ns | **−52%** |
| `SkipUnmatchedRoutes/405_NoMiddleware/without_skip` | 1067ns | 620.9ns | **−42%** |
| `SkipUnmatchedRoutes/405_NoMiddleware/with_skip` | 1069ns | 627.7ns | **−41%** |
| `SkipUnmatchedRoutes/405_Middleware/without_skip` | 1084ns | 654.6ns | **−40%** |
| `SkipUnmatchedRoutes/405_Middleware/with_skip` | 875.3ns | 551.4ns | **−37%** |
| `Router_Handler_CaseSensitive` | 94.8ns | 64.9ns | **−32%** |
| `Route_Match` | 19.1ns | 13.1ns | **−32%** |
| `Router_Handler_StrictRouting` | 95.0ns | 65.7ns | **−31%** |
| `Router_Handler` | 95.4ns | 66.8ns | **−30%** |
| `Router_Handler_Strict_Case` | 93.7ns | 67.9ns | **−28%** |
| `Router_NotFound` | 300.1ns | 230.6ns | **−23%** |
| `SkipUnmatchedRoutes/Unmatched/without_skip` | 303.0ns | 239.5ns | **−21%** |
| `SkipUnmatchedRoutes_Deep/Matched_1param/with_skip` | 144.2ns | 119.3ns | **−17%** |
| `SkipUnmatchedRoutes_Deep/Matched_1param/without_skip` | 110.8ns | 100.7ns | **−9%** |
| `SkipUnmatchedRoutes_Deep/Matched_3param/with_skip` | 198.6ns | 185.6ns | −7% |
| `Route_Match_Star` | 2.6ns | 2.4ns | −6% |
| `SkipUnmatchedRoutes/Unmatched/with_skip` | 190.1ns | 181.1ns | −5% |
| `SkipUnmatchedRoutes_Deep/Matched_4param/with_skip` | 224.6ns | 214.9ns | −4% |
| `SkipUnmatchedRoutes_Deep/Matched_5param/with_skip` | 242.2ns | 233.9ns | −3% |
| `SkipUnmatchedRoutes_Deep/Matched_{3,4}param/without_skip` | | | −3% |
| `Router_GitHub_API_Serial` | 165.5µs | 160.8µs | −3% |

Within noise: `App_MethodNotAllowed` +1.9%, `Route_Match_Root` flat.

Regressions, all measured and none hidden:

| Benchmark | main | this branch | | why |
|---|---|---|---|---|
| `App_RebuildTree` | 17.3µs | 21.3µs | **+23%** | builds the flat index and precomputes the filters; allocations 50 → 70 per op |
| `SkipUnmatchedRoutes_Deep/Matched_health_0param/with_skip` | 77.8ns | 82.7ns | **+6%** | a short static path in a deep app: the filter has little to reject, so its per-request word pack is not repaid |
| `Startup_Process` | 440.5µs | 460.2µs | **+5%** | precomputes filters and the index |
| `Router_Handler_Unescape` | 153.2ns | 160.3ns | **+5%** | unescaping dominates; the scan is a small share of the total |
| `Router_Chain` | 180.0ns | 186.7ns | **+4%** | six handlers on one route, so `next()` runs once and the win has nothing to amortise against |
| `SkipUnmatchedRoutes_Deep/Matched_5param/without_skip` | 202.1ns | 208.4ns | **+3%** | shapes the specialization does not claim pay one predicted branch at the top of `getMatch` |

`RebuildTree` is the notable one. It is the documented-slow path — its own doc calls it performance-intensive and unsuited to production request paths. Flagging it rather than burying it.

`Benchmark_Router_GitHub_API_Serial` is new here. Both existing corpus benchmarks use `RunParallel` and swing ±40% run to run on a small machine, which cannot resolve the few-percent deltas scan changes produce; this one is deterministic. Its `main` column was measured by porting the same benchmark onto `main`.

One change was measured and **rejected**: a trailing-byte filter, keying parametric routes on their last constant segment. It was worth −17% on the corpus benchmark but cost 6 points on `Router_Handler` and ~14 on deep-param across four implementations, so it trades the general case for one workload shape.

## Type of change

## Checklist

- [x] Ensured that new and existing unit tests pass locally with the changes — full suite green, including `-race`, and `golangci-lint` clean.
- [x] Verified that any new dependencies are essential — none added; `swar` is an existing direct dependency already imported by this package.
- [x] Aimed for optimal performance with minimal allocations in the new code — request path allocates nothing new; added allocations are confined to tree construction and quantified above.
- [ ] Followed the inspiration of the Express.js framework for new functionalities — not applicable, no new user-facing functionality.
- [ ] Updated the documentation in the `/docs/` directory — not applicable.

### How behaviour preservation is established

Each optimization is only ever allowed to skip work, so each is tested against the thing it shortcuts rather than against hand-written expectations:

- `Test_Router_ScanMatchesReference` drives real requests through the handler across all 16 combinations of `CaseSensitive`, `StrictRouting`, `SkipUnmatchedRoutes` and middleware, and requires the matched route and captured parameters to equal what a linear scan of the route stack picks with every optimization switched off. No golden file, so nothing to regenerate.
- `Test_RouteParser_ConstParamDifferential` / `_Fixture` run the specialized and generic matchers against each other and compare outcome and captured parameter; `_ConstParamShape` pins which patterns the gate claims, in both directions.
- `Test_Route_PrefixFilter_Differential` / `_Fixture` / `_Rejects` assert the filter never rejects what `Route.match` accepts, over generated patterns and the exhaustive matching fixture, with routes built through real registration rather than by hand — which is what the escaped-wildcard bug slipped past the first time.
- `Test_Route_PrefixFilter_NotStale` recomputes every registered route's filter and requires it to match what is stored, closing the staleness hazard created by moving the computation out of `buildTree`.
- `Test_RouteTree_SlotSpreadsAcrossLargeTables` builds the 131072-slot table that 65536 distinct three-byte prefixes produce and requires probe starts to reach its upper half.
- `Test_ResolveSkip_StaticSlashGate` covers the `/` gate with no wildcard or root fallback registered, since either would let tier 2 rescue a broken tier 1.

Every one of these was mutation-tested: reintroducing each fixed bug, and widening or narrowing each gate, fails at least one. Two corpus gaps that mutation testing exposed are closed. Separately, a 4992-outcome replay across route/path/config combinations is byte-identical to `main`.

https://claude.ai/code/session_01FpxTzyWHtRBgLQGRnmquvF


gaby:
RebuildTree and Startup regressions are expected because of the precomputation.

router.go
```diff
@@ -999,10 +1247,3 @@ func (app *App) buildTree() *App {
 	app.hasRoutesRefreshed = false
 	return app
 }
-
-func reuseRouteBucket(prev map[int][]*Route, key, capHint int) []*Route {
```
ReneWerner87, line 1003: pls keep this

ReneWerner87:
Small follow-up idea from reading through this, not a blocker.

With `treeIndex` serving the request path, `treeStack` is now build-time only. Its remaining readers are the assignment in `buildTree` (`router.go:1239`) and, in `buildLookahead`, the bucket enumeration (`router_skip.go:135`) plus the membership test (`router_skip.go:162`).

Since `buildSkipIndexes` is called from `buildTree` anyway, the map could be passed down as a parameter instead of living on `App`. One field less to keep in sync, and it makes the build-time-only role visible in the code rather than only in the comments.

Fine as a separate change if you would rather keep this PR focused.

router.go
gaby: Restored in f1f891f.

For the record on why it went away: the recycled array is one the currently published tree still points at, so a rebuild appends over the slice a scan in progress is iterating. I could never turn that into a wrong answer though — ~57M requests across three probes, including middle-of-stack removals and the lookahead path that dispatches without re-matching, all came back clean — and `RebuildTree` is documented as unsafe against in-flight requests anyway. Paying 45 allocations a rebuild for a hazard I can't demonstrate was the wrong call.

With it back, `App_RebuildTree` goes from +35% to +23% against main, and its allocations from 115 to 70 per op. `buildRouteTree`'s doc no longer claims the buckets are freshly allocated, and now says the recycling is exactly why publishing the tree in one pointer store doesn't add up to rebuild safety.

---

gaby:
@ReneWerner87 on the `treeStack` idea — I agree with the goal, but it collides with keeping `reuseRouteBucket`, so I've left the field in place for now.

Recycling reads the *previous* build's buckets (`prevBuckets := app.treeStack[method]`), so the map has to survive between `buildTree` calls. Making it a local that's passed down would mean dropping the recycling, which is the opposite of the other review comment. What I did instead is state the role on the field, so it doesn't only live in a comment somewhere else:

```go
// Route stack divided by HTTP methods and route prefixes. Build-time only:
// requests go through treeIndex. It survives between rebuilds because
// reuseRouteBucket recycles the previous build's buckets out of it, and
// buildLookahead reads it to index the same buckets next() will scan.
treeStack []map[int][]*Route
```

There is a version that gets you both, if you want it: `treeIndex` already holds the same bucket slices, so `reuseRouteBucket` could take the previous `*routeTree` and do an exact-hit lookup (not `lookup`, which falls back to globals on a miss and would recycle the wrong bucket). Then `treeStack` becomes a local in `buildTree`, passed to `buildSkipIndexes`, and the field goes away with the recycling intact. It changes `reuseRouteBucket`'s first parameter, which is why I didn't just do it after being asked to keep the function — say the word and I'll push it here, or leave it for a separate PR as you suggested.

On the Codecov report: the one genuinely new uncovered branch was `computePrefixFilter`'s guard for routes whose first segment constrains nothing, now covered by `Test_Route_PrefixFilter_UnconstrainedShapes`. Neither shape is reachable through the public API — `register` forces a leading `/`, so a parsed route's first segment is always a constant and its parser is never empty — but a filter left enabled for either would reject paths the route matches, so the guard is worth pinning.

The two remaining lines are the `route.mount` skips in `next()` and `nextCustom()`. Those are pre-existing branches this diff only moved, and they're unreachable in a started app: `processSubAppsRoutes` strips mount routes out of `app.stack` during startup (I checked — zero survive). Covering them would mean building an app state the framework doesn't produce, so I've left them.

---

ReneWerner87 (approved):

PR #4560, opened 2026-07-28
📚 docs(ctx): document Ctx as a never-cancelable context

sunghyun1999:
## Summary

Documents `fiber.Ctx` as what it is: a `context.Context` that carries values but can never be canceled. Adds regression tests for the two post-handler storage writes that depend on that property.

No behavior change. `ctx.go` is comment-only against the base.

## Background

This PR started as an attempt to make `Deadline()`, `Done()` and `Err()` delegate to the user context installed with `SetContext` (#4335). That does not work on a pooled `Ctx`, and the reasoning is in the thread. In short: a non-nil `Done()` makes `context.propagateCancel` and `database/sql` spawn watcher goroutines that retain the pooled `*DefaultCtx` past `release()`. When such a goroutine wakes on a channel captured from the previous request and reads `Err()` back through the recycled ctx, it observes a closed channel and a nil error, and `child.cancel(false, nil, nil)` panics in a goroutine Fiber cannot recover.

`Done() == nil` is explicitly permitted by the stdlib ("Done may return nil if this context can never be canceled"), so the current behavior is a conforming implementation, not a gap. Cancellation is available through `c.Context()`.

## Changes

- **`docs/guide/context.md`**: taught `context.WithTimeout(c, 5*time.Second)` and `doSomething(c)` on the same page where it warned to use `c.Context()`. Every example now derives from `Context()`.
- **`docs/api/ctx.md`**: states the never-cancelable property once, with a caution on passing `c` to cancellation-aware calls.
- **`ctx.go` / `ctx_interface_gen.go`**: doc comments give the real reason (pooling) instead of pointing at a fasthttp issue, and say what to use instead.
- **`ctx_test.go`**: pins `Done() == nil` and `Err() == nil` even with a canceled user context installed, so re-landing delegation has to argue with a test.
- **`middleware/session`, `middleware/idempotency`**: regression tests for the storage writes that run after the handler returns. Both middlewares pass the `fiber.Ctx` to storage there, and the memory backend gates writes on `ctx.Err()`. A handler doing `defer cancel()` on a context it installed used to lose its session silently.

## Test plan

- [x] `go test -race ./...`
- [x] `golangci-lint run ./...` clean
- [x] `go generate ./...` clean
- [x] `Deadline`/`Done`/`Err` at 100% patch coverage
- [x] New tests verified to fail if `Err()` delegates again

## Follow-ups

- v4: drop `_ context.Context = (*DefaultCtx)(nil)` (`ctx.go:38`) so the compiler rejects `c` where a context is expected.

Refs #4335


gaby:
@sunghyun1999 The documentation also needs to be updated. It is located in the `docs` directory.

ctx.go
```diff
@@ -157,31 +157,22 @@ func (c *DefaultCtx) SetContext(ctx context.Context) {
 // Deadline returns the time when work done on behalf of this context
 // should be canceled. Deadline returns ok==false when no deadline is
 // set. Successive calls to Deadline return the same results.
-//
-// Due to current limitations in how fasthttp works, Deadline operates as a nop.
-// See: https://github.com/valyala/fasthttp/issues/965#issuecomment-777268945
-func (*DefaultCtx) Deadline() (time.Time, bool) {
-	return time.Time{}, false
+func (c *DefaultCtx) Deadline() (time.Time, bool) {
+	return c.Context().Deadline()
```
gaby: **Root cause of most of the issues below: `c.Context()` is not a pure read.**

```go
func (c *DefaultCtx) Context() context.Context {
	if c.fasthttp == nil { return context.Background() }
	if ctx, ok := c.fasthttp.UserValue(userContextKey).(context.Context); ok && ctx != nil { return ctx }
	ctx := context.Background()
	c.SetContext(ctx)   // <-- writes fasthttp userValues + c.isUserContextSet
	return ctx
}
```

`context.Context` requires `Deadline`/`Done`/`Err`/`Value` to be safe for concurrent use by multiple goroutines. Routing them through `Context()` makes all three *mutating* on the unset path: `SetContext` does `c.fasthttp.SetUserValue(...)` (unsynchronised append to `RequestCtx.userValues`) and `c.isUserContextSet = true`.

The old receiver-less implementations were trivially goroutine-safe; these are not. Concretely, with `-race`:

* a background `c.Err()` racing a foreground `c.Locals(k, v)` → both touch `fasthttp.RequestCtx.userValues`;
* two concurrent `c.Err()` calls → same;
* and it isn't only a race — a goroutine that outlives the handler can call `SetContext(context.Background())` on a ctx that has already been recycled, silently wiping the user context of an unrelated in-flight request.

Suggested shape: read the user value directly and return the zero values when unset, with no write-back — or cache the user context in a `DefaultCtx` field (set in `Reset`/`SetContext`, cleared in `release`) and have all three read that one field. The field version also fixes the Done/Err skew noted below, since both would then read the same snapshot.

---

ctx.go
```diff
@@ -157,31 +157,22 @@ func (c *DefaultCtx) SetContext(ctx context.Context) {
...
 // never be canceled. Successive calls to Done return the same value.
 // The close of the Done channel may happen asynchronously,
 // after the cancel function returns.
-//
-// Due to current limitations in how fasthttp works, Done operates as a nop.
-// See: https://github.com/valyala/fasthttp/issues/965#issuecomment-777268945
-func (*DefaultCtx) Done() <-chan struct{} {
-	return nil
+func (c *DefaultCtx) Done() <-chan struct{} {
+	return c.Context().Done()
```
gaby: **A non-nil `Done()` makes the standard library retain the pooled `*DefaultCtx` past handler return.**

Today `Done() == nil` is what lets the stdlib short-circuit. Once it can be non-nil, code that takes `fiber.Ctx` as a parent starts spawning watcher goroutines that hold a reference to `c` and call `c.Done()` / `c.Err()` on their own schedule:

* `context.WithCancel(c)` / `context.WithTimeout(c)` → `context.propagateCancel` spawns a goroutine selecting on `parent.Done()`;
* `db.QueryContext(c, ...)` → `database/sql`'s `awaitDone` goroutine does the same.

Both outlive the handler in the general case. `release()` then does `c.fasthttp = nil` and `Reset()` rebinds `c.fasthttp` to the *next* request, so the watcher is reading `c.fasthttp` (ctx.go:137) while the pool mutates it — a data race, and worse, cancellation can end up sourced from a completely unrelated request's context.

This is exactly the usage the PR description advertises (`db.QueryContext(c, ...)`), so it's not a hypothetical corner. Note the doc paragraph right above the new example still says the ctx "must not be used for asynchronous operations once the handler has completed" — the whole point of a live `Done()` is that third-party libraries will do precisely that, without asking.

A cached-snapshot implementation doesn't fully solve this either (the goroutine still holds `c`), but it at least removes the pooled-field read from the hot path. Worth deciding explicitly whether Fiber wants to sanction passing `c` itself to context-aware libraries, or keep steering users to `c.Context()`.

---

ctx.go
```diff
@@ -157,31 +157,22 @@ func (c *DefaultCtx) SetContext(ctx context.Context) {
...
 }
 
 // Err mirrors context.Err, returning nil until cancellation and then the terminal error value.
-//
-// Due to current limitations in how fasthttp works, Err operates as a nop.
-// See: https://github.com/valyala/fasthttp/issues/965#issuecomment-777268945
-func (*DefaultCtx) Err() error {
-	return nil
+func (c *DefaultCtx) Err() error {
+	return c.Context().Err() //nolint:wrapcheck // interface method must match context.Context signature
```
gaby: Two separate problems with `Err()` delegation.

**1. `Done()` and `Err()` can disagree, which the stdlib treats as an internal error.**

Both methods re-read a *mutable* source. `middleware/timeout` swaps the user context twice per request:

```go
tCtx, cancel := context.WithTimeout(parent, timeout)
ctx.SetContext(tCtx)
...
cancel()                 // tCtx.Done() is now closed
ctx.SetContext(parent)   // c.Err() now reports parent's error -> nil
```

Anything that captured `c.Done()` before the swap now holds a closed channel while `c.Err()` returns `nil`. `context.propagateCancel`'s watcher goroutine handles exactly that pair, and on a closed-Done/nil-Err parent it reaches `child.cancel(false, nil, nil)`, which panics with `context: internal error: missing cancel error`. That's a background goroutine, so no `recover()` anywhere in Fiber can catch it — it takes the process down.

This is the strongest argument for a single immutable snapshot per request rather than a live delegation: with one cached field, `Done()` and `Err()` can never observe different contexts.

**2. Behaviour change for in-tree consumers of `Err()`.**

`internal/storage/memory` gates writes on the caller's context:

```go
func wrapContextError(ctx context.Context, op string) error {
	if err := ctx.Err(); err != nil {
		return fmt.Errorf("memory storage %s: %w", op, err)
	}
	return nil
}
```

and `middleware/session` saves *after* `c.Next()` using the `fiber.Ctx` itself as the storage context (`Session.resolveContext()` → `s.ctx`). So a handler that follows the new documented pattern —

```go
ctx, cancel := context.WithTimeout(...)
defer cancel()
c.SetContext(ctx)
```

— has already run its `defer cancel()` by the time the session middleware saves. `c.Err()` now returns `context.Canceled`, the store rejects the write with `memory storage set: context canceled`, and the session is silently lost. This passes on `main` and fails on this branch.

This one survives the snapshot fix, because it's a genuine semantic change rather than a race: work that legitimately runs *after* the handler's context is done now sees a cancelled ctx. It needs a deliberate decision (e.g. teach the post-handler paths to use a non-cancellable context), not just a mechanical fix.

---

docs/api/ctx.md
```diff
@@ -106,6 +106,19 @@ app.Get("/", func(c fiber.Ctx) error {
...
+Cancellation and deadlines propagate through the user context:
+
+```go title="Example"
+app.Get("/", func(c fiber.Ctx) error {
+  ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+  defer cancel()
+  c.SetContext(ctx)
+
+  // db driver will now respect the 5s timeout
+  return db.QueryContext(c, "SELECT ...")
```
gaby: Four issues in this example, and it's the exact repro for the session regression noted on `ctx.go`:

1. **It doesn't compile.** `(*sql.DB).QueryContext` returns `(*sql.Rows, error)`, so `return db.QueryContext(...)` is a two-value return in a single-value position. It also leaks the `*sql.Rows` — no `defer rows.Close()`.
2. **Wrong parent.** `context.Background()` discards everything earlier middleware put on the user context — session's `sessionIDContextKey`/`middlewareContextKey`, anything from `PassLocalsToContext`, and any deadline already installed by `middleware/timeout`. It should derive from `c.Context()`.
3. **Wrong argument.** It passes `c`, not `ctx`, to `QueryContext` — which is what drags the pooled `fiber.Ctx` into `database/sql`'s `awaitDone` goroutine. Passing `ctx` is both correct and free of that problem.
4. **Silently breaks session persistence**, per the `ctx.go` comment: `defer cancel()` fires before the session middleware saves, and the memory store rejects the write.

Suggested:

```go title="Example"
  ctx, cancel := context.WithTimeout(c.Context(), 5*time.Second)
  defer cancel()

  // db driver will now respect the 5s timeout
  rows, err := db.QueryContext(ctx, "SELECT ...")
  if err != nil {
    return err
  }
  defer rows.Close()
  // ...
  return nil
})
```

i.e. derive from `c.Context()` and hand the derived `ctx` to the driver, rather than `SetContext`-ing it and passing `c`.

---

ctx_test.go
```diff
@@ -4353,12 +4360,30 @@ func Test_Ctx_Deadline(t *testing.T) {
...
+			t.Fatal("Done channel should not be closed yet")
+		default:
+		}
+
+		// After cancel, channel should be closed
+		cancel()
+		select {
+		case <-done:
+		case <-time.After(time.Second):
+			t.Fatal("Done channel should be closed after cancel")
```
gaby: `t.Fatal` is being called from the handler goroutine — `app.Test` runs handlers off the test goroutine, so `t.Fatal` here `runtime.Goexit`s the serving goroutine rather than the test. On failure you get a confusing transport error from `app.Test` instead of the assertion message.

Use `t.Error` + `return nil` inside the handler (or record into a variable and assert after `app.Test` returns).

Separately, `time.After(time.Second)` at line 4385 collides with `app.Test`'s own default 1s timeout, so a genuine failure races between the two and reports nondeterministically. A shorter bound (say 100ms) makes the intended message win.

---

gaby (commented):
The motivation here is right — the hardcoded `nil`/zero returns genuinely violate the `context.Context` contract, and #4335 is a real bug. But delegating straight to `c.Context()` trades one contract violation for a more dangerous one, because `c.Context()` is a *mutating* accessor on a *pooled, mid-request-mutable* value. I don't think this is mergeable as-is; details inline, summarised here.

### Root cause

`Context()` lazily calls `SetContext(context.Background())` when no user context is set — writing `fasthttp.RequestCtx.userValues` and `c.isUserContextSet`. So `Deadline`/`Done`/`Err` become mutating and non-goroutine-safe, which `context.Context` explicitly forbids ("safe for concurrent use by multiple goroutines"). And because the user context can be swapped mid-request (`middleware/timeout` does it twice), the three methods each observe a moving target rather than the immutable one the stdlib assumes.

### What follows from that

1. **`ctx.go:161` — goroutine safety.** A background `Err()` racing a foreground `Locals()`, or two concurrent `Err()` calls, race on `userValues`. A goroutine outliving the handler can also `SetContext(Background)` on a recycled ctx and wipe an unrelated in-flight request's context.
2. **`ctx.go:170` — pooled ctx retained past handler return.** A non-nil `Done()` stops the stdlib short-circuit, so `context.propagateCancel` and `database/sql`'s `awaitDone` now spawn watcher goroutines holding `c` and calling `c.Done()`/`c.Err()` after `release()` nils `c.fasthttp` and `Reset()` rebinds it. Triggered by the PR's own advertised `db.QueryContext(c, ...)` pattern.
3. **`ctx.go:175` — `Done()`/`Err()` can disagree.** `middleware/timeout` does `cancel(); ctx.SetContext(parent)`, leaving a captured Done closed while `Err()` reports `nil`. `context.propagateCancel`'s watcher hits `child.cancel(false, nil, nil)` on that pair and panics with `context: internal error: missing cancel error` — in a background goroutine, so nothing in Fiber can `recover()` it.
4. **`ctx.go:175` — session persistence regression.** `middleware/session` saves after `c.Next()` using the `fiber.Ctx` as its storage context, and `internal/storage/memory.wrapContextError` gates the write on `ctx.Err()`. A handler's own `defer cancel()` has already fired by then, so the save fails with `memory storage set: context canceled` and the session is silently lost. Passes on `main`, fails on this branch.
5. **`docs/api/ctx.md:113` — the new example doesn't compile** (`QueryContext` returns two values), leaks `*sql.Rows`, derives from `context.Background()` instead of `c.Context()`, and passes `c` rather than `ctx`. It is also the exact repro for (4).
6. **`ctx_test.go:4376` — `t.Fatal` from the handler goroutine**, plus a `time.After(time.Second)` that collides with `app.Test`'s 1s default.

### Two things not in the diff that also need updating

* **`ctx_interface_gen.go` was not regenerated.** Lines 31–51 still carry the `Due to current limitations in how fasthttp works, … operates as a nop` comments for all three methods — that's what pkg.go.dev shows for `fiber.Ctx`, and `go generate ./...` now produces an uncommitted diff. CI may well catch this.
* **`docs/guide/context.md:22` and `:266`** still say `Deadline`, `Done`, and `Err` "are no-ops", now contradicting the updated `docs/api/ctx.md`.

Also worth revisiting: the retained comment "Successive calls to `Done` return the same value" (and the `Deadline` equivalent) is no longer true once `SetContext` can be called mid-request.

### Suggested direction

Snapshot the user context **once** into a `DefaultCtx` field — set in `Reset`, updated in `SetContext`, cleared in `release` — and have all three methods read that single field with no write-back. That gets you the intended delegation while making the methods pure reads that can't observe a torn `Done`/`Err` pair. (A minimal variant: read `c.fasthttp.UserValue(userContextKey)` directly and return zero values when unset, skipping the `SetContext` write.)

Two caveats on that fix: it does not stop third-party goroutines from retaining the pooled `c` (finding 2 — that likely wants a docs stance of "pass `c.Context()`, not `c`", and dropping the `db.QueryContext(c, …)` recommendation), and it does not address finding 4, which is a real semantic change to post-handler work and needs a deliberate call from the maintainers.

Happy to be wrong on any of these — findings 1, 2 and 4 were checked against the branch (race detector and a session repro), and 3 follows from the `Done`/`Err` skew plus `propagateCancel`'s handling of it.

---

sunghyun1999:
## Addressing review feedback

Thanks for the thorough review @gaby! I've pushed commits to address all the points raised.

### Core fix: snapshot approach (commit 3d37acda)

Instead of delegating through `c.Context()` (which writes to fasthttp userValues on the unset path), `Deadline`/`Done`/`Err` now read from a cached `userCtx` field on `DefaultCtx`:

- **Set** in `SetContext()`
- **Cleared** in `release()`
- **No write-back**: the three methods are pure reads, goroutine-safe

This addresses:
- Goroutine safety (finding 1): no mutating reads
- Done/Err skew (finding 3): both read the same snapshot
- The `wrapcheck` lint issue is now scoped to the nil-guarded path only

### Docs fixes

- **`docs/api/ctx.md`**: example now derives from `c.Context()`, passes `ctx` (not `c`) to `QueryContext`, handles return values properly
- **`docs/guide/context.md`**: removed "no-ops" references, qualified `Done` caveat to `fasthttp.RequestCtx.Done()`
- **`ctx_interface_gen.go`**: removed stale "nop" comments, documented `DeadlineExceeded` in `Err()`
- **`docs/whats_new.md`**: updated method descriptions

### Test fixes

- Replaced `t.Fatal` with error variable (handler runs in a separate goroutine)
- Reduced `time.After` from 1s to 100ms to avoid colliding with `app.Test` default timeout

### Benchmark CI failure

The benchmark alerts (`Benchmark_NewError` 27.89→56.95 ns/op, 0→1 allocs) are unrelated to this PR. `Benchmark_NewError` tests the `NewError()` function in `app.go`, not `DefaultCtx`. These are likely GitHub Actions runner performance fluctuations.

### Remaining consideration (finding 4: session persistence)

The session regression (post-handler `Err()` returning `context.Canceled` after `defer cancel()`) is a genuine semantic change that survives the snapshot fix. This needs a deliberate decision from maintainers on whether post-handler paths should use a non-cancellable context. Happy to discuss the right approach here.

ctx.go
```diff
@@ -154,33 +154,44 @@ func (c *DefaultCtx) SetContext(ctx context.Context) {
...
-// See: https://github.com/valyala/fasthttp/issues/965#issuecomment-777268945
-func (*DefaultCtx) Err() error {
+// Err returns nil if no user context has been set or if it has not been canceled yet.
+// After cancellation it returns the context's error.
+func (c *DefaultCtx) Err() error {
+	if ctx := c.userContext(); ctx != nil {
+		return ctx.Err() //nolint:wrapcheck // interface method must match context.Context signature
+	}
 	return nil
 }
```
ReneWerner87, line 179: This is the line that breaks post-handler consumers. `middleware/session` and `middleware/idempotency` both pass the `fiber.Ctx` itself to storage after the handler has returned, and `internal/storage/memory.wrapContextError` gates the write on `ctx.Err()`. Once a handler runs its own `defer cancel()`, this returns `context.Canceled` and the write is silently dropped. Reproducer and details in the review summary.

ctx.go
```diff
@@ -154,33 +154,44 @@ func (c *DefaultCtx) SetContext(ctx context.Context) {
...
-//
-// Due to current limitations in how fasthttp works, Done operates as a nop.
-// See: https://github.com/valyala/fasthttp/issues/965#issuecomment-777268945
-func (*DefaultCtx) Done() <-chan struct{} {
+// never be canceled.
+func (c *DefaultCtx) Done() <-chan struct{} {
+	if ctx := c.userContext(); ctx != nil {
+		return ctx.Done()
+	}
 	return nil
```
ReneWerner87, line 172: Two things still unresolved here.

Returning a non-nil channel removes the stdlib short-circuit, so `context.propagateCancel` and `database/sql`'s `awaitDone` start watcher goroutines that hold `c` and call `Err()`/`Value()` after `release()` has nilled `c.fasthttp`.

And because `middleware/timeout` calls `SetContext` twice per request, successive `Done()` calls can hand back different channels. Dropping the "Successive calls to Done return the same value" sentence from the comment does not remove the assumption from `propagateCancel`, which evaluates `parent.Done()` once. When the captured channel closes but `Err()` has already been swapped back to the uncancelled parent, it reaches `child.cancel(false, nil, nil)` and panics with `context: internal error: missing cancel error` in a goroutine Fiber cannot recover.

docs/api/ctx.md
```diff
@@ -106,6 +106,25 @@ app.Get("/", func(c fiber.Ctx) error {
 })
 ```
 
+Cancellation and deadlines propagate through the user context. Pass the derived `ctx` (not `c`) to context-aware libraries to avoid retaining the pooled `fiber.Ctx`:
+
+```go title="Example"
+app.Get("/", func(c fiber.Ctx) error {
+  ctx, cancel := context.WithTimeout(c.Context(), 5*time.Second)
+  defer cancel()
+  c.SetContext(ctx)
```
ReneWerner87: This example is the exact reproducer for the session regression: the deferred `cancel()` fires before `Session.Save()` runs, and the save is dropped.

Separately, now that the example correctly passes `ctx` rather than `c` to `QueryContext`, it no longer demonstrates the delegation this PR adds. Passing the derived `ctx` directly works on `main` today. Worth either showing a case that genuinely needs `c` to carry the cancellation, or reconsidering whether the `SetContext` call belongs in the example at all.

ReneWerner87 (changes requested):
Re-checked at `25436a8`. The snapshot rework resolved several of @gaby's points: the three methods are pure reads now, the docs no longer say "no-ops", and the example passes `ctx` rather than `c`. Thanks for turning that around quickly.

One finding from that review is still open and I was able to reproduce it, so I'd hold the merge on it.

### Blocking: silent session data loss

`Session.Save()` uses the `fiber.Ctx` itself as the storage context (`resolveContext()`, `middleware/session/session.go:341`), and `internal/storage/memory.wrapContextError` gates every write on `ctx.Err()`. The save runs after the handler returns, by which point a handler's own `defer cancel()` has already fired. With this branch `c.Err()` reports that cancellation, so the write is dropped and the session is lost with no error surfacing to the caller.

The trigger is the pattern this PR documents in `docs/api/ctx.md`:

```go
app.Get("/set", func(c fiber.Ctx) error {
    ctx, cancel := context.WithTimeout(c.Context(), 5*time.Second)
    defer cancel()
    c.SetContext(ctx)

    FromContext(c).Set("hello", "world")
    return nil
})
```

Second request with the returned cookie, `origin/main` vs this branch:

```
main:   ok
branch: FAIL  session error: memory storage set: context canceled
              expected "world", got ""
```

`go test -race ./...` is green on the branch, so nothing in the suite covers this. `middleware/idempotency` has the same shape at `idempotency.go:179` (`Storage.SetWithContext(c, ...)` after the handler), and any user middleware that persists after `c.Next()` inherits it.

This is not a bug in the delegation itself, it is an ordering problem: Fiber's own middleware consumes `fiber.Ctx` as a `context.Context` after the handler returns. Making `Err()` truthful breaks those consumers first. They need to move to `context.WithoutCancel` (or Background) for post-handler storage writes before this can land.

### Still open from the previous review

- **Pooled ctx retained past handler return.** A non-nil `Done()` removes the stdlib short-circuit, so `context.propagateCancel` and `database/sql`'s `awaitDone` spawn watcher goroutines that hold `c` and call `Err()`/`Value()` after `release()` has nilled `c.fasthttp`.
- **`Done()` is no longer stable across calls.** `middleware/timeout` calls `SetContext` twice per request, so successive `Done()` calls can return different channels. The doc comment asserting stability was removed rather than the behaviour fixed; `propagateCancel` evaluates `parent.Done()` once and assumes it holds. The resulting `Done()`/`Err()` skew reaches `child.cancel(false, nil, nil)`, which panics with `context: internal error: missing cancel error` in a background goroutine that nothing in Fiber can recover.

### Nit

`go generate ./...` is not clean on this branch: it adds `userContext() context.Context` to `Ctx` and reverts the hand-edited `Err()` comment in `ctx_interface_gen.go`. The `Err()` doc has drifted between `ctx.go` and the generated interface, and the generated one is what pkg.go.dev shows. (Harmless for external implementers, `Ctx` already carries unexported methods, but the next person to regenerate gets an unrelated diff.) No workflow verifies generated files, so CI will not flag it.

### One framing note

`Done() == nil` is explicitly permitted by the stdlib ("Done may return nil if this context can never be canceled"), so the current implementation is a legal never-cancelled context rather than a contract violation. The real gap is functional: a deadline set by `middleware/timeout` does not propagate when `c` is passed along. Worth keeping in mind when weighing the cost, especially since the corrected example now passes `ctx` and not `c`, which is the guidance the docs already give and which does not need this change to work.

gaby:
@sunghyun1999 Using opus4.6 is adding bugs, it is missing key-points from the review made by opus5 😂

sunghyun1999:
@ReneWerner87 Addressed all open findings from your review at `25436a8`. Ready for re-review at `4771d43`.

### Changes

1. **Silent session data loss (blocking):** `middleware/session` and `middleware/idempotency` now use `context.WithoutCancel` for post-handler storage writes, so a handler's `defer cancel()` no longer causes silent data loss.

2. **Pooled ctx retained past handler return:** `Done()` now caches the result and snapshots the producing context in `doneCtx`. `Err()` checks `doneCtx` first, so `propagateCancel` goroutines that outlive the handler get the correct error instead of nil (which caused the `context: internal error: missing cancel error` panic).

3. **`Done()` stability across calls:** The cached channel is stable between `SetContext` calls. `SetContext` resets `doneChan`/`doneResolved` so the next `Done()` reflects the new context, but preserves `doneCtx` so goroutines watching the old channel still get the correct `Err()`.

4. **`go generate` drift:** Regenerated `ctx_interface_gen.go` so it matches the implementation.

5. **Docs example:** Removed `c.SetContext(ctx)` from the `WithTimeout` example (the exact session regression reproducer) and clarified why `SetContext` + `defer cancel()` is dangerous for post-handler middleware.

All tests pass with `-race`.

ctx.go
ReneWerner87: Fixed in `4771d43`. Both post-handler storage writes are wrapped in `context.WithoutCancel` now, and my session repro passes on this head. Putting it at the call site rather than inside `resolveContext` is the better call, the in-handler paths (`Destroy`, `Reset`, `Regenerate`) keep their cancellation that way.

Worth adding a regression test for both. Neither middleware covers the post-handler-cancel path today, so this can come back silently.

docs/api/ctx.md
ReneWerner87: Fixed in `4771d43`, `SetContext` is gone from the example and the caveat is explicit. Resolving.

ctx.go
ReneWerner87: Not resolved by `4771d43`, and the caching moved this the wrong way.

`Done()` now writes `doneResolved`, `doneChan` and `doneCtx` on first call, so it is a mutating method again. `context.Context` requires implementations to be safe for concurrent use and these are plain unsynchronised fields. Race detector over two scenarios (four concurrent `Done()`/`Err()` calls, and a handler reading `Done()` while another goroutine calls `SetContext`, which is exactly what `middleware/timeout` does):

```
origin/main   0 races
25436a89      1 races   (the fasthttp userValues read, reported earlier)
4771d43       7 races   (new: doneResolved / doneChan / doneCtx)
```

`go test -race ./...` is green, no existing test exercises this combination.

Two things the cache also does not close:

- `release()` does not clear `doneCtx`/`doneChan`, only `Reset()` does. A pooled `DefaultCtx` keeps the previous request's user context, and everything it references, alive until that ctx object is reused.
- The `missing cancel error` window is narrowed, not closed. Once `Reset()` clears `doneCtx` for the next request, a watcher goroutine waking after that falls back to `userContext()` and reads a nil error again. (Reasoned from the code, I did not build a repro for it.)

`DefaultCtx` also grows from 728 to 752 bytes for something most requests never touch.

I would drop the caching and keep the two middleware fixes, which stand on their own merit. The underlying problem is not cacheable: the contract wants `Done()` stable for the context's lifetime, `SetContext` may swap mid-request, and the object is pooled and touched concurrently. Either Fiber owns a request-scoped cancel context that `Done()` hangs off and `SetContext` does not replace, which is a design change worth its own PR, or `Done()` stays `nil`, which the stdlib explicitly permits.

middleware/idempotency/idempotency_cancel_test.go
```diff
@@ -0,0 +1,46 @@
...
+	"time"
+
+	"github.com/gofiber/fiber/v3"
+	"github.com/gofiber/fiber/v3/middleware/idempotency"
+	"github.com/stretchr/testify/require"
+)
+
+// The response is stored after the handler returns, so a handler's own deferred
+// cancel must not reach storage through Ctx.Err().
+func Test_Idempotency_Store_After_Handler_Cancel(t *testing.T) {
```
gaby: This should be part of existing test files, not a new file for 1 test.

middleware/session/session_cancel_test.go
```diff
@@ -0,0 +1,51 @@
...
+	"testing"
+	"time"
+
+	"github.com/gofiber/fiber/v3"
+	"github.com/stretchr/testify/require"
+)
+
+// The middleware saves after the handler returns, so a handler's own deferred
+// cancel must not reach storage through Ctx.Err().
+func Test_Session_Save_After_Handler_Cancel(t *testing.T) {
```
gaby: This should be part of existing test files, not a new file for 1 test.

gaby (changes requested):

ReneWerner87:
I pushed two more commits (`f2f1866`, `ff8961c`) that take `Deadline`, `Done` and `Err` back to no-ops. That reverses the premise of your PR, so you deserve the full reasoning rather than a verdict. Revert them if you disagree.

### What settled it

I had been blaming the mid-request `SetContext` swap. That was wrong. The cause is pooling, and it survives every workaround.

Any design where `Done()` hands out a user context's channel while `Err()` reads back through the pooled `*DefaultCtx` panics on recycle:

1. A handler derives `context.WithCancel(c)`. `propagateCancel` spawns a watcher that holds `c` and captures `c.Done()` from whatever context is installed at that moment.
2. The handler returns. `release()` clears the user value and nils `c.fasthttp`, the ctx goes back to the pool, `Reset` rebinds it for request N+1.
3. The captured channel closes. The watcher calls `parent.Err()`, which now reads request N+1's unset user context and returns `nil`.
4. `child.cancel(false, nil, nil)` reaches `panic("context: internal error: missing cancel error")`, in a goroutine nothing in Fiber can recover.

I did try to save it. If `middleware/timeout` skipped its `SetContext(parent)` restore there would be one `SetContext` per request and no `Done`/`Err` skew at all. It does not help: step 2 above is sufficient on its own.

### Why not Deadline-only

I nearly kept that. `net.Dialer.DialContext` reads `ctx.Deadline()`, so delegating just that would bound dials for free without arming a goroutine. I dropped it because it is the failure mode I would least want to debug: the dial gets bounded, the query does not, and the user reasonably concludes cancellation works. A uniform no-op fails immediately and points at `c.Context()`.

### A correction to my earlier review

I held "the delegation adds 2 data races" against this branch. That framing was wrong and I should not have used it. `main` already races on the same fasthttp `userValues` slice whenever `Value()` and `Locals()` run concurrently. The delegation adds no race class; it routes those existing racy reads into goroutines the stdlib spawns on the user's behalf, which nobody wrote and nobody can see. A real problem, but a different one than I described.

Also, `database/sql`'s `awaitDone` does not reach the panic. It never calls `child.cancel`. It closes the wrong request's rows instead, which is bad but not fatal.

### What the PR delivers now

Docs and tests. `ctx.go` is comment-only against the base.

- `docs/guide/context.md` taught `context.WithTimeout(c, 5*time.Second)` and `doSomething(c)` on the same page where it warned to use `c.Context()`. That contradiction, not the `Done()` semantics, is what actually produces broken handlers. Every example now derives from `Context()`.
- `Ctx` is documented once, plainly, as a context that carries values but never cancellation. `Done() == nil` is explicitly permitted by the stdlib, so this is a conforming never-cancelable context rather than a gap.
- Two regression tests for the post-handler storage writes in session and idempotency. They pass today and turn red the moment `Err()` delegates again.

The session data loss you found is real and carries this PR on its own. It ended up needing no code fix only because `Err()` is always nil again, and the tests are what keep it that way.

### Follow-ups, not for this PR

`_ context.Context = (*DefaultCtx)(nil)` at `ctx.go:38` is the root cause. A pooled, mid-request-mutable object should not satisfy `context.Context`; without that assertion the compiler rejects `c` wherever a context is expected and this entire class disappears. That is a v4 conversation and I will open an issue for it.

Thanks for sticking with this through several rounds. Happy to be argued out of any of the above.

ctx.go
ReneWerner87: Resolved by `f2f1866`. `Done()` returns nil unconditionally now, so `context.propagateCancel` short-circuits again: no watcher goroutine, nothing retains the pooled ctx past `release()`, and there is no `Done`/`Err` pair left to skew. `ctx_test.go` pins the nil so this has to be argued with a test if it ever comes back.

Full reasoning in the PR comment.

middleware/idempotency/idempotency_cancel_test.go
ReneWerner87: Moved into `idempotency_test.go` in `df3bd74b`, the separate file is gone. It now reuses the package's `validKey` constant and the `do` helper instead of bringing its own.

middleware/session/session_cancel_test.go
ReneWerner87: Moved into `middleware_test.go` in `df3bd74b`, the separate file is gone.

gaby:
@ReneWerner87 What happened to the PR? The point was to remove the "no-op" that makes these incompatible with other tools.

ReneWerner87:
Fair question, and the short answer is that I could not find a way to remove the no-op that does not introduce a crash path. Here is the reasoning, and I am glad to be overruled on it.

### Why the no-op stays

The blocker is the pooling, not the mid-request `SetContext` that we were both looking at earlier.

A non-nil `Done()` removes the short-circuit in `context.propagateCancel`, so deriving any context from `c` starts a watcher goroutine that holds the `*DefaultCtx` itself. The handler returns, `release()` clears the user value, `Reset` rebinds the ctx for the next request. If that watcher then wakes on the channel it captured from the previous request, it calls `Err()` through the recycled ctx, reads `nil`, and `child.cancel(false, nil, nil)` panics with `context: internal error: missing cancel error`, in a goroutine Fiber cannot recover. Being careful about when `SetContext` is called does not avoid it, because the recycle step alone is enough. This is findings 2 and 3 from your first review; I had assumed they were fixable and they are not.

The snapshot approach was implemented in `4771d43`. Measured on that commit: `Done()` became a mutating method, which took two concurrency scenarios from 0 races on main to 8, and `Err()` short-circuited on a snapshot that `SetContext` never cleared, so after `middleware/timeout` restores the parent both `Done() == nil` and `Err() == context.Canceled` hold at once. Reverted in `355b68e`.

### The upstream picture

Worth putting on the record here, because it bounds what any version of this can deliver. fasthttp closed the equivalent request as not implementable (valyala/fasthttp#965), and the reason is architectural rather than a missing feature:

> With fasthttp we read the full request and body before calling your handler, and we only write the response after your handler returns. In between nothing happens to the connection so it's impossible to know if the client closed the connection. You can only know if a connection is closed by doing an operation on it.
>
> `net/http` also only closes `ctx.Done()` if you try to write something to the response or if you try to read the request body. If you don't do anything it won't notice either.

So fasthttp hands the problem up to us, and there is no client-disconnect signal for us to forward no matter what we do with these three methods. The only cancellation that can ever flow through a Fiber handler is a deadline that we or the user installed, which `middleware/timeout` and `c.Context()` already deliver.

That makes this mostly a documentation problem, and the docs were genuinely part of the cause: `docs/guide/context.md` taught `context.WithTimeout(c, ...)` and `doSomething(c)` on the same page where it warned to use `c.Context()`. That contradiction is fixed here, and I think stating the boundary clearly is worth more to users than a partial implementation of it.

### On "incompatible with other tools"

With the no-op, passing `c` behaves like passing `context.Background()` with Locals attached. Nothing breaks; the work simply is not cancelable. The stdlib allows this explicitly: *"Done may return nil if this context can never be canceled."* `c.Context()` gives the full behaviour and is one call away.

I do take the point that one call away is not the same as it just working. Two ways to close that gap, neither of which belongs in this PR:

1. Fiber owns a request-scoped cancel context that `Done()` hangs off and `SetContext` does not replace. This is the correct version. It needs a channel and a watcher per request that installs a cancellable context, plus pool-reset discipline so stale watchers cannot reach a recycled ctx. fasthttp declined the equivalent cost one layer down ("Because creating a new channel for every request is just too expensive"), and #4359 is what this class of lifetime bug cost us the last time.
2. Stop implementing `context.Context` (#4563), so the compiler rejects `c` where a context is expected and the gap surfaces at build time instead of silently.

Happy to build (1) as its own PR with benchmarks if you want it, leaving this one as docs and tests. If you would rather have the delegation as it stood, I will revert my commits, though I would want the panic path written into the PR body so it is on the record as a deliberate trade.

PR #4564, opened 2026-07-28
🐛 fix(limiter): floor sub-second expiration in sliding window

nikolauspschuetz:
## Description

`SlidingWindow` computes the window length as:

```go
```

Any **positive sub-second** `ExpirationFunc` value truncates to `0`. The window weight is then

```go
weight := float64(resetInSec) / float64(expiration) // / 0  -> +Inf/NaN
rate   := int(math.Ceil(float64(e.prevHits)*weight)) + e.currHits
```

so `rate` becomes garbage and **every request — including the first — is rejected with `429`**, regardless of `Max`. A limiter configured with, say, `ExpirationFunc: func(fiber.Ctx) time.Duration { return 500*time.Millisecond }` blocks all traffic.

## Fix

Floor the truncated window to 1 second (a sub-second window is treated as a 1-second window rather than breaking):

```go
if expiration == 0 {
    expiration = 1
}
```

## Testing

Added `TestLimiterSlidingSubSecondExpiration`, which is rejected with `429` before this change and returns `200` after. `go test ./middleware/limiter/` (full package), `go vet`, and `gofmt` are clean.

<sub>Note: `FixedWindow` shares the same `uint64(...Seconds())` truncation; I scoped this PR to `SlidingWindow` since that's the one with a clean, demonstrable failure (the NaN-rate 429). Happy to follow up on the fixed-window window-sizing behavior separately if maintainers agree it's worth addressing.</sub>


gaby:
@nikolauspschuetz The linter workflow is failing

ReneWerner87 (approved):

PR #4571, opened 2026-07-30
fix(logger): sanitize user-controlled tag values to prevent log injection (#4341)

### LINKED ISSUE #4341: 🐛 [Bug]: Logger middleware — no sanitization on user-controlled log values (log injection)
### Bug Description

The logger middleware writes user-controlled values (path, URL, user-agent, body, headers) directly to the log buffer without sanitizing control characters. An attacker can inject `\r\n` sequences to forge log lines, evade SIEM detection, or corrupt audit trails.

### How to Reproduce

1. Use the logger middleware with `${path}` or `${url}` in the format
2. Send a request with `GET /admin%0d%0a200+GET+/legitimate+HTTP/1.1`
3. The logger writes two lines — the second looks like a legitimate 200 response
4. SIEM/audit tools interpret it as a separate request

### Affected Tags

`${path}`, `${url}`, `${ua}`, `${referer}`, `${ip}` (X-Forwarded-For), `${host}`, `${body}`, `${resBody}`, `${reqHeaders}`, `${queryParams}`, `${reqHeader:X}`, `${query:X}`, `${form:X}`, `${cookie:X}` — 15+ tag functions in `logger/tags.go` write user input without sanitization.

### Contrast

The `log/context.go` subsystem has proper sanitization via `writeSanitized()` that replaces control characters. The logger middleware does not use it.

### Affected Code

`middleware/logger/tags.go:108-151` (representative):
```go
TagPath: func(output Buffer, c fiber.Ctx, _ *Data, _ string) (int, error) {
    return output.WriteString(c.Path())       // No sanitization
},
TagURL: func(output Buffer, c fiber.Ctx, _ *Data, _ string) (int, error) {
    return output.WriteString(c.OriginalURL()) // No sanitization
},
TagBody: func(output Buffer, c fiber.Ctx, _ *Data, _ string) (int, error) {
    return output.Write(c.Body())             // No sanitization
},
```

### Expected Behavior

Apply the same `isControlByte` sanitization used in `log/context.go` to all tag functions that write user-controlled data.

### Fiber Version

v3 (latest main branch)

ObeeJ:
Closes #4341

## What this PR does

Adds sanitization of all user-controlled values written by the logger middleware to prevent log injection / CRLF attacks. An attacker who can control a header, query parameter, path, body, cookie, form value, or error message could otherwise inject fake log lines.

All bytes `< 0x20` (except horizontal tab `\t`) and `0x7f` are replaced with a single space `' '`. Tabs are preserved to avoid breaking tab-delimited log formats.

## Tags sanitized

`referer`, `ip`, `ips`, `host`, `path`, `url`, `ua`, `body`, `resBody`, `reqHeaders`, `queryParams`, `error`, `reqHeader:*`, `respHeader:*`, `query:*`, `form:*`, `cookie:*`, `locals:*`

## Why the previous attempt was rejected — and how each point is fixed

| Rejection reason | Fix |
|---|---|
| Duplicate logic between `sanitizeLog` / `sanitizeLogBytes` | `sanitizeLogBytes` scans bytes directly — no string conversion on clean path |
| `sanitizeLogBytes` converted to string twice on clean path | Fixed: bytes scanned first, conversion only happens when a bad byte is found |
| `TagError` not sanitized | `data.ChainErr.Error()` is now sanitized in both color and non-color paths |
| Missing unit tests | Full table-driven tests covering empty string, `\n`, `\r`, `\x00`, DEL, ANSI sequences, tab preservation |
| Fragile mutation test | Uses `copy` before call, asserts original slice unchanged after |
| No allocation proof | `testing.AllocsPerRun(100, ...)` asserts 0 allocs for both functions on clean input |
| Insufficient injection coverage | End-to-end tests via real Fiber app + `httptest` for: header, query param, path, body, error |

## Important: Fiber percent-encoding behaviour

`c.Path()` and `c.Request().URI().QueryString()` return **percent-encoded** values — `%0a` stays as the three characters `%`, `0`, `a` and is never decoded to a raw newline. The path injection test explicitly asserts this: the encoded form is preserved in the log and no raw newline appears. The real injection vectors are values that Fiber **does** decode (query param values via `c.Query()`, headers, body, cookies, form fields).

## Performance

Clean inputs (no control bytes) return the original string/slice with **zero allocations** — proven with `testing.AllocsPerRun`:

```
Benchmark_sanitizeLog_Clean   13260548   89 ns/op    0 B/op   0 allocs/op
Benchmark_sanitizeLog_Dirty    6594576  184 ns/op  128 B/op   2 allocs/op
```

## Checklist
- [x] Tests pass: `go test ./middleware/logger/...`
- [x] Zero allocations on clean input proven with `AllocsPerRun`
- [x] No new config flags or structured logging changes
- [x] Focused change — only `tags.go` and `coverage_test.go` modified

### DIFF
--- a/middleware/logger/coverage_test.go
+++ b/middleware/logger/coverage_test.go
@@ -2,6 +2,7 @@ package logger
 
 import (
 	"bytes"
+	"errors"
 	"net/http"
 	"net/http/httptest"
 	"strings"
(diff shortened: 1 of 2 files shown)


gaby:
Duplicate of #4570

PR #4572, opened 2026-07-30
fix(limiter): floor sub-second expiration in fixed window

ObeeJ:
## Description

`FixedWindow` computes the window length as:

```go
```

Any positive sub-second `ExpirationFunc` value truncates to `0`. This causes two compounding failures:

1. `e.exp = ts + 0 = ts` — the window expires on every single request, so hits never accumulate and the limiter never fires
2. `manager.set` receives the original sub-second TTL, evicting the entry from storage before the next request arrives

The result: a limiter configured with e.g. `ExpirationFunc: func(...) time.Duration { return 500*time.Millisecond }` allows unlimited requests regardless of `Max`.

## Fix

Floor both `expiration` and `expirationDuration` to 1 s when the truncated value is 0:

```go
if expiration == 0 {
    expiration = 1
    expirationDuration = time.Second
}
```

This matches the fix introduced for `SlidingWindow` in #4564, which explicitly noted:

> _Note: FixedWindow shares the same uint64(...Seconds()) truncation; I scoped this PR to SlidingWindow since that's the one with a clean, demonstrable failure. Happy to follow up on the fixed-window window-sizing behavior separately if maintainers agree it's worth addressing._

This is that follow-up.

## Testing

Added `Test_Limiter_Fixed_SubSecondExpiration` which:
- Fails before this fix (4th request returns 200 instead of 429)
- Passes after this fix
- Confirmed with `-race`

## Checklist
- [x] Tests pass: `go test ./middleware/limiter/... -race`
- [x] No API or config changes
- [x] Focused change — only `limiter_fixed.go` and `limiter_test.go` modified

### DIFF
--- a/middleware/limiter/limiter_fixed.go
+++ b/middleware/limiter/limiter_fixed.go
@@ -3,6 +3,7 @@ package limiter
 import (
 	"fmt"
 	"sync"
+	"time"
 
 	"github.com/gofiber/fiber/v3"
 	"github.com/gofiber/utils/v2"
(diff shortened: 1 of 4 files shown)


gaby:
@ObeeJ Stop duplicating pull requests, or you will be banned.

gaby:
@ReneWerner87 Another bot

PR #4573, opened 2026-07-30
fix(limiter): floor sub-second expiration in fixed window

ObeeJ:
## Description

`FixedWindow` computes the window length as:

```go
```

Any positive sub-second `ExpirationFunc` value truncates to `0`. This causes two compounding failures:

1. `e.exp = ts + 0 = ts` — the window expires on every single request, so hits never accumulate and the limiter never fires
2. `manager.set` receives the original sub-second TTL, evicting the entry from storage before the next request arrives

The result: a limiter configured with e.g. `ExpirationFunc: func(...) time.Duration { return 500*time.Millisecond }` allows unlimited requests regardless of `Max`.

## Fix

```go
if expiration == 0 {
    expiration = 1
    expirationDuration = time.Second
}
```

This is the follow-up to #4564, which explicitly noted FixedWindow has the same truncation bug.

## Testing

- `Test_Limiter_Fixed_SubSecondExpiration` fails before this fix, passes after
- Uses `testClock` for deterministic window boundaries (no wall-clock races)
- Confirmed with `-race`

## Checklist
- [x] `go test ./middleware/limiter/... -race` passes
- [x] Only `limiter_fixed.go` and `limiter_test.go` modified
- [x] No API or config changes

### DIFF
--- a/middleware/limiter/limiter_fixed.go
+++ b/middleware/limiter/limiter_fixed.go
@@ -3,6 +3,7 @@ package limiter
 import (
 	"fmt"
 	"sync"
+	"time"
 
 	"github.com/gofiber/fiber/v3"
 	"github.com/gofiber/utils/v2"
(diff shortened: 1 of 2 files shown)


gaby:
@ObeeJ Fix your bot

ObeeJ:
Hi, I noticed the PR description for #4564 mentioned FixedWindow has the same truncation bug. I've written a fix happy to share the diff here or resubmit when the timing is better.

ObeeJ:
> @ObeeJ Fix your bot

I am not a bot.. lol 
I just want to resolve an obvious bug. Would you allow me to contribute?

PR #4576, opened 2026-07-30
fix(healthcheck): handle HEAD requests per RFC 9110

### LINKED ISSUE #4574: bug(healthcheck): HEAD requests return 404 instead of matching GET status
## Bug Description

The healthcheck middleware only checks for `GET` requests:

```go
    return c.Next()
}
```

A `HEAD` request to `/livez` or `/readyz` falls through to `c.Next()` and returns 404.

## Expected Behavior

Per RFC 9110 §9.3.2, `HEAD` must respond identically to `GET` (same status code, same headers) but with no response body. A healthy probe should return 200, an unhealthy probe should return 503.

## How to Reproduce

```go
app := fiber.New()
app.Use(healthcheck.New())

req := httptest.NewRequest(http.MethodHead, "/livez", nil)
resp, _ := app.Test(req)
fmt.Println(resp.StatusCode) // prints 404, expected 200
```

## Fiber Version

v3 (latest main)

ObeeJ:
Fixes #4574

The middleware only matched `GET`, so `HEAD /livez` and `HEAD /readyz` fell through to `c.Next()` and returned 404.

**Fix:** one line — also allow `HEAD` through the method check. fasthttp strips the response body for HEAD automatically, so no other changes needed.

**Tests:** added `Test_HealthCheck_HEAD` covering healthy (200) and unhealthy (503) probes for both endpoints, with `t.Parallel()` and `-race` confirmed.

### DIFF
--- a/middleware/healthcheck/healthcheck.go
+++ b/middleware/healthcheck/healthcheck.go
@@ -20,7 +20,7 @@ func New(config ...Config) fiber.Handler {
 			return c.Next()
 		}
 
-		if c.Method() != fiber.MethodGet {
+		if c.Method() != fiber.MethodGet && c.Method() != fiber.MethodHead {
 			return c.Next()
 		}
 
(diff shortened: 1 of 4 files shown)


ReneWerner87:
why are limiter changes part of the PR ? and you forgot the doc changes for the healthcheck middleware

PR #4590, opened 2026-08-09
🐛 fix: Delete stale entries after uncacheable revalidation

Rachit-Gandhi:
# Description

When the cache middleware uses external storage, cached metadata is decoded into a temporary `item`. If `CacheInvalidator` forces a `must-revalidate` entry stale, only that temporary copy receives the adjusted expiration. The persisted metadata retains its original fresh expiration.

If the origin then returns an uncacheable response such as `Cache-Control: private`, `Cache-Control: no-cache`, or `Vary: *`, the temporary item has already been released and the existing cleanup path does not delete the stored entry. A later request can therefore receive the stale persisted response as a fresh cache hit.

This change deletes the stale metadata and separately stored body when an external-storage revalidation produces an uncacheable response. It also removes the old heap entry after successful deletion so `MaxBytes` accounting remains consistent. Normal cache hits, misses, and cacheable revalidations are unchanged.

Follow-up to #4492.

## Changes introduced

- Delete persisted stale metadata and body after an external-storage revalidation returns `private`, `no-cache`, or `Vary: *`.
- Remove the corresponding old heap entry and stored-byte accounting after successful deletion.
- Preserve the heap entry when storage deletion fails so a later eviction can retry cleanup safely.
- Avoid per-key locks, entry snapshots, deep comparisons, and additional storage body reads.
- Add parallel regression coverage for all three uncacheable response forms, metadata and body deletion failures, `MaxBytes` accounting, and the absence of an extra body `GET`.
- Documented tradeoff: unconditional deletion can race with another request storing a replacement for the same key. The replacement can be discarded, causing a temporary cache miss, and heap accounting can temporarily retain a replacement entry until eviction. Avoiding this requires per-key serialization or versioned/CAS storage, whose permanent hot-path cost is intentionally outside this focused fix.
- 
## Type of change

- [x] Bug fix (non-breaking change which corrects existing functionality)
- [x] Code consistency (non-breaking change which improves cache reliability and storage bookkeeping)

## Checklist

- [x] Followed the inspiration of the Express.js framework for new functionalities, making them similar in usage. Not applicable; this adds no new functionality or API.
- [x] Updated the documentation in the `/docs/` directory for [Fiber's documentation](https://docs.gofiber.io/). Not applicable; there is no public API or configuration change.
- [x] Ensured that new and existing unit tests pass locally with the changes (`make test`: 4,411 tests passed, 1 skipped).
- [x] Verified that any new dependencies are essential and have been agreed upon by the maintainers/community. No dependencies were added.
- [x] Aimed for optimal performance with minimal allocations in the new code; normal cache paths receive no additional allocations or storage calls.
- [x] Provided benchmarks for the new code to analyze and improve upon; cache hit and miss allocation counts remain unchanged.


gaby:
## Review status

**3 unresolved review threads** (none outdated), all against the current head `5cdc27b`. CI is green except `codecov/patch` (75%, 2 partials); `mergeable_state` is `blocked` on required reviews.

Not yet mergeable in my view: one of the bot findings (`no-store`) is valid and squarely in scope — I reproduced it — and the fix is gated on `cfg.Storage != nil`, which leaves the same bug live for the default in-memory cache. Details below.

## Remaining review comments

- **[coderabbitai — concurrent replacement can be deleted](https://github.com/gofiber/fiber/pull/4590#discussion_r3742062275)** (`middleware/cache/cache.go:542-549`). Asks that cleanup be tied to the stale entry's *generation*, not just the key, so a replacement stored by a concurrent request isn't deleted by this `deleteKey`. **Mechanism is real** and **not addressed** at head — but this is deliberate, not an oversight: the PR description explicitly documents it as an accepted tradeoff ("unconditional deletion can race with another request storing a replacement… avoiding this requires per-key serialization or versioned/CAS storage"). Worth noting the blast radius is a temporary cache *miss*, not stale content, so it's strictly less severe than the bug being fixed. Maintainer call whether to accept the tradeoff or block on it.

- **[chatgpt-codex-connector — `no-store` revalidations leave the stale entry](https://github.com/gofiber/fiber/pull/4590#discussion_r3742077307)** (`middleware/cache/cache.go:536`). **Valid, and not addressed.** `respCacheControl.hasNoStore` returns at `middleware/cache/cache.go:525-528`, before the new cleanup block, so an external-storage revalidation answered with `Cache-Control: no-store` never deletes the persisted entry, which still carries its original fresh expiration. I reproduced this against `main`:

  ```
  step1  X-Cache=miss                      (store "stale", max-age=60)
  step2  X-Cache=unreachable  body=newer   (req max-age=0 -> revalidate; origin says no-store)
  step3  X-Cache=hit          body=stale   <-- stale served as a fresh hit
  ```

  This is the same failure mode the PR fixes for `private` / `no-cache` / `Vary: *`, just reached via a fourth directive. Leaving it out makes the fix look complete when it isn't. The vary-manifest cleanup is skipped on this path too.

- **[chatgpt-codex-connector — recycled heap index can remove a replacement](https://github.com/gofiber/fiber/pull/4590#discussion_r3742077311)** (`middleware/cache/cache.go:549`). **Mechanism confirmed, not addressed.** `indexedHeap.put` does recycle freed indices (`middleware/cache/heap.go`, "Steal index from previously removed entry"), and `removeHeapEntry` validates only `entry.idx == heapIdx && entry.key == entryKey` — so if the freed `oldHeapIdx` is re-stolen by a *replacement for the same key*, the guard passes and the live replacement is unlinked while `storedBytes` is decremented. The result is a persisted-but-untracked entry and `MaxBytes` drift. Requires `MaxBytes > 0` + external storage + concurrent revalidation of one key, so it's narrow. Note this failure shape is largely pre-existing (`removeHeapEntry(key, e.heapidx)` was already computed outside the lock on `main`); what this PR changes is that the window now spans a full origin round-trip, making it materially wider.

## Additional findings

1. **The fix is gated on `cfg.Storage != nil`, but the default in-memory cache has the same bug.** `middleware/cache/cache.go:536` — `deleteRevalidatedEntry := cfg.Storage != nil && revalidate`. Every path that sets `revalidate` also sets `e = nil`, so with `Storage == nil` the block is skipped entirely and the memory entry survives untouched. The PR description's reasoning (the storage copy is a temp decode, so `e.exp = ts - 1` is lost) only covers the `CacheInvalidator` route. The **request-driven** revalidation routes — `max-age=0`/`max-age` exceeded, `min-fresh`, and `ttl == 0 && forceRevalidate` — never mutate `e.exp` at all, so the in-memory entry stays *fresh* and is served as a hit after the origin declared the response uncacheable. Reproduced against `main` with default config (no `Storage`):

   ```
   step1  X-Cache=miss                      (store "stale", max-age=60)
   step2  X-Cache=unreachable  body=newer   (req max-age=0 -> revalidate; origin says no-cache)
   step3  X-Cache=hit          body=stale   <-- stale served as a fresh hit
   ```

   Dropping the `cfg.Storage != nil` condition and keying purely on `revalidate` appears to be enough: `deleteKey` already skips the `_body` key when `cfg.Storage == nil`, and the `release`/heap calls are guarded independently. Please also add memory-storage coverage — the two new tests are external-storage-only, so this gap is invisible to the suite.

2. **A transient storage delete failure now turns a good response into a 500.** On the new `e == nil && revalidate` path, `c.Next()` has already produced a valid uncacheable response; if `deleteKey` then fails, the handler returns an error and the client gets a 500 instead of the fresh body it was entitled to. Previously this path attempted no delete, so the client was served. `Test_CacheStorage_UncacheableRevalidationDeletionErrors` locks this in as expected behavior. Note the stale entry also still remains after the failure, so the request is failed *and* the bug is un-fixed — it's surfaced, not mitigated. Worth a maintainer decision on whether cleanup failure should fail the request or just be logged.

3. **Likely conflict with #4492.** That PR rewrites this exact block: the `if e != nil { deleteKey…; mux.Lock(); removeHeapEntry(key, e.heapidx); … }` body becomes `deleteCurrentEntry(nil, …)`, `goto continueRequest` is removed, and `sync.RWMutex` becomes `sync.Mutex`. A textual conflict is certain, but the bigger issue is semantic: #4492's `deleteCurrentEntry` unconditionally does `removeHeapEntry(key, e.heapidx)` and releases `e`, so it **cannot express** the `e == nil && revalidate` case this PR needs — it would nil-deref. Whichever lands second needs real rework, not a mechanical resolution. #4492 is already `mergeable_state: dirty`. Suggest sequencing these deliberately and folding this fix into #4492's `deleteCurrentEntry` helper (e.g. an explicit heap-index parameter) rather than resolving by hand.

4. **Nit — the `heapIdx` branch is unreachable in one direction.** `middleware/cache/cache.go:538-541`:

   ```go
   heapIdx := oldHeapIdx
   if e != nil {
       heapIdx = e.heapidx
   }
   ```

   Every assignment of `revalidate = true` is paired with `e = nil`, and `e` is not reassigned before this point, so `e != nil` and `deleteRevalidatedEntry` are mutually exclusive. The code is correct but reads as if both can hold; a comment stating the invariant (or restructuring as an if/else on `e != nil`) would make the intent clearer and would keep the invariant honest if the revalidation paths are ever changed.

5. **`codecov/patch` is failing at 75%** with 2 partials on `middleware/cache/cache.go` — the uncovered partials look like the `e != nil` sub-branches of the new condition, which item 4 suggests are unreachable. Worth confirming rather than adding a test for a path that cannot be hit.

Nothing else in the diff looked wrong to me: the delete-outside-lock / heap-update-under-lock ordering matches the surrounding code, `manager.release` is correctly re-guarded with `e != nil` to avoid a nil release on the new path, retaining the heap entry on delete failure so a later eviction retries is a sound choice, and the two new tests exercise what they claim (I verified the `contextRecorderStorage` helpers they rely on already exist on `main`).

---

Rachit-Gandhi:
@gaby @ReneWerner87 @sixcolors 
I would like some direction before expanding #4590 into a cache-concurrency redesign.

The underlying bug requires cleanup after `c.Next()`: when revalidation returns `private`, `no-cache`, `no-store`, or `Vary: *`, the previously cached representation must be removed, or it can later be served as a fresh cache hit.

Because `c.Next()` runs without the global cache mutex, another request can replace the same key before cleanup:

```text
Initial: K = old entry

Request A                       Request B
---------                       ---------
reads old K
calls origin
                                revalidates K
                                stores replacement K
origin returns no-store
Delete(K)
```

Since `fiber.Storage` only provides unconditional deletion by key, request A can delete request B's replacement. The old heap index may also have been recycled for that replacement, so removing by key and heap index alone can unlink the replacement and corrupt `MaxBytes` accounting.

A per-key mutex only around `Delete(K)` would not be sufficient: if B commits first, A would acquire the mutex afterward and still delete B's entry. A mutex-only solution would need to serialize the complete same-key revalidation lifecycle:

1. Detect that origin work is required.
2. Acquire the key mutex.
3. Recheck the cache after acquiring it.
4. Call `c.Next()` while holding the key mutex.
5. Store or delete the result.
6. Release the key mutex.

That would provide strong in-process correctness and prevent same-key cache stampedes, but it is a broader behavioral and architectural change. It introduces same-key head-of-line blocking, keyed-lock lifecycle management, cancellation and early-return concerns, and interaction with #4492's lock-flow refactor. It also cannot coordinate separate Fiber processes sharing external storage.

I prototyped generation-based protection locally, including heap-generation validation and a deterministic concurrent-replacement test, but have not pushed it. It protects replacements that completed before cleanup and prevents recycled heap indices from unlinking newer entries. However, generic `fiber.Storage` still cannot atomically express “delete this key only if its generation still matches,” so a narrow check-then-delete race remains.

Could you please discuss and help me choose the intended scope?

1. Accept the narrow concurrent-replacement race in #4590 while ensuring heap accounting cannot remove a replacement.
2. Require generation tracking in #4590.
3. Require full per-key revalidation serialization.
4. Merge the focused stale-entry fix and handle the concurrency architecture separately, possibly alongside or after #4492.

ReneWerner87:
@Rachit-Gandhi Answering the scope question you raised on 2026-09-07: the narrow option. Accept the check-then-delete race against external storage, whose worst case is one extra origin hit and never stale content served, and keep the in-process heap accounting correct. `fiber.Storage` has no conditional delete, and full per-key serialization is architecture that would have collided with #4492, now closed.

Two notes on getting there.

**The red CI was not yours.** Four key literals in `cache_test.go` predated the `v2` prefix from `cacheKeyVersion` (cache.go:63), so the suite failed on main rather than on this change. Pushed as 211d256e; the unit jobs are green and only `codecov/patch` is left.

**The heap removal is not safe yet, and the obvious guard does not cover it.** `removeHeapEntry` checks `entry.idx == heapIdx && entry.key == entryKey`, which looks sufficient but is not, because `indexedHeap.put` recycles the index of a removed entry (`idx = h.entries[:n+1][n].idx`). With two concurrent revalidations of the same key:

1. A captures `oldHeapIdx = 5`, releases the entry and goes to the origin.
2. B finishes a store: its `removeHeapEntry(key, 5)` frees slot 5, and its `put` takes slot 5 back for the replacement.
3. A finds the response uncacheable and calls `removeHeapEntry(key, 5)`. Same key, same index, so both checks pass and B's live replacement loses its accounting.

I tried the small way out first, skipping the removal when `e == nil` on the grounds that the slot is no longer this request's. That breaks `Test_CacheStorage_UncacheableRevalidationDeletesStaleEntry`, which arms a sentinel `del` error precisely to pin the removal, and pinning it is fair: a left-behind heap entry costs a later eviction a delete on a key that is already gone.

That leaves a generation on the heap entry, handed out by `put` and checked on removal, so a recycled index cannot pass for the original. It touches `heapEntry`, `put`'s signature and the places that carry `heapidx`, which is more than this PR has needed so far. Whether that belongs here, or in its own change with this one narrowed to the `e != nil` case, is your call.

PR #4593, opened 2026-08-09
🔥 feat: support per-bind JSON decoders

### LINKED ISSUE #2858: 📝 [Proposal]: Add options for active DisallowUnknownFields in the function BodyParser

Add options object for active the DisallowUnknownFields in the BodyParser for more strict parsing to a struct in cases where is required.

BodyParser middleware for JSON in express permit you to add different options that affect the parsing

Confirmed

### API Stability

DisallowUnknownFields is a native functions of the Decoder from native Go package `json`

### Feature Examples

```go
type Cat struct {
     Name
}


POST /cats
{
  “Name”: “lasagna”
   “Age”: 3
}

Response:
Error the request body not match exactly the struct of Cat
```

```
p := new(Person)
if err := c.BodyParser(p,Options{DisallowUnknownFields:true}); err != nil {
   return err
}
```

### Checklist:

RubenPari:
## Summary

Adds `Bind().WithJSONDecoder(...)` so callers can select a JSON decoder for a single bind operation without changing the app-level decoder.

This enables strict parsing patterns such as `DisallowUnknownFields` on individual binds (issue #2858), while leaving the global decoder untouched.

## Changes

- `WithJSONDecoder` on the bind API
- Tests for per-bind decoder selection / strict unknown-field behavior
- Docs in `docs/api/bind.md`

## Linked issue

Closes #2858

## Checklist

- [x] `make audit`
- [x] `make generate`
- [x] `make format`
- [x] `make lint` — 0 issues
- [x] `make test` — passed

### DIFF
--- a/bind.go
+++ b/bind.go
@@ -39,6 +39,7 @@ var bindPool = sync.Pool{
 // With WithAutoHandling(), parsing failures set HTTP 400 and return *Error instead.
 type Bind struct {
 	ctx                   Ctx
+	jsonDecoder           utils.JSONUnmarshal
 	shouldSkipErrHandling bool
 	shouldSkipValidation  bool
 }
@@ -128,8 +129,17 @@ func releasePooledBinder[T interface{ Reset() }](pool *sync.Pool, bind T) {
 
 func (b *Bind) release() {
 	b.ctx = nil
+	b.jsonDecoder = nil
 	b.shouldSkipErrHandling = true
 	b.shouldSkipValidation = false
+}
+
+// WithJSONDecoder selects the JSON decoder for this bind operation.
+// A nil decoder uses the application decoder.
+func (b *Bind) WithJSONDecoder(decoder utils.JSONUnmarshal) *Bind {
+	b.jsonDecoder = decoder
+
+	return b
 }
 
 // WithoutAutoHandling If you want to handle binder errors manually, you can use `WithoutAutoHandling`.
@@ -300,7 +310,10 @@ func (b *Bind) Query(out any) error {
 // Returns *BindError on parse failure (manual mode) or *Error with status 400 (auto-handling mode).
 func (b *Bind) JSON(out any) error {
 	bind := binder.GetFromThePool[*binder.JSONBinding](&binder.JSONBinderPool)
-	bind.JSONDecoder = b.ctx.App().Config().JSONDecoder
+	bind.JSONDecoder = b.jsonDecoder
+	if bind.JSONDecoder == nil {
+		bind.JSONDecoder = b.ctx.App().Config().JSONDecoder
+	}
 
 	defer releasePooledBinder(&binder.JSONBinderPool, bind)
 
--- a/bind.go
+++ b/bind.go
@@ -12,6 +12,7 @@ import (
 	"github.com/gofiber/fiber/v3/binder"
 	"github.com/gofiber/fiber/v3/internal/nilerror"
 	"github.com/gofiber/schema"
+	"github.com/gofiber/utils/v2"
 )
 
 // CustomBinder An interface to register custom binders.
--- a/bind.go
+++ b/bind.go
@@ -134,7 +134,7 @@ func (b *Bind) release() {
 	b.shouldSkipValidation = false
 }
 
-// WithJSONDecoder selects the JSON decoder for this bind operation.
+// WithJSONDecoder selects the JSON decoder for the current bind chain.
 // A nil decoder uses the application decoder.
 func (b *Bind) WithJSONDecoder(decoder utils.JSONUnmarshal) *Bind {
 	b.jsonDecoder = decoder
--- a/bind_test.go
+++ b/bind_test.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"mime/multipart"
 	"net/http"
 	"net/http/httptest"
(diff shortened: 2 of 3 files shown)


gaby:
**Status:** 3 review threads still unresolved (none outdated). The head commit is still `efb2d11`, the same commit all three threads were filed against, so nothing has been pushed to address them. The core change is small and correct in isolation, but the branch is 158 commits behind `main` and **will not compile after a rebase** (see first finding) — not mergeable as-is.

## Remaining review comments

- **`bind_test.go` — make the pool-cleanup assertion deterministic** (coderabbitai, unresolved, not outdated). Asks to call `b.release()` and assert on `b` directly instead of relying on `AcquireBind()` handing back the same object. Not addressed. The claim is valid, and actually understates the problem: `bindPool.New` returns `&Bind{shouldSkipErrHandling: true}`, whose zero state (`ctx == nil`, `jsonDecoder == nil`, `shouldSkipErrHandling == true`, `shouldSkipValidation == false`) is *exactly* what the test asserts — so `Test_AcquireReleaseBind` passes whether or not `release()` clears anything. Adding `t.Parallel()` in this PR makes the pool even less likely to return the same object. The suggested rewrite is the right fix.

- **`bind.go` — override persists on the per-request `*Bind`** (copilot-pull-request-reviewer, unresolved, not outdated). Not addressed. The factual part is correct and I verified it: `DefaultCtx.Bind()` (`ctx.go:880-886`) caches one `*Bind` per `Ctx`, so `WithJSONDecoder` sticks for every later `c.Bind().JSON/Body/All` in the same request. It does **not** leak across requests — `release()` clears `jsonDecoder` (`bind.go:133`) and `ctx.go:748-750` releases the bind on `Ctx` release. I'd push back on the suggested remedy though: returning a shallow copy would defeat the `bindPool`/per-`Ctx` caching and would make this method behave differently from `WithAutoHandling`, `WithoutAutoHandling` and `SkipValidation`, which are all sticky in exactly the same way. The cheap correct fix is wording: the godoc at `bind.go:139` says "for this bind operation" and the PR description says "a single bind operation", which is wrong. `SkipValidation`'s "for the current bind chain" is the accurate phrasing; the docs prose in `docs/api/bind.md` already uses it. Either align the godoc/PR description, or state explicitly that `WithJSONDecoder(nil)` resets it.

- **`bind_test.go` — add a non-leakage regression test** (copilot-pull-request-reviewer, unresolved, not outdated). Not addressed; no test asserts what a second `c.Bind()` call in the same request does. Whichever semantics the maintainers settle on above, a test pinning it down is worth having.

## Additional findings

- **Rebase will break the build.** The base commit `a2a6d29` had `"github.com/gofiber/utils/v2"` imported in `bind.go` (used by `Body`). On current `main`, `Body` was rewritten to use `bindMediaType` from `helpers.go` and the `utils` import was dropped from `bind.go` entirely. This PR adds `utils.JSONUnmarshal` at `bind.go:43` and `bind.go:140` but never touches the import block, so a rebase/merge onto `main` takes `main`'s import block and leaves `undefined: utils`. The import has to be re-added as part of the rebase. CI green on `efb2d11` says nothing about this.

- **A custom binder registered for `application/json` silently ignores the override.** `Body` checks `app.customBinders` before the content-type switch, so if someone registers a `CustomBinder` whose `MIMETypes()` contains `application/json`, `WithJSONDecoder` never reaches it. The docs claim "The override applies when JSON is decoded by `JSON`, `Body`, or `All`" without that caveat. Worth one sentence in `docs/api/bind.md`.

- **Nit — avoid the `Config()` struct copy.** `bind.go:316` uses `b.ctx.App().Config().JSONDecoder`, which copies the whole `Config` value. `XML` already does the cheaper `b.ctx.App().config.XMLDecoder`. Same package, so the direct field read works here too. (Pre-existing pattern, but this line is being touched anyway.)

- **Nit — docs placement.** `WithJSONDecoder` was added as the first entry under `## Options`, whose intro sentence reads "For more control over error handling, you can use the following methods." A decoder selector is not error handling; either move it below the three error-handling methods or widen the intro sentence.

- **API-surface question for maintainers:** only JSON gets a per-bind override; `CBOR`, `XML`, `MsgPack` and `Form` still read exclusively from app config. If per-bind decoders are a direction the project wants, it may be worth deciding the shape now (e.g. a generic option) rather than adding `WithXMLDecoder`/`WithCBORDecoder` one at a time later.

No issues found with thread safety (the `*Bind` is per-`Ctx`, never shared) or with the fallback precedence itself — per-bind → app config → `json.Unmarshal` (defaulted in `app.go:771-772`) is correct, and the `Body`/`All` paths do route through `Bind.JSON`, so the tests covering them are meaningful.

---

efectn:
I am not in favor of this change as custom JSON decoders could be used by defining simple CustomBinder struct and it use it with `Custom()` method. I think there should not be multiple ways to do the same thing. Instead, a **document/example** can be added regarding using custom JSON decoders with **CustomBinder** method, in my opinion. 

Also; in case we add per-bind JSON decoder support, it also creates a room for adding per-bind decoder for some methods such as MsgPack, XML etc.

ReneWerner87:
Following up on @efectn's point: I checked it on current main (`1d6d6088`) and I think he is right, though the documentation gap behind it is bigger than one example.

A `CustomBinder` whose `MIMETypes()` returns nil is never auto-selected, so it gives per-call strict decoding without touching the app-wide JSON path:

```go
type strictJSON struct{}

func (strictJSON) Name() string        { return "strictjson" }
func (strictJSON) MIMETypes() []string { return nil }
func (strictJSON) Parse(c fiber.Ctx, out any) error {
    d := json.NewDecoder(bytes.NewReader(c.Body()))
    d.DisallowUnknownFields()
    return d.Decode(out)
}
```

With `app.RegisterCustomBinder(strictJSON{})`, the same body gives `400 unknown field "admin"` through `c.Bind().Custom("strictjson", &p)` and `200` through `c.Bind().JSON(&p)`. That covers #2858, and it is already more than Gin (a package-level `var`) or Echo (no per-call override) offer.

What `docs/api/bind.md` does not explain is why it works. The existing `Custom` example registers a new format (YAML), where `MIMETypes` is obviously right, and the page never says what `MIMETypes` does to a format that already has a binder. Two things I measured that a reader cannot get from it today:

- A custom binder claiming `"application/json"` takes over every `Bind().Body()` call in the app. Registering a strict binder that way is a silent app-wide change, not a per-call one.
- Returning nil is what makes a binder opt-in, reachable only through `Bind().Custom(name, dest)`.

Separately, the comment above `Bind.Body` (`bind.go:397`) says custom binders are consulted "if none of the content types above are matched", but the loop runs before the switch, so they win. That needs fixing either way.

Rather than closing this, would you be up for turning it into that documentation PR? Dropping the code and adding a short "overriding a built-in format" section to `docs/api/bind.md` would close #2858 and be more useful than the API, because it also warns about the `MIMETypes` footgun.

You would need to rebase first in any case: the branch is 243 commits behind main and does not compile as is, since main dropped the `gofiber/utils/v2` import from `bind.go` that the new lines use.

ReneWerner87:
Superseded by #4664, which documents this instead of adding API.

@efectn's objection holds. A `CustomBinder` whose `MIMETypes()` returns nil is opt-in, reachable only through `Bind().Custom(name, dest)`, and gives per-call strict decoding without touching the app-wide JSON path. I ran both halves on `fe38a78e`: with such a binder registered, the same body returns `json: unknown field "admin"` through `Custom("strict", &p)` and binds normally through `Body(&p)`. A binder claiming `application/json` instead takes over every `Body()` call in the app.

What was missing was the documentation, not the feature. #4664 adds the "MIMETypes and Body" and "Overriding a built-in format" sections to `docs/api/bind.md` and corrects the doc comment on `Bind.Body`, which had the precedence backwards. It closes #2858.

Thanks for the work here. The branch is also 158 commits behind main and no longer compiles against it, so carrying it forward would have meant a rewrite in any case.

PR #4594, opened 2026-08-09
🔥 feat: render configured templates to a writer

### LINKED ISSUE #3297: 📝 [Proposal]: We need partial render!

We have amazing gofiber + htmx + sse. We can send partial template on each update via SSE
Example how it can be reached now:

```golang
app.Get("/", func (c fiber.Ctx) error {
  return c.SendStreamWriter(func(w *bufio.Writer) {
   for {
    data := getLiveData()
    buf := new(bytes.Buffer)
    c.App().Config().Views.Render(buf, "my-live-partial-template", data) // hack for partial render
    w.WriteString(buf)
    time.Sleep(5*time.Second)
   }
  })
})
```

I propose that we can add method RenderToWriter(w *bufio.Writer, templateName string, data fiber.Map, layoutName string)

```golang
app.Get("/", func (c fiber.Ctx) error {
  return c.SendStreamWriter(func(w *bufio.Writer) {
   for {
    data := getLiveData()
    c.RenderToWriter(w, "my-live-partial-template", data)
    time.Sleep(5*time.Second)
   }
  })
})
```

---

---

### API Stability

---

### Feature Examples

```go
app.Get("/", func (c fiber.Ctx) error {
  return c.SendStreamWriter(func(w *bufio.Writer) {
   for {
    data := getLiveData()
    c.RenderToWriter(w, "my-live-partial-template", data)
    time.Sleep(5*time.Second)
   }
  })
})
```

### Checklist:

RubenPari:
## Summary

Adds the ability to render app-configured view templates directly to a writer, enabling SSE/HTMX partial streaming without manual buffer wiring around `Views.Render`.

Addresses the partial-render use case from #3297.

## Changes

- App-level render-to-writer API
- Tests covering template rendering to writers
- Docs in `docs/api/app.md`

## Linked issue

Closes #3297

## Checklist

- [x] `make audit`
- [x] `make generate`
- [x] `make format`
- [x] `make lint` — 0 issues
- [x] `make test` — passed


gaby:
@codex review

gaby:
## Review status

1 unresolved review thread (not outdated), and it was filed against the current head commit `62ad5ff`, so nothing has landed since. The feature itself is small and self-contained, but I don't think it's mergeable as-is: the lock is held across writes to a caller-supplied (possibly network-backed) writer, and the `c.Render` path was not updated for the new typed-nil semantics.

## Remaining review comments

- **`chatgpt-codex-connector` on `app.go:1558` — "Guard typed-nil views in request rendering" (unresolved, not outdated).** Asks that the new nil-aware guard also be applied in the request-render path, or that the invalid config be rejected at init. **Not addressed on the current head.** The claim checks out: `res.go:942` still tests `app.config.Views != nil`, as does the domain-mount fallback at `res.go:965`, so a typed-nil engine now sails through `New()` (init skips `Load` via `isNilViews`) and then reaches `Views.Render` from a handler. For any pointer-receiver engine that touches its own state, that is a nil dereference at request time. Before this PR the same config would have blown up (or at least warned) during `init` at the `Load` call, so the PR moves the failure later and makes it a request-time panic. It is also internally inconsistent: `app.Render` returns `ErrNoViewEngineConfigured` for exactly the config that `c.Render` will happily dereference. Reusing `isNilViews` in both spots in `res.go` is the cheap fix.

## Additional findings

- **Lock held across writes to the caller's writer — reload starvation.** `app.Render` takes the views `RLock` and holds it for the whole `views.Render(out, ...)` call. `DefaultRes.Render` deliberately renders into a pooled `bytebufferpool` buffer under the lock, so the hold time is bounded by CPU. Here the hold time is bounded by how fast the *client* drains the writer. The documented use case in `docs/api/app.md` is exactly the bad case: rendering into a `*bufio.Writer` from `SendStreamWriter`. One slow SSE consumer blocks `ReloadViews` (`Lock()`), and because `sync.RWMutex` blocks new readers once a writer is queued, it then blocks every in-flight `c.Render` too. Rendering into a pooled buffer under the lock and copying to `out` after releasing it fixes this.
- **Partial writes on render error.** Same root cause: bytes are written straight to `out`, so a mid-template failure leaves a truncated fragment on the caller's stream and *also* returns an error. `DefaultRes.Render` never has this problem because it only flushes the buffer on success. For an SSE/HTMX partial this means emitting a half-rendered fragment. Buffer-then-copy fixes both this and the point above; if direct streaming is intentional, it needs to be called out in the doc, because callers can't undo the write.
- **No mount / domain-mount resolution, unlike both neighbours.** `app.Render` looks only at `app.config.Views` and `app.config.ViewsLayout`. `c.Render` walks `mountFields.appList` and the domain mount owner to find an engine, and `ReloadViews` (right above the new method, `app.go:900`) explicitly walks `app.mountedApps()`. So: root app with the engine on a mounted sub-app → requests render fine, `app.Render` returns `ErrNoViewEngineConfigured`; sub-app with no engine of its own → `c.Render` borrows the parent's engine, `subApp.Render` errors. That asymmetry is defensible, but it is currently undocumented and there's no test for it.
- **`nil` binding is not normalized.** `DefaultRes.Render` does `if bind == nil { bind = make(Map) }` before handing off to the engine; `app.Render` passes `nil` straight through. Engines that assume a map get different behaviour depending on which entry point you use.
- **`isNilViews(views any)` is looser than it needs to be.** Everything it's called with is a `Views`, and `reflect.ValueOf` on an `any` can never report `reflect.Interface`, so that case is dead. The `any` parameter appears to exist only so `Test_isNilViews_ReturnsTrueForNilUnsafePointer` can pass a bare `unsafe.Pointer` (which does not implement `Views`). Typing it `Views` and dropping the unreachable case would be tighter; the `UnsafePointer` kind is still reachable through a named type with methods.
- **Minor:** the error text differs between paths — `fmt.Errorf("fiber: failed to render views: %w", ...)` in `app.go:940` vs `"failed to render: %w"` in `res.go:948`. Also the doc comment on the new method doesn't mention that it returns `ErrNoViewEngineConfigured`, and the docs section doesn't either.
- **Test coverage gaps:** no-engine and typed-nil paths are well covered, and the `ReloadViews` synchronization test is a good addition. Missing: an engine that writes some output and *then* fails (would pin down the partial-write contract), and `app.Render` against a mounted-sub-app engine.

CI is green and CodeRabbit reported no actionable comments; `mergeable_state` is `blocked` pending maintainer approvals. Note that #4588 also touches `App.Render` and typed-nil view detection, so these two will likely need rebasing against each other.

---

gaby:
@codex review

gaby (approved):
👍 LGTM

ReneWerner87 (approved):

PR #4596, opened 2026-08-09
🔥 feat: add CrossOriginProtectionOnly mode to CSRF middleware

### LINKED ISSUE #3912: 📝 [Proposal]: Offer simpler CSRF middleware as in Go's http.CrossOriginProtection

In Go, there's now a low-effort, rather performant CSRF protection [available](https://pkg.go.dev/net/http@master#CrossOriginProtection), with a comprehensive [explanation](https://github.com/golang/go/issues/73626). Of course it doesn't cover all edge cases the token/session approach in fiber does, but I wonder whether it might make sense to at least provide an option to just mirror the default `http's` protection as a "cheap" alternative to the current dual-roundtrip approach.

We're just talking about a vendored middleware here (or maybe only the modification of one), so there should be no conflict with the API at all. That said, there is no CSRF protection middleware with a similar logic in express yet.

By default, the `http.CrossOriginProtection` assumes that `GET`, `HEAD`, and `OPTIONS` won't trigger a modification - as fiber's CRSF middleware does, too. But this doesn't conflict with the HTTP RFC, where those are considered "safe" methods. Additionally, it relies on certain header entries (`Sec-Fetch-Site` or, if absent, `Origin`), which are both not specified in the HTTP RFC, but still can be considered as standards ([#1](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Origin), [#2](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Sec-Fetch-Site)). 

### API Stability

Honestly, I'm relying here a bit on the authority of the Go developers. Since API stability and downwards compatibility is a main focus of the Go development, I'ld consider the approach rather safe against fundamental changes. The fiber middleware adjustments should keep close to it. The biggest question is probably whether to support `InsecureBypassPatterns` in the way `http` variant does, b/c the form of `[METHOD ][HOST]/[PATH]` is uncommon for fiber.

### Feature Examples

```go
Since this approach doesn't really pose a conflict with the existing CSRF middleware, it could be integrated as an additional option:

app.Use(csrf.New(csrf.Config{
    CrossOriginProtection: CrossOriginProtectionOpts{TrustedOrigins: []string{"http://localhost:8080}, InsecureBypassPatterns: []InsecureBypassPattern{{Method: []string{fiber.MethodPost}, Pattern: "example.com/"}},
    ...
}))
```

### Checklist:

RubenPari:
## Summary

Adds a tokenless CSRF mode (`CrossOriginProtectionOnly`) aligned with Go's `net/http.CrossOriginProtection`: reject unsafe cross-origin requests using Fetch Metadata / Origin checks, without cookies or CSRF tokens.

Complements the Sec-Fetch-Site support from #3913 and addresses the simpler protection option requested in #3912.

## Changes

- `Config.CrossOriginProtectionOnly`
- Validation path + tests (including comparison notes vs `net/http` when available)
- Docs in `docs/middleware/csrf.md`

## Linked issue

Closes #3912

## Checklist

- [x] `make audit`
- [x] `make generate`
- [x] `make format`
- [x] `make lint` — 0 issues
- [x] `make test` — passed


gaby:
@codex review

gaby:
**Status:** 9 review threads, all still unresolved (1 outdated, but duplicated by a live thread). Head is still the original commit `8ed4c3d`, so nothing has been pushed in response to any of them. GitHub also reports `mergeable_state: dirty` — the branch has conflicts with `main` and needs a rebase. Not mergeable yet; the trusted-origin normalization bug and the Go version gate are both real and need code changes.

## Remaining review comments

- **`docs/middleware/csrf.md` — scope existing security claims to token mode** (coderabbit, unresolved, not outdated). Asks that the older paragraphs claiming CSRF "relies on token-based validation" and that `Origin`/`Referer` checks "always" run in addition to token validation be qualified, since `CrossOriginProtectionOnly` does neither. **Not addressed** — the diff only appends a new section; the pre-existing paragraphs are untouched. Valid.
- **`middleware/csrf/csrf.go:94` — normalize trusted origins before exact matching** (codex P2, unresolved; same point from copilot at `csrf.go:95`, marked outdated but still applicable). `crossOriginTrustedOrigins` stores `trimmedOrigin` (raw config value) while the token path stores `normalizedOrigin`. `normalizeOrigin` accepts `https://trusted.example/` and `https://TRUSTED.example`, so those configs are valid in token mode but can never match a browser-serialized `Origin` header in the new mode. **Not addressed, and confirmed correct** — fail-closed, so not a bypass, but it silently denies a config that works in the other mode. Store `normalizedOrigin` and compare against a lowered `origin`, or reject non-canonical entries at config time.
- **`middleware/csrf/csrf_test.go:296` — Go version gate should be `go1.25`** (coderabbit + codex P2, both unresolved). **Not addressed, and confirmed correct.** `go.mod` declares `go 1.25.0` and `net/http.CrossOriginProtection` landed in Go 1.25, so the `go1.26` gate skips the parity test on the entire 1.25.x half of the CI matrix (`.github/workflows/test.yml:25`).
- **`middleware/csrf/csrf.go:98` — skip extractor validation in tokenless mode** (codex P2, unresolved). Correct as stated: `configDefault` calls `validateExtractorSecurity` (`middleware/csrf/config.go:176`) before the `CrossOriginProtectionOnly` branch is reached, so a cookie extractor matching `CookieName` panics at startup even though no extractor is used. **Not addressed.** Low impact but real.
- **`middleware/csrf/csrf_test.go:346` — `go run` without a timeout** (copilot, unresolved). **Not addressed, and worth acting on** — the CI matrix is 1.25.x/1.26.x across Linux/Windows/macOS, so once the version gate is corrected this test compiles and executes a child Go program on every runner. See "Additional findings" for a suggested alternative.
- **`middleware/csrf/csrf_test.go:229` — `Origin: http://example.com` on an `https://` request is asserted as allowed** (copilot, unresolved). **Not addressed.** The behaviour matches `net/http` (host-only comparison, deliberate fail-open), but the test name gives no hint that a scheme downgrade is being accepted. See S2 below.
- **`middleware/csrf/csrf_test.go:343` — path built with `"/"` instead of `os.PathSeparator`** (copilot, unresolved). **Not addressed.** I'd call this one a non-issue: Go on Windows accepts forward slashes in file paths and `go run` handles them, so the test is already platform-agnostic. Safe to skip with a note.

## Additional findings

**S1 (security, highest value) — enabling the mode silently discards all token protection with no diagnostic.** The early return at `middleware/csrf/csrf.go:98` bypasses handler/storage construction entirely, so `Session`, `Storage`, `Extractor`, `SingleUseToken`, `KeyGenerator`, `IdleTimeout` and every `Cookie*` field are silently ignored. Someone adding `CrossOriginProtectionOnly: true` to an existing token-based config loses Double Submit Cookie validation, the Referer fallback, and storage-backed token checks, and gets no panic, no log line, and no test failure. This middleware already panics on invalid trusted origins and on insecure extractors — the same treatment (or at minimum `log.Warnf`) should apply when the flag is combined with token-mode configuration.

**S2 (security) — the Origin fallback does a raw host comparison, and the new docs don't disclose the scheme fail-open.** `middleware/csrf/csrf.go:399` compares `originURL.Host == c.Host()` with no scheme check and no default-port or case normalization, unlike `originMatchesHost` which uses `schemehost.Match`. Two consequences:
  - `Origin: http://example.com` is accepted on an `https://example.com` request. This mirrors `net/http`, which documents the trade-off and points at HSTS, but the new docs section says nothing about it. For security-sensitive middleware that disclosure belongs in the docs, and the stdlib rationale belongs in a code comment.
  - Conversely `Origin: https://example.com:443` against `Host: example.com` is *denied* here while the token mode allows it. Fail-closed, but an inconsistency users switching modes will hit.

**S3 (security, low) — empty `Host` makes `Origin: null` pass.** `url.Parse("null")` yields `Host == ""`, so if `c.Host()` is empty (HTTP/1.0 request with no `Host` header) the comparison at `middleware/csrf/csrf.go:399` succeeds and the request is allowed. Not browser-reachable and inherited from the stdlib shape, but the guard is one condition: `if h := c.Host(); h != "" && originURL.Host == h`. The `null origin` test case only covers the non-empty-host path.

**S4 — `TRACE` and `QUERY` change classification between modes, undocumented.** The token path treats `MethodTrace` and `MethodQuery` as safe (`middleware/csrf/csrf.go:139`); `validateCrossOriginProtection` treats them as unsafe. The tests assert this deliberately and it matches the stdlib, but it is a behaviour change for anyone switching modes and isn't mentioned in the docs.

**S5 — context helpers are unavailable in this mode and the docs don't say so.** `HandlerFromContext(c)` returns `nil`, so existing code doing `csrf.HandlerFromContext(c).DeleteToken(c)` nil-derefs, and `TokenFromContext(c)` returns `""`. `Test_CSRF_CrossOriginProtectionOnly_bypasses_token_state` asserts both, but the new docs section doesn't warn about it.

**Test structure** — compiling and running a generated program via `exec.Command("go", "run", ...)` inside a unit test is a lot of machinery for parity coverage, and it makes the suite depend on a working toolchain, a writable `GOCACHE`, and process spawning on three OS runners. A `//go:build go1.25`-tagged test file calling `http.NewCrossOriginProtection().Check()` directly gets the same parity signal with none of that, and drops the timeout concern raised in the copilot thread. Codex's thread suggests the same alternative.

**Coverage gaps** worth adding: a trusted origin configured with a trailing slash or uppercase host (would have caught the `csrf.go:94` bug); a wildcard trusted origin with a mismatched scheme (`http://api.example.net` against `https://*.example.net` — currently denied, correctly, but untested); and `Sec-Fetch-Site: same-origin` combined with a mismatched `Origin` (allowed, matching stdlib, but untested).

Default is `false`, opt-in only, and the ordering of the Fetch-Metadata and Origin checks faithfully reproduces `net/http.CrossOriginProtection.Check` — no bypass found in the decision matrix itself.

---

sixcolors:
I’m wondering if this should just be its own middleware?

gaby:
@RubenPari Can you fix the merge conflicts.

sixcolors:
> I’m wondering if this should just be its own middleware?

@gaby thoughts?

gaby:
> > I’m wondering if this should just be its own middleware?
> 
> @gaby thoughts?

@sixcolors How do yo envision this as a middleware, it is barely ~100 lines.

sixcolors:
> > > I’m wondering if this should just be its own middleware?
> > 
> > 
> > @gaby thoughts?
> 
> @sixcolors How do yo envision this as a middleware, it is barely ~100 lines.

To me this looks like a separate concern from traditional CSRF token protection: when  CrossOriginProtectionOnly  is enabled, the middleware stops doing token generation/validation and instead implements browser cross-origin checks using Fetch Metadata and Origin. It’s closer to “should this unsafe cross-origin browser request be allowed?” than “is this CSRF token valid?”

That feels analogous to adding JWT support to a basic-auth middleware by branching and ignoring basic auth entirely. I’m not attached to a particular structure; and it can be done here if desired and the API and docs reflect that these are distinct mechanisms, not just two modes of the same thing.

ReneWerner87:
@RubenPari @sixcolors @gaby A measurement that may settle the "should this be its own middleware" question.

With the flag set, `New` returns early with a separate handler, before the session and storage managers are constructed. Counting what that handler reaches: `Config` carries 18 fields and the cross-origin path uses four of them, `CrossOriginProtectionOnly`, `Next`, `ErrorHandler` and `TrustedOrigins`. The other 14 are inert, so `Storage`, `Session`, `Extractor`, `CookieName`, `SingleUseToken` and `IdleTimeout` can each be set alongside the flag with no effect and no error.

Framework precedent points the same way: Django, Rails and Echo treat tokenless checks as a supplement to the token rather than a mode that replaces it, and Go's `net/http/csrf` is a standalone type.

A second constructor in this package keeps both sides of the discussion:

```go
csrf.NewCrossOriginProtection(csrf.CrossOriginConfig{
    Next:           ...,
    ErrorHandler:   ...,
    TrustedOrigins: ...,
})
```

No new package and no new doc page, and the 14 dead fields stop being reachable rather than being documented as ignored. Keeping the bool means adding validation that rejects those combinations, which is more code than the constructor costs.

One thing that is not a problem, since it sat in my own notes as one: the cross-origin path reads headers with `c.Get` where the token path uses `headerlookup.Value`. That reads like a gap under `DisableHeaderNormalizing` but is not. A lone `Sec-Fetch-Site: cross-site` with no `Origin`, where a missed header would answer 200, is denied in all six combinations of `Sec-Fetch-Site`, `sec-fetch-site` and `SEC-FETCH-SITE` with the setting on and off.

PR #3328, opened 2025-02-25
📚 Doc: Add how to use NewCtxFunc and middleware at the same time

JIeJaitt:
# Description

Add documentation that tells users how to use both NewCtxFunc and app.Use().

Fixes #3319

## Changes introduced

List the new features or adjustments introduced in this pull request. Provide details on benchmarks, documentation updates, changelog entries, and if applicable, the migration guide.

- [ ] Benchmarks: Describe any performance benchmarks and improvements related to the changes.
- [x] Documentation Update: Detail the updates made to the documentation and links to the changed files.
- [ ] Changelog/What's New: Include a summary of the additions for the upcoming release notes.
- [ ] Migration Guide: If necessary, provide a guide or steps for users to migrate their existing code to accommodate these changes.
- [ ] API Alignment with Express: Explain how the changes align with the Express API.
- [ ] API Longevity: Discuss the steps taken to ensure that the new or updated APIs are consistent and not prone to breaking changes.
- [ ] Examples: Provide examples demonstrating the new features or changes in action.

## Type of change

Please delete options that are not relevant.

- [x] Documentation update (changes to documentation)

## Checklist

Before you submit your pull request, please make sure you meet these requirements:

- [ ] Followed the inspiration of the Express.js framework for new functionalities, making them similar in usage.
- [x] Conducted a self-review of the code and provided comments for complex or critical parts.
- [x] Updated the documentation in the `/docs/` directory for [Fiber's documentation](https://docs.gofiber.io/).
- [x] Added or updated unit tests to validate the effectiveness of the changes or new features.
- [x] Ensured that new and existing unit tests pass locally with the changes.
- [ ] Verified that any new dependencies are essential and have been agreed upon by the maintainers/community.
- [x] Aimed for optimal performance with minimal allocations in the new code.
- [ ] Provided benchmarks for the new code to analyze and improve upon.

## Commit formatting

Please use emojis in commit messages for an easy way to identify the purpose or intention of a commit. Check out the emoji cheatsheet here: [CONTRIBUTING.md](https://github.com/gofiber/fiber/blob/master/.github/CONTRIBUTING.md#pull-requests-or-commits)

PR #4598, opened 2026-08-09
🐛 bug: pin CustomCtx overrides when middleware is present

### LINKED ISSUE #3319: 🐛 [Bug]: There is a conflict between app.NewCtxFunc and app.Use
### Bug Description

CustomCtx no effect when app.newCtxFunc and app.use simultaneously existing

### How to Reproduce

1. create custom ctx
`type Ctx struct {
	fiber.DefaultCtx
}

func (c *Ctx) JSON(data any, ctype ...string) error {
	return c.DefaultCtx.JSON(Response{
		Code:    http.StatusOK,
		Data:    data,
		Message: http.StatusText(http.StatusOK),
	}, ctype...)
}`
type Response struct {
	Code    int    `json:"code,omitempty" xml:"code,omitempty" yaml:"code,omitempty"`
	Data    any    `json:"data,omitempty" xml:"data,omitempty" yaml:"data,omitempty"`
	Message string `json:"message,omitempty" xml:"message,omitempty" yaml:"message,omitempty"`
}

2. create app with newCtxFunc:
`              app := fiber.New(fiber.Config{
			JSONEncoder: sonic.Marshal,
			JSONDecoder: sonic.Unmarshal,
		})

		app.NewCtxFunc(func(app *fiber.App) fiber.CustomCtx {
			return &Ctx{
				DefaultCtx: *fiber.NewDefaultCtx(app),
			}
		})

                app.Get("/vvv", func(ctx fiber.Ctx) error {
			return ctx.JSON(fiber.Map{"a": "b"})
		})
                _ = app.Listen(":8080")
`
3. request:
`curl localhost:8080/vvv   
{"code":200,"data":{"a":"b"},"message":"OK"}%`

4. create app with newCtxFunc and app.use:
`              app := fiber.New(fiber.Config{
			JSONEncoder: sonic.Marshal,
			JSONDecoder: sonic.Unmarshal,
		})

		app.NewCtxFunc(func(app *fiber.App) fiber.CustomCtx {
			return &Ctx{
				DefaultCtx: *fiber.NewDefaultCtx(app),
			}
		})

                app.Use(func(c fiber.Ctx) error {
			return c.Next()
		})

                app.Get("/vvv", func(ctx fiber.Ctx) error {
			return ctx.JSON(fiber.Map{"a": "b"})
		})
                _ = app.Listen(":8080")
`
5. request:
`curl localhost:8080/vvv
{"a":"b"}`

### Expected Behavior

app.NewCtxFunc and app.Use can coexist

### Fiber Version

v3.0.0-beta.4

### Code Snippet (optional)

```go
package main

import "log"

func main() {
  app := fiber.New()

  // Steps to reproduce

}
```

### Checklist:

RubenPari:
## Summary

On current main, `NewWithCustomCtx` already preserves custom methods through middleware via `handlerCtx`. This PR pins that behavior with a regression test (custom `JSON` envelope with and without `app.Use`) and documents the pattern under `NewWithCustomCtx`.

## Changes

- Regression test for #3319
- Docs example: custom JSON + global middleware

## Linked issue

Closes #3319

## Checklist

- [x] `make format`
- [x] `make lint` — 0 issues
- [x] targeted tests pass

### DIFF
--- a/ctx_test.go
+++ b/ctx_test.go
@@ -399,6 +399,26 @@ func (c *customCtx) Params(key string, defaultValue ...string) string { //revive
 	return "prefix_" + c.DefaultCtx.Params(key)
 }
 
+// jsonEnvelopeCustomCtx overrides JSON to wrap payloads in a standard API envelope.
+// Used to regression-test #3319 (CustomCtx methods must survive middleware / Next).
+type jsonEnvelopeCustomCtx struct {
+	DefaultCtx
+}
+
+type jsonEnvelope struct {
+	Data    any    `json:"data"`
+	Message string `json:"message"`
+	Code    int    `json:"code"`
+}
+
+func (c *jsonEnvelopeCustomCtx) JSON(data any, ctype ...string) error {
+	return c.DefaultCtx.JSON(jsonEnvelope{
+		Code:    StatusOK,
+		Data:    data,
+		Message: "OK",
+	}, ctype...)
+}
+
 // go test -run Test_Ctx_CustomCtx
 func Test_Ctx_CustomCtx(t *testing.T) {
 	t.Parallel()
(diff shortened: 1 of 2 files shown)


gaby:
**Status:** 1 unresolved review thread (not outdated, not addressed by head `a349a09`). The change is test + docs only — no production code, so no request-path or allocation impact, and Codecov confirms a 0-line delta. Close to mergeable once the docs snippet is fixed up; `mergeable_state` is currently `blocked` pending maintainer approval.

I verified the regression test does what it claims: the handler calls `c.JSON(...)` through the `Ctx` interface, so if `handlerCtx` propagation regressed the handler would receive a `*DefaultCtx` and the body would be `{"a":"b"}` — `require.JSONEq` would fail. It is also not redundant with the existing `ctx_test.go:481` test, which asserts the concrete type via `c.(*customCtx)` rather than exercising dynamic dispatch of an overridden method (the actual #3319 symptom).

## Remaining review comments

- **copilot-pull-request-reviewer on `docs/api/app.md:695`** — the new example omits `package main` and the `import` block, unlike other examples in the file, so it is not copy-pasteable. **Valid, and not addressed on head.** Every other `go title="Example"` block in `docs/api/app.md` (Mounting, Group, RouteChain, Domain, Stack, Name, GetRoute, GetRoutes, RegisterCustomBinder, RebuildTree, RemoveRoute) opens with `package main` + imports, and the new block uses `log.Fatal` and `fiber.*` with neither. Adding `package main` and an import of `log` + `github.com/gofiber/fiber/v3` resolves it.

CodeRabbit reported "no actionable comments"; nothing there needed rebutting.

## Additional findings

- **`docs/api/app.md:689` — the new example largely duplicates the one directly above it.** The preceding `NewWithCustomCtx` example already shows the full factory + `NewDefaultCtx(app)` pattern; the only new material is the `app.Use(...)` line and the `JSON` override. Consider trimming the second block to just that delta rather than a second full ~30-line program with its own `func main()`, which is also what creates the missing-imports issue above.
- **The `:::tip` asserts pointer embedding works, but nothing pins it.** "Prefer embedding `fiber.DefaultCtx` (by value or pointer)" is correct — method promotion carries the unexported `CustomCtx` methods either way — but no test in the repo embeds `*DefaultCtx`; every one uses `DefaultCtx: *NewDefaultCtx(app)`. For a PR whose stated purpose is pinning documented behavior, the `*DefaultCtx` variant is the one case actually worth adding.
- **Title/label vs. content.** The PR is titled `🐛 bug:` and carries `☢️ Bug`, but contains no fix. Release-drafter (`.github/release-drafter.yml`) categorizes from title/label, so this would land under bug fixes in the v3 notes while fixing nothing. `📚 Doc:` or a docs/test label matches the diff.
- **Prior attempt on the same issue.** #3328 ("📚 Doc: Add how to use NewCtxFunc and middleware at the same time") was closed without landing — there is no trace of it in `main`'s history and `docs/api/app.md` has no such section. Worth confirming with a maintainer why it did not land before merging the same approach.

Non-blocking: `Closes #3319` from a test/docs-only PR is defensible here, since the reported behavior is genuinely fixed on `main` by the `handlerCtx` plumbing in `ctx.go:255-267` and `ctx_interface.go:83-87`.

Checked and found nothing wrong: both custom-ctx dispatch branches in `Next()` (`ctx.go:257` same-route multi-handler, and `ctx.go:263` cross-route via `nextCustom`) are already covered by the existing custom-ctx tests, so the new test is not leaving a branch gap. The value-copy embedding pattern (`*NewDefaultCtx(app)`) is safe despite `NewDefaultCtx` self-linking `DefaultReq.c`/`DefaultRes.c`, because `Reset` re-links both on every acquire (`ctx.go:735-736`).


---

ReneWerner87 (approved):
LGTM

PR #4600, opened 2026-08-09
🐛 bug: pin ShutdownWithTimeout waits for in-flight handlers

### LINKED ISSUE #3370: 🐛 [Bug]: "ShutdownWithTimeout" & "ShutdownTimeout" not respected
### Bug Description

I created a dummy function, that just offers a 10GB file as download and send it the stop signal (`CRTL + C`) after starting the download.
Even though I started the server with `ShutdownTimeout=10 * time.Second` and stopped him with `app.ShutdownWithTimeout(10 * time.Second)` the app shut down immediately and did not wait the configured time.

### How to Reproduce

1. `dd if=/dev/urandom of=/tmp/large_files/large_file.bin bs=1M count=10240 status=progress`

2. CONFIG
```go
// Pfad zur großen Datei
const largeFilePath = "/tmp/large_files/large_file.bin"

// HandleLargeFileStream streamt eine echte 10GB Datei
// Dies hält die Verbindung für eine längere Zeit offen, abhängig von der Downloadgeschwindigkeit
func HandleLargeFileStream(c fiber.Ctx) error {
	// Datei öffnen
	file, err := os.Open(largeFilePath)
	if err != nil {
		return c.Status(fiber.StatusInternalServerError).SendString(fmt.Sprintf("Fehler beim Öffnen der Datei: %v", err))
	}
	defer file.Close()

	// Dateigröße ermitteln
	fileInfo, err := file.Stat()
	if err != nil {
		return c.Status(fiber.StatusInternalServerError).SendString(fmt.Sprintf("Fehler beim Abrufen der Dateigröße: %v", err))
	}
	fileSize := fileInfo.Size()

	// Header setzen
	c.Set("Content-Type", "application/octet-stream")
	c.Set("Content-Disposition", "attachment; filename=\"large_file.bin\"")
	c.Set("Content-Length", fmt.Sprintf("%d", fileSize))

	c.SendFile(largeFilePath)
	return nil
}
```
```go
// Fiber Listen configuration
fiberConfig := fiber.ListenConfig{
	ShutdownTimeout: 10 * time.Second,
}
```

Set up this Signal-Trap:

```go
// Set up graceful shutdown
c := make(chan os.Signal, 1)
signal.Notify(c, os.Interrupt, syscall.SIGHUP, syscall.SIGINT, syscall.SIGTERM)
go func() {
	<-c
	_ = app.ShutdownWithTimeout(cfg.Server.ShutdownTimeout)
}()
```

### Expected Behavior

I was expecting it to close all listeners but to respect ALL currently active connections (untill the timeout). But this did not happen, the app shut down immediately.

### Fiber Version

v3.0.0-beta.4

### Checklist:

RubenPari:
## Summary

Pins the v3 behavior for #3370: with a sufficient timeout, `ShutdownWithTimeout` waits for in-flight handlers, including streaming responses (`SendStreamWriter`), rather than returning immediately.

Also documents how shutdown timeouts apply and how they interact with `ListenConfig.ShutdownTimeout`.

## Changes

- Streaming in-flight handler regression test
- Docs under Server Shutdown

## Linked issue

Closes #3370

## Checklist

- [x] `make format`
- [x] `make lint` — 0 issues
- [x] targeted tests pass

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -1491,6 +1491,84 @@ func Test_App_ShutdownWithTimeout(t *testing.T) {
 	}
 }
 
+// Test_App_ShutdownWithTimeout_WaitsForInFlightHandler pins #3370: with a
+// generous timeout, shutdown waits for active handlers (including stream writers)
+// instead of returning immediately.
+func Test_App_ShutdownWithTimeout_WaitsForInFlightHandler(t *testing.T) {
+	t.Parallel()
+
+	app := New()
+	handlerStarted := make(chan struct{})
+	handlerFinished := make(chan struct{})
+
+	app.Get("/slow", func(c Ctx) error {
+		close(handlerStarted)
+		// Stream long enough that a premature shutdown would drop the client.
+		err := c.SendStreamWriter(func(w *bufio.Writer) {
+			for range 10 {
+				if _, writeErr := w.WriteString("chunk\n"); writeErr != nil {
+					return
+				}
+				if flushErr := w.Flush(); flushErr != nil {
+					return
+				}
+				time.Sleep(100 * time.Millisecond)
+			}
+		})
+		close(handlerFinished)
+		return err
+	})
+
+	ln := fasthttputil.NewInmemoryListener()
+	go func() {
+		assert.NoError(t, app.Listener(ln))
+	}()
+	time.Sleep(50 * time.Millisecond)
+
+	type clientResult struct {
+		err error
+		n   int64
+	}
+	clientDone := make(chan clientResult, 1)
+	go func() {
+		conn, err := ln.Dial()
+		if err != nil {
+			clientDone <- clientResult{err: err}
+			return
+		}
+
+		_, err = conn.Write([]byte("GET /slow HTTP/1.1\r\nHost: example.com\r\n\r\n"))
+		if err != nil {
+			_ = conn.Close() //nolint:errcheck // best-effort cleanup after write failure
+			clientDone <- clientResult{err: err}
+			return
+		}
+		n, err := io.Copy(io.Discard, conn)
+		if closeErr := conn.Close(); err == nil {
+			err = closeErr
+		}
+		clientDone <- clientResult{n: n, err: err}
+	}()
+
+	<-handlerStarted
+	started := time.Now()
+	require.NoError(t, app.ShutdownWithTimeout(5*time.Second))
+	elapsed := time.Since(started)
+
+	// Shutdown must wait for the ~1s stream rather than returning immediately.
+	require.GreaterOrEqual(t, elapsed.Milliseconds(), int64(800))
+
+	select {
+	case <-handlerFinished:
+	case <-time.After(2 * time.Second):
+		t.Fatal("handler did not finish before/around shutdown")
+	}
+
+	res := <-clientDone
+	require.NoError(t, res.err)
+	require.Positive(t, res.n)
+}
+
 func Test_App_ShutdownWithContext(t *testing.T) {
 	t.Parallel()
 
--- a/docs/api/fiber.md
+++ b/docs/api/fiber.md
@@ -343,6 +343,14 @@ func (app *App) ShutdownWithTimeout(timeout time.Duration) error
 func (app *App) ShutdownWithContext(ctx context.Context) error
 ```
 
+:::info How long does shutdown wait?
+`Shutdown` / `ShutdownWithTimeout` / `ShutdownWithContext` wait for **in-flight handlers** (including streaming responses written with `SendStream` / `SendStreamWriter`) until they return or the timeout fires. The timeout is measured from the start of shutdown; if a handler finishes before the deadline, shutdown returns without waiting for the full timeout.
+
+Call `Shutdown*` from a separate goroutine while `Listen` / `Listener` runs, and keep the process alive until `Shutdown*` returns. Set a non-zero `ReadTimeout` / `IdleTimeout` so idle keepalive connections do not block shutdown forever.
+
+`ListenConfig.ShutdownTimeout` is applied only when Fiber runs its **built-in** graceful shutdown path (for example with `GracefulContext`). If you call `app.ShutdownWithTimeout` yourself, that argument is the one that applies.
+:::
+
 ## Helper functions
 
 ### NewError


RubenPari:
@copilot Fix the code for all comments in [this review thread](https://github.com/gofiber/fiber/pull/4600#pullrequestreview-4892524591).

When a review comment includes a suggested change, apply the suggestion exactly.

Do not make changes beyond what is described in the linked review thread.

gaby:
## Status

**2 unresolved review threads**, neither outdated. Head is still `c863241` (the only commit on the branch), so nothing has been pushed since the bot reviews — neither thread is addressed.

Not close to mergeable in its current form, for two reasons beyond the open threads:

1. GitHub reports the PR as **conflicting** (`mergeable_state: dirty`).
2. The test half is **already on `main`**. `Test_App_ShutdownWithTimeout_WaitsForStreamedBody` landed via #4615 (commit `05e86a3`, Aug 15) at `app_test.go:1549` and pins the same #3370 behavior — shutdown waiting for a streamed body written after the handler returned. It is inserted at exactly the spot this PR inserts into (between `Test_App_ShutdownWithTimeout` and `Test_App_ShutdownWithContext`), which is almost certainly the conflict.

Suggestion: rebase and re-scope. The docs change is **not** superseded — `docs/api/fiber.md` on `main` still has no such text, and it's the more valuable half. The only thing the new test adds over `main`'s is that it goes through the public `Ctx.SendStreamWriter` rather than `c.Response().SetBodyStreamWriter` directly; if that's worth keeping, it's a small addition to the existing test, not a second near-duplicate one.

## Remaining review comments

- **coderabbitai — `app_test.go:1519`, "assert completion of the streamed response"**: not addressed, and the core claim is **correct**. `SendStreamWriter` (`res.go:1274`) only calls `SetBodyStreamWriter` and returns, so `close(handlerFinished)` fires before a single byte is streamed. I ran a replica of this exact test with instrumentation: `handlerFinished` closed **3.6µs** after `ShutdownWithTimeout` was entered, while shutdown itself took 1.10s. The `select { case <-handlerFinished: ... case <-time.After(2*time.Second): t.Fatal(...) }` block therefore cannot fail and asserts nothing. The suggested fix (decode the body, assert all ten `chunk\n` values) is the right shape.

- **copilot — `app_test.go:1526`, fixed `time.Sleep(50ms)` for server readiness**: not addressed. The flakiness concern is real in general but **does not apply here**, and `require.Eventually` is not the fix. `fasthttputil.InmemoryListener.Dial` queues the conn without needing `Accept`, and the actual synchronization point is `<-handlerStarted`, which guarantees the server is serving before shutdown is called. I removed the sleep from the replica and ran `-race -count=5`: all pass. The sleep can simply be **deleted**. The second half of that comment (disabling the startup banner) is a reasonable nit — the banner does print in these tests today.

- **copilot suppressed comment on `app_test.go:1540` ("send `Connection: close` or `io.Copy` may hang")**: listed in the review body but not filed as a thread. It is **wrong for fasthttp v1.73**. In the replica, `io.Copy` received EOF and shutdown returned `nil` in 1.10s with default (zero) `ReadTimeout`/`IdleTimeout` — fasthttp's shutdown loop closes idle keep-alive connections. The Fiber doc comments it cites (`app.go:1315`, `app.go:1327`) are stale on this point.

## Additional findings

- **`require.Positive(t, res.n)` is inert.** `io.Copy` counts the status line and headers, so `n > 0` holds even with a zero-length body — it cannot detect a truncated stream. Combined with the vacuous `handlerFinished` select, the **only** assertion doing real work is `elapsed >= 800ms`. That one does pin the behavior (measured 1.10s), but it would still pass on a stream cut off at, say, 900ms.

- **Docs paragraph contradicts the PR's own test.** `docs/api/fiber.md` adds "Set a non-zero `ReadTimeout` / `IdleTimeout` so idle keepalive connections do not block shutdown forever." Both this PR's test and `main`'s `Test_App_ShutdownWithTimeout_WaitsForStreamedBody` use a plain `New()` with zero timeouts and a keep-alive connection, and shutdown returns cleanly. The sentence is inherited from the stale `app.go` comments; it should be softened or verified before being promoted into user-facing docs.

- **"in-flight handlers" is the wrong term for what is being pinned.** Both the test comment and the new docs paragraph say shutdown waits for *handlers* to return. With `SendStream`/`SendStreamWriter` the handler has already returned (in microseconds); what shutdown waits on is the **response body write on a still-open connection**. Worth rewording, since the distinction is exactly what #3370 was about and is what `main`'s test comment already calls out.

- **`ListenConfig.ShutdownTimeout` wording, minor.** The docs say it applies "for example with `GracefulContext`". Per `listen.go:268` and `listen.go:424`, `gracefulShutdown` is started *only* when `cfg.GracefulContext != nil` — it is the sole path, not an example. Also worth noting `ShutdownTimeout == 0` falls back to `app.Shutdown()` and waits indefinitely (`listen.go:720-723`).

- **`assert.NoError(t, app.Listener(ln))` in an unjoined goroutine** (`app_test.go` listener goroutine): if `Listener` returns after the test function completes, testify's `t.Errorf` panics with "Log in goroutine after test has completed". Pre-existing pattern in this file, low priority, but it's copied into new code.

---

ReneWerner87:
Thanks for this, and sorry it sat here as long as it did.

The behavior you pin is worth pinning, and #3370 deserved a regression test. The problem is on our side: `main` already grew one while this PR was waiting. `Test_App_ShutdownWithTimeout_WaitsForStreamedBody` landed on 2026-08-15 via #4615 and cites the same issue:

```go
// A streamed body is written after the handler has already returned, so the
// shutdown tests above, which block inside the handler, say nothing about it.
// Reported as #3370 against v3.0.0-beta.4: a download was cut off instead of
// being allowed to finish within the timeout.
```

Same issue, same mechanism. Your test drives it through `c.SendStreamWriter` and that one through `c.Response().SetBodyStreamWriter`, but `SendStreamWriter` is a single line over `SetBodyStreamWriter` (`res.go:1257`), so both exercise the same path.

Your PR was open on 2026-08-09, six days before ours. That order is on us, not on you, and I would rather say so plainly than close this as a duplicate without explanation.

Closing as already covered. Your other shutdown PR, #4599, is not covered by anything on `main`: it pins that `c.Context()` survives graceful shutdown while `RequestCtx.Done()` closes, which is exactly the distinction a previous PR tried to remove. I have rebased it onto `main` and hardened the timing, and it is ready for review.

ReneWerner87:
Correction to my previous comment: "already covered" was true of the test, not of the whole PR. The documentation half was not covered by anything on `main`, and I should have checked that before closing.

It is now carried over to #4599 as 37f09897, with you as the commit author. I verified each claim before taking it:

- `ListenConfig.ShutdownTimeout` really is read only in `gracefulShutdown` (`listen.go:720`), which only ever starts when `GracefulContext` is set (`listen.go:268` and `listen.go:424`). Nothing on `main` documented that, and it is the part most likely to surprise someone.
- The keepalive advice matches fasthttp's own note on `ShutdownWithContext`: "does not close keepalive connections so it's recommended to set ReadTimeout and IdleTimeout".

One thing I did not carry over verbatim: taking your whole `docs/api/fiber.md` would have reverted three entries `main` gained since August 9th (`SkipUnmatchedRoutes`, `CertClientFile`, and a TLSConfig note), so only the block itself moved.

Thanks for the writeup. It stays as you wrote it.

PR #4608, opened 2026-08-12
perf: optimize cache middleware key generation and allocation

gaby:
# Description

This PR optimizes the cache middleware to reduce memory allocations and improve performance by refactoring key generation and buffer management. The changes consolidate key building into single pooled buffers, eliminate intermediate string allocations, and optimize delimiter escaping to avoid heap allocations on every request.

### Performance Improvements
- **Buffer consolidation**: Cache keys are now assembled in a single pooled buffer instead of being concatenated multiple times, eliminating intermediate string allocations
- **Optimized escaping**: Replaced `strings.Replacer`-based delimiter escaping with append-based implementations that avoid allocating intermediate strings
- **Vary key optimization**: The vary suffix is now appended directly to the key buffer instead of being returned as a separate string and concatenated
- **Auth hash optimization**: Authorization hash is appended directly to the key buffer instead of being returned as a string
- **Header formatting**: Response headers (Date, Cache-Control, Age) are now formatted into local arrays instead of allocating fresh slices
- **Early exit optimization**: Added early return in `parseVary()` when the Vary header is empty to avoid unnecessary allocations

### Code Quality
- Added comprehensive tests (`Test_EscapeKeyDelimiters_AppendFormsMatch`, `Test_Cache_KeyFormatIsStable`) to ensure cache key format stability across refactoring
- Introduced `keyRecordingStorage` test helper to verify the exact keys used by the middleware
- Refactored key generation to support both default and custom generators efficiently
- Added detailed comments explaining allocation optimization strategies

### Benchmarks
The changes reduce allocations on the cache hit path by consolidating multiple string concatenations and intermediate allocations into single buffer operations. Specific improvements:
- Eliminated 2-3 allocations per cache key generation (one per string concatenation)
- Eliminated 1 allocation per delimiter escaping operation
- Eliminated 1 allocation per vary key generation
- Eliminated 1 allocation per auth hash generation
- Eliminated 1 allocation per response header formatting

## Type of change

## Checklist

- [x] Conducted a self-review of the code and provided comments for complex or critical parts
- [x] Added unit tests to validate the effectiveness of the changes (`Test_EscapeKeyDelimiters_AppendFormsMatch`, `Test_Cache_KeyFormatIsStable`, updated `Test_appendAuthHash`, `Test_appendVaryKey`)
- [x] Ensured that new and existing unit tests pass locally with the changes
- [x] Aimed for optimal performance with minimal allocations in the new code
- [x] Verified cache key format stability through pinning tests

## Test Plan

The changes are covered by:
- New test `Test_EscapeKeyDelimiters_AppendFormsMatch` verifies that append-based escaping produces identical bytes to the reference implementation
- New test `Test_Cache_KeyFormatIsStable` pins the exact cache keys generated for various request scenarios
- Updated tests for `appendAuthHash` and `appendVaryKey` verify the new append-based APIs
- Existing cache tests continue to pass, ensuring backward compatibility
- Benchmark tests updated to use cacheable status codes and verify cache hits

https://claude.ai/code/session_01DsGg5ckK1hkruPKChMEj2U

### DIFF
--- a/middleware/cache/cache.go
+++ b/middleware/cache/cache.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"math"
 	"slices"
+	"strconv"
 	"sync"
 	"time"
 
@@ -20,9 +21,23 @@ import (
 	"github.com/gofiber/fiber/v3/internal/headerlookup"
 )
 
-// buffer size for hexpool
 // hexLen is the hex-encoded length of a SHA-256 sum, shared by the auth and vary hashers.
 const hexLen = sha256.Size * 2
+
+// Sizes for the local arrays the response headers are formatted into, so the
+// values reach the header store — which copies them — without a heap
+// allocation each. Both are upper bounds; an append past one still produces the
+// same bytes, only more slowly.
+const (
+	// httpDateLen holds an IMF-fixdate, "Mon, 02 Jan 2006 15:04:05 GMT".
+	httpDateLen = 32
+	// maxUintDigits holds the decimal form of any uint64.
+	maxUintDigits = 20
+)
+
+// publicMaxAge is the Cache-Control this middleware writes on a hit when the
+// entry carries none of its own, less the delta-seconds appended after it.
+const publicMaxAge = "public, max-age="
 
 // cache status
 // unreachable: when cache is bypass, or invalid
@@ -155,15 +170,6 @@ func New(config ...Config) fiber.Handler {
 	heap := &indexedHeap{}
 	// count stored bytes (sizes of response bodies)
 	var storedBytes uint
-	// Pool for hex encoding buffers
-	hexBufPool := &sync.Pool{
-		New: func() any {
-			buf := make([]byte, hexLen)
-			return &buf
-		},
-	}
-	hashAuthorization := makeHashAuthFunc(hexBufPool)
-	buildVaryKey := makeBuildVaryKeyFunc(hexBufPool)
 
 	// Delete key from both manager and storage
 	deleteKey := func(ctx context.Context, dkey string) error {
@@ -258,9 +264,16 @@ func New(config ...Config) fiber.Handler {
 			return c.Next()
 		}
 
-		// Get key from request
-		baseKey := cacheKeyVersion + "|" + requestMethod + "|" + cfg.KeyGenerator(c)
-		manifestKey := baseKey + "|vary"
+		// Get key from request. Assembled in one pooled buffer so the whole key
+		// costs a single allocation: the pieces used to be concatenated one at a
+		// time, and every join copied the key built so far into a fresh string.
+		keyBufPtr := acquireKeyBuffer()
+		keyBuf := (*keyBufPtr)[:0]
+		keyBuf = append(keyBuf, cacheKeyVersion...)
+		keyBuf = append(keyBuf, '|')
+		keyBuf = append(keyBuf, requestMethod...)
+		keyBuf = append(keyBuf, '|')
+		keyBuf = appendGeneratedKey(keyBuf, c, &cfg)
 		if hasAuthorization {
 			// Read at the point of use: the result aliases the header's storage, and
 			// cfg.KeyGenerator is user code that may recycle the slot. Every field
@@ -270,9 +283,21 @@ func New(config ...Config) fiber.Handler {
 			// the same as two lines carrying one each, and the two are different
 			// principals, so a comma would put them in one partition.
 			authLines := fieldname.Lines(&c.Request().Header, fiber.HeaderAuthorization, canonical)
-			authHash := hashAuthorization(authLines)
-			baseKey += "|auth=" + authHash
-			manifestKey = baseKey + "|vary"
+			keyBuf = append(keyBuf, "|auth="...)
+			keyBuf = appendAuthHash(keyBuf, authLines)
+		}
+		baseLen := len(keyBuf)
+
+		// baseKey and manifestKey differ only by the suffix, so one string holds
+		// both: slicing a string off the front of another shares its bytes rather
+		// than copying them, which the separate concatenation could not do.
+		var baseKey, manifestKey string
+		if cfg.DisableVaryHeaders {
+			baseKey = string(keyBuf)
+		} else {
+			keyBuf = append(keyBuf, "|vary"...)
+			manifestKey = string(keyBuf)
+			baseKey = manifestKey[:baseLen]
 		}
 		key := baseKey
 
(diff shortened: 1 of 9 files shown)


gaby:
## On the 3 "slower" benchmarks — they are a benchmark fix, not a code regression

The benchmark job matches by name, but this PR changes **what those benchmarks do**, so base and head are not measuring the same workload.

### Why

`Benchmark_Cache`, `_Storage`, `_AdditionalHeaders` and `_MaxSize` all answered **418**. That status is not in the middleware's `cacheableStatusCodes` on `main` (it mirrors the heuristically-cacheable set in RFC 9110 §15.1). So on `main` every one of those benchmarks got `X-Cache: unreachable` on every iteration — **nothing was ever stored, the expiration heap was never touched, and the eviction loop never ran.**

The bot's own base numbers confirm it: `MaxSize/Disabled` at 1923 ns/op and 352 B/op is the cost of deciding *not* to cache. This PR changes those handlers to 200 so the benchmarks exercise the store and eviction paths they are named for. More work per op, so a higher ns/op — that is the benchmark starting to do its job.

The same artifact inflates the other direction: the "520x faster" on `Benchmark_Cache` is not real either. On `main` that handler re-read `README.md` from disk on every iteration because the response was never cached; now it is served from the entry. Please don't read that number as a win — it is the same apples-to-oranges comparison, just with a favourable sign.

### Apples-to-apples

Base middleware with **this PR's benchmark file** vs. head, 6 runs each, same machine:

| Benchmark | allocs/op | B/op | ns/op |
|---|---|---|---|
| `MaxSize/Disabled` | 9 → **5** | 809–881 → **520–604** | 4285–5909 → **2881–3231** |
| `MaxSize/Unlim` | 9 → **5** | 1115–1134 → **793–878** | 5536–6489 → 3933–6859 |
| `MaxSize/LowBounded` | 10 → **6** | 785 → **497** | 2548–4203 → **2245–2718** |

Allocations and bytes are deterministic and improve in all three. On ns/op, `Disabled` and `LowBounded` improve clearly; `Unlim` overlaps within noise on my box (it grows the heap without bound, so GC timing dominates) — I'd call that flat rather than claim a win.

The hit-path numbers in the bot's report are unaffected by any of this and stand on their own: **4 → 1 allocs** on `Benchmark_Cache` / `_AdditionalHeaders`, **5 → 1** on `_QueryMethod`, **8 → 5** on `_Storage`.

### What I'd suggest

Nothing to fix in the middleware — but the gate is red and that's a maintainer call:

1. **Accept and merge.** The baseline re-establishes on `main` and future comparisons are meaningful again.
2. **If you'd rather keep the gate clean,** I can rename the four fixed benchmarks so the bot sees them as new with no baseline, leaving the old names retired. Say the word and I'll push it.

I'd lean towards (1), since the names still describe what the benchmarks measure — they just finally measure it.

### Also pushed

- `keygen_test.go:289` — dropped the unused receiver on `keyRecordingStorage.Close` (2ff82d8). My local golangci-lint was v2.5.0 against CI's v2.12.2 and could not load several revive rules, so I missed it; I've since matched the CI version locally and the full-repo run is clean.
- `cache_test.go` — corrected the RFC citation to **9110** §15.1 (caa6bc6). CodeRabbit was right: 9111 has no §15 and defers the status codes to 9110. The same wrong citation is in the first commit message, which I've left alone rather than force-push.

---

gaby:
Both Codex findings were real and are fixed in 9e49594.

**Prime the hit benchmarks** — confirmed, and worse than just the assertion. `go test -bench '^Benchmark_Cache$' -benchtime=1x` failed with `actual: "miss"` before the fix. Beyond that, the first `b.Loop()` iteration *was* the cache-filling miss, so the reported hit-path cost carried a share of reading README.md off disk and marshalling it in. One untimed request before the loop fixes both — `b.Loop` resets the timer when it first returns, so the prime stays out of the numbers.

The hit path reads ~20% cheaper now that its own setup is out of the average:

| Benchmark | before prime | after prime |
|---|---|---|
| `Benchmark_Cache` | 1320 ns/op | **1078 ns/op** |
| `Benchmark_Cache_QueryMethod` | 1420 ns/op | **1173 ns/op** |
| `Benchmark_Cache_AdditionalHeaders` | 1590 ns/op | **1267 ns/op** |

All still 1 alloc/op. Applied to `_Storage` too, and `_QueryMethod` gained the `X-Cache: hit` assertion the others already had. `_Miss` and `_MaxSize` are left unprimed — they measure the miss/store path deliberately.

**Vary slice capacity** — also correct, and my regression. The early return for the empty field was the point of that change, but dropping to a nil slice meant `append` regrew the backing array, so a three-name manifest cost three allocations where it had cost one. Now sized from the separators actually present rather than the old fixed 8, bounded by the cap the loop already enforces, so it allocates once and never grows. Measured with `testing.AllocsPerRun`:

```
""                                         -> 0 allocs   (the fast path, unchanged)
"a,b,c,d,e,f,g,h,i,j"                      -> 1 alloc    (was 5 via growth)
```

Worth noting the exact-sizing beats the original too: `storeVaryManifest` writes names already lowercased, so the per-lookup `loadVaryManifest` → `parseVary` round-trip parses a lowercase string and now costs exactly one allocation.

`make test` passes with `-race`, `make lint` is clean at v2.12.2.

The benchmark-check question from my previous comment is unchanged and still needs a maintainer call — those three `MaxSize` numbers are a benchmark-definition change, not a code regression.

---

gaby:
## The benchmark check: control experiment

The report re-ran on 9e49594 with the same finding (`regressions=5` metrics across the same 3 `MaxSize` benchmarks; the job logged `CHANGED: false`, so the findings have not moved). Rather than repeat my earlier argument, here is the single-variable experiment that settles it.

**Same middleware code — `main` at 9da941f, untouched — with only the benchmark definition swapped:**

| `Benchmark_Cache_MaxSize` | 418 benchmark (as on `main`) | this PR's benchmark | |
|---|---|---|---|
| `Disabled` | 1418–1519 ns · 352 B · 6 allocs | 4285–5909 ns · 809–881 B · 9 allocs | **3.4x** |
| `Unlim` | 1444–1504 ns · 352 B · 6 allocs | 5536–6489 ns · 1115–1134 B · 9 allocs | **4.0x** |
| `LowBounded` | 1447–1532 ns · 352 B · 6 allocs | 2548–4203 ns · 785 B · 10 allocs | **2.2x** |

Not one line of middleware changed between those two columns. The entire 2.2–4.0x is the benchmark starting to store entries and run the eviction loop instead of returning `unreachable`. The bot's flagged 1.66–2.65x is *smaller* than that, because this PR's optimisations claw part of it back.

Put the two comparisons side by side for `MaxSize/Unlim`:

```
main + 418 benchmark      1444-1504 ns   352 B    6 allocs   <- what the bot compares against
main + this benchmark     5536-6489 ns  1115 B    9 allocs   <- same code, benchmark fixed
this PR                   3933-6859 ns   793 B    5 allocs   <- the actual change
```

The bot compares row 1 to row 3. The honest comparison is row 2 to row 3, and there the PR wins on every deterministic metric.

I'd still rather not rename the benchmarks unless you want it — the names describe what they measure, and renaming discards their history on `main`. Happy to do it if you'd prefer the gate green; otherwise this needs someone to accept the check.

## Also pushed: 3b76214

Codex flagged that my `parseVary` capacity fix used `strings.Count` over the whole field, while the walk below stops at `maxVaryHeaders` — so an over-long `Vary` got one full pass before being refused, which is exactly the work that cap exists to avoid. Correct, and my regression. Reverted to a fixed capacity, which restores the early exit and still keeps what the change was for:

```
""                                        -> 0 allocs   (absent/empty, almost every response)
"accept, accept-encoding, accept-language" -> 1 alloc    (was 3 with a nil slice)
"a,b,c,d,e,f,g,h"                          -> 1 alloc
```

`make test` passes with `-race`; `make lint` clean.

---

ReneWerner87 (approved):

PR #4619, opened 2026-08-15
🐛 fix(domain): expand nested mounts when cloning a sub-app

### LINKED ISSUE #4618: 🐛 [Bug]: nil pointer dereference when a domain-mounted sub-app contains a nested mount
### Bug Description

Mounting a sub-app on a domain router panics on the first request when that sub-app itself mounts another app:

```go
app.Domain("api.example").Use("/api", sub)   // where sub.Use("/v1", child)
```

`domainRouter.Use` clones the sub-app's routes so the original is not mutated. The clone is produced by `App.copyRoute`, which builds a fresh `Route` and does not carry over the unexported `group` field. For an ordinary route that is harmless, but a **mount placeholder** route carries its target app in `route.group.app`. The cloned placeholder therefore reaches startup with `mount == true` and `group == nil`, and `processSubAppsRoutes` dereferences it:

https://github.com/gofiber/fiber/blob/0724cee80bec75c0d20cedd708797facb1f9c159/mount.go#L199

```go
subRoutes := make([]*Route, len(route.group.app.stack[m]))  // route.group is nil
```

The panic surfaces from `startupProcess`, so it fires on the first request (or `Listen`), not at registration time.

### How to Reproduce

```go
package main

import (
	"net/http/httptest"

	"github.com/gofiber/fiber/v3"
)

func main() {
	ok := func(c fiber.Ctx) error { return c.SendString("ok") }

	child := fiber.New()
	child.Get("/x", ok)

	sub := fiber.New()
	sub.Use("/v1", child) // the sub-app itself mounts an app

	app := fiber.New()
	app.Domain("api.example").Use("/api", sub)

	// panics inside startupProcess
	_, _ = app.Test(httptest.NewRequest(fiber.MethodGet, "/api/v1/x", nil))
}
```

### Expected Behavior

The nested route is registered under the domain and served only for a matching host — as a plain nested mount already is, and as a flat domain mount already is. No panic.

### Actual Behavior

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0]

github.com/gofiber/fiber/v3.(*App).processSubAppsRoutes(...)
	mount.go:199
github.com/gofiber/fiber/v3.(*App).mountStartupProcess.func2()
	mount.go:124
github.com/gofiber/fiber/v3.(*App).startupProcess(...)
	app.go:1701
github.com/gofiber/fiber/v3.(*App).Test(...)
	app.go:1435
```

### Scope

Isolated against `main` at `0724cee`; only the third case fails, so this is specific to the combination of a domain mount and a nested sub-app rather than to either feature alone:

| Case | Result |
|:-----|:-------|
| `app.Use("/api", sub)` where `sub.Use("/v1", child)` — plain nested mount, no domain | ✅ 200 |
| `app.Domain("api.example").Use("/api", sub)` — sub-app has no nested mount | ✅ routes, host filtering applies |
| `app.Domain("api.example").Use("/api", sub)` where `sub.Use("/v1", child)` | 💥 panic |

### Possible Fix

Two directions, both in `domainRouter.Use`:

1. Carry `group` across the clone for mount placeholders, and recursively clone/wrap the mounted descendants so the nested app's handlers also get the domain filter. Without the recursion, a fix that only restores `group` would expand the nested app's handlers verbatim and serve them on every host — trading the panic for a host-filtering hole, so the two parts belong together.
2. Expand the sub-app's own mounts before applying the domain wrapper, so no placeholder survives into the clone.

### Fiber Version

`main` @ `0724cee` (v3.5.0). Reproduced on Go 1.25 and 1.26, linux/amd64.

### Code Snippet (optional)

_See "How to Reproduce" above._

### Checklist

lazerg:
# Description

`domainRouter.mount` clones the sub-app's routes into a wrapper app so the original is left alone. The clone comes from `App.copyRoute`, which does not carry over the unexported `group` field. That is fine for a normal route, but a mount placeholder keeps the app it points at in `route.group.app`, so if the sub-app has mounts of its own the clone reaches startup with `mount == true` and `group == nil`, and `processSubAppsRoutes` dereferences it.

The plain (non-domain) case does not hit this because every mounted app expands its own placeholders before the parent copies its stack. The wrapper app is never in `mountFields.appList`, so nothing expands its placeholders and one survives into the parent stack.

Instead of cloning placeholders, the wrapper is now built by expanding the sub-app's mounts in place, recursively, applying the placeholder prefix the same way `processSubAppsRoutes` does. The handlers of the nested apps go through `wrapHandlers` too, so they keep the host filter rather than being served on every host.

Fixes #4618

## Changes introduced

- Added `domainRouter.cloneRoutes`, which returns a sub-app's routes for one method with domain-filtered handlers and mounted apps expanded in place.
- `domainRouter.mount` builds the wrapper app through it.

## Type of change

- [x] Bug fix (non-breaking change which corrects existing functionality)

## Checklist

- [x] Conducted a self-review of the code.
- [x] Added a unit test, `Test_Domain_UseMountNested`, which panics on `main` and passes here. It covers two levels of nesting and checks that the nested routes still 404 on a non-matching host.
- [x] Ensured that new and existing unit tests pass locally, `go test -race ./...` and `make lint` are clean.
- [x] No new dependencies.
- [x] The expansion happens once at mount time, so request-time routing is unchanged.

### DIFF
--- a/domain.go
+++ b/domain.go
@@ -431,13 +431,7 @@ func (d *domainRouter) mount(prefix string, subApp *App) Router {
 	subApp.mutex.Lock()
 	defer subApp.mutex.Unlock()
 	for m := range subApp.stack {
-		for _, route := range subApp.stack[m] {
-			clonedRoute := subApp.copyRoute(route)
-			if len(clonedRoute.Handlers) > 0 {
-				clonedRoute.Handlers = d.wrapHandlers(clonedRoute.Handlers)
-			}
-			wrapperApp.stack[m] = append(wrapperApp.stack[m], clonedRoute)
-		}
+		wrapperApp.stack[m] = append(wrapperApp.stack[m], d.cloneRoutes(subApp, m)...)
 	}
 
 	d.app.mutex.Lock()
@@ -471,6 +465,34 @@ func (d *domainRouter) mount(prefix string, subApp *App) Router {
 	}
 
 	return d
+}
+
+// cloneRoutes returns the routes app has registered for a single HTTP method,
+// cloned with domain-filtered handlers. Apps mounted on app are expanded in
+// place rather than cloned as mount placeholders: copyRoute does not carry over
+// the unexported group field, so a placeholder surviving into the clone would
+// leave processSubAppsRoutes dereferencing a nil group at startup.
+func (d *domainRouter) cloneRoutes(app *App, m int) []*Route {
+	routes := make([]*Route, 0, len(app.stack[m]))
+
+	for _, route := range app.stack[m] {
+		if route.mount {
+			mountedApp := route.group.app
+			for _, mountedRoute := range d.cloneRoutes(mountedApp, m) {
+				routes = append(routes, app.addPrefixToRoute(route.path, mountedRoute, mountedApp.config.RegexHandler, mountedApp.customConstraints...))
+			}
+
+			continue
+		}
+
+		clonedRoute := app.copyRoute(route)
+		if len(clonedRoute.Handlers) > 0 {
+			clonedRoute.Handlers = d.wrapHandlers(clonedRoute.Handlers)
+		}
+		routes = append(routes, clonedRoute)
+	}
+
+	return routes
 }
 
 // Get registers a route for GET methods.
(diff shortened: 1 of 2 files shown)


gaby:
@lazerg The team is already working on this, please ask first to be assigned issues.

PR #4620, opened 2026-08-15
🐛 fix: nil pointer dereference when a domain-mounted sub-app contains a nested mount

### LINKED ISSUE #4618: 🐛 [Bug]: nil pointer dereference when a domain-mounted sub-app contains a nested mount

gaby:
# Description

Mounting a sub-app on a domain router panicked on the first request when that sub-app itself mounted another app:

```go
app.Domain("api.example").Use("/api", sub)   // where sub.Use("/v1", child)
```

`domainRouter.mount` clones the sub-app's routes into a wrapper app so the original is left alone. `copyRoute` does not carry the unexported `group` field over, and a **mount placeholder** keeps its target app in `route.group.app` — so the cloned placeholder reached startup with `mount` set and `group` nil, and `processSubAppsRoutes` dereferenced it. The panic surfaces from `startupProcess`, i.e. on the first request or `Listen`, not at registration.

The clone is now recursive: an app the sub-app has mounted is expanded in its placeholder's position, so no placeholder survives into the wrapper and the nested handlers go through `wrapHandlers` like the rest. Restoring `group` alone would have expanded those handlers verbatim and served them on every host — trading the panic for a host-filtering hole — so the two halves belong together.

Fixes #4618

## How this grew

The panic itself is `5e31ca3`, and it is small. Everything else came out of review rounds — CodeRabbit and Codex between them raised findings on almost every commit, and each one that reproduced got its own fix and its own test. Every test added here was run against the unfixed code first to confirm it fails there.

Two things are worth a reviewer's attention up front:

**The work splits cleanly in two.** The panic fix and the mount-expansion fixes have been quiet since `b088e52`. The host-scoped `ErrorHandler`/`Views` feature (`28891f4`) has drawn a finding in every round since, and accounts for most of the diff and nearly all of the churn. If you would rather ship #4618 on its own, reverting `28891f4` and its follow-ups leaves the panic fixed and the branch much smaller. Say the word and I will do the split; the feature can come back as its own PR where a behaviour change to config resolution gets the attention it deserves.

**Fixing it well meant recording provenance rather than inferring it.** Three separate classes of bug here came from guessing which app a request belongs to from its host and path. Routes now carry the app they were cloned from, the constraints they were parsed with, and whether their mount registers automatic HEAD routes; mount records carry the apps they are registered inside and the patterns that reach them. Each time an inference was replaced with a recorded fact, a whole class of finding closed rather than one instance of it.

## What changed

**The #4618 panic** — `5e31ca3`. Each app in the mount tree is snapshotted under its own lock and cloned once, mounted apps flattened once and reused across method indexes; no two app locks are ever held at once, and the walk carries the branch it came from so a mount cycle terminates. Route values are copied under the source app's lock (`5212e93`), since registering a route on a path an app already serves appends to that route's handlers in place.

**Expanding a mount by method, not by stack index** — `3b134a6`. `processSubAppsRoutes` read `route.group.app.stack[m]`, where `m` indexes the *parent's* `RequestMethods`. That table is per-app configurable, so the two need not agree: a sub-app serving fewer methods panicked with an index out of range, and one listing the same methods in another order had its routes served under the wrong verb — a GET reaching the POST handler.

**Automatic HEAD routes** — `525c305`, `85b4f43`, `6885aa3`, `49c746c`. Domain-mounted GET routes answered HEAD with 405 where a plain mount answered 200, because the wrapper is deliberately outside the parent's app list and nothing generated them. Generating them then exposed the opposite problem: a sub-app that omits HEAD from `RequestMethods` contributes no HEAD middleware, so a synthesized HEAD route would run its handler with the `Use` gate missing — `HEAD /api/secret` returning 200 where GET returned 401. The opt-out is recorded per route as the routes are cloned, covering the app that opted out and the apps it stands in front of, and nothing else. Deduplication is per mounted app inside a wrapper (`e4a24ea`), where several apps' routes sit side by side under different patterns.

**Custom constraints** — `77e4771`, `2f3e62b`. Expanding a mount re-parses the sub-app's routes against the app they move into, and an app two mounts deep had its constraints dropped: `findConstraintHandler` returns nil for a name it has never seen, the constraint is silently discarded, and the route goes on matching while validating nothing. On `main` that means a value every constraint rejects is accepted. Constraints travel with each route now, per route rather than merged into a shared registry — two apps mounted side by side may name one constraint differently, and neither binds the other.

**Route params and mount-path matching** — `b088e52`. `addPrefixToRoute` rebuilt the path without rebuilding `Params`, and mount paths were compared literally where the router would normalize them.

**Host-scoped config** — `28891f4` and follow-ups. Mounting registered the sub-app in `mountFields.appList`, keyed by path alone: two sub-apps mounted at the same path on different domains displaced each other, and a sub-app's `ErrorHandler` and view engine answered on hosts its own pattern rejects. Domain mounts now live in a list of their own carrying the patterns they were mounted behind. Resolving them took the rest of the rounds: the owner comes from the route that ran, inheritance follows recorded mount ancestry rather than path coverage, a mount is superseded only by one that did not serve the request, and the view engine and layout are resolved in one walk that stops where rendering does.

**Locking** — `b77c2ef`, `3009488`, `5212e93`. Two app locks are never held at once; `ReloadViews` walks the mount tree taking each app's lock in turn; each app's mount path is written under its own lock.

## Type of change

- [x] Documentation Update: [`docs/api/app.md`](https://github.com/gofiber/fiber/blob/main/docs/api/app.md) — the `Domain` note now covers nested mounts and the host-scoped `ErrorHandler`/`Views` resolution.
- [x] Examples: the reproduction in #4618 now serves `/api/v1/x` on the matching host and 404s elsewhere.
- [ ] Benchmarks: not applicable — the new work is all registration-time. Request-time behaviour is unchanged apart from a hostname check in `ErrorHandler` and `Render`, and only for apps that have domain mounts.
- [ ] Migration Guide: not needed.

## Checklist

- [x] Added or updated unit tests to validate the effectiveness of the changes.
- [x] Verified that any new dependencies are essential — none were added.

Full suite, `go test -race`, `golangci-lint` (repo-pinned v2.12.2), `gofumpt` and `go vet` are clean, along with `go test -count=3` on the root package — the repeated CI job caught a bug a single local run had missed, since map iteration order decides which of two paths through the mount tree is recorded. Patch coverage is complete.

## Known limitations, left alone deliberately

All three are pre-existing and reproduce on `main` with ordinary `Use`, with no domain router involved. Fixing any of them changes how `Use` behaves generally, which seemed worth a maintainer's decision rather than a unilateral divergence on the domain path:

- **A nested app's `Config.RegexHandler` is dropped** when its routes are expanded. `processSubAppsRoutes` re-parses a composed path with the *mounted* app's compiler, so the compiler of the app that wrote the prefix is lost at the first level of nesting — `mid` alone, with no parent, already fails to apply its own. Unlike the constraint list it is a single function, and a composed path can carry segments from several apps, so it cannot simply be merged.
- **`App.mount` and `Group.mount` write the mounted app's `mountPath` while holding only the mounting app's lock**, which races a route registration on the mounted app. The domain path was fixed here; the ordinary one is untouched.
- **A cyclic mount graph** (`a.Use("/b", b)` with `b.Use("/a", a)`) hangs at startup in `appendSubAppLists`. The walks added here terminate on such a graph, but that does not make the app servable.

---
_Generated by [Claude Code](https://claude.ai/code/session_017fX7w3sDadh6r9fhuuXR2e)_

### DIFF
--- a/app.go
+++ b/app.go
@@ -20,6 +20,7 @@ import (
 	"net/http/httputil"
 	"os"
 	"reflect"
+	"slices"
 	"strconv"
 	"strings"
 	"sync"
@@ -1589,6 +1590,7 @@ func (app *App) ErrorHandler(ctx Ctx, err error) error {
 
 	var (
 		mountedErrHandler  ErrorHandler
+		mountedErrApp      *App
 		mountedPrefixParts int
 	)
 
@@ -1601,6 +1603,7 @@ func (app *App) ErrorHandler(ctx Ctx, err error) error {
 			if mountedPrefixParts <= parts {
 				if subApp.configured.ErrorHandler != nil {
 					mountedErrHandler = subApp.config.ErrorHandler
+					mountedErrApp = subApp
 				}
 
 				mountedPrefixParts = parts
@@ -1618,7 +1621,11 @@ func (app *App) ErrorHandler(ctx Ctx, err error) error {
 		// above them both still encloses the owner, and an owner configuring
 		// no handler of its own inherits it, as a nested mount inherits the
 		// handler of the mount it sits in.
-		if owner.ties(mountedPrefixParts) {
+		//
+		// Unless the mount is one of the apps the owner sits in: a domain
+		// mount at the root of a plain one covers exactly the paths that mount
+		// does, and is as deep as it without being beside it.
+		if owner.ties(mountedPrefixParts) && !slices.Contains(owner.ancestors, mountedErrApp) {
 			mountedErrHandler = nil
 		}
 
--- a/app.go
+++ b/app.go
@@ -1599,7 +1599,7 @@ func (app *App) ErrorHandler(ctx Ctx, err error) error {
 
 		if app.mountCoversPath(prefix, ctx.Path()) {
 			// Count slashes instead of splitting - more efficient
-			parts := strings.Count(prefix, "/") + 1
+			parts := mountDepth(prefix)
 			if mountedPrefixParts <= parts {
 				if subApp.configured.ErrorHandler != nil {
 					mountedErrHandler = subApp.config.ErrorHandler
@@ -1611,27 +1611,16 @@ func (app *App) ErrorHandler(ctx Ctx, err error) error {
 	}
 
 	// Sub-apps mounted on a domain answer only for a matching host, so their
-	// error handler cannot be picked by path alone. They are considered after
-	// the plain mounts, which lets a domain mount win a tie on path depth: it
-	// matched the host too, so it is the more specific of the two.
-	//
-	// Between two domain mounts at the same depth the first one registered
-	// wins, matching the order their routes are matched in — a pattern that
-	// only overlaps a later one still keeps its own handler.
-	domainPrefixParts := -1
-
-	for i := range app.mountFields.domainAppList {
-		mount := &app.mountFields.domainAppList[i]
-
-		if app.mountCoversPath(mount.path, ctx.Path()) && mount.matchesHost(ctx.Hostname()) {
-			if parts := mount.prefixParts(); mountedPrefixParts <= parts && domainPrefixParts < parts {
-				if mount.app.configured.ErrorHandler != nil {
-					mountedErrHandler = mount.app.config.ErrorHandler
-				}
-
-				mountedPrefixParts = parts
-				domainPrefixParts = parts
-			}
+	// error handler cannot be picked by path alone. The owner is considered
+	// after the plain mounts, which lets a domain mount win a tie on path
+	// depth: it matched the host too, so it is the more specific of the two.
+	if owner, depth := app.domainMountOwner(ctx); owner != nil && mountedPrefixParts <= depth {
+		// The owner decides, as it does among the plain mounts: one without a
+		// handler of its own falls through to this app's, rather than to the
+		// handler of a shallower mount that did not serve the request.
+		mountedErrHandler = nil
+		if owner.configured.ErrorHandler != nil {
+			mountedErrHandler = owner.config.ErrorHandler
 		}
 	}
 
--- a/app.go
+++ b/app.go
@@ -1606,13 +1606,10 @@ func (app *App) ErrorHandler(ctx Ctx, err error) error {
 		mountedPrefixParts int
 	)
 
-	normalizedPath := utils.AddTrailingSlashString(ctx.Path())
-
 	for _, prefix := range app.mountFields.appListKeys {
 		subApp := app.mountFields.appList[prefix]
-		normalizedPrefix := utils.AddTrailingSlashString(prefix)
 
-		if prefix != "" && strings.HasPrefix(normalizedPath, normalizedPrefix) {
+		if app.mountCoversPath(prefix, ctx.Path()) {
 			// Count slashes instead of splitting - more efficient
 			parts := strings.Count(prefix, "/") + 1
 			if mountedPrefixParts <= parts {
@@ -1637,9 +1634,8 @@ func (app *App) ErrorHandler(ctx Ctx, err error) error {
 
 	for i := range app.mountFields.domainAppList {
 		mount := &app.mountFields.domainAppList[i]
-		normalizedPrefix := utils.AddTrailingSlashString(mount.path)
 
-		if mount.path != "" && strings.HasPrefix(normalizedPath, normalizedPrefix) && mount.matchesHost(ctx.Hostname()) {
+		if app.mountCoversPath(mount.path, ctx.Path()) && mount.matchesHost(ctx.Hostname()) {
 			if parts := mount.prefixParts(); mountedPrefixParts <= parts && domainPrefixParts < parts {
 				if mount.app.configured.ErrorHandler != nil {
 					mountedErrHandler = mount.app.config.ErrorHandler
--- a/app.go
+++ b/app.go
@@ -1613,10 +1613,15 @@ func (app *App) ErrorHandler(ctx Ctx, err error) error {
 	// after the plain mounts, which lets a domain mount win a tie on path
 	// depth: it matched the host too, so it is the more specific of the two.
 	if owner := app.domainMountOwner(ctx); owner.outranks(mountedPrefixParts) {
-		// The owner decides, as it does among the plain mounts: one without a
-		// handler of its own falls through to this app's, rather than to the
-		// handler of a shallower mount that did not serve the request.
-		mountedErrHandler = nil
+		// A plain mount as deep as the owner is a sibling that did not serve
+		// the request, so its handler is dropped rather than inherited. One
+		// above them both still encloses the owner, and an owner configuring
+		// no handler of its own inherits it, as a nested mount inherits the
+		// handler of the mount it sits in.
+		if owner.ties(mountedPrefixParts) {
+			mountedErrHandler = nil
+		}
+
 		if owner.app.configured.ErrorHandler != nil {
 			mountedErrHandler = owner.app.config.ErrorHandler
 		}
--- a/app.go
+++ b/app.go
@@ -1629,17 +1629,24 @@ func (app *App) ErrorHandler(ctx Ctx, err error) error {
 	// error handler cannot be picked by path alone. They are considered after
 	// the plain mounts, which lets a domain mount win a tie on path depth: it
 	// matched the host too, so it is the more specific of the two.
+	//
+	// Between two domain mounts at the same depth the first one registered
+	// wins, matching the order their routes are matched in — a pattern that
+	// only overlaps a later one still keeps its own handler.
+	domainPrefixParts := -1
+
 	for i := range app.mountFields.domainAppList {
 		mount := &app.mountFields.domainAppList[i]
 		normalizedPrefix := utils.AddTrailingSlashString(mount.path)
 
 		if mount.path != "" && strings.HasPrefix(normalizedPath, normalizedPrefix) && mount.matchesHost(ctx.Hostname()) {
-			if parts := mount.prefixParts(); mountedPrefixParts <= parts {
+			if parts := mount.prefixParts(); mountedPrefixParts <= parts && domainPrefixParts < parts {
 				if mount.app.configured.ErrorHandler != nil {
 					mountedErrHandler = mount.app.config.ErrorHandler
 				}
 
 				mountedPrefixParts = parts
+				domainPrefixParts = parts
 			}
 		}
 	}
--- a/app.go
+++ b/app.go
@@ -20,7 +20,6 @@ import (
 	"net/http/httputil"
 	"os"
 	"reflect"
-	"slices"
 	"strconv"
 	"strings"
 	"sync"
@@ -1616,16 +1615,15 @@ func (app *App) ErrorHandler(ctx Ctx, err error) error {
 	// after the plain mounts, which lets a domain mount win a tie on path
 	// depth: it matched the host too, so it is the more specific of the two.
 	if owner := app.domainMountOwner(ctx); owner.outranks(mountedPrefixParts) {
-		// A plain mount as deep as the owner is a sibling that did not serve
-		// the request, so its handler is dropped rather than inherited. One
-		// above them both still encloses the owner, and an owner configuring
-		// no handler of its own inherits it, as a nested mount inherits the
-		// handler of the mount it sits in.
+		// A plain mount the owner supersedes did not serve the request, so its
+		// handler is dropped rather than inherited. One enclosing the owner
+		// still does, and an owner configuring no handler of its own inherits
+		// it, as a nested mount inherits the handler of the mount it sits in.
 		//
 		// Unless the mount is one of the apps the owner sits in: a domain
 		// mount at the root of a plain one covers exactly the paths that mount
 		// does, and is as deep as it without being beside it.
-		if owner.ties(mountedPrefixParts) && !slices.Contains(owner.ancestors, mountedErrApp) {
+		if owner.supersedes(mountedPrefixParts, mountedErrApp) {
 			mountedErrHandler = nil
 		}
 
--- a/app.go
+++ b/app.go
@@ -20,6 +20,7 @@ import (
 	"net/http/httputil"
 	"os"
 	"reflect"
+	"slices"
 	"strconv"
 	"strings"
 	"sync"
@@ -901,9 +902,19 @@ func (app *App) ReloadViews() error {
 	app.mutex.Lock()
 	defer app.mutex.Unlock()
 
-	apps := map[string]*App{"": app}
+	apps := []*App{app}
 	if app.mountFields != nil {
-		apps = app.mountFields.appList
+		apps = apps[:0]
+		for _, subApp := range app.mountFields.appList {
+			apps = append(apps, subApp)
+		}
+		// Domain mounts are kept out of appList, but their view engines still
+		// have to be reloaded.
+		for i := range app.mountFields.domainAppList {
+			if mount := &app.mountFields.domainAppList[i]; !slices.Contains(apps, mount.app) {
+				apps = append(apps, mount.app)
+			}
+		}
 	}
 
 	var reloaded bool
@@ -1585,7 +1596,7 @@ func (app *App) init() *App {
 // the app, which if not set is the DefaultErrorHandler.
 func (app *App) ErrorHandler(ctx Ctx, err error) error {
 	// Fast path: no mounted sub-apps, so no prefix lookup is needed
-	if len(app.mountFields.appListKeys) == 0 {
+	if len(app.mountFields.appListKeys) == 0 && len(app.mountFields.domainAppList) == 0 {
 		return app.config.ErrorHandler(ctx, err)
 	}
 
@@ -1606,6 +1617,25 @@ func (app *App) ErrorHandler(ctx Ctx, err error) error {
 			if mountedPrefixParts <= parts {
 				if subApp.configured.ErrorHandler != nil {
 					mountedErrHandler = subApp.config.ErrorHandler
+				}
+
+				mountedPrefixParts = parts
+			}
+		}
+	}
+
+	// Sub-apps mounted on a domain answer only for a matching host, so their
+	// error handler cannot be picked by path alone. They are considered after
+	// the plain mounts, which lets a domain mount win a tie on path depth: it
+	// matched the host too, so it is the more specific of the two.
+	for i := range app.mountFields.domainAppList {
+		mount := &app.mountFields.domainAppList[i]
+		normalizedPrefix := utils.AddTrailingSlashString(mount.path)
+
+		if mount.path != "" && strings.HasPrefix(normalizedPath, normalizedPrefix) && mount.matchesHost(ctx.Hostname()) {
+			if parts := mount.prefixParts(); mountedPrefixParts <= parts {
+				if mount.app.configured.ErrorHandler != nil {
+					mountedErrHandler = mount.app.config.ErrorHandler
 				}
 
 				mountedPrefixParts = parts
--- a/app.go
+++ b/app.go
@@ -20,7 +20,6 @@ import (
 	"net/http/httputil"
 	"os"
 	"reflect"
-	"slices"
 	"strconv"
 	"strings"
 	"sync"
@@ -902,21 +901,10 @@ func (app *App) ReloadViews() error {
 	app.mutex.Lock()
 	defer app.mutex.Unlock()
 
-	apps := []*App{app}
-	if app.mountFields != nil {
-		for _, subApp := range app.mountFields.appList {
-			if !slices.Contains(apps, subApp) {
-				apps = append(apps, subApp)
-			}
-		}
-		// Domain mounts are kept out of appList, but their view engines still
-		// have to be reloaded.
-		for i := range app.mountFields.domainAppList {
-			if mount := &app.mountFields.domainAppList[i]; !slices.Contains(apps, mount.app) {
-				apps = append(apps, mount.app)
-			}
-		}
-	}
+	// Walks the mount metadata rather than the flattened app list: a domain
+	// mount nested inside a plain one only reaches the root's list once the
+	// app has started, and ReloadViews may be called before that.
+	apps := app.mountedApps()
 
 	var reloaded bool
 	for _, targetApp := range apps {
--- a/app.go
+++ b/app.go
@@ -1622,7 +1622,7 @@ func (app *App) ErrorHandler(ctx Ctx, err error) error {
 			mountedErrHandler = nil
 		}
 
-		if handler := app.domainMountConfig(ctx, owner, appHasErrorHandler); handler != nil {
+		if handler := domainMountConfig(owner, appHasErrorHandler); handler != nil {
 			mountedErrHandler = handler.config.ErrorHandler
 		}
 	}
--- a/app.go
+++ b/app.go
@@ -1612,13 +1612,13 @@ func (app *App) ErrorHandler(ctx Ctx, err error) error {
 	// error handler cannot be picked by path alone. The owner is considered
 	// after the plain mounts, which lets a domain mount win a tie on path
 	// depth: it matched the host too, so it is the more specific of the two.
-	if owner, depth := app.domainMountOwner(ctx); owner != nil && mountedPrefixParts <= depth {
+	if owner := app.domainMountOwner(ctx); owner.outranks(mountedPrefixParts) {
 		// The owner decides, as it does among the plain mounts: one without a
 		// handler of its own falls through to this app's, rather than to the
 		// handler of a shallower mount that did not serve the request.
 		mountedErrHandler = nil
-		if owner.configured.ErrorHandler != nil {
-			mountedErrHandler = owner.config.ErrorHandler
+		if owner.app.configured.ErrorHandler != nil {
+			mountedErrHandler = owner.app.config.ErrorHandler
 		}
 	}
 
--- a/app.go
+++ b/app.go
@@ -1622,8 +1622,8 @@ func (app *App) ErrorHandler(ctx Ctx, err error) error {
 			mountedErrHandler = nil
 		}
 
-		if owner.app.configured.ErrorHandler != nil {
-			mountedErrHandler = owner.app.config.ErrorHandler
+		if handler := app.domainMountConfig(ctx, owner, appHasErrorHandler); handler != nil {
+			mountedErrHandler = handler.config.ErrorHandler
 		}
 	}
 
--- a/app.go
+++ b/app.go
@@ -898,12 +898,10 @@ func (app *App) RegisterCustomBinder(customBinder CustomBinder) {
 // ReloadViews reloads the configured view engine by invoking its Load method.
 // It returns an error if no view engine is configured or if reloading fails.
 func (app *App) ReloadViews() error {
-	app.mutex.Lock()
-	defer app.mutex.Unlock()
-
 	// Walks the mount metadata rather than the flattened app list: a domain
 	// mount nested inside a plain one only reaches the root's list once the
-	// app has started, and ReloadViews may be called before that.
+	// app has started, and ReloadViews may be called before that. The walk
+	// takes each app's own lock in turn, so this must not hold one itself.
 	apps := app.mountedApps()
 
 	var reloaded bool
--- a/app.go
+++ b/app.go
@@ -904,9 +904,10 @@ func (app *App) ReloadViews() error {
 
 	apps := []*App{app}
 	if app.mountFields != nil {
-		apps = apps[:0]
 		for _, subApp := range app.mountFields.appList {
-			apps = append(apps, subApp)
+			if !slices.Contains(apps, subApp) {
+				apps = append(apps, subApp)
+			}
 		}
 		// Domain mounts are kept out of appList, but their view engines still
 		// have to be reloaded.
--- a/app_test.go
+++ b/app_test.go
@@ -2479,6 +2479,42 @@ func Test_App_ReloadViews_MountedViews_MultipleApps(t *testing.T) {
 	require.Equal(t, initialLoadsB+1, viewB.loads)
 }
 
+func Test_App_ReloadViews_MountedViews_ConcurrentMount(t *testing.T) {
+	t.Parallel()
+	view := &countingView{}
+	subApp := New()
+	app := New(Config{Views: view})
+	app.Use("/sub", subApp)
+
+	// The walk down the mount tree reads every app's mount metadata, so it has
+	// to take each app's own lock rather than only the one it started from.
+	initialLoads := view.loads
+
+	var (
+		wg        sync.WaitGroup
+		reloadErr error
+	)
+
+	wg.Go(func() {
+		for range 200 {
+			if err := app.ReloadViews(); err != nil {
+				reloadErr = err
+				return
+			}
+		}
+	})
+
+	wg.Go(func() {
+		for range 200 {
+			subApp.Use("/nested", New())
+		}
+	})
+
+	wg.Wait()
+	require.NoError(t, reloadErr)
+	require.Equal(t, initialLoads+200, view.loads)
+}
+
 func Test_App_ReloadViews_MountedViews_SharedEngineBlocksSiblingRender(t *testing.T) {
 	t.Parallel()
 
--- a/constraint.go
+++ b/constraint.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"reflect"
 	"regexp"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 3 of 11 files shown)


gaby:
Thanks — the automated reviews caught a real one. Summary of what I acted on and what I did not, with the checks I ran for each.

## Fixed

**HEAD bypassing middleware** (`85b4f43`). Reproduced exactly as described: a sub-app whose `RequestMethods` excludes HEAD registers its `Use` middleware for its own methods only, so a synthesized HEAD route ran the GET handler with nothing in front of it — `GET /api/secret` → 401, `HEAD /api/secret` → 200 with the handler's headers.

Two places had to change, because the synthesis happens twice:

- the domain wrapper no longer synthesizes unless every app behind the mount registers automatic HEAD routes itself;
- `ensureAutoHeadRoutesLocked` now skips a route whose path belongs to a mounted app that registers none. Without this the parent's own pass re-created the unprotected route once the mount had been expanded into its stack. **This affects plain mounts identically** — `app.Use("/api", sub)` with the same sub-app leaked the same way — so the fix is in the shared path rather than the domain one. On `main` that configuration panics before it can leak (`index out of range`), which is why it has not been seen.

`DisableHeadAutoRegister` on a nested app is honoured by the same rule.

**Lock nesting** (`b77c2ef`). Correct, and reproducible: `app.Domain("h").Use("/self", app)` hung on the second acquire. The sub-app's mount lists are now snapshotted under its own lock and released before the parent's is taken, so two app locks are never held at once. The self-mount test asserts bounded completion rather than `NotPanics`, since a deadlock is what a regression would look like.

**Views matched with `strings.Contains`** (`b77c2ef`). Fixed before the second review landed — it now prefix-matches the request path, as `ErrorHandler` does for the same entries. `/elsewhere?next=/api/view` no longer picks the mounted engine.

**Domain mounts registered after the outer mount** (`85b4f43`). Real regression from this PR: `app.Use("/mid", mid)` followed by `mid.Domain(...).Use("/child", child)` left the child's `ErrorHandler` and `Views` unreachable while its routes worked. Collection moved into the startup discovery pass, so the late registration is picked up.

**Overlapping domain patterns** (`85b4f43`). Between two domain mounts at one path whose patterns both match, the first registered now keeps its error handler, matching the order their routes are matched in.

## Not fixed here, with evidence

**Parametric mount prefixes.** `addPrefixToRoute` rewrites the parser but not `Route.Params`, so a mount at `/v1/:version` yields a route that never matches. This is not specific to nesting or to domains — a single plain mount reproduces it on this branch and on `main`:

```
flat plain param prefix (app.Use("/v1/:version", child)):  404
plain nested param prefix:                                  404
domain nested param prefix:                                 404
```

**Case-insensitive mount paths.** `app.Use("/API", sub)` and `app.Domain(...).Use("/API", sub)` both fail the prefix comparison for `/api/boom` and fall through to the root handler — identical on plain and domain mounts, and identical on `main`.

Both are pre-existing defects in the shared mount path. They are worth their own issue; folding speculative fixes for them into a bugfix PR would widen the blast radius well past what #4618 needs. Happy to open one, or to fix them here if a maintainer prefers.

**Same-named constraints in nested apps.** Correct that a collision resolves to the outer app's implementation. Worth noting what it replaces: before this PR the nested app's constraint was dropped entirely, so the route validated *nothing* and accepted every value. The merge only shadows on an exact name collision and never makes a previously-correct case wrong. A complete fix needs per-route constraint provenance, which is a larger change than this PR should carry.

**Domain parameters across nested domain mounts.** Real: two `wrapHandlers` layers write the same `Locals` key, so the outer pattern's parameters are overwritten. This shape is newly reachable — it panicked before — and merging parameter sets needs a defined policy for duplicate names, so I would rather do it deliberately than as a follow-on here.

Every fix above is pinned by a test that fails without it. Full suite, `-race`, `golangci-lint`, `gofumpt` and `go vet` are clean; codecov reports the patch fully covered.

---

gaby:
@claude lets fix the 2 pre-existing issues found in this branch. Push changes to this branch.

gaby:
Both pre-existing issues are fixed and pushed as `b088e52`.

**Parametric mount prefixes.** `addPrefixToRoute` rewrote the path and the parser but left `Route.Params` alone, and `Params` is what decides whether the router matches a route by pattern or by string equality — so every route of a sub-app mounted at `/v1/:version` fell through to an exact-string compare and 404'd. They are rebuilt from the prefixed path now, read from its unprettified form so parameter names keep the case they were registered with (`:Version` stays addressable as `c.Params("Version")`). The extra parse is skipped when prettifying is a no-op.

**Case-insensitive mount paths.** A mount is stored as the caller spelled it, so `app.Use("/API", sub)` on a case-insensitive app served `/api/x` while the sub-app's `ErrorHandler` and view engine did not answer for it — the lookups compared `/API` against a normalized request path and missed. Both lookups now normalize each side with the routing rules of the app the mount was registered on, so config resolution follows the same rules the routes do. A case-sensitive app still keeps `/API` and `/api` apart.

One related change came with it: `Render` selected a mounted app's views with `strings.Contains` over the whole URL, so `/elsewhere?next=/api/view` rendered with the mount's engine. It matches the request path by prefix now, like the error-handler lookup beside it.

All three are covered by tests on both the plain and the domain path, each verified to fail without the fix:

| Test | Covers |
|:--|:--|
| `Test_App_Mount_ParametricPrefix` | flat mount at `/v1/:Version`, mixed-case param name |
| `Test_App_Mount_ParametricPrefixNested` | prefix applied twice, by an outer mount |
| `Test_Domain_UseMountParametricPrefix` | same under a domain mount, host filtering intact |
| `Test_App_Mount_CaseInsensitivePath` | `/API` mount answering `/api`, and a case-sensitive app not |
| `Test_Domain_UseMountCaseInsensitivePath` | same for a domain mount, ErrorHandler and Views |
| `Test_App_Mount_ViewsPathOnly` | mount path in a query string no longer selects its views |

Full suite, `-race`, `golangci-lint`, `gofumpt` and `go vet` are clean, and codecov shows the patch fully covered.

---

gaby:
All five findings from the latest Codex review reproduced, and are fixed in `f97252d`.

**Automatic HEAD routes were withheld by path.** Two ways that went wrong: a mount at `/` covers every path in the app above it, and a mount sharing a path with a route that app registered itself covers that one too — so a sub-app registering no automatic HEAD routes took the parent's HEAD routes down with it, on any host, for a domain mount whose pattern the request never matched. Provenance is now recorded per route as a mount is expanded, and travels with the clones as they move up a level, so only the mount's own routes are withheld. The domain wrapper carries the answer for its whole tree, which let the path-ownership guesswork and two helpers go.

**Domain mounts were recorded once per path leading to them.** An app list holds the descendants of the apps it lists, so the startup walk arrives at the same mount by every route through the graph: six ordinary layers produced 32 copies of one mount, doubling per layer, and every error-handler and view lookup scanned all of them. They are deduplicated as they are recorded now, at registration and at startup alike.

**`ReloadViews` before the first request.** It read the flattened app lists, and a domain mount nested inside an ordinary one only reaches the root's list at startup — so the child's engine was missed, and with nothing else configured the call returned `ErrNoViewEngineConfigured`. It walks the mount metadata itself now.

**Views borrowed from a later overlapping mount.** Selection skipped mounts without an engine *before* ranking them, so a mount that owned the route but configured no views handed the request to a later overlapping mount's engine (in the reproduction, a 500 from a template the other app never had). Ownership is settled first — by depth, then registration order — matching what the error-handler lookup already did; an owner without views falls through to the root.

**Constraints named in a mount prefix.** The combined path is re-parsed with the *mounted* app's constraints, which never include the mounting app's, so a constraint written into the prefix was dropped. Harmless while the route matched nothing; with route parameters now rebuilt it became an active pattern matching everything the constraint existed to reject — `/bar/x` returned 200. The re-parse gets both sets, the mounted app's winning where a name is defined twice.

Five tests added, each verified to fail without its fix. Full suite, `-race`, `golangci-lint`, `gofumpt` and `go vet` clean; patch coverage complete.

Worth noting for whoever reviews: the last three rounds have all been in the automatic-HEAD area, and each fix there exposed the next edge. The current version stopped guessing from paths and records provenance instead, which is what the earlier rounds were approximating. If the maintainers would rather keep this PR to the #4618 panic, the HEAD work (`525c305`, `85b4f43`, `6885aa3`, and the HEAD part of this commit) is separable from the nil-deref fix and the host-scoping work.

---

gaby:
The three findings on `f97252d` reproduced and are fixed in `59f0fb6` — a parametric mount path never matched the lookup (`/:tenant` compared literally), the root mount counted as deep as a one-segment mount and won ties against it, and views were resolved by checking domain mounts before plain ones instead of ranking them, so a mount at `/api` answered for one at `/api/admin`. Mount paths are parsed once when recorded and matched as patterns; depth counts covered segments, which is none for the root; and both lookups now go through one owner resolution.

**A note on where this PR stands, for whoever picks it up.**

Twelve commits in, the pattern is worth naming. The last four review rounds have all landed in one commit's blast radius — `28891f4`, host-scoped `ErrorHandler`/`Views`. That commit alone has now needed ten follow-up fixes: collision handling, host leak, sub-app mutation, error-handler ties, deduplication, `ReloadViews` traversal, view ownership, parametric paths, root depth, and plain-vs-domain ranking.

The reason is structural, not incidental: it introduces a second mount-resolution path beside `mountFields.appList`, and every rule the original has — depth, ties, path matching, discovery timing — has to be re-derived and kept consistent with it. Each round has found another rule that had not been.

Split by stability, the branch is two very different halves:

**Stable — no findings in the last four rounds:**
- `5e31ca3` the #4618 panic itself
- `3b134a6` method-table expansion
- `77e4771` constraint propagation
- `b088e52` route params and mount-path normalization
- `b77c2ef` lock nesting, `6b5e9cc` docs

**Churning:**
- `28891f4` host-scoped config, plus its follow-ups in `f97252d` and `59f0fb6`
- the automatic-HEAD work (`525c305`, `85b4f43`, `6885aa3`), which took three rounds to settle on per-route provenance

My recommendation: merge the stable half for #4618, and take the host-scoping feature as its own PR where it can get the review attention a behavior change to `ErrorHandler`/`Views` resolution deserves. It is a clean revert — `git revert 28891f4` plus the two follow-up commits — and I am happy to do the split myself if you would like it that way. If you would rather keep it all together, that is fine too; it is green and covered, and I will keep working the review rounds.

Full suite, `-race`, `golangci-lint`, `gofumpt` and `go vet` clean on `59f0fb6`.

---

gaby:
Both findings on `59f0fb6` reproduced and are fixed.

**`3009488` — data race walking the mount tree.** `ReloadViews` walks the mount metadata down the tree, but held only the root's mutex while reading every descendant's `appList`; `App.mount` writes those lists under the *child's* mutex, so a reload racing a mount on an already-mounted sub-app was an unsynchronised map access. Each app's lists are now read under that app's own lock, one at a time, and `ReloadViews` drops its outer lock so the walk can take the receiver's. Regression test under `-race`.

**`0e549dc` — owner resolved from the route that ran.** Two sub-apps mounted at one path behind overlapping patterns both cover a request and both match its host, so ranking by depth and registration order gave the first one every request under that path — including routes the second one registered (`/api/second` on `admin.example.com` got the first app's `ErrorHandler`). Which app each cloned route came from is now recorded through the domain wrapper and through the mount expansion that follows it, including synthesized HEAD routes, and the route that ran names its owner.

The mount still has to cover the path and match the host for that to apply. Without the host check this over-reached in the opposite direction: a domain-wrapped handler that declines a host leaves its route as the one that ran, so a request the mount passed on would have been attributed to it — three existing tests caught that, and a fourth test now pins it. An owner resolved this way outranks a deeper plain mount, which fixes the same misattribution in the other direction; errors raised before any mounted route matched resolve as before.

Four tests added, three verified to fail without the fix (the fourth guards the host check above). Full suite, `-race`, `golangci-lint`, `gofumpt` and `go vet` clean; patch coverage complete.

The split offer in my previous comment stands — this round is the eleventh follow-up in `28891f4`'s blast radius — but nothing here is blocked on that decision.

---

gaby:
All three findings on `0e549dc` reproduced, and are fixed in `83dc1b4`.

**Domain mounts behind an ordinary descendant.** `outer.Use("/mid", mid)` followed by `mid.Domain(...).Use("/child", inner)` left no record of `inner` anywhere the root could see: those records live only on the app they were registered on, and ordinary mounting does not carry them upward. The routes reached `inner` and its handler ran, but the config lookup could only choose between `outer` and `mid` — `/api/mid/child/boom` fell to the default 500. The snapshot taken when a domain mount is registered now walks the sub-app's descendants and collects theirs as well, composing the paths and the patterns; a request still has to satisfy every pattern on the way in, which the second half of the test pins. Worth noting this shape crashes on `main` — it is the #4618 nil-deref — so it is new ground rather than a regression.

**A sub-app that had already started.** One used on its own before being domain-mounted has had its mount placeholders replaced by its children's routes, and those routes carried no provenance, so the fallback took them all for the sub-app: a child error used the outer 556 handler instead of the child's 555. This one is a regression from `0e549dc`, which recorded ownership for domain mounts only — `3009488` answered it correctly by inference. Ordinary mounts are recorded as owners now too, which is also the more honest reading of the map: it answers "which mounted app did this route come from", and only the domain lookup consults it.

**`ViewsLayout` without an engine.** A mount that configures a layout and inherits the engine above it was dropped from selection rather than ranked, so the layout was lost and the parent engine rendered unwrapped — `<h1>Hello</h1>` where a plain mount at the same path gives `<h1>Hello</h1><h1>I'm main</h1>`. That is a regression from `28891f4`: on `main` the domain mount sat in `appList`, so the path scan found it and applied the layout. Selection keeps the owner either way now; the layout applies as an ordinary mount's does, and only the engine falls through.

Three tests added, each verified to fail without its fix. Full suite, `-race`, `golangci-lint`, `gofumpt` and `go vet` clean; patch coverage complete.

Two of these three landed in `28891f4`'s blast radius again, which is the eleventh and twelfth follow-up on that commit. The recommendation in my earlier comment stands unchanged, and so does the offer to do the split — the panic fix and the mount fixes have now gone six rounds without a finding, while the host-scoped `ErrorHandler`/`Views` work has drawn one in every round. Happy either way; just say which.

---

gaby:
All three findings on `84a0828` reproduced, and are fixed in `b25b3bb`. They turned out to be one mistake wearing three faces: a domain mount was being treated as something other than an ordinary mount at its path.

**Late ordinary descendants.** `outer.Use("/mid", mid)` followed by `mid.Use("/child", child)` leaves `child` out of `outer`'s flattened list — that list is filled in when an app is mounted, and startup repairs it only for apps reachable through `appList`, which a domain-mounted one is not. So the previous round's walk started from a stale snapshot. It now goes through each app's own mount metadata in turn, which is current for the mounts registered on it, composing paths and patterns as it descends and terminating on a cycle. `/api/mid/child/boom` gets the child's 589 rather than the default 500.

**Inheriting the enclosing handler.** `root.Use("/mid", mid)` with `mid.Domain(...).Use("/child", child)`, where `child` configures no handler: the plain scan correctly found `mid`'s, and the domain branch then cleared it unconditionally, so the error fell to the root's. Ordinary nesting inherits here — the plain control returns 590 where the domain version returned 500. Only a plain mount *as deep as* the owner is dropped now: that one is a sibling that did not serve the request, not an ancestor.

**Borrowing a sibling's engine.** The same shape in the other direction. An owner with no engine let a plain mount at its own path render, so a domain route asking for `index.tmpl` was rendered by the plain sibling's engine and failed on a template it never had. The scan skips that mount and continues to the ones enclosing them both.

The unifying change is that ranking is now depth-only. The `exact` flag I added in `0e549dc` did two jobs: pick the right app among overlapping domain mounts, and outrank every plain mount. The first is real and stays — it lives in the owner lookup, which prefers the app the route that ran came from. The second was the bug behind two of these three, and it is gone; the route's owner now carries the depth of the mount it was reached through and ranks against the plain mounts like any other mount at that depth.

Five tests added — three for the findings, two covering the branches they introduced (a dropped sibling handler, and a cyclic mount graph). Each of the first three verified to fail without its fix. Full suite, `-race`, `golangci-lint`, `gofumpt` and `go vet` clean; patch coverage complete.

That is rounds seven through nine for `28891f4`. The offer to split stands.

---

gaby:
Both findings on `b25b3bb` reproduced, and are fixed in `e4a24ea` — along with a third the second one implies.

**Inheriting from an enclosing domain mount.** A domain-mounted app with an `ErrorHandler` and a child without one: the child owns the route, only the child's config was consulted, and the error fell to the root. The plain control returns 594 where the domain version returned 595. Each setting is now taken from the owner or, when it configures none, from the deepest domain mount enclosing it — mounts as deep as the owner still skipped, since those are the siblings the last round was about.

**The same gap for views**, which the finding does not mention but the shape guarantees: a child of a domain-mounted app rendered through the *root's* engine rather than its parent's, failing on a template the root never had (500 against the plain control's 200). Fixed by the same walk, layout included.

**Automatic HEAD across a wrapper.** Deduplication was by path alone, but a wrapper holds the routes of several mounted apps side by side under different patterns. An explicit `HEAD /x` from a nested mount stood in for the outer app's `GET /x`, and on a hostname the nested pattern rejects it declined — `HEAD /api/x` returned 404 while `GET /api/x` returned 200. Deduplication is now per mounted app where routes are host-scoped, which is the wrapper and nothing else; ordinary apps keep the path-keyed behavior exactly, so a parent's `HEAD` still suppresses a mounted `GET`'s companion as before.

Four tests added, each verified to fail without its fix, plus one pinning that inheritance follows only the mounts a request actually passed through — a domain mount on a sibling path contributes nothing. Full suite, `-race`, `golangci-lint`, `gofumpt` and `go vet` clean; patch coverage complete.

Rounds ten and eleven for `28891f4`, and by my count its twenty-first follow-up fix overall. I will keep answering rounds for as long as they come, but I do not think review iteration is converging here, and I would rather not keep growing a PR whose stable half has been ready since `b088e52`. The split offer stands: `git revert` of the host-scoping commits leaves #4618 fixed and reviewable on its own, and I will open the follow-up PR with this work intact the moment anyone says the word.

---

gaby:
All three findings on `e4a24ea` reproduced, and are fixed in `5212e93`.

**Inheritance followed coverage, not ancestry.** Two independently registered mounts overlapping by host and path — a wildcard at `/api`, an exact-domain mount at `/api/child` — made the first answer for the second, which it encloses in neither sense (598 where the root should give 599). Each record now carries the apps it was registered inside, collected as the mount tree is walked, and that chain is what a setting is inherited along. The depth-and-coverage guesswork is gone from the lookup entirely, which is the third time this PR that replacing an inference with recorded structure has closed a class of finding rather than an instance.

**A plain route attributed to a domain mount.** With `/api/boom` on an ordinary mount and `/api/other` on a domain mount at the same prefix, an error from the plain route used the domain app's handler: the route named an app the domain records did not describe, and the lookup fell back to inferring an owner from host and path. Recorded ownership is conclusive now — only a request that ran no mounted route at all, a 404 or an error raised above them, falls back to the mount covering its path.

**The clone race is real, and there is a second one underneath it that is not mine.** Copying route pointers under `src.mutex` and reading the routes after releasing it does race a concurrent registration, since registering on a path the app already serves appends to that route's handlers in place. Routes are copied by value under the lock now.

Fixing that exposed a second race in the same test, and it reproduces on `main` with ordinary `Use`: `mount()` writes the mounted app's `mountPath` while holding only the *mounting* app's lock, and a route registration on the mounted app reads it under its own. `App.mount` and `Group.mount` both do this today. I fixed the domain path, since that code is this PR's, by writing each app's mount path under its own lock — but the plain-mount twin is untouched and pre-existing, and it seemed wrong to change `Use`'s locking here without being asked.

Three tests added, each verified to fail — two on the assertion, one on the race detector — without its fix. Full suite, `-race`, `golangci-lint`, `gofumpt` and `go vet` clean; patch coverage complete.

---

gaby:
Both findings on `5212e93` reproduced, and are fixed in `9fe062d`.

**Constraints in a nested mount prefix.** `mid.Use("/:name<onlyfoo>", child)` inside a domain-mounted tree: the composed path is recorded by the root, re-parsed with only the root's constraints, and `onlyfoo` — registered on `mid` — was dropped, leaving a bare parameter that matches anything. `/api/mid/bar/nope` was covered by `child` and answered by its handler where the root should own the 404. The constraints of the apps a path is composed from now travel with the record, theirs winning where a name is defined twice, which is the rule mount expansion already uses.

**Nondeterministic ancestry.** Confirmed, and it is as unpleasant as it sounds — 40 identical runs of the same request gave `mid`'s handler 30 times and the root's 10 times. An app list holds the descendants of the apps it lists, so the walk reaches a mount both through its real parent and as an entry of the app above that one; only the first arrives with the complete chain, and map iteration order picked the winner. The longer chain now wins, the other being a prefix of it by construction. I kept the traversal rather than restricting it to direct mounts, because after startup the placeholders that describe direct mounts are gone from the stack, so the flattened list is the only complete source — the fix is in choosing between the two arrivals rather than avoiding one.

Two tests added, both verified to fail without their fix; the ancestry one runs the same request forty times, which failed reliably before and passes at `-count=3`. Full suite, `-race`, `golangci-lint`, `gofumpt` and `go vet` clean; patch coverage complete.

---

gaby:
CI caught my previous commit being half a fix, and `8bab1ee` finishes it.

`Test_Domain_UseMountNestedPrefixConstraint` — the test I added in `9fe062d` — failed all three passes of the `repeated` job while passing locally. The cause is the same duplicate-arrival problem that commit was about, one field over: the two walks that reach a mount differ in more than the ancestor chain. The arrival that skipped the app whose prefix carries `onlyfoo` also lost that app's constraints, so its path was parsed with the constraint missing and matched everything the parameter did. Deduplication kept the first record and upgraded only `ancestors`, leaving the parser to whichever arrival map iteration produced first. The complete record replaces the partial one outright now.

It reproduces ten times in sixty runs, which is why one local pass at `-count=3` was not enough to see it — the repeated job is doing exactly its job here. Verified after the fix: sixty runs of both tests clean, the whole root package at `-count=3` clean, full suite, `-race`, `golangci-lint`, `gofumpt` and `go vet` clean.

---

gaby:
One of the two findings on `8bab1ee` is real and fixed in `6355c9a`; the other is pre-existing behaviour that ordinary mounts share, so I have not changed it.

**Layout inheritance past the engine — fixed.** Confirmed exactly as described: a nested child with its own `Views` and no `ViewsLayout` rendered `<h1>child</h1><h1>I'm main</h1>` under a domain mount against `<h1>child</h1>` for the same tree mounted ordinarily. I had resolved the engine and the layout with two independent walks, so the layout kept going after the engine was found. They are one walk now, ending at the app that renders; a layout still travels outward when nothing below configures an engine, which is the case that walk was added for.

**The prefix compiler — pre-existing, and not domain-specific.** The finding holds, but not as a difference this PR introduces. Both mounts return 404 for the case-insensitive example — and on `main`, so does `mid` **standalone**, with no outer app and no domain router anywhere:

```go
mid := fiber.New(fiber.Config{CaseSensitive: true, RegexHandler: caseInsensitiveCompiler})
mid.Use("/:name<regex(^foo$)>", child)
mid.Test(httptest.NewRequest("GET", "/FOO/x", http.NoBody)) // 404 on main
```

`processSubAppsRoutes` re-parses the composed path with `route.group.app.config.RegexHandler` — the mounted app's — so the compiler of the app that wrote the prefix is dropped at the first level of nesting, before any of this PR's code is reached. `domainRoutes` uses `src.config.RegexHandler` for the identical reason and produces the identical result, which is the behaviour I want it to have: a domain mount should route like an ordinary one.

Making the prefix owner's compiler win is possible, but `parseRoute` takes one compiler for the whole path while a composed path can carry segments from several apps, so it is a change to how `Use` compiles nested prefixes generally — worth doing deliberately in its own change, on both paths at once, rather than diverging the domain path here. Happy to open that separately if a maintainer wants it; the constraints case was tractable because a parse takes a *set* of them and a compiler is a single function.

One test added, verified to fail without its fix. Full suite, `-race`, `golangci-lint`, `gofumpt` and `go vet` clean, plus the root package at `-count=3` and sixty runs of the new test after last round's lesson; patch coverage complete.

---

gaby:
Reproduced and fixed in `49c746c`.

A default outer app with `GET /top`, and an unrelated child under `/v1` that sets `DisableHeadAutoRegister`:

```
domain mount:  HEAD /api/top = 405   HEAD /api/v1/deep = 405
ordinary mount: HEAD /api/top = 200   HEAD /api/v1/deep = 405   ← the correct answers
```

One accumulator carried the answer for the whole tree, so any app in it opting out — or leaving HEAD out of its `RequestMethods` — took the routes away from every other app behind the same mount. The opt-out is now recorded on the routes it actually covers as they are cloned: the app's own, and those of the apps it stands in front of, since its middleware is what a synthesized HEAD route would run without. It says nothing about siblings or about the apps it is mounted inside, which keep theirs. The marks travel to the app the mount is registered on, which is the same per-route mechanism ordinary mounts use, so the two paths now agree exactly.

Worth saying explicitly because this area is where the middleware-bypass bug lived earlier in the review: the narrower opt-out does not widen it. The gate case is a sub-app that omits HEAD from `RequestMethods` and puts a `Use` in front — that app and everything mounted behind it are still marked, so `HEAD /api/secret` still 405s rather than running the handler bare. `Test_Domain_UseMountAutoHeadKeepsMiddleware` and its plain-mount twin pin it, and I ran both twenty times over.

One test added, verified to fail without the fix. Full suite, `-race`, `golangci-lint`, `gofumpt` and `go vet` clean, plus `-count=3` on the root package and sixty runs of the auto-HEAD tests; coverage checked against `main` across the whole PR — no uncovered added lines.

---

gaby:
Both findings on `49c746c` reproduced, and are fixed in `9c73df4`.

**Exponential walk.** Confirmed at exactly the numbers reported — a chain of ordinary mounts produced 2^depth records: 32 for 6 apps, 256 for 9, 1024 for 11. The walk now remembers the longest chain each app has been reached at and walks it again only for a longer one, which is sound because only the longest is kept: the same eleven apps produce 29 records, and the bound is quadratic in depth rather than exponential.

I kept the flattened list as the source rather than traversing placeholders, for the reason from two rounds ago — after startup the placeholders are gone, so it is the only complete view — but the duplicate arrivals it causes are now collapsed as they happen instead of at the end.

**Equal-depth ancestor.** Also confirmed: `root.Use("/mid", mid)` with `mid.Domain(...).Use("/", child)` returned 500 against the plain equivalent's 555. A domain mount at the root of an ordinary one is exactly as deep as it, and the tie rule read that as standing beside it. It is passed over now only when it is not one of the apps the owner sits inside.

That alone did not fix it, which was the more interesting half: the records propagated up a level had no ancestry pointing at the app they were registered on, because a walk starts *at* the app being mounted and that app is outside it. It is prepended as the records move up, so `mid` is now visible as `child`'s ancestor from the root — which is also what makes the inheritance correct rather than just the tie check.

Two tests added, both verified to fail without their fix; the walk one asserts a quadratic bound at depth 12, where the old code produced 4096 records. Full suite, `-race`, `golangci-lint`, `gofumpt` and `go vet` clean, plus `-count=3` on the root package and sixty runs of both new tests; coverage checked against `main` across the whole PR — nothing uncovered.

---

gaby:
Both findings on `9c73df4` reproduced, and are fixed in `72b12fc`.

**Matcher-blind deduplication.** A regression from last round's memoization, and a fair catch: keying the walk by app and path alone collapsed one app mounted at `/` for both `a.example.com` and `b.example.com` into a single record. The `b` routes ran and the error fell to the root — 500 against `a`'s 555. The patterns a mount was reached through are part of what makes it that mount, and are part of the key now.

**Depth outranking the route.** Also confirmed: a domain app at `/api` serving `/admin/boom`, with an unrelated plain app at `/api/admin` that serves only `/other`, sent the domain route's error to the plain app's handler because its prefix was one segment longer.

This is the piece I removed in `b25b3bb`, and I think it is worth saying why it belongs back in a narrower form. The `exact` flag there did two jobs, and I dropped both when one of them proved wrong. Ranking against *shallower* mounts is where it was wrong — those are enclosing, and the request inherits from them. Ranking against *deeper unrelated* ones is where it was right: depth stands in for ownership only when nothing better is known, and here the route names its owner outright. So `outranks` consults the route again, while `supersedes` decides whether a plain mount is replaced or inherited from — at or below the owner's depth and not one of its ancestors. The rounds that removed it stay fixed; I re-ran those tests specifically.

Two tests added, both verified to fail without their fix, and the second pins the other direction too: the deeper mount still answers for the routes it does serve. Full suite, `-race`, `golangci-lint`, `gofumpt` and `go vet` clean, plus `-count=3` on the root package and sixty runs of both new tests; coverage checked against `main` across the whole PR — nothing uncovered.

---

gaby:
Reproduced and fixed in `2f3e62b`. This one is worth the numbers, because all three states differ. Two sibling apps registering `same`, one accepting `foo` and one `bar`, mounted under a shared parent:

| | `/a/foo` | `/a/nope` | `/b/bar` | `/b/nope` |
|---|---|---|---|---|
| `main` | 200 | **200** | 200 | **200** |
| `72b12fc` | 200 | 404 | **404** | 404 |
| `2f3e62b` | 200 | 404 | 200 | 404 |

On `main` the nested constraints are dropped entirely, so every value is accepted — including the ones the constraints exist to reject. My earlier commit fixed that but merged the two registries by name, so whichever was merged first bound both siblings: the valid `/b/bar` started answering 404, and Codex's other direction is the sharper one — a broader first implementation would have admitted what the stricter one rejects.

The set a route was parsed with is now recorded on the route and travels with it as it is cloned a level up, with the mounting app's merged underneath: the prefix is that app's and may name a constraint only it knows, while the route's own win. That also fixed a case I would not have predicted — the first version of the test failed with 405 on the domain path, because the synthesized HEAD routes were still being re-parsed from the merged registry and matched a path their GET twin rejected. They carry the GET route's set now.

One test added covering both mount kinds and both directions, verified to fail without the fix. Full suite, `-race`, `golangci-lint`, `gofumpt` and `go vet` clean, plus `-count=3` on the root package and twenty runs of the constraint tests; coverage checked against `main` across the whole PR — nothing uncovered.

Correcting myself on the previous round, for the record: I said in chat that the review of `72b12fc` came back clean. The finding arrived a few seconds behind the review body and I had already called it. The review was not clean, and this is it.

---

gaby:
@claude update the PR description

gaby:
Updated. The old description covered the first five commits; there are twenty-seven now, so I rewrote it around what the branch actually contains rather than appending to it.

What is new in there for a reviewer:

- **How it grew, and where the churn is.** The panic fix and the mount-expansion work have been quiet since `b088e52`; the host-scoped config feature (`28891f4`) has drawn a finding in every round since and accounts for most of the diff. The split offer is stated in the description itself now, not just buried in the thread.
- **The through-line.** Three separate classes of bug came from inferring which app a request belongs to from host and path. Routes now carry their originating app, the constraints they were parsed with, and their mount's stance on automatic HEAD; mount records carry their ancestry and patterns. That framing explains most of the commit list better than the commits do individually.
- **Known limitations**, promoted to their own section and expanded from two to three — the nested `RegexHandler`, the `mountPath` write under the wrong lock, and the cyclic mount graph. All three reproduce on `main` with ordinary `Use` and no domain router; each says so explicitly, with the reason I left it rather than diverging the domain path.
- The security-relevant HEAD case (`HEAD /api/secret` returning 200 where GET returned 401) is now described in the body rather than only in the commit that fixed it.

Head is `2f3e62b`, CI fully green including the three `repeated` jobs, patch coverage complete.

---

ReneWerner87 (approved):

Issue #4604, opened 2026-08-11
Consolidate sanitization/normalization helpers and remove double processing


ReneWerner87:
Core and middleware carry around 40 duplicated or divergent transform sites (origin/host normalization, comma-list scanning, ETag parsing, control-char handling, quoted-string production, URL composition guards), grouped into 13 clusters below. Goal: one implementation per concern, at most one pass per value, and the fasthttp trust boundary written down at the call sites that depend on it. Suggested order: clusters 1, 2 and 5 first (most call sites).

New helpers go to `internal/` packages (no new public API); only generic byte primitives belong in gofiber/utils.

### Fix first

Three small bug-ish fixes that can land as standalone PRs before any consolidation:

- [ ] router.go `buildRouteURL` substitutes path params with no escaping; `urlnorm.RootedPath` afterwards defuses a leading `//`, but a param containing `?` or `#` still restructures the URL, while `Redirect().Route()` escapes query values via `utils.AppendQueryEscape`. Escape substituted params with path-segment rules (or reject structure-altering values).
- [x] client/hooks.go `parserRequestURL` substitutes path params via unanchored `strings.ReplaceAll(uri, ":"+key, val)`: `:id` also matches inside `:idx`, and unescaped values containing `/`, `?` or `#` restructure the request target. Anchor at segment boundaries and percent-encode values.
- [x] extractors/extractors.go `FromParam` runs `url.PathUnescape` on `c.Params()`. With `UnescapePath: false` (default) that is the single decode; with `UnescapePath: true` the path was already decoded once in `configDependentPaths`, so params get decoded twice (`%2520` arrives as a space). Decide which layer decodes and make the count independent of the config flag.

### What fasthttp v1.73.0 already guarantees (server path)

- Every exported header-value write (`Set`, `Add`, `SetCanonical`, `SetCookie`, `SetContentType`, special setters) funnels through `initHeaderValueBytes` -> `removeNewLines`: CR and LF are replaced with a space, silently; header keys are CR/LF-filtered too; trailer keys are the one different shape (validated and rejected instead). No exported API bypasses the value filter. Fiber-side CRLF stripping before a header write is therefore browser-parity normalization at most, never injection defense.
- Parsed request-header values reject C0 bytes except HTAB, and reject DEL; bytes >= 0x80 pass. A violating request is answered 400 and the connection closed; obs-folds are collapsed to one space. A request target containing any CTL byte is rejected too.
[...]

PR #4626, opened 2026-08-19
🩹 Fix: Decode a path parameter exactly once in FromParam

Rohilalala:
# Description

Fix-first item 3 of #4604.

`extractors.FromParam` always ran `url.PathUnescape` on `c.Params()`. With the default `UnescapePath: false` that is the single decode and it is correct. With `UnescapePath: true` the router already decoded the path in `configDependentPaths`, so the value was decoded a second time here — the number of decodes depended on a config flag the extractor never looked at.

Effect: a client sending a literal `%20` (on the wire as `%2520`) reached the extractor as a space.

    UnescapePath: false → "%2520" → router leaves it → FromParam decodes once → "%20"   ✅
    UnescapePath: true  → "%2520" → router decodes → "%20" → FromParam decodes again → " "   ❌

The decode now happens only when the router left the value raw, so the count is exactly one under either setting. A value with no percent sign returns immediately, which skips both the config read and the decode on the common path.

This matters beyond the extractor package: `FromParam` backs the parameter-based token lookups used by middleware such as keyauth and csrf, so under `UnescapePath: true` a token containing a percent-escape was compared in a form the client never sent.

## Changes introduced

- [ ] Benchmarks: no benchmark added; the change removes work on the common path (an early return before the config read) and adds none.
- [ ] Documentation Update: not needed, the documented behavior ("retrieves a value from a URL parameter") is what now actually happens.
- [ ] Changelog/What's New: happy to add an entry if you would like one.
- [ ] Migration Guide: not needed. Only affects apps running `UnescapePath: true` with percent-escaped parameter values, which were receiving over-decoded values.
- [ ] API Alignment with Express: n/a, no public API change.
- [x] API Longevity: no exported surface changed.
- [ ] Examples: n/a.

Test added: `Test_FromParam_DecodesOnce` runs the same request (`/test/%2520`) under both `UnescapePath` settings and requires the same extracted value, `"%20"`. It fails on `main` for the enabled case.

## Type of change

## Checklist

- [x] Conducted a self-review of the code and provided comments for the reasoning.
- [ ] Documentation update not required for this change.
- [x] Added a unit test that fails before the change and passes after.
- [x] Ensured that new and existing unit tests pass locally (`go test ./... -race -shuffle=on`).
- [x] No new dependencies.
- [x] Aimed for optimal performance: the common case now returns before reading the config.
- [x] `make lint` (golangci-lint v2.12.2) reports 0 issues; `gofumpt -l .` is clean.

### DIFF
--- a/extractors/extractors.go
+++ b/extractors/extractors.go
@@ -28,6 +28,7 @@ package extractors
 import (
 	"errors"
 	"net/url"
+	"strings"
 	"unsafe"
 
 	"github.com/gofiber/fiber/v3"
@@ -469,6 +470,19 @@ func FromParam(param string) Extractor {
 		value := c.Params(param)
 		if value == "" {
 			return "", ErrNotFound
+		}
+		// Without a percent sign there is nothing to decode, and this is
+		// the common case, so it skips the config read below.
+		if !strings.Contains(value, "%") {
+			return value, nil
+		}
+		// UnescapePath already decoded the path once in the router, so
+		// decoding again here would spend the client's escaping twice: a
+		// literal "%20" sent as "%2520" would arrive as a space. Decode
+		// only when the router left the value raw, which keeps the number
+		// of decodes at one whatever the config says.
+		if c.App().Config().UnescapePath {
+			return value, nil
 		}
 		unescapedValue, err := url.PathUnescape(value)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


ReneWerner87 (approved):
LGTM

PR #4629, opened 2026-08-21
🐛 bug: escape generated route parameter values

evgenyponomarev:
## Summary

- percent-encode every value substituted by `buildRouteURL` with `utils.AppendPathEscape`, preventing `/`, `?`, `#`, `%`, controls, and similar bytes from restructuring generated URLs
- keep `Route.URL`, `GetRouteURL`, and `Redirect.Route` byte-identical by fixing the shared helper and updating their existing security regressions, including greedy parameters
- document automatic path-segment encoding across all three public APIs and update the URL-normalization comments that previously described raw substitution

## Behavior change

Parameter values are now treated as one URL path segment. For example, `Map{"id": "a/b?c#d"}` emits `a%2Fb%3Fc%23d`; this also applies to greedy `*` and `+` values. Ordinary unreserved parameter values are unchanged.

## Validation

- affected route and redirect tests pass with race detection
- the full root package passes with `-race -count=1 -shuffle=on`
- `go vet ./...` passes
- `govulncheck ./...` reports no reachable vulnerabilities with Go 1.25.13
- generation and gofumpt produce no additional changes
- golangci-lint v2.12.2 reports 0 issues
- the repository-wide race run completed 5,182 tests; its only remaining failure is an unrelated TEST-NET dial assertion because this sandbox routes `192.0.2.1:1`

Related #4604

AI disclosure: OpenAI Codex assisted with implementation, test and documentation updates, and validation. I reviewed the resulting diff and remain responsible for the contribution.


gaby:
@evgenyponomarev We are doing a massive refactor of `redirect` in #4621 so lets wait until that is merge first.

ReneWerner87:
Ran the generated URLs back through the app that generated them: register a named route, call `Route.URL`, send that exact URL to the same app, read `Params` in the handler. Checked against a raw socket and a real `net/http` client as well, not just `app.Test`, since `%2F` in a path segment is not `/` per RFC 3986 and both agree with fasthttp on every case.

The diagnosis holds, and main is worse than the description says. On `/user/:name`, `a?b` comes back as `a` and `a#b` as `a`, silently, because the value opens a query or a fragment. `100%` makes `Route.URL` produce a target fasthttp rejects outright. The doc note this PR removes never warned about either. Escaping is the right instinct.

Where it does not hold is the greedy case, and that one this PR introduces rather than inherits. Default config, `/user/*`:

| value | main | this PR |
| --- | --- | --- |
| `docs/index.html` | `docs/index.html` | `docs%2Findex.html` |
| `a/b` | `a/b` | `a%2Fb` |
| `john` | `john` | `john` |

A `/*` route goes from three working values to one. In a wildcard `/` is not a delimiter to neutralize, it is the content the route exists to accept, so this breaks static-file style routes and `Redirect().Route` onto a wildcard. `redirect_test.go:324` and `router_test.go:4499` move their expectations to the encoded output, which is why CI stays green: the suite asserts the generated string and never a round trip.

`UnescapePath: true` does not rescue it, and the reason matters for picking a direction: it decodes the whole path before segmentation, so `%2F` becomes a separator again and re-splits. The PR's `/user/a%2Fb` then 404s on `/user/:name`. Escape-on-generate and `UnescapePath` are not inverse operations for `/`.

What would make this mergeable, as far as I can see: escape everything except `/` inside a greedy segment. `segment.IsGreedy` is available a few lines above the change. That fixes `a b`, `a?b`, `a#b` and `100%`, which are the cases actually broken today, and leaves `/*` working. The non-greedy `/` case cannot round-trip either way without changing how the router decodes, so I would leave it and document it rather than encode it.

Happy to be talked out of this if you see a way to keep the greedy escaping and still match. Please add a round-trip test either way, the current expectations cannot catch this class.

ReneWerner87:
Numbers for the record, against main `aadd456b` on darwin/arm64, 8 interleaved rounds, with this branch's `redirect_test.go` and `router_test.go` copied onto the base so only production code differs.

The request path does not move. `Benchmark_Router_Handler` and `_Strict_Case` are unchanged, and `Benchmark_Route_Match`, which this change cannot reach, drifts +1.21%, which sets the noise floor. `Benchmark_Redirect_Route` is +3.32% (131.1 -> 135.5 ns, p=0.002) and `_WithQueries` +1.97%. Allocations are identical everywhere, 1/op and 16 B/op on the redirect benchmarks, since the escaping appends into the pooled buffer.

Those benchmarks pass values that need no escaping, so they under-report what URL generation now costs. Measuring `Route.URL` directly:

| | main | this PR | |
|---|---|---|---|
| plain param | 76.7n | 83.6n | +9.0% |
| greedy, 3 segments, plain | 128.8n | 163.6n | +27.0% |
| param needing escaping | 86.8n | 160.8n | +85.2% |
| greedy needing escaping | 136.3n | 207.0n | +51.9% |

Decomposed with a constant 23-byte payload split into a varying number of segments, so the base stays flat at ~126 ns: about 12 ns is the scan that proves nothing needs escaping, which the old `WriteString` did not do and which correctness requires; about 9 ns more is greedy-specific even when the value holds no slash; and 6 to 7 ns comes per additional segment.

The last two are avoidable. `appendGreedyPathEscape` walks the value looking for `/`, then `AppendPathEscape` walks each piece again against the noEscape table, so the bytes are scanned twice and every segment costs a call. A `utils` variant that treats `/` as a safe byte would make it one pass: at six segments that is +52.8 ns today against roughly +12 ns, independent of segment count. I am taking that to gofiber/utils separately. It need not hold this PR, since the cost lands per generated URL rather than per request, and it buys four cases that main loses data on.

PR #4632, opened 2026-08-21
🧹 chore: consolidate entity-tag parsing into internal/etag

BitWeaverDev:
# Description

Core and the ETag middleware carried two entity-tag parsers. The core one is quote-aware; the middleware one split `If-None-Match` on every comma, which its own comment admitted mis-parses an opaque-tag containing one:

```go
// RFC 9110 allows a comma inside an opaque-tag, so this split can
// mis-parse such tags. It fails open and Fiber's ETags never contain
// commas, so it is acceptable here.
```

This moves the core implementation into `internal/etag` and deletes both copies of the duplicated logic, so one parser serves `Ctx.Fresh()` and the middleware.

This is cluster 6 of the consolidation issue:

> **6. ETag:** `middleware/etag` `isNoneMatch` splits If-None-Match naively on commas (its own comment admits commas inside opaque-tags mis-parse) while core `isEtagStale` is quote-aware; `etagWeakMatch` duplicates core `matchEtag`/`matchEtagStrong`, which additionally factor through `normalizeEtag`. **Target:** internal/etag Parse/Match/Format used by both `Fresh()` and the middleware.

Related #4604

## Changes introduced

- **New `internal/etag`** (no new public API): `Parse`, `Match`, `MatchStrong` moved verbatim from `helpers.go`, plus `AnyMatch` for a raw `If-None-Match` field value — the quote-aware splitter that was the body of `app.isEtagStale`, inverted to report a match rather than staleness.
- **`helpers.go`**: `normalizeEtag`, `matchEtag`, `matchEtagStrong` and the body of `isEtagStale` deleted. `isEtagStale` survives as the wrapper applying `app.toString`, which is what honors `Immutable`, so the `Ctx.Fresh()` call site is untouched.
- **`middleware/etag`**: `isNoneMatch` and `etagWeakMatch` deleted, along with the caveat above.

Net **+207 / −167**. All 30 existing HTTP-level middleware tests pass **unedited** — that is the intended proof that behavior is unchanged.

### Behavior changes: none

The middleware's comma split is a latent defect, not an observable one, and I would rather say so than overclaim a fix:

- RFC 9110 `etagc` excludes DQUOTE, so a fragment produced by splitting a valid tag on an interior comma can never be quoted at both ends — and `etagWeakMatch` required both ends quoted, so the naive split already failed closed.
- Middleware-generated tags (`"<len>-<crc32>"`) never contain a comma.
- The middleware early-returns when the response already carries an `ETag` header, so a handler-supplied tag with a comma never reaches the comparison.

No request could observe a difference. `Fresh()` was already correct. This matches the issue's acceptance criterion that outputs stay byte-identical unless a divergence was a bug.

### Note on `MatchStrong`

`matchEtagStrong` has no production caller today — it was exercised only by `helpers_test.go`. I kept it in the new package as the RFC 9110 §8.8.3.1 counterpart an `If-Match` implementation would need, rather than silently dropping it. Happy to remove it if you would prefer not to carry it.

### Follow-up, deliberately not in this PR

`middleware/compress/compress.go` does a fourth ad-hoc inspection (`strings.HasPrefix(tag, "W/")`). Routing it through `Parse` would change behavior for malformed ETags — an unquoted tag would stop being regenerated — which is a policy decision rather than a consolidation, so it is left alone.

- [x] Benchmarks: see below — 0 allocs unchanged, sec/op within noise.
- [x] Documentation Update: none needed; no public API or behavior change.
- [x] Changelog/What's New: none needed; internal refactor.
- [ ] Migration Guide: not applicable.
- [ ] API Alignment with Express: not applicable.
- [x] API Longevity: no public surface added — the new package is under `internal/`, as the issue requires.
- [ ] Examples: not applicable.

## Type of change

## Benchmarks

`benchstat`, 10 runs each, baseline measured in the same session to control for drift (an initial naive before/after showed a uniform ~5% rise on benchmarks that never reach the ETag path, which was machine noise):

```
                           │   sec/op    │   sec/op     vs base               │
Benchmark_Ctx_Fresh_StaleEtag-10      354.2n ± 3%   351.8n ± 1%  -0.68% (p=0.005 n=10)
Benchmark_Ctx_Fresh_WithNoCache-10    77.70n ± 0%   76.61n ± 1%  -1.39% (p=0.000 n=10)
Benchmark_Ctx_Fresh_LastModified-10   107.4n ± 1%   107.0n ± 0%  -0.42% (p=0.034 n=10)
Benchmark_Etag-10                     79.19n ± 0%   78.50n ± 1%  -0.87% (p=0.000 n=10)
```

`B/op` and `allocs/op` are 0 before and after on all four ("all samples are equal").

## Checklist

- [x] Verified that any new dependencies are essential — none added.

## Validation

- `internal/etag`: **100.0%** statement coverage
- full suite: `go test ./... -race -count=1 -shuffle=on` passes
- `golangci-lint` v2.12.2: **0 issues**
- `go vet ./...`, `gofumpt`: clean, no diff
- `govulncheck ./...`: no reachable vulnerabilities

### DIFF
--- a/helpers.go
+++ b/helpers.go
@@ -29,6 +29,7 @@ import (
 
 	"github.com/gofiber/fiber/v3/binder"
 	"github.com/gofiber/fiber/v3/internal/contextvalue"
+	etagpkg "github.com/gofiber/fiber/v3/internal/etag"
 	"github.com/gofiber/fiber/v3/internal/mediatype"
 	"github.com/gofiber/fiber/v3/log"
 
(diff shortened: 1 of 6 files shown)


ReneWerner87 (approved):
LGTM

PR #4642, opened 2026-08-29
🩹 fix: treat Vary field names as case-insensitive

### LINKED ISSUE #4641: 🐛 [Bug]: Vary dedups field names byte-exactly, so a differing case is listed twice
### Bug Description

`Res.Vary()` decides whether a field name is already listed by comparing bytes, so the same field name in a different case is added again. Field names are case-insensitive (RFC 9110 Section 5.1), so `Vary: accept, Accept` names one thing twice.

The awkward part is that Fiber already reads `Vary` case-insensitively in two places, so it will not recognise on the way in what it just wrote on the way out:

* `middleware/compress` checks `Vary` for `Accept-Encoding` with a case-insensitive token match, and skips adding its own entry when it finds one.
* `middleware/cache` `parseVary` lowercases every name before building the variant key, so `accept` and `Accept` are the same dimension there. A response carrying both hashes that one request header twice, and both count toward `maxVaryHeaders` (32).

This is a conformance bug rather than a security one. Nothing is mis-served, since caches lowercase field names too, so the duplicate is redundant rather than wrong. What it costs is a malformed-looking `Vary` on the wire and a value Fiber contradicts itself about.

It is easy to hit without doing anything unusual, because several built-ins add `Vary` themselves in canonical case. `Res.AutoFormat()` calls `Vary("Accept")`, so a handler that spelled it `c.Vary("accept")` gets both.

### How to Reproduce

1. Run the snippet below against v3.5.0.
2. Observe `Vary: "accept, Accept"` on the first line, from a handler that named one field.

The second line is the direct form, without `AutoFormat` involved.

### Expected Behavior

`Vary` should treat its members as case-insensitive field names, so the second call is recognised as already listed and the header stays `Vary: accept` (first spelling wins, matching how the method already keeps existing members in place).

Actual output:

```
Vary: "accept, Accept"
Vary: "Accept-Encoding, accept-encoding"
```

Scope note: this is specific to `Vary`, whose members are defined to be field names. The generic `Res.Append()` also dedups byte-exactly, but that is defensible, since it is used for `Link`, `Cache-Control` and others whose members are not all case-insensitive. Only `Vary` has the guarantee that makes the folded comparison correct. `Vary("*")` collapsing is unaffected and already works.

I ran into this while consolidating the comma-list scanners in #4637, and flagged it in that PR description as belonging in its own change rather than being folded into a refactor, so filing it here. Happy to send the fix if you agree with the shape.

### Fiber Version

v3.5.0 (also on current `main`, e7229b1b)

### Code Snippet (optional)

```go
package main

import (
	"fmt"
	"log"
	"net/http/httptest"

	"github.com/gofiber/fiber/v3"
	"github.com/valyala/fasthttp"
)

func main() {
	app := fiber.New()

	// A handler that varies on Accept, spelled lower case, then hands off to
	// AutoFormat, which internally calls Vary("Accept").
		c.Vary("accept")
		return c.AutoFormat("hello")
	})

	req := httptest.NewRequest("GET", "/", nil)
	req.Header.Set("Accept", "text/plain")
	resp, err := app.Test(req)
	if err != nil {
		log.Fatal(err)
	}
	fmt.Printf("Vary: %q\n", resp.Header.Get("Vary")) // "accept, Accept"

	// Direct form.
	c.Vary("Accept-Encoding")
	c.Vary("accept-encoding")
	fmt.Printf("Vary: %q\n", string(c.Response().Header.Peek("Vary"))) // "Accept-Encoding, accept-encoding"
	app.ReleaseCtx(c)
}
```

### Checklist:

official-burak:
## Description

`Res.Vary()` compared field names byte for byte, so `Vary("accept")` followed by `Vary("Accept")` (or `AutoFormat`, which adds `Accept`) produced `Vary: accept, Accept`. RFC 9110 Section 5.1 treats field names as case-insensitive.

Vary now dedups with ASCII case folding and keeps the first spelling. Generic `Append()` stays byte-exact, because those lists are not all field names (Link, Cache-Control). That is documented on Append. `Append("Vary", ...)` can still list `accept` and `Accept` together; use `Vary()` for this header.

Fixes #4641

## Changes introduced

- `headerlist.AppendUniqueFold` for case-insensitive list members
- `Res.Vary()` uses it; `Append()` still uses byte-exact `AppendUnique`
- Tests for the AutoFormat path, mixed-case calls, and later field lines
- Append godoc and `docs/api/ctx.md` note the byte-exact comparison

## Benchmark

Review A/B, interleaved `-count=8`: `Ctx_Vary` 39.55n ± 4% to 41.20n ± 2% (+4.20%, p=0.000). Allocs unchanged. That is EqualFold vs `==`, about 1.65ns.

## Type of change

- Bug fix (non-breaking change which fixes an issue)

## Checklist

- Added or updated unit tests
- New and existing unit tests pass locally
- Documentation in `/docs/` updated


BitWeaverDev (commented):
Thanks for picking this up. I filed #4641, so I checked the fix against the tree rather than just reading the diff. It is correct, and the shape is the one I would have written.

Everything below is verified locally on `e48d7ab9`: all 50 packages pass, `-race` clean on the root and `internal/headerlist`, `golangci-lint` reports 0 issues, and the package stays at 100% statement coverage.

## What works well

Splitting `AppendUnique` / `AppendUniqueFold` mirrors the `Contains` / `ContainsFold` split the package already had, so the new function lands where a reader expects it. A `fold bool` parameter would have been the obvious alternative and a worse one.

I was ready to complain about the `present func(...)` parameter on a hot path, but the compiler removes it:

```
headerlist.go:217:23: inlining call to appendIfAbsent
headerlist.go:224:23: inlining call to appendIfAbsent
```

Both callers inline it, so the indirection is devirtualized and the abstraction costs nothing. Extracting the loop was the right call over duplicating it.

Leaving generic `Append()` byte-exact is also right. `Link` and `Cache-Control` members are not all case-insensitive, so folding there would be wrong.

## One question about scope

The same header can still be made non-conforming through `Append`. Verified on this branch:

```
Vary("accept")          + Vary("Accept")           -> "accept"           fixed
Append("Vary","accept") + Append("Vary","Accept")  -> "accept, Accept"   still duplicates
Vary("accept")          + Append("Vary","Accept")  -> "accept, Accept"   still duplicates
Append("Vary","accept") + Vary("Accept")           -> "accept"           fixed
```

The case-insensitivity belongs to the header rather than to the method, so someone following the `Append` docs can still emit `Vary: accept, Accept`. Your reasoning for not folding generic `Append` is sound, so I am not asking you to change it blindly. `Append` does already receive `field`, so routing the known case-insensitive list headers through the folded comparison is a small change if the maintainers want it. Saying so in the `Append` godoc would also close it honestly. Either way it is worth a decision rather than silence, since #4641 reads as if the header is fixed.

## Adjacent, probably follow-up

`middleware/compress.appendVaryAcceptEncoding` is now a hand rolled twin of what `Vary()` does: check `*`, fold check presence, append `", "`. It could just call `c.Vary(fiber.HeaderAcceptEncoding)`.

That is not only tidiness. compress reads through `GetRespHeader`, which returns the first field line only, while `Vary()` goes through `peekJoinedResponseHeader` and combines them all (RFC 9110 Section 5.3). So compress can miss an `Accept-Encoding` sitting on a second `Vary:` line and add a duplicate. Switching it over would fix that for free. Fine as a separate PR.

While looking: `middleware/basicauth` does `c.Set(fiber.HeaderVary, fiber.HeaderAuthorization)` on the 401 path, and `Set` overwrites, so any existing `Vary` is dropped. Same family, also pre-existing, also a separate PR.

## Benchmark, worth putting in the description

Interleaved A/B against the base, `-count=8`:

```
Ctx_Vary-10   39.55n ± 4%  ->  41.20n ± 2%   +4.20% (p=0.000, n=8)
B/op and allocs/op: identical
```

Real and reproducible, but it is the cost of `EqualFold` over `==`, not the abstraction, and it is about 1.65ns. Fine for a correctness fix in my view. I mention it only because benchmarks are gated here and it is better stated up front than found later.

## Minor

The test acquires five contexts without `ReleaseCtx`, which matches `Test_Ctx_Vary_Wildcard` right above it, so it is consistent with the file and I would leave it.

`Test_AppendUniqueFold` repeats most of `Test_AppendUnique`'s table. Readable as it is, no need to change.

One correction that is mine, not yours: the package doc still opens with "Seven scanners ... three in the core", which #4637 made wrong when it folded in `transferEncodingLineHasBody` (it is eight, four in the core). You are editing that paragraph anyway, so it is a convenient place to fix, but it is my leftover and not your obligation.

official-burak:
Thanks for checking this against the tree, especially those Append cases. I am leaving generic Append byte-exact on purpose; Link and Cache-Control members are not all field names, so folding there would be the wrong default. I added that to the Append godoc and the API docs so #4641 does not read as if every Vary path is covered. Compress and basicauth look right as follow-ups, I can do those separately if maintainers want them.

ReneWerner87 (approved):

Issue #3159, opened 2024-10-08
🤗 [Question]: How do we get multiple values for a key in form submission?


ngocchien:
### Question Description

I have a field in my form, its name is the description

Example: 
```
<input name="item" value="item1">
<input name="description" value="description1, this is a description for item one, the first item is very nice, bla bla bla">
<input name="item" value="item2">
<input name="description" value="description2">
<input name="item" value="item3">
<input name="description" value="description3">
```

On my server side, I define a struct with the information:
```
FormCreate struct {
		Item    []string                   `json:"item" form:"item"`
		Description  []string         `json:"description" form:"description"`
	}
formData = new(FormCreate)
c.Bind().Body(formData)
```
I have a problem with binding data from the form to my formData.
My expectation data for formData.Description is []string{"description, this is a description for item one, the first item is very nice, bla bla bla",  "description2", "description3"} **(This has 3 values for the slice)**. 
But currently, I received for formData.Description after binding data is []string{"description1", "this is a description for item one", "the first item is very nice", "bla bla bla", "description2", "description3"} **(This has 6 values for the slice)**.

It seems that the Bind() function has split the value of the description field of the first item into a slice by the "," , so I  received the data for formData.Description is []string{"description1", "this is a description for item one", "the first item is very nice", "bla bla bla", "description2", "description3"}.

Who can resolve my problem?
Thank you so much!

### Code Snippet (optional)

```go
package main

import "github.com/gofiber/fiber/v3"
import "log"

func main() {
  app := fiber.New()

  // An example to describe the question
  app.Post("/create-post", handleCreate) 
  log.Fatal(app.Listen(":3000"))
}

function handleCreate(c fiber.Ctx) error{
   FormCreate struct {
		Item    []string                   `json:"item" form:"item"`
		Description  []string         `json:"description" form:"description"`
	}
  var (
    formData = new(FormCreate)
  )
   if c.Method() == http.MethodPost {
	err = c.Bind().Body(formData)
   }
   
   return c.Render("index", fiber.Map{
        "formData": formData,
    })
}

```

### Checklist:

- [X] I agree to follow Fiber's [Code of Conduct](https://github.com/gofiber/fiber/blob/master/.github/CODE_OF_CONDUCT.md).
- [X] I have checked for existing issues that describe my questions prior to opening this one.
- [X] I understand that improperly formatted questions may be closed without explanation.

michal-laskowski:
For JSON it will work
<details>
[...]

Issue #2859, opened 2024-02-14
🐛 [Bug]: Incorrect Parsing of Slice by `QueryParser()` with Embedded Structs


ecoderat:
### Bug Description

When the "Products" field is within the "Person" struct, the `QueryParser()` correctly parses it as a slice. However, when I separate the "Products" field into another struct and embed it, the `QueryParser()` incorrectly parses it as a single string instead of a slice.

### How to Reproduce

```go
package main

import (
	"log"

	"github.com/gofiber/fiber/v2"
)

type Person struct {
	Name     string   `query:"name"`
	Pass     string   `query:"pass"`
	A
}

type A struct {
	Products []string `query:"products"`
}


func main() {
	app := fiber.New(fiber.Config{
			EnableSplittingOnParsers: true,
		})

	app.Get("/", func(c *fiber.Ctx) error {
		p := new(Person)

		if err := c.QueryParser(p); err != nil {
			return err
		}

		log.Println(p.Name)
		log.Println(p.Pass)
		log.Println(p.Products)

		return c.JSON(p)
	})

	app.Listen(":3000")
}
```

```bash
$ curl "http://localhost:3000/?name=john&pass=doe&products=shoe,hat" | jq

{
    "Name": "john",
    "Pass": "doe",
    "Products": [
        "shoe,hat"
    ]
}
```

### Expected Behavior

```bash
$ curl "http://localhost:3000/?name=john&pass=doe&products=shoe,hat" | jq

{
    "Name": "john",
    "Pass": "doe",
    "Products": [
        "shoe",
        "hat"
    ]
}
```

### Fiber Version

v2.49.2

### Checklist:

- [X] I agree to follow Fiber's [Code of Conduct](https://github.com/gofiber/fiber/blob/master/.github/CODE_OF_CONDUCT.md).
- [X] I have checked for existing issues that describe my problem prior to opening this one.
- [X] I understand that improperly formatted bug reports may be closed without explanation.

ReneWerner87:
Duplicate of https://github.com/gofiber/fiber/issues/2833#issuecomment-1928931741

https://docs.gofiber.io/api/ctx#queryparser
![image](https://github.com/gofiber/fiber/assets/7063188/da6cc2c4-1aa8-4419-b04c-d696c6534485)
```go
        // fiber.Config{EnableSplittingOnParsers: false} - default
        log.Println(p.Products)    // ["shoe,hat"]
        // fiber.Config{EnableSplittingOnParsers: true}
[...]

Issue #2833, opened 2024-02-06
🐛 [Bug]: The documentation of `QueryParser()` is incorrect.


your-diary:
### Bug Description

[The documentation of `QueryParser()`](https://github.com/gofiber/fiber/blob/main/docs/api/ctx.md#queryparser) includes this example:

```go
// Field names should start with an uppercase letter
type Person struct {
    Name     string     `query:"name"`
    Pass     string     `query:"pass"`
    Products []string   `query:"products"`
}

app.Get("/", func(c fiber.Ctx) error {
        p := new(Person)

        if err := c.QueryParser(p); err != nil {
            return err
        }

        log.Println(p.Name)     // john
        log.Println(p.Pass)     // doe
        log.Println(p.Products) // [shoe, hat]

        // ...
})
// Run tests with the following curl command

// curl "http://localhost:3000/?name=john&pass=doe&products=shoe,hat"
```

But this is incorrect in many ways:

- The function passed to `app.Get()` receives `fiber.Ctx`. It should be `*fiber.Ctx`.

- The output of `log.Println(p.Products)` is described as `[shoe, hat]`. The actual output is `[shoe,hat]` with no space between the elements.

- The curl example is incorrect. It passes `shoe,hat` as a single string rather than as an array.

### How to Reproduce

```go
package main

import (
	"log"

	"github.com/gofiber/fiber/v2"
)

type Person struct {
	Name     string   `query:"name"`
	Pass     string   `query:"pass"`
	Products []string `query:"products"`
}

func main() {
	app := fiber.New()

	app.Get("/", func(c *fiber.Ctx) error {
		p := new(Person)

		if err := c.QueryParser(p); err != nil {
			return err
		}

		log.Println(p.Name)
		log.Println(p.Pass)
		log.Println(p.Products)

		return c.JSON(p)
	})

	app.Listen(":3000")

}
```

```bash
$ curl "http://localhost:3000/?name=john&pass=doe&products=shoe,hat" | jq

{
    "Name": "john",
    "Pass": "doe",
    "Products": [
        "shoe,hat"
    ]
}
```

### Expected Behavior

```bash
$ curl "http://localhost:3000/?name=john&pass=doe&products=shoe,hat" | jq

{
    "Name": "john",
    "Pass": "doe",
    "Products": [
        "shoe",
        "hat"
    ]
}
```

### Fiber Version

2.52.0

### Checklist:

[...]

PR #4648, opened 2026-09-02
feat: add Bind().WithSplitting() for per-call override of EnableSplittingOnParsers

miladev95:
### Summary

`EnableSplittingOnParsers` is currently a single, global `fiber.Config` flag that controls whether comma-separated values are split into slice elements for `Query`, `Header`, `RespHeader`, `Cookie`, and `Form` binding. This forces an all-or-nothing choice: turning it on splits everywhere, leaving it off disables splitting everywhere.

This PR adds a chainable per-call override so a single request can opt in or out without flipping the global config:

```go
// Force splitting for this request only:
c.bind().WithSplitting(true).Query(&q)

// Suppress splitting even if the app config has it enabled:
c.bind().WithSplitting(false).Query(&q)

```
The override applies to every binding method on the same Bind chain (Query, Header, RespHeader, Cookie, Form) and is scoped to that bind instance — it does not leak across requests that reuse the same pool entry.

### Changes
`bind.go`
- Add tri-state overrideSplitting *bool field to Bind. nil falls back to the app's EnableSplittingOnParsers config; non-nil forces the value for the current chain. (*bool is required to distinguish "use config" from "explicitly false".)
- Add Bind.WithSplitting(enable bool) *Bind chainable method.
- Add internal Bind.splittingEnabled() helper that resolves the effective value.
- Replace the five direct reads of b.ctx.App().config.EnableSplittingOnParsers in Header, RespHeader, Cookie, Query, and Form with b.splittingEnabled().
- Reset overrideSplitting = nil in Bind.release() so the override does not leak across pooled Bind instances.
- Reorder Bind struct fields per betteralign (saves 8 bytes per instance).

`bind_test.go`
- Add Test_Bind_WithSplitting with 5 parallel subtests covering:
1. Override on with app config off.
2. Override off with app config on.
3. No override honors the app config.
4. Override does not leak across reused pool entries.
5. Override applies to Header binding (not just Query).

`docs/api/bind.md`
- Add a "Per-Bind Override" section under the comma-separated values docs describing the new method, the affected binding methods, and the per-instance scoping guarantee.
WhyUsers have asked for the ability to split in some routes but not others without maintaining two app configurations. The existing SkipValidation(bool) / WithAutoHandling() chainable methods establish the pattern; WithSplitting follows the same convention.

### Behavior
- Default (no WithSplitting call): identical to current code — uses fiber.Config{EnableSplittingOnParsers}.
- WithSplitting(true): splits comma-separated values for the current chain regardless of the app config.
- WithSplitting(false): suppresses splitting for the current chain regardless of the app config.
- Override is reset when the Bind instance is released back to bindPool (Bind.release()), matching the lifecycle of shouldSkipValidation and shouldSkipErrHandling.

### Verification
All programmatic checks pass:
make audit # ✅ go mod verify, go vet, govulncheck — clean
make generate    # ✅
make betteralign # ✅ (pre-existing path.go / router.go warnings on main are unrelated and not touched)
make format # ✅
make lint        # ✅ 0 issues
make test # ✅ 5515 tests, 1 skipped in 47.317s
The new test Test_Bind_WithSplitting passes alongside the full suite.

## Compatibility
Purely additive — no breaking changes. Existing callers that never invoke WithSplitting see identical behavior.

### DIFF
--- a/bind.go
+++ b/bind.go
@@ -38,7 +38,10 @@ var bindPool = sync.Pool{
 // By default (manual mode), parsing failures are returned as *BindError; use errors.As to extract source and field details.
 // With WithAutoHandling(), parsing failures set HTTP 400 and return *Error instead.
 type Bind struct {
-	ctx                   Ctx
+	ctx Ctx
+	// overrideSplitting tri-state: nil falls back to the app's
+	// EnableSplittingOnParsers config; non-nil forces the value for this chain.
+	overrideSplitting     *bool
 	shouldSkipErrHandling bool
 	shouldSkipValidation  bool
 }
@@ -130,6 +133,7 @@ func (b *Bind) release() {
 	b.ctx = nil
 	b.shouldSkipErrHandling = true
 	b.shouldSkipValidation = false
+	b.overrideSplitting = nil
 }
 
 // WithoutAutoHandling If you want to handle binder errors manually, you can use `WithoutAutoHandling`.
@@ -154,6 +158,27 @@ func (b *Bind) SkipValidation(skip bool) *Bind {
 	b.shouldSkipValidation = skip
 
 	return b
+}
+
+// WithSplitting overrides the app's EnableSplittingOnParsers config for the
+// current bind chain. When enabled, comma-separated values are split into
+// individual elements for fields whose target type is a slice.
+// Pass true to force splitting on, false to force it off; the override applies
+// until the Bind instance is released back to the pool.
+func (b *Bind) WithSplitting(enable bool) *Bind {
+	b.overrideSplitting = &enable
+
+	return b
+}
+
+// splittingEnabled reports the effective comma-splitting setting for this bind
+// chain, falling back to the app's EnableSplittingOnParsers config when no
+// per-chain override has been set.
+func (b *Bind) splittingEnabled() bool {
+	if b.overrideSplitting != nil {
+		return *b.overrideSplitting
+	}
+	return b.ctx.App().config.EnableSplittingOnParsers
 }
 
 // Check WithAutoHandling/WithoutAutoHandling errors and return it by usage.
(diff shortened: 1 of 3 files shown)


bind.go
```diff
@@ -38,7 +38,10 @@ var bindPool = sync.Pool{
 // By default (manual mode), parsing failures are returned as *BindError; use errors.As to extract source and field details.
 // With WithAutoHandling(), parsing failures set HTTP 400 and return *Error instead.
 type Bind struct {
-	ctx                   Ctx
+	ctx Ctx
+	// overrideSplitting tri-state: nil falls back to the app's
```
gaby: Remove this comment

bind.go
```diff
@@ -156,6 +160,27 @@ func (b *Bind) SkipValidation(skip bool) *Bind {
 	return b
 }
 
+// WithSplitting overrides the app's EnableSplittingOnParsers config for the
+// current bind chain. When enabled, comma-separated values are split into
+// individual elements for fields whose target type is a slice.
```
gaby: This is too verbose

bind.go
```diff
@@ -156,6 +160,27 @@ func (b *Bind) SkipValidation(skip bool) *Bind {
...
+// Pass true to force splitting on, false to force it off; the override applies
+// until the Bind instance is released back to the pool.
+func (b *Bind) WithSplitting(enable bool) *Bind {
+	b.overrideSplitting = &enable
+
+	return b
+}
+
+// splittingEnabled reports the effective comma-splitting setting for this bind
+// chain, falling back to the app's EnableSplittingOnParsers config when no
```
gaby: This is too verbose

bind_test.go
```diff
@@ -3309,3 +3309,100 @@ func Test_Bind_Form_ContentTypeNormalization(t *testing.T) {
 		require.Equal(t, StatusOK, resp.StatusCode)
 	})
 }
+
+// Test_Bind_WithSplitting covers the per-chain override of the
+// EnableSplittingOnParsers config flag. It verifies:
+//  1. With no override, the chain inherits the app config.
+//  2. WithSplitting(true) forces splitting even when the app config is off.
+//  3. WithSplitting(false) suppresses splitting even when the app config is on.
```
gaby: Remove this whole comment, should only be the 1st line

gaby (changes requested):

miladev95:
@gaby please review again

ReneWerner87:
Closing this for now, on the API question rather than the code.

`EnableSplittingOnParsers` is an app-level setting, and a per-call override adds a second way to say the same thing. The description says users have asked for it, but the PR links no issue and a search of the tracker turns up nothing open on it: #2624, #2859, #3159 and #2833 are all closed and about the app-level parsing behaviour, not a per-call switch. Without that, there is nothing to weigh the extra API surface against.

If you have a case where the app-level setting cannot work, an issue describing it would settle the question quickly, and this branch is easy to revive from there. As it stands it also conflicts with main and the four review comments are still open.

PR #4653, opened 2026-09-03
🧹 chore: consolidate CORS and CSRF origin matching into internal/origin

BitWeaverDev:
# Description

Cluster 1 of #4604: the origin toolkit shared by CORS and CSRF.

Both middleware carried their own copy. `normalizeOrigin` differed by a single scheme check. The `subdomain` type, its matcher `matchNormalized`, and the wildcard walk `matchSubdomainOrigin` were character for character identical in `middleware/cors/utils.go` and `middleware/csrf/helpers.go`. The config-time loop that splits an allow-list entry on `://*.` was the same algorithm written twice, once with `strings.Cut` and once with `strings.Index`.

This adds `internal/origin` as the single implementation and names the one genuine difference instead of leaving it to be inferred from whichever copy you happen to open:

- **CORS accepts any scheme**, because `Access-Control-Allow-Origin` is echoed for schemes beyond http(s) and the browser is the party enforcing the check.
- **CSRF accepts only http and https**, because it decides whether a request may mutate state, and a non-web scheme has no business doing so.

That is now `SchemePolicy`, passed explicitly at both call sites as `corsSchemes` and `csrfSchemes`, each with a comment saying why it is what it is.

**Behavior is unchanged.** Details on how that was checked are below.

Related #4604

## Changes introduced

Net **-204 lines** of production code (`+36/-240` at the call sites), replaced by a 179-line package. The duplicated test files collapse from 536 lines to 370, and the new package sits at **97.7% statement coverage**.

- **New:** `internal/origin` with `Normalize`, `ParsePattern`, `Subdomain`, `MatchNormalized`, `MatchAny`, and `SchemePolicy`.
- **Removed:** `middleware/cors/utils.go` entirely, and the origin half of `middleware/csrf/helpers.go` (its constant-time compares stay).
- **Also removed:** csrf's `subdomain.match`, which #4604 flags as dead production code. Its only callers were in `helpers_test.go`.

`ParsePattern` returns a `Pattern` struct rather than four values. `revive`'s `function-result-limit` flagged the four-result version, and it was right, so this is not a `//nolint`.

The package is imported as `originpkg` because both middleware already use `origin` as a loop variable and a parameter name.

## Verification

This is security code, so equivalence is demonstrated rather than asserted. The pre-consolidation implementations were kept as reference functions and fuzzed against the new ones:

```
FuzzEquivNormalize        3,452,878 execs   byte-identical to both originals
FuzzEquivMatchNormalized  9,845,230 execs   identical
```

`Normalize` was compared against **both** old versions, under `AnyScheme` against the CORS original and under `WebSchemesOnly` against the CSRF original. That scaffolding is not part of the diff, since carrying copies of the deleted code would defeat the point, but it is easy to reproduce and I can push it to a branch if you would rather see it run in CI.

Beyond that: all 51 packages pass, `-race` clean on cors, csrf and the new package, `golangci-lint` (repo-pinned v2.12.2) reports 0 issues.

The migrated tests keep every case from both files. Where the two tables disagreed, the disagreement is now the point: `Test_Normalize_PolicyDelta` pins `app://example.com` as valid under `AnyScheme` and invalid under `WebSchemesOnly`, so changing that has to be deliberate.

## One deliberate behavior note

CORS had a **second** panic, `"[CORS] Invalid origin format after normalization:"`, guarding a `strings.Cut(normalizedOrigin, "://")` that cannot fail once `Normalize` has accepted the value, because `Normalize` builds the string with `"://"` in it. That branch is gone and such an entry now reports the same message as every other rejected one. No reachable input changes behavior; only the text of an unreachable panic does.

## Benchmarks

`benchstat` over the CORS handler and preflight benchmarks, interleaved A/B, `-count=8`:

```
                                      │ before      │ after                      │
_CORS_NewHandler-10                     223.3n ± 1%   223.9n ± 1%  ~ (p=0.700)
_CORS_NewHandler_Parallel-10            37.89n ± 3%   38.88n ± 3%  +2.61% (p=0.047)
_CORS_NewHandlerSingleOrigin-10         223.6n ± 2%   223.0n ± 2%  ~ (p=0.959)
_CORS_NewHandlerWildcard-10             126.7n ± 2%   127.5n ± 1%  ~ (p=0.138)
_CORS_NewHandlerPreflight-10            615.2n ± 1%   627.4n ± 2%  +1.97% (p=0.005)
_CORS_NewHandlerPreflightSingleOrigin   621.2n ± 2%   629.4n ± 1%  +1.32% (p=0.007)
_CORS_NewHandlerPreflightWildcard       558.7n ± 2%   563.4n ± 1%  +0.83% (p=0.019)

allocs/op and B/op: identical on all twelve, "all samples are equal"
```

**Allocations are unchanged everywhere.** The small timing cost is a cross-package call plus one policy branch on the request path, and it is real rather than noise on the three preflight numbers. I did not chase it with a `//go:noinline` dance or by inlining the policy check back into each middleware, since that would rebuild the duplication this removes, but say the word if you would rather have the nanosecond.

Two subdomain benchmarks moved into the new package alongside the code they measure, so the two benchmark sets do not line up exactly; `benchstat` notes this.

## Type of change

## Checklist

- [x] Added or updated unit tests. Every case from both original files is preserved, plus new coverage for `ParsePattern`, the `WebSchemesOnly` policy, and the policy delta itself.
- [x] Ensured that new and existing unit tests pass locally. All 51 packages, plus `-race` on the affected ones.
- [x] Verified that any new dependencies are essential. None added.
- [ ] Updated the documentation in `/docs/`. Not applicable: `internal/` is not public API and no exported behavior changed.
- [ ] Followed the inspiration of the Express.js framework. Not applicable, internal helper.


gaby:
@codex review

gaby (approved):
LGTM

ReneWerner87 (approved):

PR #4655, opened 2026-09-05
🔥 feat: filter route candidates with a probe constant and add DisableFlashMessages

gaby:
# Description

This branch is the outcome of a performance audit of the router, done with the SIMD/SWAR helpers in `gofiber/utils/v2` in mind.

A CPU profile of `Benchmark_Router_GitHub_API_Serial` put 39% of the router's time inside `getMatch` and 10% in the `strings.Index` call in `findParamLen`. Instrumenting the candidate scan over the GitHub API fixture (239 requests, 203 routes) showed why: after the leading-byte prefix filter and the slash-count bounds, **4.6 candidates per request still reached the full segment walk, and nearly all of them failed on the first constant after a parameter**. `/repos/:owner/:repo/keys/:id` against `/repos/a/b/issues/1` matches two constants and two parameters, including a substring search over the remainder, before it learns that `/keys/` is not there.

The branch adds a third precomputed filter that tells such candidates apart with one masked word compare, replaces the substring search in `findParamLen` with a byte scan and a prefix compare, keeps `Route.match` small on the paths every request takes, and fixes a benchmark helper whose `b.Helper()` call was a quarter of the benchmark's samples.

The same audit found that the per-request scan for the `fiber_flash` cookie costs more on a browser-sized request than routing a simple route does, and that no cheaper scan exists. So the branch also adds `Config.DisableFlashMessages`, an opt-out for deployments whose clients keep no cookies across redirects.

Everything is backward compatible: the only public API change is the new config field, and it defaults to the current behavior.

Fixes # (no linked issue)

### The probe constant filter

Each parametric route now precomputes a *probe*: the first constant segment that follows a parameter, packed into a word the way the existing prefix filter packs the path head, plus where to look for it. A non-greedy parameter ends at the next `/`, so once the leading constant has matched, the first parameter starts right after it, and every parameter and lone `/` between it and the probe accounts for exactly one more slash to pass over. `Route.matchParams` finds that slash with an inline `swar.MatchByteMask` scan from the end of the leading constant (a parameter spans a few bytes, so the slash is nearly always in the first word loaded) and rejects the candidate with a single masked compare before `getMatch` walks anything.

`computeProbe` builds a probe only where the segments guarantee the position: no greedy, optional, or adjacent (fixed-length) parameter before the constant, every compare part led by `/`, and no constant whose trailing `/` `getMatch` may drop. Routes outside that shape simply keep the current behavior.

On the GitHub API fixture the scan now looks like this (3,419 candidates over 239 requests):

| Stage | Candidates removed |
|---|---|
| leading-byte prefix filter | 438 |
| slash-count bounds | 1,801 |
| **probe compare (new)** | **875** |
| full `getMatch` walk | 227 (was 1,102) |

Full segment walks drop from **4.6 to 0.9 per request**, and the walks that remain are essentially the routes that match.

Two designs were measured and rejected on the way, which is why the probe locates its slash per candidate instead of using per-request state:

- Indexing every slash of the detection path while it is derived, in a fused SWAR pass (copy + case fold + `MatchByteMask` per word), cost more per word than the `memmove` plus stdlib vector `bytes.Count` it replaced (38.6 vs 23.5 ns at 44 bytes).
- Locating the slashes lazily on the first probe cost about 20 ns on a matched deep route, where the probe can only pass (`SkipUnmatchedRoutes_Deep/Matched_3param` +12%).

Searching from the leading constant needs no per-request state and beat both. Replacing the `strings.IndexByte` call in that search with the inline SWAR scan took the corpus benchmark from −32.8% to −38.1% in isolation.

### `findParamLen` slash-stop

For a non-greedy parameter whose compare part starts with `/`, the parameter can only end at the first `/` of the remainder, and only if the compare part sits right there. The `strings.Index` over the remainder plus the `IndexByte` back-check for an earlier slash become one `IndexByte` and one prefix compare. Where the compare part is absent the function now answers `0` instead of `len(s)`; the constant after the parameter failed either way, and the slash-stop rows in the shared route-matching fixture pin the match-level outcomes.

### Keeping `match` small

With the new checks written inline, `Route.match` had grown enough to slow the middleware and static-endpoint compares that every request pays for, even though those never reach a probe (interleaved A/B runs put `Router_Handler_Unescape` at +3.3% and `SkipUnmatchedRoutes/Unmatched` at +1.8%). The parametric branch moved into `matchParams`; the split costs the parametric scan about 1.5% and returns those paths to parity or better.

### Benchmark hygiene

`acquireDefaultCtxForRouterBenchmark` called `b.Helper()` on every iteration of the router benchmarks. `b.Helper()` walks the stack, and in the profile it accounted for a quarter of `Benchmark_Router_GitHub_API_Serial`'s samples: 210 µs/op with it, 155 µs/op without, for identical router code. The baseline below already excludes it, so the numbers measure the router.

### Numbers

`benchstat`, `main` vs this branch, 10 runs each, 4-core Xeon @ 2.80 GHz, Go 1.26:

| Benchmark | main | branch | delta |
|---|---|---|---|
| Router_GitHub_API_Serial | 158.2 µs ± 2% | 104.6 µs ± 3% | **−33.9%** |
| Path_matchParams `/api/:param/fixedEnd` match | 40.62 ns ± 2% | 35.52 ns ± 1% | −12.6% |
| Path_matchParams `/api/:param/fixedEnd` no match | 28.40 ns ± 2% | 24.18 ns ± 2% | −14.8% |
| RoutePatternMatch `/api/:param/fixedEnd` no match | 625.1 ns ± 4% | 588.8 ns ± 1% | −5.8% |
| App_MethodNotAllowed | 557.3 ns ± 1% | 523.3 ns ± 1% | −6.1% |
| SkipUnmatchedRoutes_Deep/MethodNotAllowed_deep (without / with skip) | 615.7 / 549.7 ns | 597.5 / 533.8 ns | −3.0% / −2.9% |
| SkipUnmatchedRoutes/405_NoMiddleware/without_skip | 944.4 ns ± 2% | 917.7 ns ± 1% | −2.8% |
| SkipUnmatchedRoutes/Matched/without_skip | 164.9 ns ± 2% | 161.0 ns ± 2% | −2.4% |
| Router_Next_Default | 69.07 ns ± 2% | 67.50 ns ± 3% | −2.3% |
| Router_Handler | 110.1 ns ± 6% | 110.4 ns ± 4% | ~ |
| Route_Match (direct call, `/user/keys/:id`) | 16.26 ns ± 2% | 17.90 ns ± 1% | +1.6 ns (the `matchParams` call) |

The single-route handler rows (`Router_Handler*`, `Matched_*param`) drift by ±3–5% between runs in either direction; interleaved A/B runs showed none of them as a real regression. The full-matrix geomean over 70 benchmarks is −0.6%, dominated by micro-benchmarks that exercise only `getMatch`, which is unchanged for the routes they use. The win scales with the route set: the more parametric routes share a prefix, the more candidates the probe removes.

### Verification

- `Test_Route_Match_SlashBoundsDifferential` now covers the probe as well: for every generated pattern and path, the filtered `Route.match` must agree with a raw `getMatch`.
- The shared route-matching fixture in `path_testcases_test.go` gains rows for the slash-stop shape: a parameter followed by a multi-byte constant that is present, missing, found only after a later slash, cut short, extended, or carrying an optional trailing slash. Every fixture-driven test (`Test_Path_matchParams`, `Test_Route_Match_SlashBoundsConsistency`, `Test_RoutePatternMatch`, `Test_Route_PrefixFilter_Fixture`) runs them.
- `Test_RouteParser_Probe` pins which patterns get a probe and what it holds; `Test_ConstProbe_Rejects` pins the compare, including the tail and end-of-path cases; `Test_wordAt` and `Test_packConst` cover the word packing.
- The existing `Test_Router_ScanMatchesReference` drives real requests through the handler across every routing config combination and compares against an unoptimized stack scan.
- Independent review passes ran randomized differentials of 17.6M and 38.6M (pattern, path, use) triples plus an exhaustive small-alphabet sweep with zero divergences.

### What the scan costs today

`hasFlashCookie` runs `bytes.Contains` over every request's raw header block looking for `fiber_flash=`. The existing handler benchmarks understate this because their hand-built requests carry no raw header block at all. On parsed requests:

| Request shape | Raw headers | Cost per request |
|---|---|---|
| API client (bearer token, JSON) | ~300 B | ~16 ns |
| Browser (typical Chrome header set) | ~560 B | ~45 ns |

It grows with header size, so large cookie jars or long auth headers push it higher. For scale, routing a simple request is ~100 ns.

### Why an option, and not a cheaper scan

Flash messages exist for the post-redirect-get pattern: `Redirect().With()` stores messages in a session-only, `HttpOnly` cookie that only a client keeping cookies across the redirect sends back, which in practice means a browser (an HTTP client with a cookie jar that follows redirects can too). A client sees a flash cookie only if the app itself calls `With()`. For deployments whose clients keep no cookies the scan is pure overhead, and the alternatives were measured before adding a config field:

- `simd.Memmem` is slower than `bytes.Contains` at header sizes (85 vs 61 ns on browser headers).
- Searching only the `Cookie` header through fasthttp's `Peek` gains nothing on browser headers (44.7 vs 45.6 ns), costs more on API headers (31 vs 16 ns), and once cookies have been collected `Peek("Cookie")` re-serializes them (3 allocations, 160 ns).
- Auto-detecting "this process never called `With()`" is unsafe: with several instances or a rolling deploy, a redirect issued by one instance lands on another, which would silently drop the messages.

A config flag is the only deterministic way to remove the cost.

### What `DisableFlashMessages` does

`fiber.Config{DisableFlashMessages: true}` (default `false`, so nothing changes for existing apps):

- skips the header scan and the cookie parsing in both the default and the custom-context request handlers;
- keeps `processFlashMessages` from setting the cookie on redirect, so `Redirect().With()` sets none, and makes `WithInput()` a no-op so it no longer parses the form;
- leaves `Messages()` and `OldInput()` empty, and leaves an incoming `fiber_flash` cookie alone rather than expiring it.

Benefits for API-only and service-to-service deployments: the per-request scan disappears (16–45 ns per request, more with large headers), `WithInput` can no longer copy a submitted form into a cookie by accident, and no `Set-Cookie: fiber_flash` ever leaves the service. One thing to know before flipping it on an app whose clients did keep cookies: a `fiber_flash` cookie issued before the flag is no longer expired on the next request, so it lives until the client's session ends.

### Numbers

`Benchmark_Redirect_FlashCookieCheck` drives the handler with a parsed browser-style request, which the other handler benchmarks miss:

| | ns/op |
|---|---|
| `DisableFlashMessages: false` (default) | 107.2 ± 3% |
| `DisableFlashMessages: true` | 76.2 ± 7% |

That is 31 ns, or 29% of the handler path on that request shape, with 0 allocations either way.

## Changes introduced

- [x] Benchmarks: full `benchstat` matrix above (`main` vs branch, 10 runs); new `Benchmark_Redirect_FlashCookieCheck` for the flash scan; `Benchmark_Router_GitHub_API_Serial` no longer measures `b.Helper()`.
- [x] Documentation Update: `docs/api/fiber.md` gains the `DisableFlashMessages` row in the config table; `docs/api/redirect.md` notes the option in the Flash Message section; `docs/extra/internal.md` describes the candidate-filtering stage of route matching, including the diagram.
- [x] Changelog/What's New: `docs/whats_new.md` gains a "Disabling flash messages" subsection under Redirect.
- [ ] Migration Guide: not needed; the option defaults to the current behavior and the routing changes are internal.
- [ ] API Alignment with Express: Express has no built-in flash messages (they come from middleware such as `connect-flash`), so there is no counterpart to align with; the option follows Fiber's own `Disable*` config family.
- [x] API Longevity: a single boolean in the existing `DisableHeadAutoRegister` / `DisableKeepalive` family, default `false`, with a `json` tag matching its neighbors; the routing work adds no public API.
- [x] Examples: `docs/api/redirect.md` and `docs/whats_new.md` show `fiber.New(fiber.Config{DisableFlashMessages: true})`.

## Type of change

## Checklist

- [x] Ensured that new and existing unit tests pass locally with the changes (`make test`: 5,519 tests with `-race -shuffle=on`; `make lint`, `make format`, `make betteralign`, `go generate`, `go mod verify`, `go vet` all clean; `govulncheck` reports only known Go 1.26.0 standard-library advisories that `main` shares).
- [x] Verified that any new dependencies are essential and have been agreed upon by the maintainers/community (no new dependencies).
- [x] Aimed for optimal performance with minimal allocations in the new code (0 allocations on every touched path).

https://claude.ai/code/session_01LRd49yHeAwJknQhJuS9wLi

---
_Generated by [Claude Code](https://claude.ai/code/session_01LRd49yHeAwJknQhJuS9wLi)_


docs/api/fiber.md
```diff
@@ -54,6 +54,7 @@ app := fiber.New(fiber.Config{
...
+| <Reference id="disableflashmessages">DisableFlashMessages</Reference>                 | `bool`                                                          | Disables redirect flash messages: `Redirect().With` and `WithInput` set no cookie, `Messages` and `OldInput` report nothing, and the scan of every request's headers for an incoming `fiber_flash` cookie is skipped. Flash messages only reach browsers that follow a redirect, so APIs and other non-browser deployments can turn them off. | `false`                                                                |
```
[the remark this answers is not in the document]
gaby: Agreed, the criterion is cookie retention across the redirect, not the client type. Reworded in 9850d20 in the four places that said "browsers": the `Config` comment in `app.go`, this table row, the note in `docs/api/redirect.md`, and the `docs/whats_new.md` section now all say that only a client that keeps cookies across the redirect (a browser, typically) receives flash messages, and that deployments whose clients keep no cookies can turn them off.

---

path_test.go
```diff
@@ -989,6 +995,276 @@ func Test_RouteParser_ConstParamShape(t *testing.T) {
...
+	require.Equal(t, uint64(0xFFFFFF), mask)
+
+	word, mask = packConst("/abcdefghij")
+	require.Equal(t, pathHeadWord("/abcdefg"), word)
+	require.Equal(t, ^uint64(0), mask)
+}
+
+// legacyFindParamLen is findParamLen before the slash-stop shortcut: a
+// substring search, rejected when a '/' preceded the hit.
+func legacyFindParamLen(s string, segment *routeSegment) int {
```
ReneWerner87: Too Much Logic in a Test

path_test.go
```diff
@@ -989,6 +995,276 @@ func Test_RouteParser_ConstParamShape(t *testing.T) {
...
+	}
+
+	return len(s)
+}
+
+// legacyGetMatch is getMatch's generic walk over legacyFindParamLen, without
+// the "/const/:param" specialization.
+//
+//nolint:revive // flag-parameter: mirrors getMatch's signature
+func legacyGetMatch(parser *routeParser, detectionPath, path string, params *[maxParams]string, partialCheck bool) bool {
```
ReneWerner87: Too Much Logic in a Test

path_test.go
gaby: Agreed, those two functions were copies of production code living in the test file. Removed in b472698 together with the differential test that needed them. The shapes it generated are now plain rows in the shared route-matching fixture in `path_testcases_test.go` (a parameter followed by a multi-byte constant that is present, missing, found only after a later slash, cut short, extended, or carrying an optional trailing slash), so `Test_Path_matchParams`, `Test_Route_Match_SlashBoundsConsistency`, `Test_RoutePatternMatch` and `Test_Route_PrefixFilter_Fixture` all cover them with no extra logic.

---

path_test.go
gaby: Removed in b472698 together with `legacyFindParamLen`; the cases it served as the oracle for are fixture rows now (see the thread above).

---

ReneWerner87 (approved):

PR #4666, opened 2026-09-07
⚡ perf: adopt utils/v2 helpers and cut per-request allocations in the hot path

gaby:
# Description

`gofiber/utils/v2` v2.5.0 landed a set of scan, parse and format helpers that this repository was not yet using, and profiling the request path turned up a few places where Fiber does avoidable work per request. This PR adopts those helpers where they measure faster, and fixes four independent per-request costs the profiles pointed at:

1. **Hand-rolled SWAR scans** that duplicate a helper `utils` now ships with a better implementation of.
2. **A double pass over the request path** in `configDependentPaths`, which copied the path and then folded the copy.
3. **Two per-request allocations in the CORS middleware**, both resolvable at configuration time.
4. **A heap-allocated slice on every `Locals` write**, caused by a variadic argument crossing the `Ctx` interface.

Every change is behavior-preserving. The swaps whose contracts differ in the detail (`SplitHostPort`'s failure mode, `Duration.String`'s column width, `PathUnescape`'s clean fast path) were checked against the stdlib over their edge cases first, and `PathUnescape` was left alone because it lost. Nothing in the public API changes.

No linked issue — this came out of a profiling pass rather than a bug report.

## Changes introduced

- [x] **Benchmarks**: measured throughout; see [Performance](#performance) below for the full breakdown, methodology and the changes that were rejected.
- [ ] **Documentation Update**: none needed — no public API, configuration or behavior changes.
- [x] **Changelog/What's New**: suggested entry —
  > **Performance** — the CORS middleware is now allocation-free on both the simple-request and preflight paths, the access log no longer allocates for `${status}`, `${bytesSent}`, `${bytesReceived}` or `${latency}`, request locals are stored without a heap allocation, and the router's per-request path preparation reads the path once instead of twice.
- [ ] **Migration Guide**: not applicable — no breaking changes.
- [ ] **API Alignment with Express**: not applicable — no API surface added or changed.
- [x] **API Longevity**: no public API was added, removed or re-signed. `Res.Vary`, `Req.Locals` and `Res.Set` keep their exact signatures and semantics; the new helpers (`appendCopyLowerASCII`, `setLocal`, `hasDotDotSegment`, `writeScratch`, `writePadding`, `appendIntTag`, `appendDurationTag`, `appendDurationPadded`, `writeColored`, `writeColoredInt`, `appendDurationColumns`, `writeColumnRight`, `writeColumnLeft`, `headerLists`, `vary`) are all unexported, and `internal/ctxlocal` is an internal package. `StoreInContext` keeps its signature and, for a custom `Ctx`, still dispatches through the interface so an overridden `Locals` is the one that runs.
- [x] **Examples**: not applicable — nothing user-facing to demonstrate. Unit tests cover the new helpers directly (`Test_appendCopyLowerASCII` in `helpers_test.go`).

### Commit-by-commit

| Commit | What it does |
| --- | --- |
| `🧹 chore: adopt gofiber/utils/v2 high-performance helpers in hot paths` | Control-byte scans → `utils.IndexControl`/`IndexControlExceptTab`; `net.SplitHostPort` → `utils.SplitHostPort`; `strings.ContainsAny` → `utils.IndexAny2`; `strings.Split` → `utils.SplitTrimSeq`; single-byte `Cut` → `utils.CutByte`; `strconv` → `utils.AppendUint`/`AppendInt`/`FormatInt`/`ParseInt`; `strings.EqualFold` → `utils.EqualFold`; ASCII fast path in `sanitizeFilename`; `slices.Contains(strings.Split(...))` → a shared `CutByte` walk in `middleware/static` |
| `🧹 chore: render the logger's numeric columns through utils.Append*` | `${latency}`, `${status}` and `${method}` off `fmt`; fixes an escape that put `appendInt`'s scratch on the heap |
| `⚡ perf: build the path and its detection form in one pass` | `appendCopyLowerASCII` writes `c.path` and `c.detectionPath` from one read of the original |
| `⚡ perf: make the CORS middleware allocation-free` | Hoisted `Vary` field lists; `strings.Join` of the configured header lists moved into `New` |
| `⚡ perf: store request locals without the variadic's heap slice` | `setLocal` calls the concrete `*DefaultCtx` method so the one-element `...` slice stays in the frame |
| `test: cover the logger column writers and setLocal's fallback` | Error propagation through a failing `Buffer`, the column-width boundary, and both `setLocal` paths |
| `fix(cors): keep a configured but empty AllowHeaders from echoing the request` | Guards the list headers on whether the list was configured, not on whether its joined form is empty |
| `fix(cors,csrf): keep the shared Vary field lists away from a custom Ctx` | The hoisted field lists reach `Vary` through a helper that clones for anything but `*DefaultCtx` |
| `test(logger): pin both latency columns against the fmt forms they replaced` | The 20 000-duration parity check, previously run from scratch, joins the suite |
| `⚡ perf(logger): write the colored default format in place` | The colored default line off `fmt.Fprintf`: 11 allocs → 1, −62% |
| `⚡ perf: store middleware locals without the variadic's heap slice` | `internal/ctxlocal.Set` brings the concrete-`Locals` path to the extractors, session, idempotency and earlydata |
| `test(extractors): cover ExtractWithSource on a bare chain` | The one shape that reaches `extractChainWithSource` — and its cycle guard — directly |

### Methodology

All figures are `benchstat` over **interleaved** runs — old and new alternating in one loop, so machine drift affects both sides equally — on `linux/amd64`. Sample counts are given per table; end-to-end router numbers use `GOMAXPROCS=1` to cut scheduler noise. Micro-benchmarks were taken with `-count 5`.

This matters: a first, non-interleaved sweep at `n=3` reported regressions in `adaptor` (+1.8%), `hostauthorization` (+3.0%) and `encryptcookie` (+4.3%). **All three inverted at `n=12`–`16` and are in fact improvements.** Every number below survived proper sampling; `p` values are shown where a claim rests on significance.

### End-to-end

**Root package** (`benchstat`, n=12):

| Benchmark | before | after | delta |
| --- | --- | --- | --- |
| `Ctx_Hostname` | 21.73n ± 6% | 19.70n ± 2% | **−9.34%** (p=0.001) |
| `Ctx_Subdomains` | 104.65n ± 8% | 97.04n ± 3% | **−7.28%** (p=0.000) |
| `Ctx_Range/bytes=500-` | 133.5n ± 6% | 124.3n ± 3% | **−6.89%** (p=0.000) |
| `Router_Next_Default` | 45.07n ± 3% | 42.01n ± 6% | **−6.79%** (p=0.001) |
| `Communication_Flow` | 52.42n ± 3% | 49.51n ± 3% | **−5.56%** (p=0.004) |
| `Ctx_Range/bytes=-700` | 132.7n ± 6% | 125.4n ± 4% | **−5.50%** (p=0.001) |
| **geomean** | 136.7n | 131.1n | **−4.11%** |

**`middleware/cors`** (n=10) — every allocation removed:

| Benchmark | before | after | delta |
| --- | --- | --- | --- |
| `CORS_NewHandler` | 459.6n ± 6% | 418.6n ± 2% | **−8.92%** (p=0.000) |
| `CORS_NewHandler_Parallel` | 147.2n ± 6% | 105.9n ± 19% | **−28.08%** (p=0.000) |
| `CORS_NewHandlerPreflight` | 1.300µ ± 3% | 1.093µ ± 1% | **−15.89%** (p=0.000) |
| `CORS_NewHandlerPreflight_Parallel` | 377.5n ± 3% | 288.2n ± 7% | **−23.68%** (p=0.000) |
| `CORS_NewHandlerPreflightWildcard` | 1188.5n ± 3% | 975.6n ± 4% | **−17.92%** (p=0.000) |
| **geomean** | 369.6n | 307.2n | **−16.87%** |
| simple `B/op` | 16 B, 1 alloc | **0 B, 0 allocs** | −100% (p=0.000) |
| preflight `B/op` | 104 B, 3 allocs | **0 B, 0 allocs** | −100% (p=0.000) |

The parallel variants gain most, which is exactly what dropping the allocations buys: no per-request trip through the allocator under contention.

**`middleware/logger`** (n=10):

| Benchmark | before | after | delta |
| --- | --- | --- | --- |
| `WithBytesAndStatus` | 507.2n ± 4% | 410.7n ± 2% | **−19.04%** (p=0.000) |
| | 3 allocs/op | **0 allocs/op** | −100% (p=0.000) |
| `WithTagParameter` | 598.6n ± 4% | 489.1n ± 3% | **−18.30%** (p=0.000) |
| | 3 allocs/op | **0 allocs/op** | −100% (p=0.000) |
| `WithBody` | 257.4n ± 4% | 246.0n ± 2% | **−4.43%** (p=0.001) |
| **geomean** | 609.1n | 588.3n | **−3.42%** |
| `DefaultFormatForceColors` (second sweep, n=12) | 1605.5n ± 1% | 611.5n ± 2% | **−61.91%** (p=0.000) |
| | 208 B, 11 allocs | **8 B, 1 alloc** | −96% / −91% (p=0.000) |
| `DefaultFormat` / `DisableColors` (control) | 572n | 574n | ~ (p=0.35 / 0.71) |

**`middleware/static`** — `sanitizePath` (n=12):

| Benchmark | before | after | delta |
| --- | --- | --- | --- |
| `SanitizePath/nilFS_-_slashes` | 616.8n ± 4% | 501.4n ± 2% | **−18.70%** (p=0.000) |
| `SanitizePath/nilFS_-_urlencoded` | 580.9n ± 4% | 487.4n ± 3% | **−16.10%** (p=0.000) |
| `SanitizePath/dirFS_-_urlencoded` | 681.5n ± 3% | 573.8n ± 2% | **−15.80%** (p=0.000) |
| **geomean** | 625.0n / 193.3 B / 5.65 allocs | 519.5n / 140.4 B / 4.64 allocs | **−16.88% / −27.40% / −17.79%** |

**Other packages** (n=12–16):

| Benchmark | before | after | delta |
| --- | --- | --- | --- |
| `basicauth/Middleware_BasicAuth` | 848.0n ± 2% | 821.1n ± 3% | **−3.18%** (p=0.008) |
| | 48 B, 3 allocs | 32 B, 2 allocs | **−33.33%** (p=0.000) |
| `adaptor/HTTPHandler` | 3.001µ ± 4% | 2.848µ ± 2% | **−5.10%** (p=0.006) |
| `adaptor/FiberApp` | 3.685µ ± 2% | 3.551µ ± 4% | **−3.65%** (p=0.000) |
| `encryptcookie/Invalid_Cookie_Parallel` | 2.177µ ± 4% | 2.068µ ± 4% | **−5.01%** (p=0.028) |
| `encryptcookie/Empty_Cookie` | 1.099µ ± 7% | 1.047µ ± 5% | **−4.69%** (p=0.020) |
| `encryptcookie/Valid_Cookie_Parallel` | 4.083µ ± 3% | 3.943µ ± 2% | **−3.43%** (p=0.000) |
| `hostauthorization/matchHost_Mixed` | 18.26n ± 3% | 17.80n ± 1% | **−2.52%** (p=0.002) |
| `hostauthorization` geomean | 204.9n | 201.4n | **−1.73%** |
| `idempotency/Idempotency/hit` (second sweep, n=8) | 592 B, 8 allocs | 576 B, **7 allocs** | −12.5% allocs (p=0.000) |

A full sweep of **every** benchmark under `middleware/...` was run old-vs-new. No package regressed.

### Where the time went — per change

**`utils.IndexControlExceptTab` in `internal/logtemplate`.** Every sanitized access-log value (path, IP, error, headers) runs this scan. `utils` unrolls two words per branch and folds the whole test into one arithmetic mask, against the single-word loop it replaced:

| input | before | after |
| --- | --- | --- |
| 12-byte path | 10.8ns | **7.9ns** (−27%) |
| 45-byte path | 29.3ns | **17.7ns** (−40%) |
| 120-byte user agent | 76.9ns | **42.8ns** (−44%) |

The same helper replaces `middleware/proxy`'s own SWAR loop (~35 lines deleted) and a byte-at-a-time loop in `client`'s redirect handling.

**`appendCopyLowerASCII` (`configDependentPaths`).** The path was copied into `c.path`, then that copy was folded into `c.detectionPath` — every byte read twice, two capacity checks, two loop set-ups. Under the default configuration the detection path is exactly the case fold of the path, so both are now written from one pass:

| path length | before | after |
| --- | --- | --- |
| 5 bytes | 9.2ns | **6.2ns** (−33%) |
| 19 bytes | 12.7ns | **6.9ns** (−46%) |
| 41 bytes | 16.2ns | **11.4ns** (−30%) |
| 70 bytes | 19.8ns | **16.0ns** (−19%) |

The three non-default configurations keep the existing two-step, since `UnescapePath` rewrites `c.path` before the fold can run.

**`utils.SplitHostPort`.** Same rules as `net.SplitHostPort`, but a miss reports `ok == false` instead of allocating a `*net.AddrError` — and a miss is the common case for a bare hostname:

| input | before | after |
| --- | --- | --- |
| `example.com` (miss) | 53ns, 1 alloc | **9.2ns, 0 allocs** (−83%) |
| `example.com:8080` (hit) | 21.0ns | **17.8ns** (−15%) |

**`utils.IndexAny2` for `strings.ContainsAny`:** 18ns → **6.9ns** on a 16-byte host, 47.9ns → **8.8ns** on a 54-byte one (−82%).

**Logger columns through `utils.Append*`.** `${latency}` rendered with `fmt.Fprintf("%13v")`, which reflects over the `Duration`, builds its `String` and boxes it. And `appendInt` already used `utils.AppendInt` into a stack scratch — but then handed the slice to `output.Write`. `Buffer` is an interface, so the compiler has to assume `Write` keeps the pointer, and the array was moved to the heap on every `${status}`, `${bytesSent}` and `${bytesReceived}`:

| | before | after |
| --- | --- | --- |
| `${latency}` | ~205ns, 2 allocs | **~39ns, 0 allocs** |
| `appendInt` (a status code) | 47ns, 1 alloc (24 B) | **21ns, 0 allocs** |
| `${status}` colored (3 codes) | 465ns | **202ns** |
| `${method}` colored | 85ns | **19ns** |

The `${latency}` padding still counts **runes**, so the column keeps the width `%13v` produced when a sub-millisecond latency renders `µs` — pinned by a parity test over 20 000 random durations against `fmt`'s own output.

**Colored default format (second sweep).** The default format has two branches, and only the uncolored one had been rewritten; the colored one — taken whenever the stream is a terminal — was still a single `Fprintf` over eleven arguments plus a format string concatenated per call around the error padding. It now writes the same columns in place with the same helpers, so the two branches cost the same: **1605 → 612 ns, 208 → 8 B, 11 → 1 alloc** (the survivor is `c.IP()`, shared with the uncolored branch). Two details of `fmt`'s output are kept deliberately rather than unified: `fmt` pads by runes, so the latency and path columns count runes (a `µs` latency and a non-ASCII path keep their width) where the uncolored branch has always counted bytes. `Test_Logger_DefaultFormat_WithColor_MatchesFmt` keeps the old format string as its oracle and pins the line byte for byte over exactly those cases.

**CORS.** `Vary` takes its field names variadically, and a fresh `...` argument list is a slice the compiler must heap-allocate: `Vary`'s result reaches the header store, so escape analysis marks the elements as leaking even though fasthttp copies the bytes out. Passing a package-level slice hands the callee the existing backing array instead. That was the single allocation on the simple path and three of the four on preflight.

One backing array shared by every request must not reach an implementation that could write to it, and `Ctx` is an interface: `NewWithCustomCtx` can install a `Vary` that sorts or otherwise rewrites the field list it is handed, which would corrupt the names used by later responses and race with the requests running alongside. Both middlewares therefore call `Vary` through a helper that type-asserts to `*fiber.DefaultCtx` — whose `Vary` is in this repo and only reads the list — and clones for anything else. The default path keeps the shared array and its zero allocations; interleaved benchstat over the six CORS handler benchmarks shows no change in time, `B/op` or `allocs/op`. (`res.go`'s `varyAccept` needs no such helper: both callers are methods on `*DefaultRes`, so the `Vary` they reach is fixed at compile time.)

Moving the joins into `New` also moved the guards from `len(cfg.X) > 0` onto the joined string, which is a different question — `AllowHeaders: []string{""}` is a configured list that joins to `""`, and reading it as absent made the preflight echo `Access-Control-Request-Headers` where it used to authorize nothing. `headerLists` now records presence alongside each joined form and the guards test that, restoring the original semantics for all three lists. The rest was `strings.Join` over `AllowMethods`, `AllowHeaders` and `ExposeHeaders`, rebuilding identical strings on every request; none can change once `New` returns, so they are joined there, beside the `maxAge` conversion that already worked that way.

**Request locals.** The same variadic-through-an-interface problem, on a much hotter path:

| | ns | B | allocs |
| --- | --- | --- | --- |
| `c.Locals(k, v)` via the `Ctx` interface | 47.2 | 16 | 1 |
| `dc.Locals(k, v)` on `*DefaultCtx` | **11.4** | **0** | **0** |
| fasthttp `SetUserValue` (the floor) | 9.1 | 0 | 0 |
| `StoreInContext` before | 50.2 | 16 | 1 |
| `StoreInContext` after | **17.0** | **0** | **0** |

The value itself has to escape — fasthttp keeps it for the request — but the slice holding it does not. `setLocal` makes the concrete call the shared path and falls back to the interface call for a custom `Ctx`. It covers every middleware that records something per request through `StoreInContext` (basicauth, keyauth, csrf, session, requestid, paginate), the domain router's two per-request stores, and the generic `Locals[V]` setter.

The second sweep extends it to the packages that write locals directly and cannot reach an unexported helper — the extractors' chain guards and winner stack, session's per-request ID and extractor records and its release-time clear, idempotency's and earlydata's markers — through `internal/ctxlocal.Set`, the same assertion-and-fallback behind an internal package (`*fiber.DefaultCtx` is exported, and the concrete call inlines across packages just as well: 50.2 ns / 16 B / 1 alloc through the interface, 16.4 ns / 0 B / 0 allocs through `Set`). Twenty call sites across the four packages. `Benchmark_Session` does not move because it reuses one ctx and `Store.Get` reads the cached session-ID local before it would reach those lines; on the middleware path every request is a fresh ctx and takes them.

**Smaller swaps**, each measured before adoption: `utils.CutByte` for single-byte `Cut` (8–22% faster, 18 call sites); `utils.AppendUint` / `AppendInt` in `middleware/cache` (−11% / −21%); `utils.SplitTrimSeq` in `middleware/paginate` (90ns/2 allocs → 62ns/1 alloc); an ASCII fast path in `sanitizeFilename` (16.6→12.6ns at 13 bytes, 111→47ns at 90).

### Changes measured and rejected

Recorded so the next person does not repeat them:

- **`hasFlashCookie` narrowed to the Cookie field lines.** The shipped `bytes.Contains` over `RawHeaders()` costs ~105ns on an 890-byte browser header block. A `VisitAll` variant that searches only the cookie lines is **slower** (156ns): the per-header closure cost exceeds the smaller haystack. Left as is.
- **Routing `Res.Set` through `SetCanonical`.** Would skip fasthttp's key normalization *and* one redundant copy of every header value. But `normalizeHeaderKey` also neutralizes CR/LF in the key and `SetCanonical` does not, and `isCanonicalName` accepts `"X-Foo\r\nevil"`. That is a header-injection vector for ~20ns. Not worth it.
- **`utils.AppendPathUnescape`.** Loses to `url.PathUnescape`, which returns the input unchanged (no allocation) when there is nothing to decode: 21ns/0 allocs against 43ns/1 alloc on a clean 12-byte path.
- **`bytes.UnsafeToLower` in the cache's `isIgnoredHeader`.** 23% *slower* — its scan-then-convert pair loses to a single-pass loop on short header names.
- **`utils.HasPrefixFold`** for the `EqualFold(h[:n], scheme)` sites — a wash, so left alone.
- **`utils.IndexNonQuotable`** in `internal/quotedstring` — it permits HTAB, which would change pinned escaping behavior.
- **`utils.IndexFold` / `ParseNativeInt`** — existing comments in `constraint.go` and `cachecontrol.go` record earlier benchmarks where these lost. Honored rather than re-litigated.
- **`Subdomains` label split via `utils.CutByte`** instead of `strings.SplitSeq` (second sweep) — a profile pointed at the iterator, but interleaved benchstat (n=12) showed 197.8 → 197.0 ns, p=0.30. Not carried.
- **`proxy/security.go`'s two `net.SplitHostPort` calls** — they run on the dial path, where the split succeeds and allocates nothing; the swap would only change the wrapped `*net.AddrError` text on the failure path. Left alone.

### Not addressed

- **`c.IP()`** is now the largest remaining per-request allocation on the logger path (`net.IP.String`), but removing it means returning request-scoped memory from a getter that today returns a keepable heap string — a contract change that wants maintainer sign-off, not a perf patch.
- **`middleware/session` (192 allocs/op)** is ~70% `encoding/gob` recompiling its type engine per call. Reuse would break the self-describing blob format that stored sessions depend on.
- **Boxing a value into `any` for `Locals`** — a string or struct value still costs one allocation for the interface conversion itself (it is what fasthttp stores). That is the allocation left on every `Locals` write this PR touched; removing it would mean changing what `Locals` stores.

## Type of change

## Checklist

- [x] Followed the inspiration of the Express.js framework for new functionalities, making them similar in usage. — no new functionality; all public APIs keep their current shape.
- [x] Conducted a self-review of the code and provided comments for complex or critical parts. — each swap carries a comment stating the measured reason, so a future reader can tell an intentional choice from an accident.
- [x] Updated the documentation in the `/docs/` directory. — nothing to update: no public API, configuration or behavior change.
- [x] Added or updated unit tests to validate the effectiveness of the changes or new features. — `Test_appendCopyLowerASCII` pins the new helper against the two operations it replaces at every length across the SWAR word boundaries, on fresh and reused destinations; a parity test over 20 000 random durations pins `${latency}` against `fmt`'s output; `middleware/cors`'s direct-call tests were updated for the new parameter and still cover the nil-config path. The logger's column writers are tested against a `Buffer` that fails mid-column, since `Buffer` is public and `Config.LoggerFunc` is handed one; `setLocal`, `ctxlocal.Set`, `Vary` field sharing and a configured-but-empty `AllowHeaders` each have a test that fails without its change; the colored default line is pinned byte for byte to the `Fprintf` it replaced.
- [x] Ensured that new and existing unit tests pass locally with the changes. — see Verification below.
- [x] Verified that any new dependencies are essential and have been agreed upon. — **no new dependencies**; `gofiber/utils/v2` is already required, and `main` has since moved it to v2.5.1 (#4668), which this branch has merged and been re-verified against.
- [x] Aimed for optimal performance with minimal allocations in the new code. — four per-request allocation sites removed outright; the framework's own request path stays at zero allocations.
- [x] Provided benchmarks for the new code to analyze and improve upon. — see [Performance](#performance).

### Verification

Run on the current head:

| Check | Result |
| --- | --- |
| `go test ./...` | 53 packages, **0 failures** |
| `go test ./... -race -count=1 -shuffle=on` | 53 packages, **0 failures, no races** |
| `go vet ./...` | clean |
| `golangci-lint fmt --diff` (gofumpt + goimports, repo config) | clean |
| `golangci-lint run ./...` (v2.12.2, repo config) | **0 issues** |
| `go mod verify` | all modules verified |
| Coverage of the lines this PR adds | **99.1%** (331/334) |

The uncovered added lines are the `..`-segment rejection in `middleware/static`, which `pathpkg.Clean` already makes unreachable, and two branches of the extractors' nested-chain winner stack (`truncateChainWinStack` shrinking a longer stack, `popChainWinningSource` leaving a non-empty one) that the suite did not reach before this PR either; all three only appear in the diff because the statements on them were re-spelled.

One gate could not be run here: `make betteralign` has no build that fits — its releases either require Go 1.27 or are built with Go 1.24, and this module targets 1.26. No struct layouts changed apart from `headerLists`, which is constructed once per `New`.

## Commit formatting

Commits use the repository's emoji prefixes — `⚡ perf:` for the three behavior-preserving optimizations and `🧹 chore:` for the two `utils/v2` adoption passes, matching the `⚡ perf:` / `🧹 chore:` usage already in `main`'s history. The follow-ups use the plain `test:` and `fix(scope):` forms that `main` also carries (`test:`, `refactor(router):`, `perf(router):`), and the second sweep's two optimizations return to `⚡ perf:`.

https://claude.ai/code/session_01CA3Pup9oZ7ruZCkMHEhJLu


middleware/cors/cors.go
```diff
@@ -19,6 +19,33 @@ const corsSchemes = originpkg.AnyScheme
...
+// Vary takes its field names variadically, and a fresh "..." argument list is
+// a slice the compiler has to heap-allocate: Vary's result reaches the header
+// store, so escape analysis marks the elements as leaking even though fasthttp
+// copies the bytes. Passing a package-level slice with "..." hands the callee
+// the existing backing array instead, which removed the only allocation on the
+// simple-request path and three of the four on preflight. Vary never mutates
+// what it is given, so sharing these across requests is safe.
+var (
+	varyOrigin           = []string{fiber.HeaderOrigin}
+	varyPreflight        = []string{fiber.HeaderAccessControlRequestMethod, fiber.HeaderAccessControlRequestHeaders, fiber.HeaderOrigin}
```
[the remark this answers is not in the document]
gaby: Fixed in 785f325.

Both middlewares now go through a `vary` helper that type-asserts to `*fiber.DefaultCtx` — whose `Vary` is right here and only reads the field list — and clones for anything else, so a custom `Ctx` never receives the package-level array. `middleware/csrf/csrf.go`'s `varyCookie` got the same treatment.

The default path keeps the shared array and its zero allocations: interleaved benchstat over the six CORS handler benchmarks (12 runs each, `GOMAXPROCS=1`) shows no change in time and identical `B/op` and `allocs/op`.

`res.go`'s `varyAccept` is not affected — both callers are methods on `*DefaultRes`, so the `Vary` they reach is fixed at compile time rather than dispatched through the interface. Its comment now says that instead of making a claim about `Vary` in general.

Added `Test_CORS_VaryDoesNotShareFieldsWithCustomCtx` and `Test_CSRF_VaryDoesNotShareFieldsWithCustomCtx`, which install a custom context whose `Vary` overwrites every element it is handed; both fail on the previous code and pass now.

---

middleware/cors/cors.go
```diff
@@ -191,20 +233,20 @@ func New(config ...Config) fiber.Handler {
...
-		if len(cfg.AllowMethods) > 0 {
-			c.Set(fiber.HeaderAccessControlAllowMethods, strings.Join(cfg.AllowMethods, ", "))
+		if lists.allowMethods != "" {
+			c.Set(fiber.HeaderAccessControlAllowMethods, lists.allowMethods)
 		}
-		if len(cfg.AllowHeaders) > 0 {
-			c.Set(fiber.HeaderAccessControlAllowHeaders, strings.Join(cfg.AllowHeaders, ", "))
+		if lists.allowHeaders != "" {
+			c.Set(fiber.HeaderAccessControlAllowHeaders, lists.allowHeaders)
 		} else {
```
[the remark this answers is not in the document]
gaby: Correct, and fixed in fac5098.

`headerLists` now records whether each list was configured alongside its joined form, and the guards test that flag instead of the string. That restores exactly the `len(cfg.X) > 0` semantics the code had before the join was hoisted out of the request path, while keeping the join in `New`.

`AllowMethods` and `ExposeHeaders` had the same flaw, though for those the difference is an omitted header rather than a permissive one — all three are back on presence.

`Test_CORS_ConfiguredEmptyAllowHeaders` pins both sides: an unconfigured list still echoes `Access-Control-Request-Headers`, `AllowHeaders: []string{""}` no longer does. It fails on the previous code.

---

middleware/cors/cors.go
```diff
@@ -191,20 +233,20 @@ func New(config ...Config) fiber.Handler {
...
 		// Set Preflight headers
-		if len(cfg.AllowMethods) > 0 {
-			c.Set(fiber.HeaderAccessControlAllowMethods, strings.Join(cfg.AllowMethods, ", "))
+		if lists.allowMethods != "" {
+			c.Set(fiber.HeaderAccessControlAllowMethods, lists.allowMethods)
 		}
-		if len(cfg.AllowHeaders) > 0 {
-			c.Set(fiber.HeaderAccessControlAllowHeaders, strings.Join(cfg.AllowHeaders, ", "))
+		if lists.allowHeaders != "" {
+			c.Set(fiber.HeaderAccessControlAllowHeaders, lists.allowHeaders)
```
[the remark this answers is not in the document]
gaby: Fixed in fac5098 — see the reply on the Codex thread for the same finding.

`headerLists` now carries a `has*` flag per list, set from `len(cfg.X) > 0` in `New`, and the response guards test that rather than the joined string. `AllowMethods` and `ExposeHeaders` were guarded the same way and moved back to presence too. `Test_CORS_ConfiguredEmptyAllowHeaders` covers both branches and fails without the change.

---

gaby:
@codex review

internal/ctxlocal/ctxlocal.go
```diff
@@ -0,0 +1,24 @@
...
+//
+// Locals takes its value variadically. Reached through the Ctx interface the
+// compiler cannot see that it only reads that argument, so the one-element
+// "..." slice is heap-allocated on every call; the concrete method inlines and
+// keeps it in the frame, which measured 50ns and a 16-byte allocation against
+// 11ns and none. A custom Ctx fails the assertion and keeps the interface call,
+// so an overridden Locals is still the one that runs. fiber's own
+// StoreInContext does the same in-package; this is the form middleware and the
+// extractors can reach.
+func Set(c fiber.Ctx, key, value any) any {
```
gaby, line 19: @ReneWerner87 Thoughts on this?

ReneWerner87, line 19: not sure, is the check really needed ? it performs the same method usage?

gaby: It is the same method, but the two calls do not compile the same way — the check is the whole point of the function.

`Locals(key any, value ...any)` is variadic, so every call builds a one-element `[]any` for `value`. Whether that slice lives on the stack or the heap is decided by escape analysis:

- `c.Locals(key, value)` with `c` a `fiber.Ctx` **interface**: the callee is not known at compile time, so the compiler has to assume it might keep the slice, and heap-allocates it on every call.
- `dc.Locals(key, value)` with `dc` a `*fiber.DefaultCtx`: the callee is known, its `value` parameter is proven to only be read (`value[0]` is stored, the slice itself is not), so the slice stays in the caller's frame.

The compiler says so directly. Two one-line wrappers, `-gcflags=-m`:

```
viaInterface(c fiber.Ctx, k, v any)         { c.Locals(k, v) }   // ... argument escapes to heap
viaConcrete(dc *fiber.DefaultCtx, k, v any) { dc.Locals(k, v) }  // ... argument does not escape
```

and the same two wrappers benchmarked side by side (`GOMAXPROCS=1`, `-count=3`, all runs within ±2%):

```
Benchmark_ViaInterface   30.6–31.7 ns/op   16 B/op   1 allocs/op
Benchmark_ViaConcrete     7.9 ns/op         0 B/op   0 allocs/op
```

So the assertion is what removes the allocation; the fallback branch only exists so a custom `Ctx` from `NewWithCustomCtx` still gets its own `Locals` called (there is a test for that). It is the same trick `StoreInContext` uses in package `fiber` via `setLocal`; `internal/ctxlocal.Set` is that helper for the packages that cannot reach an unexported one.

If you would rather not have the type assertion at all, the alternative is a non-variadic setter on the `Ctx` interface (e.g. `SetLocal(key, value any)`), which gets the same result through dynamic dispatch with no allocation — but that is a public API addition, so I left it as your call. Happy to do it as a follow-up commit on this PR if you prefer that shape.

---

gaby: That's a wild optimization, it does change the UX. Since users are used to `c.Local`.

@ReneWerner87 Do we do this, or revert it?

gaby: One clarification so the decision is made on the right facts: as merged into this PR, nothing user-facing changes. `c.Locals(k, v)` keeps its signature and behaviour and users keep calling it exactly as they do today. `internal/ctxlocal.Set` is an internal package that only Fiber's own middleware and the extractors call (twenty sites), and `setLocal` in package `fiber` backs `StoreInContext` the same way; a custom `Ctx` still gets its own `Locals` called. The exported `SetLocal` on the `Ctx` interface was only an alternative I offered above in case the type assertion itself is unwanted — that one *would* be new API, and it is not in the PR.

So the choice is between two self-contained shapes, either of which is one commit on my side:

- **Keep**: no API or behaviour change; removes one heap allocation per `Locals` write on the middleware paths (basicauth, keyauth, csrf, session, requestid, paginate, idempotency, earlydata, the extractors' chain guards). Measured end to end on basicauth: 48 → 32 B/op, 3 → 2 allocs/op.
- **Revert**: drop `internal/ctxlocal`, `setLocal` and the type assertion, back to plain `c.Locals` everywhere. The rest of the PR (utils adoption, path fold, CORS, logger) is unaffected.

Say which and I'll push it.

---

ReneWerner87: thx for the explaination, really helpful

ReneWerner87 (approved):

Issue #2714, opened 2023-11-07
🚀 [v3 Maintenance]: Perform and Document Benchmarks for v2 vs v3 Features for Release Notes


ReneWerner87:
### Maintenance Task Description

This issue is to track the performance benchmarks comparing the existing v2 features against the new v3 features. The objective is to quantify the performance improvements or regressions and to document these findings for inclusion in the v3 release notes.

## Tasks
- [ ] Identify key features and performance metrics to benchmark between v2 and v3.
- [ ] Execute the benchmarks, ensuring to keep the tests fair and consistent across both versions.
- [ ] Analyze the benchmark results to identify any significant performance changes.
- [ ] Document the benchmarking process and results in a clear and concise manner.
- [ ] Determine the best way to present this data in the v3 release notes.

## Expected Outcome
- A comprehensive benchmark report highlighting the performance comparisons between v2 and v3 features.
- Informed decisions on whether to display benchmark data in the release notes based on the significance of the performance findings.

### Checklist:

- [X] I have confirmed that this maintenance task is currently not being addressed.
- [X] I understand that this task will be evaluated by the maintainers and prioritized accordingly.
- [X] I am available to provide further information if needed.

fzlzjerry:
I'd like to work on a reproducible v2/v3 comparison for this. I plan to compare v2.52.15 and v3.5.0, with the current v3 commit shown separately, using independent modules so dependency resolution doesn't silently force both majors onto the same fasthttp version.

The initial matrix will cover routing, parameter/query access, JSON binding and responses, and representative middleware. I'll validate equivalent outputs before timing, use the same Go version and CPU settings, interleave repeated runs, and include raw samples plus latency/allocation results rather than a blanket “v3 is faster” claim. This is separate from #4666's before/after optimization measurements within v3.

Are there particular feature comparisons you'd prioritize for the release-note summary?

PR #4669, opened 2026-09-08
📒 docs: add reproducible v2/v3 feature benchmark comparison

fzlzjerry:
# Description

Refs #2714.

Adds a reproducible comparison of v2.52.15, v3.5.0, and a separately labeled development snapshot. Each version has an independent module graph and benchmark binary, avoiding a shared module silently raising v2's fasthttp dependency. No production Fiber code or root module dependencies change.

The [report](https://github.com/fzlzjerry/fiber/blob/3f69d4a202af27badf8c9ada387ee026f100f7ae/benchmarks/versioncompare/REPORT.md) covers 22 scenarios across routing, request access, JSON binding/responses, and middleware. It includes all 792 observations from 12 balanced rounds, module metadata, functional checks, and benchstat output. Results are mixed, so the release-note recommendation avoids an aggregate speedup claim. For example, query binding uses fewer allocations, while the tested CORS paths and middleware locals have additional costs.

## Changes introduced

- [x] Benchmarks: native v2/v3 API adapters, verified response bodies/required headers, explicit timing boundaries, and interleaved sampling with raw-data integrity checks.
- [x] Documentation: [reproduction instructions](https://github.com/fzlzjerry/fiber/blob/3f69d4a202af27badf8c9ada387ee026f100f7ae/benchmarks/versioncompare/README.md), the measured report, and an entry in the existing benchmark documentation.
- [x] Release-note guidance: a scoped example and limitations, not a general throughput claim. Default 404 behavior and the raw fasthttp floor are separate controls.
- [x] Spelling configuration: file-scoped technical vocabulary for the fixtures/docs; generated results and the dictionary configuration are excluded from prose checks without changing raw bytes.
- [x] Examples/tests: race-enabled correctness fixtures and 17 runner tests. The new path-scoped CI workflow validates fixtures without collecting performance numbers.

Most added lines are the retained result bundle, marked as generated for review. The harness and report are not marked generated. Raw Go/benchstat output is preserved byte-for-byte, including its padded CPU label.

## Local validation

Go 1.26.8, Linux/amd64:

| Command | Result |
| --- | --- |
| `make audit` | Pass; zero reachable findings, one non-reachable module finding |
| `make generate` | Pass |
| `GOFLAGS=-buildvcs=false make betteralign` | Pass |
| `make format` | Pass |
| `make lint` | Pass; 0 issues |
| `make test` | Pass; 5,813 reported tests, 1 Windows-only skip |

Also passed: all 22 functional scenarios for each of the three versions with the race detector, the 17 Python tests, actionlint v1.7.12, Markdown checks, CSpell 10.2.2 and codespell 2.4.3 on the changed files, and `git diff --check`. Re-summarizing the exported source-only bundle validates all 36 raw hashes and all 792 observations.

Tooling notes: the pinned alignment/lint tools fail internally with Go 1.27 on unchanged upstream, so these checks use Go 1.26.8; alignment also needed VCS stamping disabled in the local worktree. Generation reproduces existing comment-only drift in two interfaces, which is left out of this PR.

Measurements use a shared VM with CPU affinity, not an exclusive CPU. The report retains variation, unfavorable results, and the distinction between in-process operation time and end-to-end latency.

## Type of change

- [x] Documentation update

## Checklist

- [x] Reviewed the implementation and the complete branch diff.
- [x] Updated `/docs/` and added runnable correctness checks.
- [x] New fixtures and the existing test suite pass locally.
- [x] No library API changes, new runtime dependencies, or migration requirements.


ReneWerner87:
@gofiber/maintainers I don't think this belongs in the main branch, since it's code that compares different versions.
should we move this to another repo or branch?

fzlzjerry:
@ReneWerner87 The runner and result bundle can live separately. They already use independent modules with pinned published versions/pseudo-versions, so they do not depend on a local Fiber checkout.

Once the team agrees on a repository or branch, I'll move the harness, its correctness workflow, and the recorded results there, preserving the version pins and raw samples. This PR can then be reduced to a documentation link and a short scoped summary, if you'd like that in the main documentation. I'll hold further changes here until the destination is settled.

ReneWerner87:
jo saw it , thx

gaby:
I'm not sure about this, but;
- The PR shouldn't include benchmarks already run, the CI should do that for consistency.
- The table should show a ratio of difference.
- There's too much fluff on the markdown file, trim it down.
- We should be using the native "benchstat" from golang to compare benchmarks, not a random python script. See https://pkg.go.dev/golang.org/x/perf/cmd/benchstat

fzlzjerry:
@gaby For the relocated version, I'll keep raw runs as CI artifacts rather than checked-in files, use benchstat's comparison table with ratios, and cut the prose. The comparisons already use Go's `benchstat`; I'll remove the extra Python summary and use a small Go/shell runner so the path from benchmark output to comparison is explicit.

I'll apply that in the repository or branch the team chooses, with only a documentation link and short summary left in main if wanted.

ReneWerner87:
Thanks for the patience, @fzlzjerry. Here's how we'd like to continue:

**Where:** a new repository, `gofiber/benchmarks`, set up like our other repos.

**Shape** (covering @gaby's points):
- `v2/` and `v3/` are separate Go modules with the Fiber version pinned in `go.mod`. Dependabot bumps only that pin, so each version keeps the dependencies it ships with.
- `harness_test.go` (timing and response checks) is identical in both modules, the runner stops if they differ. `scenarios_test.go` holds your scenarios in each version's native API.
- `compare.sh` runs the prebuilt test binaries interleaved and compares them with `benchstat`. No Python.
- No checked-in results. CI runs the comparison on every push and PR; the benchstat table goes into the job summary, the raw output into an artifact. The numbers here already show why: since #4666, `cors_simple` is at 0 allocs and `cors_preflight` at 2 on main.
- A short README instead of the report.

I've set that up based on your scenarios and pushed it as the initial commit with you as co-author: https://github.com/gofiber/benchmarks. Reviews and follow-up scenarios are welcome as PRs there.

**This PR:** could you reduce it to a short section in `docs/extra/benchmarks.md` that links to `gofiber/benchmarks`, without numbers? Then we can merge it and close #2714.

ReneWerner87:
Thanks again @fzlzjerry. To save you the extra round, we added the documentation link ourselves in #4683, so I'm closing this one.

Your scenarios live on in https://github.com/gofiber/benchmarks, where you're credited as co-author of the initial commit and in the README. Once gofiber/benchmarks#1 is merged, CI publishes every run to https://gofiber.github.io/benchmarks/. Follow-up scenarios and improvements are very welcome as PRs there.

PR #4677, opened 2026-09-12
🐛 fix(limiter): bound the rate-limit key before it reaches storage

BitWeaverDev:
# Description

Cluster 8 of #4604 flags the asymmetry directly: middleware/cache bounds and escapes storage keys built from client-influenced input (`boundKeySegment`); `middleware/limiter` never picked up the same discipline, and it builds a storage key from client-influenced input on every single request.

`cfg.KeyGenerator(c)` is handed straight to the manager, and from there straight to storage, with no bound at all.

The default `KeyGenerator` returns `c.IP()`. With `app.Config.TrustProxy` and `ProxyHeader` set — the standard configuration for anything sitting behind a reverse proxy or load balancer — and `EnableIPValidation` left at its documented default of `false`, `c.IP()` returns the raw `ProxyHeader` value verbatim. That behavior is already documented, in the "Behavior note" on `ProxyHeader` in `docs/api/fiber.md`:

> With the default `EnableIPValidation = false`, `c.IP()` will return the raw header value (the whole chain) rather than a single parsed client IP.

That raw value is an attacker-controlled string with no length limit, and it lands directly as a storage key. A custom `KeyGenerator` reading any other header, cookie, or query value is exposed the same way — this isn't specific to `c.IP()`, it's specific to handing an unbounded, request-derived string to storage.

Left unbounded, a client can grow the backing store — or the unbounded in-memory map on the no-`Storage` default path — by an amount of its own choosing, on every request, simply by sending a fresh oversized value each time.

Worth noting: `manager.go` already treats this key as sensitive enough to redact from logs by default (`shouldRedactKeys`, `logKey`). It just never extended that same wariness to the storage layer itself, where the actual cost lives.

Related #4604

## Fix

Bounded at the one seam both window strategies (`FixedWindow`, `SlidingWindow`) call through — `manager.get` and `manager.set` — rather than in each strategy file. That means a future third strategy can't reintroduce the gap by forgetting to bound the key itself, and the in-memory fallback path (no `Storage` configured) is covered along with external storage, since both branches sit downstream of the same two functions.

```go
func boundKey(key string) string {
	if len(key) <= maxKeyLength && !strings.HasPrefix(key, hashPrefix) {
		return key
	}
	hash := sha256.Sum256(utils.UnsafeBytes(key))
	return hashPrefix + hex.EncodeToString(hash[:])
}
```

Hashing rather than truncating or rejecting keeps distinct oversized inputs mapped to distinct buckets, so the client that triggers this is still rate-limited correctly rather than exempted (truncation could let a client tack on a rotating suffix that changes only past the truncation point) or itself denied service by the fix.

**Not sharing an implementation with cache's `boundKeySegment`.** That function's job includes escaping delimiters (`|`, `:`, `\`) between multiple concatenated segments — it's assembling one key out of several dimensions (method, path, headers, cookies...), and needs escaping precisely because those pieces are joined together. `limiter`'s key is a single opaque string with nothing to delimit. Only the length-bound-with-hash-fallback half of cache's pattern applies here, so that's the only half this borrows. `192` reuses cache's own `maxKeyDimensionSegmentLength` value — not a fresh arbitrary number, but the same policy the codebase already committed to for this exact class of decision.

The reserved-prefix guard (`!strings.HasPrefix(key, hashPrefix)`) mirrors cache's own collision defense for the same reason: without it, a short key that happens to start with `"sha256:"` would pass through unhashed, and could then be confused for — or deliberately chosen to land in — the bucket a genuinely oversized key hashes to.

## Verification

- **`Test_Limiter_OversizedKeyGeneratorIsBounded`** drives the real middleware (both `FixedWindow` and `SlidingWindow`, as subtests) with an oversized `KeyGenerator` and inspects what actually reached a recording storage double: never the raw key, always `<= 192` bytes, always under the reserved `hashPrefix`. It also confirms the *same* oversized key is still recognized as the same client on a second request (still lands in one bucket, still gets rate-limited), so the fix doesn't silently exempt or misattribute whoever triggers it.
- **Mutation-verified**: reverted the two `boundKey` call sites, confirmed the test above fails on *both* strategies (the raw key's length shows up in the failure output), restored the fix, confirmed it passes again.
- **`Test_BoundKey_ReservedPrefixCannotBeForged`** pins the collision guard specifically: a short key claiming the reserved prefix must still be re-hashed.
- Plus boundary tests on `boundKey` directly: exactly at the limit (unchanged), one over (hashed), empty, determinism, and that two different oversized inputs hash to two different buckets.

All 53 packages pass, `-race` clean on `limiter`, `golangci-lint` (repo-pinned v2.12.2) reports **0 issues** repo-wide, coverage on the package is **94.6%**.

## Benchmarks

`benchstat`, interleaved A/B against `upstream/main`, `-count=8`, on the two existing limiter benchmarks:

```
                         │ before       │ after                              │
_Limiter_Custom_Store-10    190.6n ± 2%   190.6n ± 2%   ~ (p=0.820)
_Limiter-10                 135.3n ± 1%   137.0n ± 1%   +1.22% (p=0.014)

B/op and allocs/op: identical on both, "all samples are equal"
```

Allocations are unchanged on both. `_Limiter_Custom_Store` shows no measurable difference at all. `_Limiter` is a real, small cost — one length check and one prefix check, run twice per request (once in `get`, once in `set`) — on the path a well-behaved short key already takes. About 1.7ns. Given what it buys, I think that's the right trade, but flagging it plainly rather than letting CI find it first.

## Type of change

## Checklist

- [x] Added or updated unit tests to validate the effectiveness of the changes. See Verification above; the key test is mutation-verified, not just present.
- [x] Verified that any new dependencies are essential. None added — `crypto/sha256` and `encoding/hex` are stdlib, already used by `middleware/cache` for the identical purpose.
- [ ] Updated the documentation in `/docs/`. Not applicable: this closes a robustness gap in the default behavior, it doesn't add or change public API or configuration.

---

This PR was written in part with the assistance of generative AI.


gaby:
@codex review

gaby (approved):
LGTM

PR #4701, opened 2026-09-22
🐛 bug: drain in-flight requests before Listen returns

### LINKED ISSUE #3370: 🐛 [Bug]: "ShutdownWithTimeout" & "ShutdownTimeout" not respected

ReneWerner87:
## Summary

`ListenConfig.ShutdownTimeout` had no observable effect for the caller: `Listen`/`Listener` returned while the graceful shutdown they started was still draining, so the caller's `main` exited and cut the connections the timeout was meant to protect. Reported in #3370 as a 10 GB download being killed on `SIGINT`.

Root cause is in fasthttp, not in the timeout: `workerPool.Stop()` deliberately does not wait for busy workers (`workerpool.go:81-83`), so `Server.Serve()` returns as soon as the listener closes. Both Fiber entry points end with `return app.server.Serve(ln)`, so the shutdown goroutine was still running at that point.

`ShutdownWithContext` itself was never the problem: it loops until `s.open == 0`, and a connection actively writing a response has `idleConnTime == 0`, so `closeIdleConns` leaves it alone.

## Changes

- `App.startGracefulShutdown` replaces the duplicated `GracefulContext` wiring in `Listen` and `Listener`. It returns the cleanup both defer: stop the watcher, then wait for an in-progress shutdown to finish. `close(stop)` has to run before the wait, or a watcher still parked in its `select` could never exit.
- `Test_Listen_GracefulShutdown_DrainsBeforeReturn` asserts the in-flight handler has completed by the time `Listener` returns. It fails on `main`.
- One sentence in `docs/api/fiber.md` stating the new contract: on the built-in path the caller may exit as soon as `Listen` returns, while an explicit `Shutdown*` call still needs the process kept alive.

## Behavior notes

- No change when `GracefulContext` is unset.
- `ShutdownTimeout` still bounds the wait (default 10s). `ShutdownTimeout: -1` means wait indefinitely, which now also blocks `Listen` indefinitely, which is what that setting asks for.
- Prefork is unchanged: each process still shuts down on its own, and a child now drains before its `Listen` returns.

Closes #3370

## Checks

`go test -race ./...` 6049 tests green, `golangci-lint` v2.12.2 clean, `go vet` and `go mod verify` clean.


ReneWerner87:
On the outside-diff hint for `listen_test.go:1544-1549` (assert response completion): changed in e22a828d, though not as suggested. A flag read at the moment `Listener` returns races with the client goroutine, which reads the response on its own schedule, so that assertion would flake. The test now waits for the client's result after `Listener` returns and requires `200 ok`, which fails if the shutdown cuts off the drained response.

gaby (approved):

